PR #6, opened 2015-04-21
Enhanced multigrid solvers, including for Stokes-IB

boyceg:


ibtk/include/ibtk/PETScMatUtilities.h
```diff
@@ -159,18 +130,40 @@ class PETScMatUtilities
...
+                                      int dof_index_idx,
+                                      const std::vector<int>& num_fine_dofs_per_proc,
+                                      const std::vector<int>& num_coarse_dofs_per_proc,
+                                      SAMRAI::tbox::Pointer<SAMRAI::hier::PatchLevel<NDIM> > fine_patch_level,
+                                      SAMRAI::tbox::Pointer<SAMRAI::hier::PatchLevel<NDIM> > coarse_patch_level,
+                                      const AO& coarse_level_ao,
+                                      const int coarse_ao_offset = 0);
+
+    /*!
+     * \brief Construct a diagonal matrix that scales the rows of adjoint of
```
boyceg: Adjoint or transpose? Transpose, right?

ibtk/include/ibtk/IndexUtilities.h
```diff
@@ -96,6 +96,29 @@ class IndexUtilities
                                                   const double* dx,
                                                   const SAMRAI::hier::Index<NDIM>& ilower,
                                                   const SAMRAI::hier::Index<NDIM>& iupper);
+	
+	/*!
+	 * \brief Convert (i,j,k,d) index for a DOF to an integer mapping on a patch level.
```
boyceg: I'd like these comments to be a bit more clear about what this routine is doing.

ibtk/include/ibtk/PETScMatUtilities.h
knepley: On Tue, Apr 21, 2015 at 10:22 AM, Boyce Griffith <notifications@github.com>
wrote:

> In ibtk/include/ibtk/PETScMatUtilities.h
> <https://github.com/IBAMR/IBAMR/pull/6#discussion_r28788672>:
>
> > +    /*!
> > +     * \brief Construct a parallel PETSc Mat object corresponding to data
> > +     * prolongation from a coarser level to a finer level.
> > +     */
> > +    static void
> > +    constructPatchLevelProlongationOp(Mat& mat,
> > +                                      int dof_index_idx,
> > +                                      const std::vector<int>& num_fine_dofs_per_proc,
> > +                                      const std::vector<int>& num_coarse_dofs_per_proc,
> > +                                      SAMRAI::tbox::Pointer<SAMRAI::hier::PatchLevel<NDIM> > fine_patch_level,
> > +                                      SAMRAI::tbox::Pointer<SAMRAI::hier::PatchLevel<NDIM> > coarse_patch_level,
> > +                                      const AO& coarse_level_ao,
> > +                                      const int coarse_ao_offset = 0);
> > +
> > +    /*!
> > +     * \brief Construct a diagonal matrix that scales the rows of adjoint of
>
> Adjoint or transpose? Transpose, right?
>
For real matrices, they are the same.

  Matt

> —
> <https://github.com/IBAMR/IBAMR/pull/6/files#r28788672>.
>



-- 
What most experimenters take for granted before they begin their
experiments is infinitely more interesting than any results to which their
experiments lead.
-- Norbert Wiener

ibtk/include/ibtk/IndexUtilities.h
```diff
@@ -96,6 +96,29 @@ class IndexUtilities
...
+	 * \param num_cells Number of data array cells for a patch level.
+	 *
+	 * \param depth Data depth.
+	 *
+	 * \param offset Component (axis) offset.
+	 *
+	 * \return The linear mapping of an AMR index to a continuous non-negative 
+	 * integer space.
+	 */
+	static int getIntegerMapping(const SAMRAI::hier::Index<NDIM>& i,
```
boyceg: Why getIntegerMapping()?

Suggest: getIntegerIndex()

ibtk/include/ibtk/IndexUtilities.h
```diff
@@ -96,6 +96,29 @@ class IndexUtilities
...
+								 const SAMRAI::hier::Index<NDIM>& domain_lower,
```
boyceg: Formatting problems.

ibtk/include/ibtk/PETScMatUtilities.h
```diff
@@ -159,18 +130,40 @@ class PETScMatUtilities
                                               int dof_index_idx,
                                               SAMRAI::tbox::Pointer<SAMRAI::hier::PatchLevel<NDIM> > patch_level);
 
-    static inline void ib_4_interp_fcn(const double r, double* const w)
```
boyceg: Please add some comments about how this works / what this is doing.

ibtk/include/ibtk/PETScMatUtilities.h
```diff
@@ -159,18 +130,40 @@ class PETScMatUtilities
...
     } // ib_4_interp_fcn
 
     static const int ib_4_interp_stencil = 4;
 
+    /*!
+     * \brief Construct a parallel PETSc Mat object corresponding to data
+     * prolongation from a coarser level to a finer level.
+     */
+    static void
+    constructPatchLevelProlongationOp(Mat& mat,
```
boyceg: constructProlongationOp() ?

ibtk/include/ibtk/PETScMatUtilities.h
```diff
@@ -159,18 +130,40 @@ class PETScMatUtilities
...
+                                      const AO& coarse_level_ao,
+                                      const int coarse_ao_offset = 0);
+
+    /*!
+     * \brief Construct a diagonal matrix that scales the rows of adjoint of
+     * prolongation matrix to get a suitable restriction matrix.
+     *
+     * \NOTE We store the diagonal enteries into a Vec rather than a Mat.
+     */
+    static void constructRestrictionScalingMat(Mat& P, Vec& L);
```
boyceg: Pick: XXX{Restriction,Prolongation}Mat() or XXX{Restriction,Prolongation}Op()

ibtk/include/ibtk/PETScVecUtilities.h
```diff
@@ -121,6 +122,16 @@ class PETScVecUtilities
     static void constructPatchLevelDOFIndices(std::vector<int>& num_dofs_per_proc,
                                               int dof_index_idx,
                                               SAMRAI::tbox::Pointer<SAMRAI::hier::PatchLevel<NDIM> > patch_level);
+	/*!
+	 * \brief Create an application ordering object (AO) that creates a mapping between
+	 * global PETSc indices and SAMRAI integer indices on a SAMRAI::hier::PatchLevel.
+	 */
```
boyceg: Say a little more about what SAMRAI integer indices are. Where do these come from?

ibtk/include/ibtk/PETScVecUtilities.h
```diff
@@ -121,6 +122,16 @@ class PETScVecUtilities
...
+	static void constructPatchLevelAO(AO& ao,
+									  std::vector<int>& num_dofs_per_proc,
```
boyceg: Formatting...

ibtk/include/ibtk/PETScMatUtilities.h
amneetb: Hmm? It's a standard ib_4 function, nothing special. I just fixed it. The one you had earlier worked with a shift of 1, which was confusing. We fixed that shift of 1 (i instead of i+1) in the code.

boyceg: > On Apr 21, 2015, at 11:39 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/include/ibtk/PETScMatUtilities.h <https://github.com/IBAMR/IBAMR/pull/6#discussion_r28790513>:
> 
> > @@ -159,18 +130,40 @@ class PETScMatUtilities
> >                                                int dof_index_idx,
> >                                                SAMRAI::tbox::Pointer<SAMRAI::hier::PatchLevel<NDIM> > patch_level);
> >  
> > -    static inline void ib_4_interp_fcn(const double r, double* const w)
> Hmm? It's a standard ib_4 function, nothing special. I just fixed it. The one you had earlier worked with a shift of 1, which was confusing. We fixed that shift of 1 (i instead of i+1) in the code.
> 
We should include some comments about what this is computing. It took us at least a few minutes to decode this earlier in the semester, didn't it?

ibtk/include/ibtk/PETScMatUtilities.h
amneetb: Makes sense if you think of it as an Op that goes from one level to another, but one can think of it as belonging to the level of origin. Both are fine to me. I can rename it if you prefer...

boyceg: > On Apr 21, 2015, at 11:47 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/include/ibtk/PETScMatUtilities.h <https://github.com/IBAMR/IBAMR/pull/6#discussion_r28791527>:
> 
> >          return;
> >      } // ib_4_interp_fcn
> >  
> >      static const int ib_4_interp_stencil = 4;
> >  
> > +    /*!
> > +     * \brief Construct a parallel PETSc Mat object corresponding to data
> > +     * prolongation from a coarser level to a finer level.
> > +     */
> > +    static void
> > +    constructPatchLevelProlongationOp(Mat& mat,
> Makes sense if you think of it as an Op that goes from one level to another, but one can think of it as belonging to the level of origin. Both are fine to me. I can rename it if you prefer...
> 
I don't care which one you use --- I like Op just fine --- the point was that you called a related function XXXMat().

-- Boyce

ibtk/include/ibtk/PETScMatUtilities.h
amneetb: I had thought about it, but scaling didn't sound like an operator to me at that time. But I can switch to 'Op' for consistency...

ibtk/include/ibtk/PETScMatUtilities.h
amneetb: Your comment seemed to suggest removing patch level from constructPatchLevelProlongationOp ==> constructProlongationOp and not between Op and Mat switch...

ibtk/include/ibtk/PETScMatUtilities.h
amneetb: Yeah, it took us a while to figure it out. Ok, I will add comments to that affect..

ibtk/include/ibtk/IndexUtilities.h
amneetb: I will touch these files in XCode for formatting...

ibtk/include/ibtk/PETScMatUtilities.h
boyceg: > On Apr 21, 2015, at 11:54 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/include/ibtk/PETScMatUtilities.h <https://github.com/IBAMR/IBAMR/pull/6#discussion_r28792311>:
> 
> >          return;
> >      } // ib_4_interp_fcn
> >  
> >      static const int ib_4_interp_stencil = 4;
> >  
> > +    /*!
> > +     * \brief Construct a parallel PETSc Mat object corresponding to data
> > +     * prolongation from a coarser level to a finer level.
> > +     */
> > +    static void
> > +    constructPatchLevelProlongationOp(Mat& mat,
> Your comment seemed to suggest removing patch level from constructPatchLevelProlongationOp ==> constructProlongationOp and not between Op and Mat switch...
> 

Yes.

I guess the question is: what would be the alternative to a "patch level prolongation operator" here?

ibtk/include/ibtk/PETScMatUtilities.h
amneetb: I will adjoint (transpose) in the comment :-)

amneetb: I will put adjoint (transpose) in the comment...

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +618,460 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+            }
+        }
+    }
+
+    // Create an empty matrix.
+    ierr = MatCreateAIJ(PETSC_COMM_WORLD, m_local, n_local, PETSC_DETERMINE, PETSC_DETERMINE, PETSC_DEFAULT, &d_nnz[0],
+                        PETSC_DEFAULT, &o_nnz[0], &mat);
+    IBTK_CHKERRQ(ierr);
+
+// Set some general matrix options.
```
boyceg: I think these are now default options in PETSc.

ibtk/include/ibtk/IndexUtilities.h
amneetb: Index seems to suggest that something exists in a collection of things, while mapping suggests something may or may not exist. Here mapped values do not exist for any vector or operator (SAMRAI or PETSc). Hence, I preferred mapping. I can change to index if mapping makes things obscure...

ibtk/include/ibtk/PETScMatUtilities.h
amneetb: Can there be something like "multi patch level prolongation operator", that takes entries from multiple levels to another level(s)?

boyceg: > On Apr 21, 2015, at 12:17 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/include/ibtk/PETScMatUtilities.h <https://github.com/IBAMR/IBAMR/pull/6#discussion_r28794932>:
> 
> >          return;
> >      } // ib_4_interp_fcn
> >  
> >      static const int ib_4_interp_stencil = 4;
> >  
> > +    /*!
> > +     * \brief Construct a parallel PETSc Mat object corresponding to data
> > +     * prolongation from a coarser level to a finer level.
> > +     */
> > +    static void
> > +    constructPatchLevelProlongationOp(Mat& mat,
> Can there be something like "multi patch level prolongation operator", that takes entries from multiple levels to another level(s)?
> 

Not with properly nested levels and compact interpolation stencils.

ibtk/include/ibtk/IndexUtilities.h
boyceg: > On Apr 21, 2015, at 12:14 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/include/ibtk/IndexUtilities.h <https://github.com/IBAMR/IBAMR/pull/6#discussion_r28794620>:
> 
> > +	 * \brief Convert (i,j,k,d) index for a DOF to an integer mapping on a patch level.
> > +	 *
> > +	 * \param i AMR index representing the (i,j,k) data cell index.
> > +	 *
> > +	 * \param domain_lower Lower index of the domain.
> > +	 *
> > +	 * \param num_cells Number of data array cells for a patch level.
> > +	 *
> > +	 * \param depth Data depth.
> > +	 *
> > +	 * \param offset Component (axis) offset.
> > +	 *
> > +	 * \return The linear mapping of an AMR index to a continuous non-negative 
> > +	 * integer space.
> > +	 */
> > +	static int getIntegerMapping(const SAMRAI::hier::Index<NDIM>& i,
> Index seems to suggest that something exists in a collection of things, while mapping suggests something may or may not exist. Here mapped values do not exist for any vector or operator (SAMRAI or PETSc). Hence, I preferred mapping. I can change to index if mapping makes things obscure...
> 

How about mapIndexToInteger() or mapIndexToUniqueInteger() or getIntegerIndex()?

To me, getIntegerMapping() sounds like it is giving you an object that stores the mapping from indexes to integers.

amneetb: mapIndexToInteger() sounds good..

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +618,460 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+    ierr = MatGetSize(P, &M, &N);
+    IBTK_CHKERRQ(ierr);
+    std::vector<double> column_sum_inv(N);
+    ierr = MatGetColumnNorms(P, NORM_1, &column_sum_inv[0]);
+    IBTK_CHKERRQ(ierr);
+
+    for (int k = 0; k < N; ++k)
+    {
+        const double sum = column_sum_inv[k];
+        if (sum > MathUtilities<double>::equalEps(sum, 0.0))
```
boyceg: This line looks suspect.

amneetb: Indeed! If (!MathUtilities<double>::equalEps(sum, 0.0)), does this look right?

boyceg: > On Apr 22, 2015, at 9:58 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/math/PETScMatUtilities.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r28874134>:
> 
> > +    // of PETSc which appears to be faster than row sum call from the documentation.
> > +    // We might have a column of all zeros for pressure DOFs. Therefore,
> > +    // care should be taken for this case.
> > +
> > +    int ierr;
> > +    PetscInt M, N;
> > +    ierr = MatGetSize(P, &M, &N);
> > +    IBTK_CHKERRQ(ierr);
> > +    std::vector<double> column_sum_inv(N);
> > +    ierr = MatGetColumnNorms(P, NORM_1, &column_sum_inv[0]);
> > +    IBTK_CHKERRQ(ierr);
> > +
> > +    for (int k = 0; k < N; ++k)
> > +    {
> > +        const double sum = column_sum_inv[k];
> > +        if (sum > MathUtilities<double>::equalEps(sum, 0.0))
> Indeed! If (!MathUtilities::equalEps(sum, 0.0)), does this look right?
> 
If you want to test whether sum is different from zero, then yes.

amneetb: yes, that's the intent.

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +618,460 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+
+    // Determine the non-zero matrix structure for constant refine.
+    std::vector<int> d_nnz(m_local, 0), o_nnz(m_local, 0);
+    for (PatchLevel<NDIM>::Iterator p(fine_patch_level); p; p++)
+    {
+        Pointer<Patch<NDIM> > fine_patch = fine_patch_level->getPatch(p());
+        const Box<NDIM>& fine_patch_box = fine_patch->getBox();
+        Pointer<CellData<NDIM, int> > dof_fine_data = fine_patch->getPatchData(dof_index_idx);
+        const unsigned depth = dof_fine_data->getDepth();
+        std::vector<int> samrai_petsc_map(depth), local_row(depth);
```
boyceg: Similar comment as in the interface --- use map for an actual mapping. These should be something like "idx", shouldn't they, or are they really "maps"?

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +618,460 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+#endif
+
+    // Determine matrix-coefficients
+    for (PatchLevel<NDIM>::Iterator p(fine_patch_level); p; p++)
+    {
+        Pointer<Patch<NDIM> > fine_patch = fine_patch_level->getPatch(p());
+        const Box<NDIM>& fine_patch_box = fine_patch->getBox();
+        Pointer<CellData<NDIM, int> > dof_fine_data = fine_patch->getPatchData(dof_index_idx);
+        const unsigned depth = dof_fine_data->getDepth();
+        std::vector<int> samrai_petsc_map(depth);
```
boyceg: Same comment as above.

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +618,460 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+
+    // Determine the non-zero matrix structure for the refine operator.
+    std::vector<int> d_nnz(m_local, 0), o_nnz(m_local, 0);
+    for (PatchLevel<NDIM>::Iterator p(fine_patch_level); p; p++)
+    {
+        Pointer<Patch<NDIM> > fine_patch = fine_patch_level->getPatch(p());
+        const Box<NDIM>& fine_patch_box = fine_patch->getBox();
+        Pointer<SideData<NDIM, int> > fine_dof_data = fine_patch->getPatchData(dof_index_idx);
+        const unsigned depth = fine_dof_data->getDepth();
+        std::vector<int> samrai_petsc_map(2 * depth), local_row(depth);
```
boyceg: Also here.

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +618,460 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+        }
+    }
+
+    // Create an empty matrix.
+    ierr = MatCreateAIJ(PETSC_COMM_WORLD, m_local, n_local, PETSC_DETERMINE, PETSC_DETERMINE, PETSC_DEFAULT, &d_nnz[0],
+                        PETSC_DEFAULT, &o_nnz[0], &mat);
+    IBTK_CHKERRQ(ierr);
+
+// Set some general matrix options.
+#if !defined(NDEBUG)
```
boyceg: I think you can omit these.

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +618,460 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+#endif
+
+    // Determine the matrix-coefficients
+    for (PatchLevel<NDIM>::Iterator p(fine_patch_level); p; p++)
+    {
+        Pointer<Patch<NDIM> > fine_patch = fine_patch_level->getPatch(p());
+        const Box<NDIM>& fine_patch_box = fine_patch->getBox();
+        Pointer<SideData<NDIM, int> > fine_dof_data = fine_patch->getPatchData(dof_index_idx);
+        const unsigned depth = fine_dof_data->getDepth();
+        std::vector<int> samrai_petsc_map(2 * depth);
```
boyceg: Same as above.

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +618,460 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+                        I_U, coarse_domain_lower, coarse_num_cells[axis], d, coarse_ao_offset + data_offset);
+                }
+                AOApplicationToPetsc(coarse_level_ao, 2 * depth, &samrai_petsc_map[0]);
+
+                for (unsigned d = 0; d < depth; ++d)
+                {
+                    int row = (*fine_dof_data)(i_s, d);
+                    int col[2] = { samrai_petsc_map[d], samrai_petsc_map[depth + d] };
+
+                    // To avoid calling refine(), the term in the bracket is same as:
```
boyceg: Why do you want to avoid calling refine?

ibtk/src/math/PETScMatUtilities.cpp
amneetb: I will. But when I read these lines, it somehow assures me that whatever nonzero spacing I have figured out, I am confident about it . It makes me happy to see these files, somehow :-)

ibtk/src/math/PETScMatUtilities.cpp
amneetb: It's not a map in the sense of std::map, but is a mapping from SAMRAI indices to PETSc indices and vice-versa. Both are OK to me.

ibtk/src/math/PETScMatUtilities.cpp
amneetb: Wee-bee of optimization ;-). Is the present formula confusing?

boyceg: > On Apr 23, 2015, at 10:20 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/math/PETScMatUtilities.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29019887>:
> 
> > +                const CellIndex<NDIM> I_U = I_L + offset;
> > +                for (unsigned d = 0; d < depth; ++d)
> > +                {
> > +                    samrai_petsc_map[d] = IndexUtilities::getIntegerMapping(
> > +                        I_L, coarse_domain_lower, coarse_num_cells[axis], d, coarse_ao_offset + data_offset);
> > +                    samrai_petsc_map[depth + d] = IndexUtilities::getIntegerMapping(
> > +                        I_U, coarse_domain_lower, coarse_num_cells[axis], d, coarse_ao_offset + data_offset);
> > +                }
> > +                AOApplicationToPetsc(coarse_level_ao, 2 * depth, &samrai_petsc_map[0]);
> > +
> > +                for (unsigned d = 0; d < depth; ++d)
> > +                {
> > +                    int row = (*fine_dof_data)(i_s, d);
> > +                    int col[2] = { samrai_petsc_map[d], samrai_petsc_map[depth + d] };
> > +
> > +                    // To avoid calling refine(), the term in the bracket is same as:
> Wee-bee of optimization ;-). Is the present formula confusing?
> 

https://shreevatsa.wordpress.com/2008/05/16/premature-optimization-is-the-root-of-all-evil/ <https://shreevatsa.wordpress.com/2008/05/16/premature-optimization-is-the-root-of-all-evil/>

ibtk/src/math/PETScMatUtilities.cpp
boyceg: > On Apr 23, 2015, at 10:17 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/math/PETScMatUtilities.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29019748>:
> 
> > +    const int i_fine_lower =
> > +        std::accumulate(num_fine_dofs_per_proc.begin(), num_fine_dofs_per_proc.begin() + mpi_rank, 0);
> > +    const int i_fine_upper = i_fine_lower + m_local;
> > +    const int j_coarse_lower =
> > +        std::accumulate(num_coarse_dofs_per_proc.begin(), num_coarse_dofs_per_proc.begin() + mpi_rank, 0);
> > +    const int j_coarse_upper = j_coarse_lower + n_local;
> > +
> > +    // Determine the non-zero matrix structure for constant refine.
> > +    std::vector<int> d_nnz(m_local, 0), o_nnz(m_local, 0);
> > +    for (PatchLevel<NDIM>::Iterator p(fine_patch_level); p; p++)
> > +    {
> > +        Pointer<Patch<NDIM> > fine_patch = fine_patch_level->getPatch(p());
> > +        const Box<NDIM>& fine_patch_box = fine_patch->getBox();
> > +        Pointer<CellData<NDIM, int> > dof_fine_data = fine_patch->getPatchData(dof_index_idx);
> > +        const unsigned depth = dof_fine_data->getDepth();
> > +        std::vector<int> samrai_petsc_map(depth), local_row(depth);
> It's not a map in the sense of std::map, but is a mapping from SAMRAI indices to PETSc indices and vice-versa. Both are OK to me.
> 
In that case it is fine --- just wanted to check.

ibtk/src/math/PETScMatUtilities.cpp
boyceg: https://shreevatsa.wordpress.com/2008/05/16/premature-optimization-is-the-root-of-all-evil/

ibtk/src/math/PoissonUtilities.cpp
```diff
@@ -994,32 +628,31 @@ void PoissonUtilities::computeSCMatrixCoefficients(Pointer<Patch<NDIM> > patch,
...
+                                                   const PoissonSpecifications& poisson_spec,
+                                                   const std::vector<RobinBcCoefStrategy<NDIM>*>& bc_coefs,
+                                                   double data_time,
+                                                   bool homogeneous_bc)
 {
-    const int depth = static_cast<int>(bc_coefs.size());
+    const int depth = rhs_data.getDepth();
 #if !defined(NDEBUG)
-    TBOX_ASSERT(rhs_data.getDepth() == depth);
+    TBOX_ASSERT((int)bc_coefs.size() == depth);
```
boyceg: Please use C++ casts.

ibtk/src/solvers/impls/CCPoissonBoxRelaxationFACOperator.cpp
```diff
@@ -0,0 +1,1189 @@
...
+
+// Type of extrapolation to use at physical boundaries; used only to evaluate
+// composite grid residuals.
+static const std::string BDRY_EXTRAP_TYPE = "LINEAR";
+
+// Whether to enforce consistent interpolated values at Type 2 coarse-fine
+// interface ghost cells; used only to evaluate composite grid residuals.
+static const bool CONSISTENT_TYPE_2_BDRY = false;
+
+struct IndexComp : std::binary_function<Index<NDIM>, Index<NDIM>, bool>
```
boyceg: Need a common impl of these comparison objects. (I am very guilty here.)

ibtk/src/math/PoissonUtilities.cpp
amneetb: Yep, that's the right thing to do, but C casts are so succinct to use. I will change it. I almost first use static_cast<> and then I realize how ugly they look and then covert it to C cast :-)

boyceg: > On Apr 23, 2015, at 10:37 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/math/PoissonUtilities.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29020359>:
> 
> >  #if !defined(NDEBUG)
> > -    TBOX_ASSERT(rhs_data.getDepth() == depth);
> > +    TBOX_ASSERT((int)bc_coefs.size() == depth);
> Yep, that's the right thing to do, but C casts are so succinct to use. I will change it. I almost first use static_cast<> and then I realize how ugly they look and then covert it to C cast :-)
> 
C++ casts are intentionally ugly, and easy to grep for. Use C++ casts please.

include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h
```diff
@@ -150,11 +156,113 @@ class IBImplicitStaggeredHierarchyIntegrator : public IBHierarchyIntegrator
 
 private:
     /*!
-     * \brief Default constructor.
-     *
-     * \note This constructor is not implemented and should not be used.
-     */
-    IBImplicitStaggeredHierarchyIntegrator();
+	 * \brief A StaggeredStokesSolver that does nothing.
```
boyceg: Why is this needed? Should we change the design to avoid needing to do this?

include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h
```diff
@@ -233,10 +351,25 @@ class IBImplicitStaggeredHierarchyIntegrator : public IBHierarchyIntegrator
...
+	std::vector<std::vector<int> > d_num_dofs_per_proc;
+	int d_u_dof_index_idx, d_p_dof_index_idx;
+	SAMRAI::tbox::Pointer<SAMRAI::pdat::SideVariable<NDIM, int> > d_u_dof_index_var;
+	SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, int> > d_p_dof_index_var;
+
+	// Solvers and associated vectors.
+	bool d_solve_for_position;
     SAMRAI::tbox::Pointer<StaggeredStokesSolver> d_stokes_solver;
     SAMRAI::tbox::Pointer<StaggeredStokesOperator> d_stokes_op;
+	SAMRAI::tbox::Pointer<StaggeredStokesIBLevelRelaxationFACOperator> d_fac_op;
```
boyceg: These should be encapsulated in the Stokes Solver/Operator?

include/ibamr/IBImplicitStrategy.h
```diff
@@ -140,6 +146,15 @@ class IBImplicitStrategy : public IBStrategy
         const std::vector<SAMRAI::tbox::Pointer<SAMRAI::xfer::RefineSchedule<NDIM> > >& f_prolongation_scheds,
         double data_time) = 0;
 
+    /*!
+     * Construct the IB interpolation operator.
+     */
+    virtual void constructInterpOperator(Mat& J,
```
boyceg: Let's use consistent naming as above (e.g. InterpOp or InterpMat or whatever).

include/ibamr/IBMethod.h
```diff
@@ -289,6 +294,15 @@ class IBMethod : public IBImplicitStrategy
         double data_time);
 
     /*!
+     * Construct the IB interpolation operator.
+     */
+    void constructInterpOperator(Mat& J,
```
boyceg: Let's use consistent naming as above (e.g. InterpOp or InterpMat or whatever).

include/ibamr/KrylovLinearSolverStaggeredStokesSolverInterface.h
```diff
@@ -89,6 +89,12 @@ class KrylovLinearSolverStaggeredStokesSolverInterface : public StaggeredStokesS
      */
     void setVelocityPoissonSpecifications(const SAMRAI::solv::PoissonSpecifications& U_problem_coefs);
 
+	/*!
+	 * \brief Set if velocity and pressure have nullspace.
+	 */
+	void setComponentsHaveNullspace(const bool has_velocity_nullspace,
```
boyceg: (1) For velocity, is a boolean sufficient, or do we need to be passing in nullspace vectors?
(2) setComponentNullspaces().

include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h
```diff
@@ -1,7 +1,7 @@
 // Filename: IBImplicitStaggeredHierarchyIntegrator.h
-// Created on 07 Apr 2012 by Boyce Griffith
+// Created on 07 Apr 2012 by Boyce Griffith and Amneet Bhalla
```
boyceg: You created this class in 2012? ;-)

include/ibamr/StaggeredStokesIBLevelRelaxationFACOperator.h
```diff
@@ -0,0 +1,677 @@
...
+	 */
+	StaggeredStokesIBLevelRelaxationFACOperator& operator=(const StaggeredStokesIBLevelRelaxationFACOperator& that);
+
+    /*!
+     * \name Methods for executing, caching, and resetting communication
+     * schedules.
+     */
+    //\{
+
+    /*!
```
boyceg: Do we need a common base class for Vanka-like FAC implementations?

include/ibamr/StaggeredStokesIBLevelRelaxationFACOperator.h
```diff
@@ -0,0 +1,677 @@
...
+	 */
+	std::vector<std::vector<int> > d_num_dofs_per_proc;
+	int d_u_dof_index_idx, d_p_dof_index_idx;
+	SAMRAI::tbox::Pointer<SAMRAI::pdat::SideVariable<NDIM, int> > d_u_dof_index_var;
+	SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, int> > d_p_dof_index_var;
+
+	/*
+	 * Jacobian of the elasticity force at the finest patch level.
+	 */
+	Mat d_mat_A;
```
boyceg: d_A_mat

include/ibamr/StaggeredStokesIBLevelRelaxationFACOperator.h
```diff
@@ -0,0 +1,677 @@
...
+
+	/*
+	 * IB interpolation operator J for the finest patch level.
+	 */
+	Mat d_mat_J;
```
boyceg: d_J_mat

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
```diff
@@ -108,6 +109,18 @@ namespace
 {
 // Version of IBImplicitStaggeredHierarchyIntegrator restart file data.
 static const int IB_IMPLICIT_STAGGERED_HIERARCHY_INTEGRATOR_VERSION = 1;
+
+static void ib_4_interp_fcn(const double r, double* const w)
```
boyceg: Does this need to be reimplemented here?

include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h
amneetb: We need to get info like u_problem_coef, nullspace info etc from INSHierarchyIntegrator, but we do not need to create any StokesSolver within ins hierarchy integrator. I don't think we need to change design, because to me provision for registering external stokes solver with ins integrator seems to be precisely for these situations...

include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h
amneetb: You mean within NoOpStokesSolver that I am creating? That seems fine but I just wanted to avoid cumbersome referencing e.g, d_stokes_solver->d_fac_op etc

include/ibamr/IBImplicitStrategy.h
amneetb: Sure. I will change it...

include/ibamr/KrylovLinearSolverStaggeredStokesSolverInterface.h
amneetb: (1) I think so, because the implementations are capable of generating vectors themselves.
(2)  setComponentNullspaces() sounds like we are passing vectors, but I just to pass booleans..

include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h
amneetb: I can remove my name or change date. Both are OK to me.

include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h
boyceg: > On Apr 24, 2015, at 12:28 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29023070>:
> 
> > @@ -150,11 +156,113 @@ class IBImplicitStaggeredHierarchyIntegrator : public IBHierarchyIntegrator
> >  
> >  private:
> >      /*!
> > -     * \brief Default constructor.
> > -     *
> > -     * \note This constructor is not implemented and should not be used.
> > -     */
> > -    IBImplicitStaggeredHierarchyIntegrator();
> > +	 * \brief A StaggeredStokesSolver that does nothing.
> We need to get info like u_problem_coef, nullspace info etc from INSHierarchyIntegrator, but we do not need to create any StokesSolver within ins hierarchy integrator. I don't think we need to change design, because to me provision for registering external stokes solver with ins integrator seems to be precisely for these situations...
> 

You can also reset the Stokes operator and PC with a standard Stokes Krylov solver. Does initialization not work correctly for the IB part?

include/ibamr/KrylovLinearSolverStaggeredStokesSolverInterface.h
boyceg: > On Apr 24, 2015, at 12:33 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In include/ibamr/KrylovLinearSolverStaggeredStokesSolverInterface.h <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29023188>:
> 
> > @@ -89,6 +89,12 @@ class KrylovLinearSolverStaggeredStokesSolverInterface : public StaggeredStokesS
> >       */
> >      void setVelocityPoissonSpecifications(const SAMRAI::solv::PoissonSpecifications& U_problem_coefs);
> >  
> > +	/*!
> > +	 * \brief Set if velocity and pressure have nullspace.
> > +	 */
> > +	void setComponentsHaveNullspace(const bool has_velocity_nullspace,
> (1) I think so, because the implementations are capable of generating vectors themselves.
> (2) setComponentNullspaces() sounds like we are passing vectors, but I just to pass booleans..
> 
OK.

include/ibamr/StaggeredStokesIBLevelRelaxationFACOperator.h
amneetb: I feel so, because I repeating some of the stuff like setComponentsHaveNullspace( )

include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h
boyceg: > On Apr 24, 2015, at 12:31 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29023144>:
> 
> >      SAMRAI::tbox::Pointer<StaggeredStokesSolver> d_stokes_solver;
> >      SAMRAI::tbox::Pointer<StaggeredStokesOperator> d_stokes_op;
> > +	SAMRAI::tbox::Pointer<StaggeredStokesIBLevelRelaxationFACOperator> d_fac_op;
> You mean within NoOpStokesSolver that I am creating? That seems fine but I just wanted to avoid cumbersome referencing e.g, d_stokes_solver->d_fac_op etc
> 
Just get this stuff from the Stokes solver unless there is a possibility that the FAC operator will be different than whatever is in the Stokes solver. E.g. it would be useful to test this out with different solvers, not just FAC using StaggeredStokesIBLevelRelaxationFACOperator.

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
```diff
@@ -606,12 +894,41 @@ PetscErrorCode IBImplicitStaggeredHierarchyIntegrator::compositeIBJacobianSetup(
...
+        d_ib_implicit_ops->computeLinearizedResidual(d_X_current, X_new);
+        d_ib_implicit_ops->setLinearizedPosition(X_new);
+
+        ierr = VecDestroy(&X_new);
+        IBTK_CHKERRQ(ierr);
+    }
+    return ierr;
 } // compositeIBJacobianSetup
 
 PetscErrorCode IBImplicitStaggeredHierarchyIntegrator::compositeIBJacobianApply_SAMRAI(Mat A, Vec x, Vec f)
```
boyceg: I think that "composite" was used here to indicate that this is the Jacobian for the "composite" system (u,p,X) or (u,p,F). In the solve_for_velocity branch, this is not a composite Jacobian but an Eulerian Jacobian operator. I would split this up into different functions for the different algorithmic options. (Notice that there is almost no code shared between the two branches.)

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
```diff
@@ -694,105 +1059,114 @@ PetscErrorCode IBImplicitStaggeredHierarchyIntegrator::compositeIBPCApply(Vec x,
...
+        ierr = PetscObjectStateIncrease(reinterpret_cast<PetscObject>(component_y_vecs[0]));
+        IBTK_CHKERRQ(ierr);
+        ierr = PetscObjectStateIncrease(reinterpret_cast<PetscObject>(y));
+        IBTK_CHKERRQ(ierr);
+    }
+    else
+    {
+        Pointer<SAMRAIVectorReal<NDIM, double> > f_g = PETScSAMRAIVectorReal::getSAMRAIVector(x);
+        Pointer<SAMRAIVectorReal<NDIM, double> > u_p = PETScSAMRAIVectorReal::getSAMRAIVector(y);
+        bool converged = d_fac_pc->solveSystem(*u_p, *f_g);
```
boyceg: Please allow for more general solvers here.

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
amneetb: (1) I don't want to add the entire header file just for ib_4 function definition
(2) We might be testing different delta functions with preconditioners and this seems to be the right place for their definition...

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
amneetb: I thought about it, but then I wanted to avoid having so many different static functions in this class. There are a lot already. One possibility is to remove keyword composite from them. But I am also OK with having different functions like  {IBJacobianSetup_velocity, IBJacobianSetup_position} if that makes it cleaner?

boyceg: > On Apr 24, 2015, at 12:51 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp:
> 
> > +        Pointer<VariableContext> current_ctx = d_ins_hier_integrator->getCurrentContext();
> > +        Pointer<Variable<NDIM> > u_var = d_ins_hier_integrator->getVelocityVariable();
> > +        const int u_current_idx = var_db->mapVariableAndContextToIndex(u_var, current_ctx);
> > +        const int u_new_idx = u->getComponentDescriptorIndex(0);
> > +        d_hier_velocity_data_ops->linearSum(d_u_idx, -0.5, u_new_idx, -0.5, u_current_idx);
> > +
> > +        d_ib_implicit_ops->interpolateLinearizedVelocity(
> > +            d_u_idx, getCoarsenSchedules(d_object_name + "::u::CONSERVATIVE_COARSEN"),
> > +            getGhostfillRefineSchedules(d_object_name + "::u"), half_time);
> > +        d_ib_implicit_ops->computeLinearizedResidual(d_X_current, X_new);
> > +        d_ib_implicit_ops->setLinearizedPosition(X_new);
> > +
> > +        ierr = VecDestroy(&X_new);
> > +        IBTK_CHKERRQ(ierr);
> > +    }
> > +    return ierr;
> >  } // compositeIBJacobianSetup
> >  
> >  PetscErrorCode IBImplicitStaggeredHierarchyIntegrator::compositeIBJacobianApply_SAMRAI(Mat A, Vec x, Vec f)
> I thought about it, but then I wanted to avoid having so many different static functions in this class. There are a lot already. One possibility is to remove keyword composite from them. But I am also OK with having different functions like {IBJacobianSetup_velocity, IBJacobianSetup_position} if that makes it cleaner?
> 
Yes.

include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h
amneetb: Where exactly should I reset the Stokes Op and PC? I don't recall the class and the file...

boyceg: > On Apr 27, 2015, at 1:06 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29166867>:
> 
> > @@ -150,11 +156,113 @@ class IBImplicitStaggeredHierarchyIntegrator : public IBHierarchyIntegrator
> >  
> >  private:
> >      /*!
> > -     * \brief Default constructor.
> > -     *
> > -     * \note This constructor is not implemented and should not be used.
> > -     */
> > -    IBImplicitStaggeredHierarchyIntegrator();
> > +	 * \brief A StaggeredStokesSolver that does nothing.
> Where exactly should I reset the Stokes Op and PC? I don't recall the class and the file...
> 

Use INSStaggeredHierarchyIntegrator::setStokesSolver() to set the solver.

There is not currently an API currently for resetting the Stokes operator itself; if you look in the current implementation, it is specialized for standard Stokes problems, although I don't think that this is really a limitation for our present purposes --- you just need to ensure that the modified IB-Stokes operator is used in the solver, and that the modified problem does not require a modification of the RHS. (For what we are doing, it is enough to modify the solver.)

amneetb: Yeah, the SNES function works with inhomogeneous BC incorporated into Stokes Op, so we don't need to worry about subtracting inhomogeneous BCs from RHS.

boyceg: > On Apr 27, 2015, at 1:42 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29170545>:
> 
> > @@ -150,11 +156,113 @@ class IBImplicitStaggeredHierarchyIntegrator : public IBHierarchyIntegrator
> >  
> >  private:
> >      /*!
> > -     * \brief Default constructor.
> > -     *
> > -     * \note This constructor is not implemented and should not be used.
> > -     */
> > -    IBImplicitStaggeredHierarchyIntegrator();
> > +	 * \brief A StaggeredStokesSolver that does nothing.
> Yeah, the SNES function works with inhomogeneous BC incorporated into Stokes Op, so we don't need to worry about subtracting inhomogeneous BCs from RHS.
> 
This isn't really about boundary conditions; this is about the extra stuff in the IB-Stokes momentum operator (i.e., (rho/dt)*I - (mu/2)*L - (dt/4)*S*A*J). I think that the code is set up so that you can pass in a Newton solver as the Stokes solver. Such a solver can handle inhomogeneous boundary conditions.

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +596,731 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+        }
+#endif
+    }
+
+#if !defined(NDEBUG)
+    TBOX_ASSERT(counter == n_subdomains);
+#endif
+    return;
+} // constructPatchLevelASMSubdomains_cell
+
```
amneetb: Moving the request over here. For physical boundaries checking the negativity of dof_idx should suffice?

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +596,731 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+    for (PatchLevel<NDIM>::Iterator p(patch_level); p; p++)
+    {
+        Pointer<Patch<NDIM> > patch = patch_level->getPatch(p());
+        const Box<NDIM>& patch_box = patch->getBox();
+        const IntVector<NDIM>& patch_lower = patch_box.lower();
+        const IntVector<NDIM>& patch_upper = patch_box.upper();
+        Pointer<CellData<NDIM, int> > dof_data = patch->getPatchData(dof_index_idx);
+        const int data_depth = dof_data->getDepth();
+
+        // Compute number of nonoverlapping subdomains
```
boyceg: Little helper function to avoid dup'ed code? (We may wish to change the way that subdomains are set up at some point in the future.)

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +596,731 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+        IntVector<NDIM> subdomains = cells / box_size;
+        for (int d = 0; d < NDIM; ++d)
+        {
+            if (subdomains(d) == 0)
+            {
+                subdomains(d) = 1;
+            }
+        }
+
+#if (NDIM == 2)
```
boyceg: See if you can make this NDIM independent.

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +596,731 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+        const int Nx = subdomains(0);
+        const int Ny = subdomains(1);
+
+        int j_lower = patch_lower(1);
+        for (int J = 1; J <= Ny; ++J)
```
boyceg: Not a big deal, but I would use C/C++ indexing conventions here (which also make it easier to convert from (I,J,K) to (i,j,k) indices).

ibtk/src/math/PETScMatUtilities.cpp
boyceg: DOF indices are set to -1 in the physical boundary and c-f interface ghost cell regions.

ibtk/src/math/PETScMatUtilities.cpp
amneetb: I kept (I,J,K) for coarse box and (i,j,k) for fine box indexing. Does that not come up conspicuously?

boyceg: > On Apr 30, 2015, at 1:08 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/math/PETScMatUtilities.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29448748>:
> 
> > +        cells += patch_upper - patch_lower;
> > +        IntVector<NDIM> subdomains = cells / box_size;
> > +        for (int d = 0; d < NDIM; ++d)
> > +        {
> > +            if (subdomains(d) == 0)
> > +            {
> > +                subdomains(d) = 1;
> > +            }
> > +        }
> > +
> > +#if (NDIM == 2)
> > +        const int Nx = subdomains(0);
> > +        const int Ny = subdomains(1);
> > +
> > +        int j_lower = patch_lower(1);
> > +        for (int J = 1; J <= Ny; ++J)
> I kept (I,J,K) for coarse box and (i,j,k) for fine box indexing. Does that not come up conspicuously?
> 
I meant: count starting from 0 instead of 1.

ibtk/src/math/PETScMatUtilities.cpp
amneetb: yeah. So, if they are, then the above code should do the right thing for c-f and physical boundaries?

boyceg: Should. (My father always used to call should "the s-word".)

ibtk/src/math/PETScMatUtilities.cpp
amneetb: That can be done easily :-)

ibtk/src/math/PETScMatUtilities.cpp
amneetb: Not that I can think of easily :(. My guess is one would use SAMRAI constructs to first partition the patch and then iterate over the subdomains.

ibtk/src/math/PETScMatUtilities.cpp
amneetb: Yes. I will first write code for sc-quantity and see what I have duplicated in both and move the common part to helper function...

ibtk/src/math/PETScMatUtilities.cpp
boyceg: One thing that you could do is to make a helper function that will actually do the basic partitioning for you --- e.g. pass in a patch Box, a subdomain size, and a overlap size, and it will return the interior and overlap boxes. I am 99% sure that this can be done in a NDIM-independent manner.

ibtk/src/math/PETScMatUtilities.cpp
boyceg: A helper function to compute explicitly the subdomain boxes would go a long way towards making the rest of the routine NDIM-independent. I would think about this before doing the side-centered case.

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +731,599 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+            ++is_counter;
+        }
+    }
+
+#if !defined(NDEBUG)
+    TBOX_ASSERT(is_counter == n_subdomains);
+#endif
+
+    return;
+} // constructPatchLevelASMSubdomains_cell
```
amneetb: Please check now...

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -80,6 +84,141 @@ class RobinBcCoefStrategy;
...
+
+#endif
+
+#if !defined(NDEBUG)
+    TBOX_ASSERT(counter == n_subdomains);
+#endif
+
+    return;
+} // partitionPatchBox
+}
```
amneetb: This is the helper function. Made changes to indexing as well...

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +623,827 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+        }
+    }
+
+#if !defined(NDEBUG)
+    TBOX_ASSERT(is_counter == n_subdomains);
+#endif
+
+    return;
+} // constructPatchLevelASMSubdomains_side
+
```
amneetb: Ok, it's in the repo now. Take a look when you get a chance. 

(1) Is there some extra work that needs to be done for multiplicative Schwarz, or these would be used as it is?
(2) For testing, I can use CCPoissonPETScLevelSolver class with PCASM as preconditioner. How is the convergence rate expected to behave?

boyceg: (1) Regarding MSM, I don't think there will be any extra work, although perhaps that code will be more permissive about whether it is necessary to have nonoverlapping interior partitionings within local patches.

(2) Here is a test: If you make the subdomain size = patch size and an overlap of one cell, then I think that you should be able to replicate the box smoother FAC operator; performance should be nearly identical if you use the same subdomain solvers in both FAC implementations. If you make the subdomain size = 1 grid cell, then you should get something that is equivalent to using Jacobi as the smoother (which means that you probably should use damping).

amneetb: I see. I believe I can directly use -pc_type jacobi from petsc? KSP for subdomain = 1 grid cell should be of richardson type?

boyceg: Choice of KSP is independent of the PC type. If you want to mimic the basic CC solver, I think that you should use Richardson for the KSP and SOR for the preconditioner.

ibtk/src/solvers/impls/PETScLevelSolver.cpp
```diff
@@ -307,6 +314,19 @@ void PETScLevelSolver::deallocateSolverState()
...
+
+	d_petsc_extern_mat = op;
+	d_extern_mat_nz_pattern = nonzero_pattern;
+
+	return;
+}// addLinearOperator
+
 /////////////////////////////// PROTECTED ////////////////////////////////////
 
 void PETScLevelSolver::init(Pointer<Database> input_db, const std::string& default_options_prefix)
```
amneetb: There is no pc_type option here and neither is one used in PETScLevelSolver::initializeSolverState. That should be changed?

boyceg: > On May 2, 2015, at 1:45 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/solvers/impls/PETScLevelSolver.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29549190>:
> 
> > @@ -307,6 +314,19 @@ void PETScLevelSolver::deallocateSolverState()
> >      return;
> >  } // deallocateSolverState
> >  
> > +void PETScLevelSolver::addLinearOperator(Mat& op, MatStructure nonzero_pattern)
> > +{
> > +#if !defined(NDEBUG)
> > +	TBOX_ASSERT(!d_is_initialized);
> > +	TBOX_ASSERT(op);
> > +#endif
> > +
> > +	d_petsc_extern_mat = op;
> > +	d_extern_mat_nz_pattern = nonzero_pattern;
> > +
> > +	return;
> > +}// addLinearOperator
> > +
> >  /////////////////////////////// PROTECTED ////////////////////////////////////
> >  
> >  void PETScLevelSolver::init(Pointer<Database> input_db, const std::string& default_options_prefix)
> There is no pc_type option here and neither is one used in PETScLevelSolver::initializeSolverState. That should be changed?
> 
If you want to.

Also, please remove addLinearOperator() from the public interface (and change the name to setLinearOperator()).

amneetb: Hmmm.. that routine is actually adding another linear operator to the one the derived class is building...'add' sounds more apt, doesn't it?

boyceg: > On May 2, 2015, at 2:24 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/solvers/impls/PETScLevelSolver.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29549506>:
> 
> > @@ -307,6 +314,19 @@ void PETScLevelSolver::deallocateSolverState()
> >      return;
> >  } // deallocateSolverState
> >  
> > +void PETScLevelSolver::addLinearOperator(Mat& op, MatStructure nonzero_pattern)
> > +{
> > +#if !defined(NDEBUG)
> > +	TBOX_ASSERT(!d_is_initialized);
> > +	TBOX_ASSERT(op);
> > +#endif
> > +
> > +	d_petsc_extern_mat = op;
> > +	d_extern_mat_nz_pattern = nonzero_pattern;
> > +
> > +	return;
> > +}// addLinearOperator
> > +
> >  /////////////////////////////// PROTECTED ////////////////////////////////////
> >  
> >  void PETScLevelSolver::init(Pointer<Database> input_db, const std::string& default_options_prefix)
> Hmmm.. that routine is actually adding another linear operator to the one the derived class is building...'add' sounds more apt, doesn't it?
> 
In any case, I don't know if this belongs in the public interface for this abstract base class. This is kind of a weird operation.

amneetb: How about supporting setAdditive{Multiplicative}LinearOperator in the class? The intention of these 'extraneous' Ops will be do some extra work on top of the main linear operator employed in the KSP, e.g., adding diagonal matrix to improve the conditioning of the existing matrix. This will avoid having to create many different classes that differ from the classical KSP matrix in some respect...

boyceg: > On May 2, 2015, at 2:56 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/solvers/impls/PETScLevelSolver.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29549789>:
> 
> > @@ -307,6 +314,19 @@ void PETScLevelSolver::deallocateSolverState()
> >      return;
> >  } // deallocateSolverState
> >  
> > +void PETScLevelSolver::addLinearOperator(Mat& op, MatStructure nonzero_pattern)
> > +{
> > +#if !defined(NDEBUG)
> > +	TBOX_ASSERT(!d_is_initialized);
> > +	TBOX_ASSERT(op);
> > +#endif
> > +
> > +	d_petsc_extern_mat = op;
> > +	d_extern_mat_nz_pattern = nonzero_pattern;
> > +
> > +	return;
> > +}// addLinearOperator
> > +
> >  /////////////////////////////// PROTECTED ////////////////////////////////////
> >  
> >  void PETScLevelSolver::init(Pointer<Database> input_db, const std::string& default_options_prefix)
> How about supporting setAdditive{Multiplicative}LinearOperator in the class? The intention of these 'extraneous' Ops will be do some extra work on top of the main linear operator employed in the KSP, e.g., adding diagonal matrix to improve the conditioning of the existing matrix. This will avoid having to create many different classes that differ from the classical KSP matrix in some respect...
> 

I'm not convinced that this is something that belongs in this interface.

If you need to modify the operator following construction, you can always grab the KSP and then modify the operator yourself. If you want to modify the operator that is constructed by a solver, it seems like the right place to do this is in a subclass.

knepley: On Sat, May 2, 2015 at 2:29 PM, Boyce Griffith <notifications@github.com>
wrote:

> In ibtk/src/solvers/impls/PETScLevelSolver.cpp
> <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29550034>:
>
> > @@ -307,6 +314,19 @@ void PETScLevelSolver::deallocateSolverState()
> >      return;
> >  } // deallocateSolverState
> >
> > +void PETScLevelSolver::addLinearOperator(Mat& op, MatStructure nonzero_pattern)
> > +{
> > +#if !defined(NDEBUG)
> > +	TBOX_ASSERT(!d_is_initialized);
> > +	TBOX_ASSERT(op);
> > +#endif
> > +
> > +	d_petsc_extern_mat = op;
> > +	d_extern_mat_nz_pattern = nonzero_pattern;
> > +
> > +	return;
> > +}// addLinearOperator
> > +
> >  /////////////////////////////// PROTECTED ////////////////////////////////////
> >
> >  void PETScLevelSolver::init(Pointer<Database> input_db, const std::string& default_options_prefix)
>
>  On May 2, 2015, at 2:56 PM, Amneet Bhalla <notifications@github.com>
> wrote: In ibtk/src/solvers/impls/PETScLevelSolver.cpp <
> https://github.com/IBAMR/IBAMR/pull/6#discussion_r29549789>: > @@ -307,6
> +314,19 @@ void PETScLevelSolver::deallocateSolverState() > return; > } //
> deallocateSolverState > > +void PETScLevelSolver::addLinearOperator(Mat&
> op, MatStructure nonzero_pattern) > +{ > +#if !defined(NDEBUG) > +
> TBOX_ASSERT(!d_is_initialized); > + TBOX_ASSERT(op); > +#endif > + > +
> d_petsc_extern_mat = op; > + d_extern_mat_nz_pattern = nonzero_pattern; > +
> > + return; > +}// addLinearOperator > + > ///////////////////////////////
> PROTECTED //////////////////////////////////// > > void
> PETScLevelSolver::init(Pointer<Database> input_db, const std::string&
> default_options_prefix) How about supporting
> setAdditive{Multiplicative}LinearOperator in the class? The intention of
> these 'extraneous' Ops will be do some extra work on top of the main linear
> operator employed in the KSP, e.g., adding diagonal matrix to improve the
> conditioning of the existing matrix. This will avoid having to create many
> different classes that differ from the classical KSP matrix in some
> respect...
> I'm not convinced that this is something that belongs in this interface.
> If you need to modify the operator following construction, you can always
> grab the KSP and then modify the operator yourself. If you want to modify
> the operator that is constructed by a solver, it seems like the right place
> to do this is in a subclass.
>

Note that PETSc already has a mechanism for this:


http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Mat/MATSHELL.html

if you do not want to modify your existing operator. However, the
operations you
describe above could also probably be encapsulated in a preconditioner. I
think I
prefer that since it gives you the flexibility to easily return to the
original problem.

   Matt


> —
> <https://github.com/IBAMR/IBAMR/pull/6/files#r29550034>.
>



-- 
What most experimenters take for granted before they begin their
experiments is infinitely more interesting than any results to which their
experiments lead.
-- Norbert Wiener

amneetb: We (maybe it's just me) are thinking (guessing) that for PCASM to work properly, there is an explicit need to add two matrices, Mat A and Mat B, which have different nonzero patterns, via MATAXPY. If that is not a requirement, we can use MATSHELL. 

P.S: In the application code, A is the Stokes operator and B is the elasticity operator that takes fluid velocity and generates elastic force. B's nonzero pattern is a function of Lagrangian mesh etc...

boyceg: > On May 2, 2015, at 2:56 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/solvers/impls/PETScLevelSolver.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29549789>:
> 
> > @@ -307,6 +314,19 @@ void PETScLevelSolver::deallocateSolverState()
> >      return;
> >  } // deallocateSolverState
> >  
> > +void PETScLevelSolver::addLinearOperator(Mat& op, MatStructure nonzero_pattern)
> > +{
> > +#if !defined(NDEBUG)
> > +	TBOX_ASSERT(!d_is_initialized);
> > +	TBOX_ASSERT(op);
> > +#endif
> > +
> > +	d_petsc_extern_mat = op;
> > +	d_extern_mat_nz_pattern = nonzero_pattern;
> > +
> > +	return;
> > +}// addLinearOperator
> > +
> >  /////////////////////////////// PROTECTED ////////////////////////////////////
> >  
> >  void PETScLevelSolver::init(Pointer<Database> input_db, const std::string& default_options_prefix)
> How about supporting setAdditive{Multiplicative}LinearOperator in the class? The intention of these 'extraneous' Ops will be do some extra work on top of the main linear operator employed in the KSP, e.g., adding diagonal matrix to improve the conditioning of the existing matrix. This will avoid having to create many different classes that differ from the classical KSP matrix in some respect...
> 

I prefer not to expose Mat in the public interface for this class. Why can't these modifications be done in the subclass as needed?

amneetb: If you are ok with it, I can add an additional class. When we were initially discussing it, you were suggesting (or least that's what I thought) to pass SAJ to Stokes class, which would add SAJ to it. So, I took that approach.

boyceg: > On May 2, 2015, at 10:45 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/solvers/impls/PETScLevelSolver.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29552476>:
> 
> > @@ -307,6 +314,19 @@ void PETScLevelSolver::deallocateSolverState()
> >      return;
> >  } // deallocateSolverState
> >  
> > +void PETScLevelSolver::addLinearOperator(Mat& op, MatStructure nonzero_pattern)
> > +{
> > +#if !defined(NDEBUG)
> > +	TBOX_ASSERT(!d_is_initialized);
> > +	TBOX_ASSERT(op);
> > +#endif
> > +
> > +	d_petsc_extern_mat = op;
> > +	d_extern_mat_nz_pattern = nonzero_pattern;
> > +
> > +	return;
> > +}// addLinearOperator
> > +
> >  /////////////////////////////// PROTECTED ////////////////////////////////////
> >  
> >  void PETScLevelSolver::init(Pointer<Database> input_db, const std::string& default_options_prefix)
> If you are ok with it, I can add an additional class. When we were initially discussing it, you were suggesting (or least that's what I thought) to pass SAJ to Stokes class, which would add SAJ to it. So, I took that approach.
> 

Yes, that sounds like what I am suggesting: the subclass sets up the system. :-)

boyceg: > On May 2, 2015, at 10:37 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> In ibtk/src/solvers/impls/PETScLevelSolver.cpp <https://github.com/IBAMR/IBAMR/pull/6#discussion_r29552452>:
> 
> > @@ -307,6 +314,19 @@ void PETScLevelSolver::deallocateSolverState()
> >      return;
> >  } // deallocateSolverState
> >  
> > +void PETScLevelSolver::addLinearOperator(Mat& op, MatStructure nonzero_pattern)
> > +{
> > +#if !defined(NDEBUG)
> > +	TBOX_ASSERT(!d_is_initialized);
> > +	TBOX_ASSERT(op);
> > +#endif
> > +
> > +	d_petsc_extern_mat = op;
> > +	d_extern_mat_nz_pattern = nonzero_pattern;
> > +
> > +	return;
> > +}// addLinearOperator
> > +
> >  /////////////////////////////// PROTECTED ////////////////////////////////////
> >  
> >  void PETScLevelSolver::init(Pointer<Database> input_db, const std::string& default_options_prefix)
> We (maybe it's just me) are thinking (guessing) that for PCASM to work properly, there is an explicit need to add two matrices, Mat A and Mat B, which have different nonzero patterns, via MATAXPY. If that is not a requirement, we can use MATSHELL.
> 
MATSHELL doesn't work here; it needs to be the sum of the matrices as a Mat. (The issue is not the operations that are being performed, but where in the interface the operations belong. I think that the right place is in the private initialization method.)

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -806,10 +623,878 @@ void PETScMatUtilities::constructPatchLevelSCInterpOp(Mat& mat,
...
+        ISRestoreIndices(is_nonoverlap[0][i], &indices);
+        counter += length;
+    }
+
+    ISCreateGeneral(PETSC_COMM_SELF, total_length, &total_indices[0], PETSC_COPY_VALUES, &is_total);
+    ISSort(is_total);
+    ISView(is_total, PETSC_VIEWER_STDOUT_WORLD);
+    ISDestroy(&is_total);
+
```
amneetb: A combined nonoverlapping IS is getting printed here...

ibtk/src/solvers/impls/PETScLevelSolver.cpp
```diff
@@ -256,6 +336,27 @@ void PETScLevelSolver::initializeSolverState(const SAMRAIVectorReal<NDIM, double
...
+        }
+    }
+
+    // Reset class data structures
+    ierr = KSPGetTolerances(d_petsc_ksp, &d_rel_residual_tol, &d_abs_residual_tol, NULL, &d_max_iterations);
+    PCType pc_type;
+    ierr = PCGetType(ksp_pc, &pc_type);
+    d_pc_type = pc_type;
+
+    // Set the nullspace,
```
amneetb: PCASM getting setup here...

amneetb:
@boyceg Try tests/Stokes-IB/test2 with the command line given in README. The timestep size is 10 times bigger than examples/IB/explicit/ex2.

boyceg:
> @boyceg <https://github.com/boyceg> Try tests/Stokes-IB/test2 with the command line given in README. The timestep size is 10 times bigger than examples/IB/explicit/ex2.

I thought that AMR wasn't working for Stokes+IB?

Two comments:

(1) In the input file, the stiffness and damping parameters are determined from the time step size, so to keep the structure fixed in place, it is necessary to increase the spring stiffness.

(2) Turning off damping seems to yield better performance --- it could be that we are not handling that part of the equation properly in the implicit solver.

amneetb:
We will have to test it out with number of cases. It seems to be doing good job (or maybe not?) with this example/setting.

PR #8, opened 2015-04-27
Support for SAMRAI 3.X

boyceg:


boyceg:
@amneetb Can you please start reviewing for cast correctness?

Note that CPP_CAST == dynamic_cast in debug mode and static_cast in opt mode, so that it *must* be used *only* for cases that do not require dynamic casting; likewise BOOST_CAST will do a static cast in opt mode, and thus *must* be used *only* for cases that do not require dynamic casting. (When do we *need* to use dynamic casting? If we use a cast to determine the type of a variable, then that needs to be a dynamic cast. If we already know what the type has to be, then we probably should use CPP_CAST or BOOST_CAST, as appropriate.)

I have tried to systematically change casts to use CPP_CAST or BOOST_CAST where appropriate. It is probably worthwhile to review all nontrivial casts in the code.

boyceg:
@amneetb Note that the library finally builds, although for now I have *disabled* ConstraintIBMethod and related examples. I have not yet tested any examples; I am sure massive breakages await.

PR #10, opened 2015-04-27
Update PETSc support to version 3.5.X

boyceg:


boyceg:
@knepley This is an update of your branch to update to PETSc 3.5. There were some merge problems with your branch and the current master, and so I decided to wipe it out and start anew from master.

@amneetb Notice that KSPSetOperators() no longer wants to be told SAME_PRECONDITIONER. This functionality is achieved by calling KSPSetReusePreconditioner() on the KSP.

PR #13, opened 2015-04-27
Add proper restart support for IBFE models

boyceg:


boyceg:
@walterjunjun reported some discrepancy in the Cartesian grid data upon restart; these need to be investigated before merging.

boyceg:
Current IBFE restart code seems to yield correct restart files for IBFE/explicit/ex1.

PR #29, opened 2015-08-06
fixes to force splitting logic

boyceg:
add separate toggles for normal and tangential surface force splitting; cleaning up logic to decide when to integrate surface forces.


boyceg:
@amneetb See if this addresses your concerns.

There are now separate flags to control splitting of normal and tangential forces, and I attempt to ensure that forces are included where they are supposed to be. The control logic is a bit simpler --- and hopefully correct!

amneetb:
https://github.com/IBAMR/IBAMR/blob/6cfe3ef0844b078540f8918ea6b2bffa5b3157a7/src/IB/IBFEMethod.cpp#L731 

It operates when d_use_jump_conditions and d_split_normal_force are true

Now at 
https://github.com/IBAMR/IBAMR/blob/6cfe3ef0844b078540f8918ea6b2bffa5b3157a7/src/IB/IBFEMethod.cpp#L1681
if (d_constrained_part[part] || (!d_split_normal_force && !d_split_tangential_force))
Should there be any check on d_split_tangential_force here?

BTW: For Lee and LeVeque method we want : d_use_jump_conditions = true, d_split_normal_force = true,
d_split_tangential_force = false. Right?

boyceg:
> On Aug 6, 2015, at 1:47 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> https://github.com/IBAMR/IBAMR/blob/6cfe3ef0844b078540f8918ea6b2bffa5b3157a7/src/IB/IBFEMethod.cpp#L731 <https://github.com/IBAMR/IBAMR/blob/6cfe3ef0844b078540f8918ea6b2bffa5b3157a7/src/IB/IBFEMethod.cpp#L731>
> It operates when d_use_jump_conditions and d_split_normal_force are true
> 
Yes: At least currently, the only jump conditions are related to the normal force.
> Now at 
> https://github.com/IBAMR/IBAMR/blob/6cfe3ef0844b078540f8918ea6b2bffa5b3157a7/src/IB/IBFEMethod.cpp#L1681 <https://github.com/IBAMR/IBAMR/blob/6cfe3ef0844b078540f8918ea6b2bffa5b3157a7/src/IB/IBFEMethod.cpp#L1681>
> if (d_constrained_part[part] || (!d_split_normal_force && !d_split_tangential_force))
> Should there be any check on d_split_tangential_force here?
> 
It is not necessary to perform a check on the tangential split, but I also don't see how it does any harm.
> BTW: For Lee and LeVeque method we want : d_use_jump_conditions = true, d_split_normal_force = true,
> d_split_tangential_force = false. Right?
> 
It is only an analogy to that method, and I do not know which option for the tangential forces will be more accurate (of if it even matters).

PR #31, opened 2015-09-28
Moving least squares approach to do one-sided interpolation and spreading

amneetb:
@boyceg : Please take a look at the first shot of 2d MLS LEInteractor::interpolation routine for IB_4 kernel.

### DIFF
--- a/examples/IBFE/explicit/ex0/input2d
+++ b/examples/IBFE/explicit/ex0/input2d
@@ -14,6 +14,7 @@ MFAC = 4.0                                     // ratio of Lagrangian mesh width
 ELEM_TYPE = "QUAD9"                            // type of element to use for structure discretization
 CONVERGENCE_STUDY = FALSE                      // indicate whether we are performing a convergence study or not;
                                                // if so, attempt to make "nested" structural meshes
+ONE_SIDED_INTERACTION = TRUE
 
 // problem parameters
 SMOOTH_CASE = FALSE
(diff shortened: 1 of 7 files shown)


boyceg:
Looks good.

A few comments:

(1) You probably don't need separate functions for setting up the mask for interpolation and spreading --- note that those spec objects have a common base class.

(2) For the mask data, do you think it might be better simply to have an extra SAMRAI variable / patch data that has a boolean mask of the points that are "inside" the region of interest, which you could sample as needed in performing interpolation / spreading?

(3) Please make sure that in the end, you have a common implementation for computing the weights for interpolation and spreading.

For tests: obviously you want to verify the reproducing property. You should also verify that with no masking, you recover the original kernel function. (At least, I think that is what happens.)

amneetb:
Ok, I made changes accordingly. Also added sc-interp routine in both 2d and 3d. There is code in research directory amneetb/InternalFlows/OneSideInterpolation/main.cpp that tests the routine in 2d and 3d for a sc quantity. Please take a look. A masking region is set via muParser in the input*d files. I printed weights (w and psi) with unmasked domain and (a part of the region) masked domain. 
With the unmasked domain, psi == w and with masked region psi gets modified. I tried interpolating from a linear field and the interpolation is exact using the modified psi weights.

boyceg:
> On Oct 1, 2015, at 1:56 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> Ok, I made changes accordingly. Also added sc-interp routine in both 2d and 3d. There is code in research directory amneetb/InternalFlows/OneSideInterpolation/main.cpp that tests the routine in 2d and 3d for a sc quantity. Please take a look. A masking region is set via muParser in the input*d files. I printed weights (w and psi) with unmasked domain and (a part of the region) masked domain. 
> With the unmasked domain, psi == w and with masked region psi gets modified. I tried interpolating from a linear field and the interpolation is exact using the modified psi weights.
> 
Cool!

PR #34, opened 2016-01-14
Simplify Cartesian grid boundary condition handling

boyceg:
This is an attempt to simplify the boundary condition handling for cases in which some DOFs live on the domain boundary (e.g. node, side/face, or edge data).


boyceg:
@amneetb

boyceg:
I think this is probably the wrong approach. Let's try to do this using the existing machinery.

PR #35, opened 2016-01-14
Simplifying BC handling for operators involving DOFs on the domain boundary

boyceg:
This PR aims to make it easier to handle both homogeneous and inhomogeneous Dirichlet boundary conditions that live on the domain boundary. Operators are now able to modify both solution and RHS values to account for boundary conditions. Previously, some of these values had to be set outside of the solver, which is error prone for callers.


boyceg:
Tested to work with navier_stokes/ex0 with Dirichlet BCs.

@amneetb, please review. I hope this will help to make it easier to get correct boundary handling for the RigidIBFE branch.

PR #38, opened 2016-01-19
Adding some more functionality and fixing cell index of IB points bas…

amneetb:
…ed on lower computational domain

### DIFF
--- a/ibtk/include/ibtk/IndexUtilities.h
+++ b/ibtk/include/ibtk/IndexUtilities.h
@@ -36,10 +36,22 @@
 /////////////////////////////// INCLUDES /////////////////////////////////////
 
 #include <functional>
+#include <vector>
 
+#include "Box.h"
 #include "CellIndex.h"
 #include "Index.h"
 #include "IntVector.h"
+#include "tbox/Pointer.h"
+
+namespace SAMRAI
+{
+namespace geom
+{
+template <int DIM>
+class CartesianGridGeometry;
+}
+}
 
 /////////////////////////////// CLASS DEFINITION /////////////////////////////
 
@@ -51,11 +63,9 @@ struct CellIndexFortranOrder : std::binary_function<SAMRAI::pdat::CellIndex<NDIM
     {
         return (lhs(0) < rhs(0)
 #if (NDIM > 1)
-                ||
-                (lhs(0) == rhs(0) && lhs(1) < rhs(1))
+                || (lhs(0) == rhs(0) && lhs(1) < rhs(1))
 #if (NDIM > 2)
-                ||
-                (lhs(0) == rhs(0) && lhs(1) == rhs(1) && lhs(2) < rhs(2))
+                || (lhs(0) == rhs(0) && lhs(1) == rhs(1) && lhs(2) < rhs(2))
 #endif
 #endif
                     );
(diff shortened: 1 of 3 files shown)


boyceg:
Looks good; however, it is potentially quite confusing that getCellIndex() has two different behaviors depending on the class initialization status.

How about making two versions of this routine:

* getCellIndexGlobal() gives a "global" cell index, requires static initialization
* getCellIndexLocal() gives a "local" cell index (e.g. with respect to a given local coordinate system)

amneetb:
That sounds clean expect for a lot of changes in the library. How about keeping getCellIndex() for local
and adding getCellIndexGlobal() for the new functionality?

amneetb:
On second thought --- this will lead to two types of routines in the library, which I think can be dangerous in parallel environment. I think the way it is now is fine. One can change the behavior in the entire library from main.cpp.

boyceg:
Also, in getCellIndexGlobal, should there be a debugging check that dx is related to dx0 by an integer refinement ratio?

I don't think we can ask for the refinement ratio directly, however, as that information is not available to all callers for this routine.

amneetb:
Can that be done with floating points --- checking dx/dx0 is an integer?

amneetb:
Something like 

IntVector<NDIM> ratio ...;
TBOX_ASSERT ( MathUtilities<double>::equal_eps(dx/dx0[d],ratio[d]) ?

boyceg:
> On Jan 19, 2016, at 3:32 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> Something like
> 
> IntVector ratio ...;
> TBOX_ASSERT ( MathUtilities::equal_eps(dx/dx0[d],ratio[d]) ?
> 
I would write this as MathUtilities::equal_eps(dx[d],ratio[d]*dx0[d]).

boyceg:
It is confusing for users (and developers!) to have two different implementations of getCellIndex() with exactly the same name, but with the different functionality triggered by whether a static variable has been set or not.

Are there any cases where it would not work to make IndexUtilities a lightweight class that takes a GridGeometry object in its constructor. The current getCellIndexLocal() could remain static, but getCellIndexGlobal() would be a non-static member function.

boyceg:
Static initialization could also be fine, but:

* getCellIndexGlobal() should emit an error message if the GridGeometry object is not set
* the GridGeometry object must be set internally in the library
* the _same_ GridGeometry object must be used by all callers to IndexUtilities::init()

That last condition makes it seem like the better design would be to just make a lightweight, stateful version of IndexUtilities.

amneetb:
Please review now. If you find them OK and merge it to master, I will merge master to boyceg/MG-smoothers

boyceg:
> On Jan 19, 2016, at 4:39 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> Please review now. If you find them OK and merge it to master, I will merge master to boyceg/MG-smoother
> 
Looks fine, except that IndexUtilities::init() needs to ensure that if it is reinitialized, it does not change the mappings. E.g. if the class is initialized, init() needs to emit an error if dx0, x_lower, x_upper, ilower, or iupper change. (I think it would be better to handle this without relying on static initialization and static member variables --- I am sure that at some point a need will arise to use this function for different GridGeometries.)

boyceg:
IndexUtilities::init() *needs* to be callable multiple times. Instead of emitting an error, init() should check to see if it is being reinitialized in a manner that would change the index mappings, and only emit an error in that case.

boyceg:
@amneetb if you want to make IndexUtilities::init() static, it has to do something bad if you try to swap out the CartesianGridGeometry object. Alternatively, add a static ::is_initialized() method and emit an error if you try to reinitialize it. In any case, ::init() must not silently ignore its argument if the GridGeometry object changes.

amneetb:
How does this init() look? 
http://paste.ofcode.org/ZWu4R48FCabz4GgbxFgbzJ

boyceg:
Replaced by #39.

PR #42, opened 2016-01-22
Fix code formatting drift

boyceg:
* update .clang-format to work with clang-format 3.7
* rerun clang-format 3.7 on all source code
* add scripts to make it easy to setup pre-commit hooks to enforce clang-formatting rules

We all need to stop applying clang-format automatically on entire files (e.g. in Xcode), because doing so is creating terrible diffs.


boyceg:
@amneetb are you happy with these .clang-format settings?

boyceg:
I had high hopes for assignment operator alignment, but some of the type names are so long that there was just too much white space for my taste.

boyceg:
If you have the `git-clang-format` script (which seems to be bundled with clang-format on homebrew, but otherwise is available [here](https://github.com/askay/coding_standards_setup)), it is easy to setup a pre-commit hook that will check to see if the code to be committed adheres to clang-formatting rules.

If it doesn't, you can magically fix the formatting by running `git clang-format`. If you want to see what lines do not satisfy the `.clang-format` style file, you can do so by running `git clang-format --diff`.

There is a (dumb) script, `scripts/setup_clang_format_precommit_hook.sh`, that will set up this pre-commit hook for you.

Issue #24, opened 2015-07-29
Implicit IB doesn't honor error_on_dt_change flag


amneetb:

boyceg:
Fixed by PR #43.

PR #43, opened 2016-01-23
refactoring IB time step size change monitoring

boyceg:
* add IBHierarchyIntegrator::preprocessIntegrateHierarchy()
* move time step size change monitoring into IBHierarchyIntegrator from IBExplicitHierarchyIntegrator


boyceg:
Closes #24.

PR #44, opened 2016-01-23
Fixing Stokes-IB/test0

amneetb:
* Accounting for pre-release version of petsc - 3.6
* Accounting for new BC handling for Stokes operator.
* Adding VecAssemblyBegin/End() after VecSetValue()


boyceg:
:+1:

Issue #14, opened 2015-06-01
dependencies branch reduce-memory-use


fbusabiaga:
The branch boyceg/reduce-memory-use needd the version 3.3 or 3.4 of PETSc and the version 0.9.3
of libMesh while the master branch is already using newer versions. If it is possible, it may be convenient to have the same dependencies in all the branches to avoid the worry of having different versions of the same library.

boyceg:
> On Jun 1, 2015, at 4:24 PM, fbusabiaga <notifications@github.com> wrote:
> 
> The branch boyceg/reduce-memory-use needd the version 3.3 or 3.4 of PETSc and the version 0.9.3
> of libMesh while the master branch is already using newer versions. If it is possible, it may be convenient to have the same dependencies in all the branches to avoid the worry of having different versions of the same library.
> 

I agree in principle, but I cannot guarantee that all development branches will be kept up-to-date with master at all time.

If you are using a branch that needs to be updated to incorporate changes from master, you can do this in a local copy via, e.g.

   % git pull
   % git checkout boyceg/reduce-memory-use
   % git merge master

If the merge is clean, you (as a developer :-) can also push the updated topic branch back to GitHub, e.g.

   % git pull
   % git checkout boyceg/reduce-memory-use
   % git merge master
   % git push

amneetb:
SHA b0da59939e8d1cd5c2220b307a0d3a5a94f59725 gives the files changed in master for updating it to use  libMesh 0.9.4
https://github.com/IBAMR/IBAMR/commit/b0da59939e8d1cd5c2220b307a0d3a5a94f59725

SHA cce95f2259d2775e0fa961ef9f607629a62df733 gives files changed  gives the files changed in master for updating it to use  PETSc 3.5
https://github.com/IBAMR/IBAMR/commit/cce95f2259d2775e0fa961ef9f607629a62df733

Another easier way (or maybe not) is to configure this branch with PETSc 3.5 and libMesh 0.9.4 and let compiler tell you what lines in IBAMR branch needs an update. Most likely you can figure it out yourself, but in case you want to confirm, you can see the corresponding changes in the two SHA's. Global grep and replace can also be useful for a large and identical changes...

fbusabiaga:
Nothing urgent but I tried to merge and I found a few differences in IBAMR/ibtk/src/math/HierarchyMathOps.cpp which I cannot follow. For example in the constructor two lines change

[...]

PR #50, opened 2016-01-25
moving data allocations/deallocations closer to points of use

boyceg:
* moves SAMRAI vector allocations/deallocations closer to points of use
* moves patch data allocations closer to points of use
* removes some unneeded calls to PatchLevel::checkAllocated()
* avoids allocating some patch data in HierachyMathOps


boyceg:
This addresses Issue #14.

PR #57, opened 2016-02-09
Initial periodic shifts

amneetb:
This PR adds the feature to get initial periodic shifts and displacements for LNodes. This can be useful for certain cases of periodic boundary conditions where the initial unshifted position X0 is required.

### DIFF
--- a/ibtk/include/ibtk/LNode.h
+++ b/ibtk/include/ibtk/LNode.h
@@ -80,8 +80,10 @@ public:
     LNode(int lagrangian_nidx = -1,
           int global_petsc_nidx = -1,
           int local_petsc_nidx = -1,
-          const SAMRAI::hier::IntVector<NDIM>& periodic_offset = SAMRAI::hier::IntVector<NDIM>(0),
-          const Vector& periodic_displacement = Vector::Zero(),
+          const SAMRAI::hier::IntVector<NDIM>& initial_periodic_offset = SAMRAI::hier::IntVector<NDIM>(0),
+          const SAMRAI::hier::IntVector<NDIM>& current_periodic_offset = SAMRAI::hier::IntVector<NDIM>(0),
+          const Vector& initial_periodic_displacement = Vector::Zero(),
+          const Vector& current_periodic_displacement = Vector::Zero(),
           const std::vector<SAMRAI::tbox::Pointer<Streamable> >& node_data =
               std::vector<SAMRAI::tbox::Pointer<Streamable> >());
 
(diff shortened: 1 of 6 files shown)


boyceg:
Looks fine. Squash the commits, and then maybe have a second commit to set the initial values in IBStandardInitializer?

amneetb:
The initial value is already set in the constructor (which is called from IBStandardInitializer). No need to do a separate initialization for these data members.

boyceg:
It initializes the shifts to zero, right? Don't you want them to be nonzero if the initial position requires a shift to be inside the computational domain?

amneetb:
On Mon, Feb 8, 2016 at 10:40 PM, Boyce Griffith <notifications@github.com>
wrote:

> It initializes the shifts to zero, right?


It initializes it to whatever is passed to its constructor. See here:
https://github.com/IBAMR/IBAMR/blob/master/src/IB/IBStandardInitializer.cpp#L466-L467


-- 
--Amneet

boyceg:
Ah, I see --- looks like the values are initialized correctly already in the current PR.

However, the constructor really should let you set the values for the initial and actual displacements, since they will be equal only at the beginning of the simulation.

amneetb:
Yeah --- I just searched the library, and it seems like this is constructed only once and rest of it is registerPeriodicShifts(). I thought it was on purpose, since LNode should be constructed only once (mass conservation) at the beginning of the simulation. A duplicated LNode will get this t_0 info from another LNode. This seemed like a reasonable design to me.

boyceg:
> On Feb 9, 2016, at 1:56 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> Yeah --- I just searched the library, and it seems like this is constructed only once and rest of it is registerPeriodicShifts(). I thought it was on purpose, since LNode should be constructed only once (mass conservation) at the beginning of the simulation. A duplicated LNode will get this t_0 info from another LNode. This seemed like a reasonable design to me.
> 

The design is not unreasonable, but it really should be possible to construct LNodes with different initial and current shifts — just update the constructor and the call in IBStandardInitializer.

amneetb:
Done.

boyceg:
Looks good. Please squash. Also, please take a quick look for places in the code that are using the initial frame to compute forces — this may only be for standard beam forces, and maybe not even there — and make sure that initial positions shifted into the correct frame.

Please also test out a version of IB/ex0 or IB/ex1 that places the initial position of the structure straddling a periodic boundary --- note that this can be done easily by shifting the domain extents instead of the initial positions of the points, so it should not be a lot of work.

amneetb:
ex0 is working fine with structure straddling outside the domain. The code brings the structure inside the domain and then computes forces on it . A circle straddling out of the periodic domain on both ends in x directions looks like an hour glass in VisIt.

boyceg:
Looks good. Please squash and then I will merge.

amneetb:
Done.

PR #58, opened 2016-02-12
Fixed bug in ibtk/src/math/fortran/vclaplace2d.f.m4. Also modified in…

nishantnangia:
…put file in ibtk/examples/VCLaplace to have velocity field that has nonzero off-diagonals of (grad u + grad u ^T)

### DIFF
--- a/ibtk/examples/VCLaplace/input2d
+++ b/ibtk/examples/VCLaplace/input2d
@@ -1,15 +1,15 @@
 u {
-   function_0 = "sin(2*PI*X_0)*sin(2*pi*X_1)"  // x-component
-   function_1 = "cos(2*PI*X_0)*cos(2*pi*X_1)"  // y-component
+   function_0 = "(1.0-X_0)*X_0*(1.0-X_1)*X_1" // x-component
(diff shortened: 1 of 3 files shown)


boyceg:
Updated kernel looks good, but I get an error with the example:

griffith@Boyces-MacBook-Pro:VCLaplace$ ./main2d input2d 
WARNING: Automatically using both VisIt and Silo data writers.
In the future, it will be necessary to include "Silo" in the list of data writer in order to output Silo data.
.
P=00000:Program abort called in file ``/Users/griffith/sfw/samrai/current/darwin-clang-dbg/include/Geometry.C'' at line 137
P=00000:ERROR MESSAGE: 
P=00000:Refine operator 'CUBIC_COARSEN' not found for variable 'u_side
Abort trap: 6

nishantnangia:
Oops. I had fixed the InterpolationTransactionComponent object declarations a few days ago on my machine to get the example to run, but forgot to include those fixes on my branch. I know what needs to be done and I'll commit those changes tomorrow when I'm at my computer.

PR #66, opened 2016-02-29
switching most calls to ISCreateGeneral to use PETSC_OWN_POINTER

boyceg:


boyceg:
@amneetb Please take a look and check to see if this makes any difference in solver setup times.

Note that I did not modify the MSM case, but it would be straightforward to do so.

amneetb:
Sure. Thanks!

On Mon, Feb 29, 2016 at 2:24 PM, Boyce Griffith <notifications@github.com>
wrote:

> @amneetb <https://github.com/amneetb> Please take a look and check to see
> if this makes any difference in solver setup times.
>
> Note that I did not modify the MSM case, but it would be straightforward
> to do so.
>
> —
> <https://github.com/IBAMR/IBAMR/pull/66#issuecomment-190426221>.
>



-- 
--Amneet

PR #67, opened 2016-02-29
adding SAJ_fill parameter to IB MG smoothers

boyceg:


boyceg:
@amneetb This will let you set the fill factor for SAJ construction. For Stokes-IB/test2, it needs to be around 9 to avoid any reallocations when constructing SAJ.

Please check how this parameter affects performance.

amneetb:
Great! How did you find out about zero reallocations -- command line?
Won't this 9 be true for all cases then?

On Mon, Feb 29, 2016 at 2:38 PM, Boyce Griffith <notifications@github.com>
wrote:

> @amneetb <https://github.com/amneetb> This will let you set the fill
> factor for SAJ construction. For Stokes-IB/test2, it needs to be around 9
> to avoid any reallocations when constructing SAJ.
>
> Please check how this parameter affects performance.
>
> —
> <https://github.com/IBAMR/IBAMR/pull/67#issuecomment-190431361>.
>



-- 
--Amneet

boyceg:
Didn't I email you or text you about how to do this? Maybe not. In any case, run with -info and grep for MatPtAP or Realloc.

It looks like reallocations are needed only on one of the levels (my guess is the coarsest one). I do not know if this is a universal factor or not. I think that it will need to be larger if there is lots of "long range" coupling (long range relative to the background grid spacing) in A. Probably also different in 2D vs 3D.

We should probably put in a default that is different than 1.0.

PR #72, opened 2016-04-04
support for "augmented" linear solves

boyceg:
We need support for solving systems of equations that involve "extra" variables (in addition to Cartesian grid data).

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -184,8 +184,8 @@ am__DIST_COMMON = $(srcdir)/Makefile.in \
 	$(top_srcdir)/config/install-sh $(top_srcdir)/config/ltmain.sh \
 	$(top_srcdir)/config/make.inc.in $(top_srcdir)/config/missing \
 	config/compile config/config.guess config/config.rpath \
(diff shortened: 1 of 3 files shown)


boyceg:
@amneetb This is my stab at an augmented solver. Note that the solver is set up to make it easy to continue to use standard Cartesian grid solvers. There is are extra MatVecMult and PCApply functions to be defined in subclasses to control how the extra variables are coupled to the rest of the system.

boyceg:
Although now that I have created the PR, it occurs to me that it might be better to request users to supply the "coupling" operators directly (e.g. as Mat and PC objects).

amneetb:
Looks good. I think for initializing extra_vec's one will need to pass in Vecs of these kinds.

boyceg:
The intent is that users set these when using the subclass before calling solve. It is clunky.

boyceg:
@amneetb Revised this using MatNest. Still missing support for custom preconditioning, but I think this is less clunky than the previous design.

PR #76, opened 2016-05-18
Add support for backward Euler in the implicit IB solver

boyceg:


boyceg:
@amneetb Please review.

I have attempted to remove all hard-coded constants that depend on the time stepping scheme along with hard-coded evaluation times.

amneetb:
Looks good in a quick glance. I will carefully go through the code tomorrow and try the example case. This PR will also support BD for explicit IB, right? Also, need to change things in CIB{FE}Method class as well...

boyceg:
> On May 18, 2016, at 12:05 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> Looks good in a quick glance. I will carefully go through the code tomorrow and try the example case. This PR will also support BD for explicit IB, right?
> 
No --- backward Euler requires an implicit solver.
> Also, need to change things in CIB{FE}Method class as well...
> 
Sounds like a good idea, but you will need to do it.

PR #78, opened 2016-06-16
Global CellIndex computation in ConstraintIBMethod

nishantnangia:
Compute the CellIndex from global domain geometry rather than from patch geometry for more robust volume computation


amneetb:
Looks good. Merge it with master once you test it with 'next' branch.

PR #95, opened 2016-08-04
explicitly set homogeneous BCs for spreading in explicit IB solver

boyceg:


boyceg:
@nishantnangia see if this fixes your problem.

nishantnangia:
@boyceg this seems to fix the problem

PR #99, opened 2016-08-05
update custom domain decomposition smoother

boyceg:


boyceg:
@amneetb This custom RASM implementation works in parallel. I will do (local) MSM next.

boyceg:
@amneetb this now includes an updated MSM also. MSM works in parallel, but performance with AMR is quite poor. On the other hand, RASM works well with AMR.

boyceg:
@amneetb please review and let me know if this looks OK to you.

Note that I dropped the "use_ksp_as_smoother" flag since it was only setting the solver norm type to NONE, which is something that I think is better done explicitly using command-line PETSc flags.

PR #104, opened 2016-08-09
simplify ASM subdomain indexing at c-f interfaces

boyceg:


boyceg:
@amneetb This uses a different approach to decide whether to include a DOF index in a subdomain. If the DOF is on an "upper" patch boundary, it is included in the subdomain if that location is "owned" by the processor.

Ownership is determined by the side data synchronization routine, which is able to resolve overlapping values along patch boundaries.

I haven't tried to fix your code, but my guess is that the c-f boundary box was sticking "inside" one of the patches at a corner in the c-f interface, and so the index was being identified as "belonging" to both patches that owned it.

PR #107, opened 2016-08-09
Jenkinsfile

deleeke:
@boyceg 

Once this is merged I can change the Jenkins build script to execute this file. Once that is the case you will be able to change how the Jenkins builds configure and build IBAMR in a way that is consistent with the specific needs of the branch you are working on. 

You will need to make sure any pull requests going forward have a valid .jenkinsfile in the top directory or you will get a false fail if it cannot find the file.


boyceg:
Nice!

Is the same .jenkinsfile used for each of the toolchain tests? Can different .jenkinsfile'es be used for different checks and/or could the .jenkinsfile have different sections for different checks?

deleeke:
Right now the LLVM and GCC toolchain dependencies are set up on the VM to have identical paths except their top directory, so they can both use the same .jenkinsfile because the job just sets the $HOME environment variable before it executes the .jenkinsfile script. 

The .jenkinsfile is simply a shell script, and its name is arbitrary. 
I imagine there are many ways we could use it. 
For example instead of this, we could have a couple different scripts like .jenkins_quicktest, .jenkins_fulltest, and have the nightly build run the .jenkins_fulltest and the pull requests run .jenkins_quicktest. 

Would you rather do that? Even if they are essentially the same right now, it may be good to go ahead and get the files put in place and the jobs configured to execute their respective .jenkinsfile so that as you decide how you'd like to test the pull requests and the nightly build differently it is easy to change.

boyceg:
Having multiple files for different types of tests sounds great! Do you want to add this to this PR, or shall I go ahead and merge?

deleeke:
@boyceg 

Ok, now it is ready for merge. Currently both .jenkins* files are identical, but feel free to modify the contents as you like in the future!

Once this is merged I will activate the nightly build for the master branch.

boyceg:
Amusingly, Jenkins does not think that this is OK to merge.

deleeke:
Not sure why this failed during compilation, maybe it is because it has fallen behind a couple commits from master from yesterday? 

This is where it failed:
```
make[2]: Entering directory `/srv/jenkins_home/sharedspace/mainIBAMR_pr/ibtk/lib'
  CXX      ../src/solvers/impls/libIBTK2d_a-CCPoissonBoxRelaxationFACOperator.o
  CXX      ../src/solvers/impls/libIBTK2d_a-CCPoissonLevelRelaxationFACOperator.o
  CXX      ../src/solvers/impls/libIBTK2d_a-CCPoissonSolverManager.o
  CXX      ../src/solvers/impls/libIBTK2d_a-CCPoissonPointRelaxationFACOperator.o
  CXX      ../src/solvers/impls/libIBTK2d_a-FACPreconditioner.o
  CXX      ../src/solvers/impls/libIBTK2d_a-PETScAugmentedKrylovLinearSolver.o
  CXX      ../src/solvers/impls/libIBTK2d_a-PoissonFACPreconditioner.o
../src/solvers/impls/PETScAugmentedKrylovLinearSolver.cpp:450:1: warning: unused parameter ‘x’ [-Wunused-parameter]
 PETScAugmentedKrylovLinearSolver::initializeSolverStateAugmented(const SAMRAIVectorReal<NDIM, double>& x,
 ^
../src/solvers/impls/PETScAugmentedKrylovLinearSolver.cpp:450:1: warning: unused parameter ‘b’ [-Wunused-parameter]
../src/solvers/impls/PETScAugmentedKrylovLinearSolver.cpp: In member function ‘void IBTK::PETScAugmentedKrylovLinearSolver::resetKSPPC()’:
../src/solvers/impls/PETScAugmentedKrylovLinearSolver.cpp:602:94: error: cannot convert ‘const char*’ to ‘PetscOptions {aka _n_PetscOptions*}’ for argument ‘1’ to ‘PetscErrorCode PetscOptionsGetString(PetscOptions, const char*, const char*, char*, size_t, PetscBool*)’
     ierr = PetscOptionsGetString(d_options_prefix.c_str(), "-pc_type", pc_type_str, len, &flg);
                                                                                              ^
  CXX      ../src/solvers/impls/libIBTK2d_a-PoissonFACPreconditionerStrategy.o
make[2]: *** [../src/solvers/impls/libIBTK2d_a-PETScAugmentedKrylovLinearSolver.o] Error 1
```
I think you can still merge it because my .jenkins* files have nothing to do with that

deleeke:
I suppose this is a drawback of using the same workspace for all the pull requests

boyceg:
Yeah, PETScAugmentedKrylovLinearSolver.cpp is actually coming from another branch, so this appears to be a stale Makefile.

boyceg:
If you configure with --enable-maintainer-mode, then running make causes the configure script along with all of the Makefile's to be automatically regenerated when there are changes to configure.ac or any of the Makefile.am files.

deleeke:
Ah, so I think that I'm understanding that in the cases where one branch has new/unique files, not just changed ones, that when we use the same workspace those files are still kicking around even when it checks out a branch that doesn't have those files?

Not sure how to deal with that.

deleeke:
Oh, ok I see. Let me change that.

boyceg:
Don't change it just yet...

deleeke:
Ok let me know when. I was planning on changing it in Jenkins and in these .jenkins* files

boyceg:
I just remembered that using --enable-maintainter-mode can yield failure cases that cause "make" to enter an infinite loop.

boyceg:
I am going to go ahead and merge this PR. Let's see if we can think of a way to make the checks more robust.

PR #108, opened 2016-08-09
Extending interface of PETScLevelSolver for ASM and FieldSplit PC's

amneetb:
@boyceg : Please review.


amneetb:
One can potentially use a different routine to generate subdomains for PCTYPE SHELL.

boyceg:
This looks good. Could you avoid the repeated code for ASM and "shell" PC types?

PR #116, opened 2016-08-23
Adding bi/trilinear matrix based prolongation operator.

amneetb:
@boyceg: Please review. It took a lot of time. The input file needs to specify U_petsc_prolongation_method = "LINEAR"

1) The bug with periodic domain is coming from AO returning negative index. This is caught here for RT0 op:

https://github.com/IBAMR/IBAMR/blob/amneetb/bi-tri-linear-prolongation/ibtk/src/math/PETScMatUtilities.cpp#L1147

and here for LINEAR op

 https://github.com/IBAMR/IBAMR/blob/amneetb/bi-tri-linear-prolongation/ibtk/src/math/PETScMatUtilities.cpp#L1512

Not sure why the index (0,0) is not getting mapped (Try Stokes-IB/test1 with input2d.shell having periodic boundary(ies))

### DIFF
--- a/ibtk/include/ibtk/IndexUtilities.h
+++ b/ibtk/include/ibtk/IndexUtilities.h
@@ -108,25 +108,6 @@ public:
                                                   const SAMRAI::hier::Index<NDIM>& iupper);
 
     /*!
-     * \brief Convert (i,j,k,d) index for a DOF to an integer mapping on a patch level.
-     *
-     * \param i AMR index representing the (i,j,k) data cell index.
-     *
-     * \param domain_lower Lower index of the domain.
-     *
-     * \param num_cells Number of data array cells for a patch level.
-     *
-     * \param depth Data depth
-     *
-     * \return The linear mapping of an AMR index to a continuous non-negative
-     * integer space.
-     */
-    static int getIntegerMapping(const SAMRAI::hier::Index<NDIM>& i,
-                                 const SAMRAI::hier::Index<NDIM>& domain_lower,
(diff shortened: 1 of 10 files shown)


amneetb:
Yeah... The scheme reverts to RT0 near boundaries (in lieu of simpler implementation). This will not give exact results.

boyceg:
> On Aug 22, 2016, at 8:31 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> Yeah... The scheme reverts to RT0 near boundaries (in lieu of simpler implementation). This will not give exact results.
> 

I would suggest not doing that — use linear all the way up to the boundaries.

amneetb:
That will require physical boundary conditions...

boyceg:
> On Aug 22, 2016, at 8:34 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> That will require physical boundary conditions…
> 
You could use extrapolation, or you could use physical boundary conditions, or you could fix periodic boundary conditions.

amneetb:
Right now it's like homog Neumann.

boyceg:
> On Aug 22, 2016, at 8:36 PM, Amneet Bhalla <notifications@github.com> wrote:
> 
> Right now it's like homog Neumann.
> 
Given that this is for an experiment and not necessarily to be merged into master, I would just make it work for periodic boundaries, and then we can sort out what to do at physical boundaries if it works.

amneetb:
Sounds good.

PR #120, opened 2016-09-15
Adding Galerkin Stokes-IB MG method

amneetb:
@boyceg: Check tests/Stokes-IB/test1. Galerkin Stokes-IB is not performing better than rediscretized Stokes. input2d.shell has the necessary algorithmic controls.

### DIFF
--- a/ibtk/include/ibtk/PETScLevelSolver.h
+++ b/ibtk/include/ibtk/PETScLevelSolver.h
@@ -240,6 +240,16 @@ public:
      */
     void addLinearOperator(Mat& op);
 
+    /*!
+     * \brief Use externally provided linear operator.
+     * \NOTE this function should be called prior to initializing the solver
+     * state.
+     *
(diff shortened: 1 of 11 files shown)


ibtk/include/ibtk/PETScLevelSolver.h
```diff
@@ -240,6 +240,16 @@ class PETScLevelSolver : public LinearSolver
      */
     void addLinearOperator(Mat& op);
 
+    /*!
```
boyceg: This function (along with addLinearOperator) does not seem to be used anywhere except in the Stokes solver.

This stuff doesn't belong in the PETScLevelSolver interface unless it is used by all implementations.

boyceg: I attempt to fix this in PR #136.

boyceg:
Superseded by #136.

PR #133, opened 2016-11-08
switch VecMultiVec to VecNest

boyceg:


boyceg:
@amneetb Please check if this breaks anything in CIB.

boyceg:
Ugh, there is some breakage here.

boyceg:
OK, I was too aggressive in removing calls to PetscObjectStateIncrease() in IBImplicitStaggeredHierarchyIntegrator.

PR #142, opened 2016-11-19
Update .jenkins* files

deleeke:
@boyceg 

I changed compiler flags to just "-Wall" and replaced the samrai path with "$SAMRAI_DIR" to allow for optimized builds. All the jenkins jobs have $SAMRAI_DIR defined, so hopefully this should work.

I still have yet to get the optimized build of SAMRAI built on the VM, but this will make it possible to get it running once it is in place.

### DIFF
--- a/.jenkins_fulltest
+++ b/.jenkins_fulltest
@@ -18,7 +18,8 @@ set -o pipefail
   --with-boost=$HOME/linux/boost/1.61.0 \
   --enable-libmesh \
   --with-libmesh=$HOME/linux/libmesh/1.0.0/1.0.0-debug \
-  --with-libmesh-method=dbg
+  --with-libmesh-method=dbg \
+  --config-cache
 
 make -kj4 lib
 make -kj4 examples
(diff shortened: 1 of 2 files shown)


deleeke:
Ok, apparently I spoke too soon.

```
+ ./.jenkins_quicktest
configure: loading cache config.cache
configure: error: `CFLAGS' has changed since the previous run:
configure:   former value:  `-g -O1 -Wall'
configure:   current value: `-Wall'
configure: error: `CXXFLAGS' has changed since the previous run:
configure:   former value:  `-g -O1 -Wall'
configure:   current value: `-Wall'
configure: error: `FCFLAGS' has changed since the previous run:
configure:   former value:  `-g -O1 -Wall'
configure:   current value: `-Wall'
configure: error: in `/srv/jenkins_home/workspace/IBAMR_pr':
configure: error: changes in the environment can compromise the build
configure: error: run `make distclean' and/or `rm config.cache' and start over
Build step 'Execute shell' marked build as failure
```

boyceg:
You may need to remove config.cache when you change flags or options.

deleeke:
@boyceg 

I deleted the old config.cache in all four PR and overnight jobs, so and added back the --config-cache flag. I think it should be good to go now.

deleeke:
opt test

(this first build will probably take a while since it will be in a fresh directory)

PR #145, opened 2016-12-12
Output strain rate

abarret:
Added functionality to compute the cell-centered strain rate from a side-centered vector field, i.e. E = 0.5*(grad(u) + grad(u)^T).

Also adds the option to output the fluid strain rate to viz files, defaults to false.


deleeke:
retest this please
(I think there may have been a problem with config.cache on this last opt build)

boyceg:
In the function calls, should strain --> strain_rate?

abarret:
That's probably a good idea. Should I also change all variables to strain_rate?

boyceg:
I just did it. I will merge once Jenkins reports back clean builds.

PR #147, opened 2017-01-24
Fixing 2D torque computation and added helpers to get momentum, volume, and COM of structures

nishantnangia:
@amneetb Please take a look.

### DIFF
--- a/include/ibamr/ConstraintIBMethod.h
+++ b/include/ibamr/ConstraintIBMethod.h
@@ -179,6 +179,31 @@ public:
         return d_l_data_U_correction;
     }
 
+    /*!
+     * \brief Get the total volume for all the Lagrangian structures
+     */
+    inline const std::vector<double>& getVolumeStructure()
+    {
+        return d_vol_structure;
+    }
+
+    /*!
+     * \brief Get the total linear momentum for all the Lagrangian structures
+     */
+    inline const std::vector<std::vector<double> >& getLinearMomentumStructure()
+    {
+        return d_linear_mom_structure;
+    }
+
+    /*!
+     * \brief Get the center of mass for all Lagrangian structures
+     */
+
+    inline const std::vector<std::vector<double> >& getCOMStructure()
+    {
+        return d_center_of_mass_new;
+    }
+
 private:
     /*!
      * \brief Default constructor.
@@ -335,6 +360,11 @@ private:
     void calculateEulerianMomentum();
 
     /*!
+     * \brief Calculate total translational momentum of all Lagrangian structures
+     */
+    void calculateLinearMomentumStructure();
+
+    /*!
      * No of immersed structures.
      */
     const int d_no_structures;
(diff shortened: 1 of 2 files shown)


amneetb:
@nishantnangia Looks good.

include/ibamr/ConstraintIBMethod.h
```diff
@@ -30,8 +30,8 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-#ifndef included_IBAMR_ConstraintIBMethod
-#define included_IBAMR_ConstraintIBMethod
+#ifndef included_ConstraintIBMethod
```
boyceg: Please do not remove the namespacing here.

include/ibamr/ConstraintIBMethod.h
```diff
@@ -486,4 +526,4 @@ class ConstraintIBMethod : public IBAMR::IBMethod
 
 /////////////////////////////////////////////////////////////////////////
 
-#endif //#ifndef included_IBAMR_ConstraintIBMethod
+#endif //#ifndef included_ConstraintIBMethod
```
boyceg: Same here...

src/IB/ConstraintIBMethod.cpp
```diff
@@ -452,6 +454,7 @@ ConstraintIBMethod::postprocessSolveFluidEquations(double current_time, double n
     if (d_output_torque) calculateTorque();
     if (d_output_eul_mom) calculateEulerianMomentum();
     if (d_output_power) calculatePower();
+    calculateLinearMomentumStructure();
```
boyceg: Should this be protected by an if statement, as above?

include/ibamr/ConstraintIBMethod.h
nishantnangia: My mistake. I reverted back to original namespacing and squashed

src/IB/ConstraintIBMethod.cpp
nishantnangia: No, since the linear momentum is not outputted to a file. Rather, it is computed then stored in d_linear_mom_structure, which is returned by getLinearMomentumStructure(). This is helpful for hydrodynamic force computation since the momentum of the structure is needed at every time step.

amneetb:
@nishantnangia : Whenever in doubt follow IBAMR code files to adopt the coding convention. Looks like the files have been clang-formatted now. 

@Boyce: Can you mark on IBAMR github page about the clang-format version and automake version for all developers to adopt. All should update at the same time.

boyceg:
> @Boyce <https://github.com/Boyce>: Can you mark on IBAMR github page about the clang-format version and automake version for all developers to adopt. All should update at the same time.

I think it is too strict to require the same version. If we do incremental clang-formatting, I think that that should be good enough (assuming that the formatting is initially good, of course).

PR #153, opened 2017-04-25
Adding provisions to allow ASM preconditioning for PETSc level Poisson solvers

nishantnangia:
@boyceg, @amneetb: I moved the constructPatchLevelASMSubdomains calls for the PETSc level Poisson solvers to a generateASMSubdomains function. Moreover, I added provisions to ensure that generate_petsc_is_from_std_is is never called from PETScLevelSolver when subdomains are indexed by PETSc IS indexing directly.

This allows for ASM preconditioning to be used for PETSc level Possion solvers; it was encountering TBOX_ERROR before.

### DIFF
--- a/ibtk/include/ibtk/CCPoissonPETScLevelSolver.h
+++ b/ibtk/include/ibtk/CCPoissonPETScLevelSolver.h
@@ -134,6 +134,10 @@ public:
 
 protected:
     /*!
+     * \brief Generate IS/subdomains for Schwartz type preconditioners.
+     */
+    void generateASMSubdomains(std::vector<std::set<int> >& overlap_is, std::vector<std::set<int> >& nonoverlap_is);
+    /*!
      * \brief Compute hierarchy dependent data required for solving \f$Ax=b\f$.
      */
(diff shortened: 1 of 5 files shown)


boyceg:
Looks good.

@deleeke, perhaps you could work with Nishant to get a unit test set up for this in a separate branch. (Although I guess I still need to merge some stuff before that can happen.)

deleeke:
On Tue, Apr 25, 2017 at 11:05 AM, Boyce Griffith <notifications@github.com>
wrote:

> Looks good.
>
> @deleeke <https://github.com/deleeke>, perhaps you could work with
> Nishant to get a unit test set up for this in a separate branch. (Although
> I guess I still need to merge some stuff before that can happen.)
>
Ok, I can pull this and email Nishant about it.

As far as merging my gtest branch, if you want to be more incremental, we
could just merge the portion that allows people to link to the gtest
library (configure, the googletest m4 script, etc).
That way other developers could start writing tests for their code.
 Also, I don't know if you noticed but I already made a package for
BuildSystem to be able to download/configure/link to googletest as well so
when we migrate that won't be an issue.

Issue #144, opened 2016-11-23
Runtime error in example/IB/explicit/ex5


deleeke:
I'm getting this run-time error in IB/explicit/ex5 
This error message is from the main2d, but main3d produces an almost identical error.

This MAY be some issue with my SAMRAI...

```
/examples/IB/explicit/ex5$ ./main2d input2d

WARNING: Automatically using both VisIt and Silo data writers.
In the future, it will be necessary to include "Silo" in the list of data writer in order to output Silo data.
.
IBStandardInitializer:  Reading from input files.
  base filename: points2d
  assigned to level 0 of the Cartesian grid patch hierarchy


Global seed = 1479860631

IBStandardInitializer:  Deallocating initialization data.

Writing visualization files at timestep # 0 t=0

Writing output files at timestep # 0 t=0
P=00000:Program abort called in file ``/home/deleeke/sfw/samrai/2.4.4/linux-g++-debug/include/Patch.C'' at line 117
P=00000:ERROR MESSAGE: 
P=00000:Failed assertion: d_patch_data[id].isNull()
[deleeke-x401a:01591] *** Process received signal ***
[deleeke-x401a:01591] Signal: Aborted (6)
[deleeke-x401a:01591] Signal code:  (-6)
```

boyceg:
Really fixed by #171.

PR #159, opened 2017-08-24
Fix stochastic forcing function

abarret:
Extra stress variables in stochastic forcing function were not deallocated correctly.

Fixes issue #144


boyceg:
👍

PR #165, opened 2017-09-29
Adding fast sweeping level set method

amneetb:
@nishantnangia : Can you take a look at the fortran kernel? If it looks OK, I will do the second order.


boyceg:
Great!

What is the plan for integrating this with a fluid solver? The level set is an auxiliary field that is managed by the application code, or something else?

amneetb:
It will not be integrated with fluid solver per say. The application code will register a LS variable with hyperbolic integrator, get current context from it,  map it to patch index, and will re-initialize it time-to-time depending upon the physics of the problem. 

It is more or less self-sufficient. What do you think?

boyceg:
Sounds good. These are going to be passively advected level set functions? (Although I guess is it not that hard to add other explicit solvers for that part of the problem too.)

How frequently do these things need to be reinitialized --- after every step? If so, may eventually want to extend the hyperbolic level integrator to add in some callback functions, so that the application code doesn't have to worry about it.

amneetb:
Yes, LS will be passively advected. 

The frequency of reinitialization is highly problem dependent. Yeah, I had thought about extending the hyperbolic level integrator too.  We should eventually build that support.

nishantnangia:
Looks good so far. I'll take a closer look on the plane tomorrow.

nishantnangia:
@boyceg : I'm almost done verifying that the variable density/viscosity Stokes solver in the vc-viscous branch is working. Within the next few days, Amneet and I will start using the level set stuff in this branch to set/advect rho and mu.

Since vc-viscous and level-set are separate feature branches, we're not sure the best way to start doing this -- we can either:
1) have the current level-set branch merged into master and then rebase vc-viscous branch against master and work out of that,
2) don't merge anything into master right now and just merge level-set into vc-viscous and work out of vc-viscous (and try to get it merged into master later down the road once its fully validated)

Which of the above do you suggest, or if you have any other suggestions, could you let us know?

boyceg:
> On Oct 19, 2017, at 1:07 AM, Nishant Nangia <notifications@github.com> wrote:
> 
> @boyceg <https://github.com/boyceg> : I'm almost done verifying that the variable density/viscosity Stokes solver in the vc-viscous branch is working. Within the next few days, Amneet and I will start using the level set stuff in this branch to set/advect rho and mu.
> 
Great. I have limited time to review anything until mid-to-late next week. Try to get the PRs in good shape in the meantime, and if things look good, we can aim to merge the following week.
> Since vc-viscous and level-set are separate feature branches, we're not sure the best way to start doing this -- we can either:
> 
> have the current level-set branch merged into master and then rebase vc-viscous branch against master and work out of that,
> don't merge anything into master right now and just merge level-set into vc-viscous and work out of vc-viscous (and try to get it merged into master later down the road once its fully validated)
You could also create your own integration branch and merge in these two branches. If you encounter bugs, then you would fix them in the “upstream” feature branch. I would suggest managing it this way.

amneetb:
@nishantnangia : I think it makes a lot of sense to add level set reinitialization stuff directly into AdvDiffHierarchyIntegrator. The AdvDiffHierarchyIntegrator::preprocessHierarchyIntegrator() will make calls to reinitLS(). The reinitLS() should also take integrator step number.

Hiding LS stuff under AdvDiffHierarchyIntegrator will make VCINSIntegrator much closer to INSIntegrator --- so there is a good chance that we need just one of them  :-)

deleeke:
@jenkinsbot1024 retest this please

@boyceg  I woke up the pull request builder jobs by making them poll for changes instead of waiting on github hook, since we did not set up the new VM to allow POST from github.

Looks like the VM does not have `ccache` installed. i've now commented out the line in the bash script that does the build that actually runs the jenkins files, just to see if it will work now.

deleeke:
@jenkinsbot1024 llvm test

deleeke:
@jenkinsbot1024 llvm test

deleeke:
@jenkinsbot1024 llvm test
gcc test
opt test

boyceg:
@amneetb @nishantnangia let me know if this is ready for review.

amneetb:
@boyceg The core part (fortran) is done. There may be just cosmetic changes  to this class in the future when we do FSI cases.

PR #166, opened 2017-10-06
Delayed initial Lagrangian data movement until after all Lag data has been set up.

abarret:
Lagrangian data in the penalty IB method was being initialized *after* all other Lagrangian data had been redistributed.

@boyceg It seems that the penalty IB method is the only IBMethod class that initializes method specific Lagrangian data in the initializePatchHierarchy() function. For example, the generalized IB method initializes director vectors in the initializeLevelData() function. Would it be better to move penalty IB method Lagrangian data initialization to the initializeLevelData() function?

### DIFF
--- a/include/ibamr/PenaltyIBMethod.h
+++ b/include/ibamr/PenaltyIBMethod.h
@@ -149,6 +149,20 @@ public:
         bool initial_time);
 
     /*!
+     * Initialize data on a new level after it is inserted into an AMR patch
+     * hierarchy by the gridding algorithm.
+     *
+     * \see SAMRAI::mesh::StandardTagAndInitStrategy::initializeLevelData
+     */
+    void initializeLevelData(SAMRAI::tbox::Pointer<SAMRAI::hier::BasePatchHierarchy<NDIM> > hierarchy,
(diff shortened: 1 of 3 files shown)


boyceg:
> On Oct 6, 2017, at 11:58 AM, abarret <notifications@github.com> wrote:
> 
> Lagrangian data in the penalty IB method was being initialized after all other Lagrangian data had been redistributed.
> 
> @boyceg <https://github.com/boyceg> It seems that the penalty IB method is the only IBMethod class that initializes method specific Lagrangian data in the initializePatchHierarchy() function. For example, the generalized IB method initializes director vectors in the initializeLevelData() function. Would it be better to move penalty IB method Lagrangian data initialization to the initializeLevelData() function?
> 

I think so: I believe that initializeLevelData() will be called as the levels are constructed, and so would avoid the issue of setting up data after Lagrangian data redistribution. Please take a look and, if this seems right, make this change.

abarret:
@boyceg The mass and penalty parameters are now created in initializeLevelData(). Initialization of the location and velocity are kept in initializePatchHierarchy since these values are simply copied.

PR #168, opened 2017-11-03
Fix bugs in the sharp-interface IBFE implementation

boyceg:


amneetb:
@boyceg : Wanted to make sure that MLS stuff does not get removed from the repo :-). Rahul  is referring this PR to do MLS in his HPC code at Riken.

boyceg:
That is hilarious, I was just cleaning that stuff out!

boyceg:
In that case, I will not make any changes.

amneetb:
Yep, sometimes I just get uncanny intuitions ;-)

boyceg:
Shall I go ahead and merge this into master "as is", and clean up any dangling issues in other PRs? We are regularly using this branch for all IB/FE models, although I believe that there was an issue that I was wanting to fix before merging. :-(

boyceg:
This PR updates the required version of libMesh to 1.1.0.

amneetb:
Hmmn, IIRC correctly MLS is just there but not getting used. I created one example as indicated by 
"Adding MLS to standard IBFE/explicit/ex0" but I think it was/is reverted back. As far MLS is concerned it should not cause any error because it is not used. 

Other than that I have lost track of IBFE development.

PR #180, opened 2017-12-10
IBFE ex7 tiny fix

sach1:
Removed code that was previously enclosed in {{ #if 0, #endif }}

Example now compiles & runs successfully.

### DIFF
--- a/examples/IBFE/explicit/ex7/example.cpp
+++ b/examples/IBFE/explicit/ex7/example.cpp
@@ -225,23 +225,6 @@ bool run_example(int argc, char* argv[])
                                             2.0 * D + 0.5 * w,
                                             Utility::string_to_enum<ElemType>(elem_type));
         upper_mesh.prepare_for_use();
-        el_end = upper_mesh.elements_end();
-        for (MeshBase::const_element_iterator el = upper_mesh.elements_begin(); el != el_end; ++el)
-        {
-            Elem* const elem = *el;
-            for (unsigned int side = 0; side < elem->n_sides(); ++side)
-            {
-                const bool at_mesh_bdry = !elem->neighbor(side);
-                if (at_mesh_bdry)
-                {
-                    BoundaryInfo* boundary_info = upper_mesh.boundary_info.get();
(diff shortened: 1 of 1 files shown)


sach1:
Also want to point out that the PK1 stress function arguments keep switching between commits for all the IBFE examples (even in prior versions of main.cpp where tests were not yet set up)

It changes between this

```cpp
void
upper_PK1_stress_function(
    TensorValue<double>& PP,
    const TensorValue<double>& FF,
    const libMesh::Point& /*X*/,
    const libMesh::Point& s,
    Elem* const /*elem*/,
    const vector<const vector<double>*>& /*var_data*/,
    const vector<const vector<VectorValue<double> >*>& /*grad_var_data*/,
    double /*time*/,
    void* /*ctx*/)
```

and

```cpp
void
upper_PK1_stress_function(
    TensorValue<double>& PP,
    const TensorValue<double>& FF,
    const libMesh::Point& /*X*/,
    const libMesh::Point& s,
    Elem* const /*elem*/,
    const vector<NumericVector<double>*>& /*system_data*/,
    double /*time*/,
    void* /*ctx*/)
```
not sure if this is intentional...

boyceg:
Yes there was a change, but the arguments should not be jumping back-and-forth. Let me know if this is not the case.

The new interface (with var_data and grad_var_data) is meant to make it easier to use additional field variables in evaluating these kinds of functions, by pre-evaluating the function value or gradient of the extra variable at the location where the stress is being evaluated.

The old interface (with system_data) just passed the full NumericVector to the stress function and left it up to the stress function to evaluate the field variable at the right place.

sach1:
Oh, I understand now. Thanks for the detailed explanation.

boyceg:
Jenkins check this please.

boyceg:
ok to test

boyceg:
Jenkins check this please.

boyceg:
Hmmmm, it looks like the compiler is just crashing on the new VM...

boyceg:
retest this please

boyceg:
Hmmm, Jenkins is having trouble with this branch. I am going to go ahead and merge since it only touches the example code.

PR #183, opened 2017-12-17
Adding multiphase fluid solver/examples

amneetb:


nishantnangia:
@amneetb I added functionality to allow the user to override the computation of the volume element for ConstraintIBMethod by providing their own value. The volume of the Lagrangian structure is then (Number of lag points)*vol_element.

Can you think of a situation where we would want the user to provide the volume of the structure instead of the volume element?

include/ibamr/ConstraintIBMethod.h
```diff
@@ -196,6 +196,25 @@ class ConstraintIBMethod : public IBAMR::IBMethod
...
+
+    /*
+     * Set the volume element for each Lagrangian node for all the structures
+     */
+    inline void setVolumeElement(std::vector<double> vol_element)
+    {
+        d_vol_element = vol_element;
+        d_vol_element_is_set = std::vector<bool>(d_no_structures, true);
+    }
+
```
amneetb: Add an assertion that passed in vol_element size is d_no_structures.

amneetb:
Looks good. ConstraintIB assumes certain regularity of Lagrangian mesh and this method would be used only when the regularity is *really* needed.

src/level_set/fortran/levelsetops2d.f.m4
```diff
@@ -934,7 +1071,302 @@ c     Compute first order derivatives
...
+              enddo
+            enddo
+            nmr = hx*hy/144.d0*nmr
+            dnr = hx*hy/144.d0*dnr
+            lambda = -nmr/dnr
+
+            if (dnr .gt. zero) then
+              U(i0,i1) = C(i0,i1) + lambda*hij*H(i0,i1)
+            endif
+
```
amneetb: Is there a way to store lambda? If lambda can be back-propagated to RelaxationLSMethod, one can basically print the norm of lambda. It should go down to zero if the mass constraint is proper and the solution has converged. One can plog this information.

nishantnangia:
@amneetb How does the rearrangement of composite data calls look to you?

src/IB/IBHierarchyIntegrator.cpp
```diff
@@ -415,6 +415,16 @@ IBHierarchyIntegrator::initializePatchHierarchy(Pointer<PatchHierarchy<NDIM> > h
     // Finish Lagrangian data movement.
     d_ib_method_ops->endDataRedistribution(hierarchy, gridding_alg);
 
+    // Initialize composite hierarchy data that was not initailized by
+    // gridding algorithm's call to initializeLevelData(), which may depend on
+    // Lagrangian data.
+    const bool from_restart = RestartManager::getManager()->isFromRestart();
+    const bool initial_time = MathUtilities<double>::equalEps(d_integrator_time, d_start_time);
+    if (!from_restart)
```
amneetb: Only parent integrator should make this call. 

Something like,
if (d_parent_integrator == NULL && !from_restart)

Does this make sense?

amneetb: Although this looks like an extra check. In practice initializePatchHierarchy() and regridHierarchy() is called only by the top-most integrator. So, it looks fine. 

@boyceg What do you say?

ibtk/include/ibtk/LDataManager.h
```diff
@@ -896,6 +896,13 @@ class LDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::Sta
      */
     void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db);
 
+    /*!
+     * Register user defined Lagrangian data to be maintained
+     *
+     */
+    void registerUserDefinedLData(const std::string& data_name,
+                                  int depth);
```
boyceg: default value of 1?

include/ibamr/AdvDiffHierarchyIntegrator.h
```diff
@@ -439,6 +439,47 @@ class AdvDiffHierarchyIntegrator : public IBTK::HierarchyIntegrator
...
 
+    /*!
+     * Prepare to advance the data from current_time to new_time.
+     */
+    void preprocessIntegrateHierarchy(double current_time, double new_time, int num_cycles = 1);
+
+    /*!
+     * \brief Function to reset variables registered by this integrator
+     */
+    typedef void (*ResetPropertiesFcnPtr)(int property_idx,
```
boyceg: 🥇

include/ibamr/INSHierarchyIntegrator.h
```diff
@@ -248,23 +248,6 @@ class INSHierarchyIntegrator : public IBTK::HierarchyIntegrator
      */
     SAMRAI::solv::RobinBcCoefStrategy<NDIM>* getProjectionBoundaryConditions() const;
 
-    /*!
-     * Register a variable mass density variable with the hierarchy integrator.
-     */
-    void registerMassDensityVariable(SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > rho_var);
```
boyceg: 💯

include/ibamr/VCINSStaggeredHierarchyIntegrator.h
```diff
@@ -0,0 +1,655 @@
...
+
+    /*!
+     * Explicitly remove nullspace components from a solution vector.
+     */
+    void removeNullSpace(const SAMRAI::tbox::Pointer<SAMRAI::solv::SAMRAIVectorReal<NDIM, double> >& sol_vec);
+
+    /*!
+     * Register a variable mass density variable with the hierarchy integrator.
+     */
+    void registerMassDensityVariable(SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > rho_var);
```
boyceg: Do you need a way to give a function to evaluate the mass density?

include/ibamr/VCINSStaggeredHierarchyIntegrator.h
```diff
@@ -0,0 +1,655 @@
...
+
+    /*!
+     * Get the mass density variable registered with the hierarchy integrator.
+     */
+    SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > getMassDensityVariable() const;
+
+    /*!
+     * Register a variable viscosity variable with the hierarchy integrator.
+     */
+    void registerViscosityVariable(SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > mu_var);
```
boyceg: Do you need a way to give a function to evaluate the viscosity?

include/ibamr/VCINSStaggeredHierarchyIntegrator.h
```diff
@@ -0,0 +1,655 @@
...
+    /*!
+     * Set the interpolation type used for material properties mu
+     */
+    void setViscosityVCInterpolationType(const IBTK::VCInterpType vc_interp_type);
+
+    /*!
+     * \brief Function to reset fluid density or viscosity if they are
+     * maintained by this integrator.
+     */
+    typedef void (*ResetFluidPropertiesFcnPtr)(int property_idx,
```
boyceg: Ah, I see, this is how rho and mu are set.

include/ibamr/VCINSStaggeredHierarchyIntegrator.h
```diff
@@ -0,0 +1,655 @@
...
+
+    /*!
+     * \brief Register interface neighborhood locating functions.
+     */
+    void registerResetFluidViscosityFcn(ResetFluidPropertiesFcnPtr callback, void* ctx);
+
+    /*!
+     * \brief Supply initial conditions for the density field, if maintained by the fluid integrator.
+     */
+    void registerMassDensityInitialConditions(SAMRAI::tbox::Pointer<IBTK::CartGridFunction> rho_init_fcn);
```
boyceg: I wonder if we should think about replacing a lot of uses of CartGridFunction with callbacks?

ibtk/include/ibtk/LDataManager.h
amneetb: I think we should force user to provide depth, so that they don't accidentally use depth of 1. But I am also fine with a default depth of 1. 

(I think I do not like default arguments in general. So I am biased :-( )

boyceg:
@amneetb, @nishantnangia, shall we merge if Jenkins is happy?

boyceg:
Hmmm, Jenkins is not happy ...

boyceg:
Please also compile with all warnings turned on (-Wall) and check for warnings --- there are a bunch of warning messages that should be taken care of before merging.

nishantnangia:
Will do.

nishantnangia:
Is there any way for me to automatically see why Jenkins is unhappy?

boyceg:
Currently the Jenkins server is behind a firewall. I will email you the output separately.

src/advect/AdvectorExplicitPredictorPatchOps.cpp
```diff
@@ -912,7 +912,7 @@ AdvectorExplicitPredictorPatchOps::getNumberCellGhosts() const
         return 3;
     case XSPPM7:
         return 4;
-    case UNKNOWN_LIMITER_TYPE:
+    default:
```
nishantnangia: @boyceg : Is it okay to put the default case here like this? We added some additional limiter_types that are not supported by this class. Keeping it as 'case UNKNOWN_LIMITER_TYPE' produced a ton of warnings.

boyceg:
> On May 30, 2018, at 11:22 AM, Nishant Nangia <notifications@github.com> wrote:
> 
> @nishantnangia commented on this pull request.
> 
> In src/advect/AdvectorExplicitPredictorPatchOps.cpp <https://github.com/IBAMR/IBAMR/pull/183#discussion_r191810656>:
> 
> > @@ -912,7 +912,7 @@ AdvectorExplicitPredictorPatchOps::getNumberCellGhosts() const
>          return 3;
>      case XSPPM7:
>          return 4;
> -    case UNKNOWN_LIMITER_TYPE:
> +    default:
> @boyceg <https://github.com/boyceg> : Is it okay to put the default case here like this? We added some additional limiter_types that are not supported by this class. Keeping it as 'case UNKNOWN_LIMITER_TYPE' produced a ton of warnings.
> 

Yes.

nishantnangia:
I have one more quick thing to commit to the INSVC classes, so please hold on merging with master until after that.

boyceg:
Let's do this.

boyceg:
I am also going to delete Nishant's branch, which is fully merged into master.

If for some reason we need to get it back, just do `git checkout -b nishant/high-density-ratio 615c1c2aa69ffd833960c0ecfd3fa8275ffaa7f8`.

PR #184, opened 2017-12-19
Fixed reset of advection-diffusion velocity after cycle 0.

abarret:
@boyceg Should I make a note in the documentation that if a function to set the velocity is not specified, the variable-context pairs of u_var and getCurrent/Scratch/NewContext() must be specified?


boyceg:
What do you mean specified? You mean that the values need to be set? Yes, the docs should probably should make that clear --- that it is the responsibility of the caller to ensure that the current and new contexts are set correctly if a function is not provided. If so, in debug mode, we should make sure to initialize the new context with junk (e.g. quiet NaN) to try to catch errors.

Can you make these changes in the branch?

PR #185, opened 2017-12-21
constrain tangential components of edge DOFs

boyceg:


boyceg:
@xxjaehoxx This branch will contain the code to handle the edge/side DOFs.

boyceg:
This doesn't seem to be as generalizable as I had hoped ... let's try some other stabilization approaches.

PR #186, opened 2017-12-30
Hooks to initialize composite hierarchy data in HierarchyIntegrator 

amneetb:
In some situations, a composite data needs to be defined that requires the entire hierarchy to be constructed a priori. HierarchyIntegrator/StandardTagAndInitializeStrategy::initializeLevelData() is not suitable for that, which expects to construct a level, define a data on it, tag cells on the level based upon the newly initialized data and then finally create the next finer level. Thus, a level-by-level data initialization is assumed in this routine. 

For some algorithms like level set, or a variable that comes out of the solution of a linear solver, one needs to initialize the composite hierarchy data in a single pass. This hook allows one to do that.


boyceg:
In initializeCompositeHierarchyData() / initializeCompositeHierarchyDataSpecialized() --- do you need to pass in the hierarchy? Shouldn't these methods just operate on the patch hierarchy managed by the class?

amneetb:
Good point. I will remove it, it seems redundant to pass it.

ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -420,6 +420,15 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
      */
     void registerApplyGradientDetectorCallback(ApplyGradientDetectorCallbackFcnPtr callback, void* ctx = NULL);
 
+    /*!
+     * Perform data initialization after the entire hierarchy has been constructed.
+     *
+     * \note Subclasses should not override the implementation of this function
```
boyceg: Please remove this comment --- the method is not virtual, so cannot be over-ridden in the subclass.

boyceg:
This looks fine. Please update the comments and see if we can get Jenkins to recheck.

boyceg:
Jenkins retest this please.

boyceg:
Jenkins retest this please

boyceg:
I went ahead and updated the comments, clang-formatted, and squashed.

ibtk/src/utilities/HierarchyIntegrator.cpp
```diff
@@ -263,6 +263,12 @@ HierarchyIntegrator::initializePatchHierarchy(Pointer<PatchHierarchy<NDIM> > hie
             done = !d_hierarchy->finerLevelExists(level_number);
             ++level_number;
         }
+
+        // Initialize composite hierarchy data that was not initailized by
+        // gridding algorithm's call to initializeLevelData().
+        initializeCompositeHierarchyData(d_integrator_time, initial_time);
+
```
amneetb: I think we have removed this call from here in MF branch. The reason is that it works fine is the data initialization is not dependent upon lag data. Some initiialization of composite hier may require lag data also

ibtk/src/utilities/HierarchyIntegrator.cpp
```diff
@@ -501,15 +486,15 @@ HierarchyIntegrator::regridHierarchy()
...
-                          << "\n"
+                          << "    old volume = " << old_volume << "\n"
+                          << "    new volume = " << new_volume << "\n"
                           << "  this may indicate overlapping patches in the AMR grid hierarchy.");
     }
 
+    // Reinitialize composite grid data.
+    const bool initial_time = false;
+    initializeCompositeHierarchyData(d_integrator_time, initial_time);
+
```
amneetb: And also from here. The top integrator makes this call explicitly.

ibtk/src/utilities/HierarchyIntegrator.cpp
boyceg: This function should be called automatically by HierarchyIntegrator somewhere, so that implementations do not need to add that call explicitly.

ibtk/src/utilities/HierarchyIntegrator.cpp
boyceg: Same comment as above. I don't want to inject this into all of the implementations.

amneetb: Here is a correct sequence:

https://github.com/IBAMR/IBAMR/blob/amneetb/multiphase-flows/src/IB/IBHierarchyIntegrator.cpp#L403-L427

I would also love to do this automatically, but now sure how.

amneetb: I prefer the name compositeHierarchy(), in which the intent is clear--- fill data only when entire hierarchy has been created. The level-by-level filling is tagging and creating levels on the fly and hierarchy is far from being done.

boyceg: Could you just do the LData initialization in the initializeCompositHierarchyData() class implementation?

boyceg: E.g. IBHierarchyIntegrator::initializeCompositeHierarchyDataSpecialized() does the data redistribution?

amneetb: You mean application specific or IBHierachyIntegrator specific? 

LData is created level-by-level basis, and cells are tagged on the level and next level created. It does not seem like a compositeHierarchyData() problem. You of course are refering to something else that I am not getting, because this is how you designed.

amneetb: One can potentially stick LData-->begin/endRedistribution() into IBHierarchyIntegrator::initializeCompositeHierarchyDataSpecialized(), but I think LData is fundamentally a level-by-level init problem, and not a compositeHierarchyInit problem. The former plays in role in deciding how the hierarchy shapes up, while the later is helpless in that regard.

boyceg: Sorry, I meant Lagrangian data redistribution --- not LData initialization.

Looking at the implementation, it seems like you should be able to put the redistribution in the initializeCompositeHierarchyDataSpecialized() method, and then it will be ready to use for any Eulerian data setting that relies on the updated distributions being available.

amneetb: Hmm, OK. If redistribution is just relying on existing hierarchy then it is OK. I have not looked too carefully into that routine.

boyceg: At the initial time there is only one patch hierarchy to deal with. :-)

It is at later times, when the hierarchy actually changes, that you need to split up the redistribution stuff into two phases.

boyceg: You would need to only do the data redistribution at initial time, since other times the redistribution "straddles" regridding.

amneetb: Yeah, that was my impression too. LData-->begin/endRedistribution can alter the hierarchy itself. The routine initCompositeHierarchy() means the hierarchy is not disrupted. But I guess the IBHierarchyIntegrator::initCompositeHierSpecialized() knows that to do these things in a specific order. 

I can stick begin/end dataRedistribution in the IBHierarchyIntegrator::initCompositeHierSpecialized() method.

boyceg:
Jenkins retest this please

ibtk/src/utilities/HierarchyIntegrator.cpp
boyceg: OK, great.

With that change, I think that the interface set up in this PR should work in the other branch too. I will merge once Jenkins is happy.

amneetb: Well, I am confused. The current sequence now is 

1) HierarchyIntegrator::initPatchHier
2) LData-->begin/end
3) HierarchyIntegrator::InitComposite


Now it would be 
HierarchyIntegrator::initPatchHier-->which calls HierarchyIntegrator::InitCompositeSpecialized()

How would HierarchyIntegrator::InitCompositeSpecialized() look without being recurcive?

boyceg:
> On Feb 8, 2018, at 2:39 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> @amneetb commented on this pull request.
> 
> In ibtk/src/utilities/HierarchyIntegrator.cpp <https://github.com/IBAMR/IBAMR/pull/186#discussion_r166851093>:
> 
> >                            << "  this may indicate overlapping patches in the AMR grid hierarchy.");
>      }
>  
> +    // Reinitialize composite grid data.
> +    const bool initial_time = false;
> +    initializeCompositeHierarchyData(d_integrator_time, initial_time);
> +
> Well, I am confused. The current sequence now is
> 
> HierarchyIntegrator::initPatchHier
> LData-->begin/end
> HierarchyIntegrator::InitComposite
> Now it would be
> HierarchyIntegrator::initPatchHier-->which calls HierarchyIntegrator::InitCompositeSpecialized()
> 
> How would HierarchyIntegrator::InitCompositeSpecialized() look without being recurcive?
> 
I believe that your implementation of initializeCompositeHierarchyData() already calls initializeCompositeHierarchyDataSpecialized(). So just move the redistribution call at initial_time into initializeCompositeHierarchyDataSpecialized().

PR #189, opened 2018-01-10
Added check for PETSc version.

abarret:
Fixes the compile issue for PETSc versions lower than 3.8


abarret:
Didn't realize that function existed...

PR #190, opened 2018-01-12
Wave propagation convective operator

abarret:
Adding in wave propagation convective operators for advection-diffusion solver and staggered Navier-Stokes solver.

They are currently tested and 2nd order for the following examples in 2d and 3d:
adv_diff/ex0
adv_diff/ex1
adv_diff/ex2
navier_stokes/ex0 (both regular and forced flow examples)

@boyceg Should I add in the collocated version for completeness? Also, I'm not sure about naming conventions for Fortran files.


amneetb:
Awesome!

boyceg:
@abarret please do add a collocated version.

Naming convention is not very clear for FORTRAN files, but usually the "package name" will come first (e.g. advect_ for advection-related, adv_diff_ for advection-diffusion, etc.). Then something short but descriptive. I would maybe use: advect_wp_convective_opXX.f.m4 (if it is designed to work with the advection-only solvers, which would be cool!) or adv_diff_wp_convective_opXX.f.m4 if it really only works for the advection-diffusion solver.

boyceg:
Each FORTRAN subroutine within an application needs to have a unique name --- and we are basically using FORTRAN 77 here, so there is no namespacing or anything fancy.

boyceg:
@abarret did you receive a build log from Jenkins after the GCC toolchain build failed?

boyceg:
Git may need to know about your email address for Jenkins to send you the build logs -- see instructions [here](https://help.github.com/articles/setting-your-commit-email-address-in-git/) on how to set up your email address.

abarret:
I did not get an email about the failed build. My email is already set up correctly on both Git and GitHub.

boyceg:
@abarret please use .f.m4 for the FORTRAN kernels --- don't commit the generated .f files to the repo. Please try to strip out the .f files using 'git rebase' and replace them with the original .f.m4 files.

abarret:
@boyceg The Fortran files have been replaced with .f.m4 files, and I've rebased to remove those initial commits with the Fortran files.

@amneetb It turns out the WENO5_interp() function was actually more accurate than the version I was using, so I rewrote the fortran kernel to use the WENO5 function. Adapting WENO5 to be WENOk might be a bit of work, and I'm not sure how I would write it in Fortran. We wouldn't want to calculate interpolation coefficients at each function call, but I'm not sure how to avoid recomputing values in a Fortran routine (i.e. making variables 'static').

boyceg:
> @amneetb <https://github.com/amneetb> It turns out the WENO5_interp() function was actually more accurate than the version I was using, so I rewrote the fortran kernel to use the WENO5 function. Adapting WENO5 to be WENOk might be a bit of work, and I'm not sure how I would write it in Fortran. We wouldn't want to calculate interpolation coefficients at each function call, but I'm not sure how to avoid recomputing values in a Fortran routine (i.e. making variables 'static').
> 

If the only issue is named constants, that is easy to deal with. In Fortran 77, use the parameter keyword. I do not know if it is possible to initialize an array of values in a parameter statement in Fortran 77, but it definitely can be done in Fortran 90 (and it is permissible to use some basic Fortran 90 constructs in IBAMR).

boyceg:
> On Feb 28, 2018, at 11:07 AM, Aaron Barrett <notifications@github.com> wrote:
> 
> I did not get an email about the GCC Toolchain build failing. What was the issue?
> 
I will try to setup Jenkins to mail the apparent perpetrator of errors. Right now the emails just go to me and @deleeke.

It looks like Jenkins is now set up to try to build Doxygen documentation, but the documentation generation failed, and so the build state was also set to "failed". Sorry about that!

deleeke:
I just tried selecting "Send to Developers" AND "Send to Culprits" which from this documentation:
https://plugins.jenkins.io/email-ext

Seems to mean it will email everyone who committed since the last successful build.

Hopefully that works. If not, there may be some issue around some new security settings in Jenkins :man_shrugging:
 
https://stackoverflow.com/questions/45350740/jenkins-email-plugin-doesnt-send-email-to-user-who-broke-the-build

boyceg:
Aaron, this appears ready to merge as soon as you resolve the conflict with Makefile.in.

amneetb:
I tried WAVE_PROP, but it seems to support only ADVECTIVE differencing. Is there a limitation on CONSERVATIVE differencing with WAVE_PROP?

abarret:
It just hasn't been implemented. Doing a conservative differencing involves solving a more complicated Riemann problem, and that seemed like overkill for the problem I was interested in since we weren't looking at a conservation law.

abarret:
@boyceg The conflict is resolved and I've squashed the branch into a single commit.

boyceg:
> On May 23, 2018, at 6:13 PM, Aaron Barrett <notifications@github.com> wrote:
> 
> It just hasn't been implemented. Doing a conservative differencing involves solving a more complicated Riemann problem, and that seemed like overkill for the problem I was interested in since we weren't looking at a conservation law.
> 
Could you add a comment that points to what would need to be done to switch to conservation form?

Also please make sure the comments for the class indicate that only advective form is supported by the current implementation.

PR #197, opened 2018-02-08
disabling gtest for Jenkins CI tests

boyceg:
Our current CI server is getting bogged down with the full gtests. For now, let's just run those in the nightly builds.


boyceg:
jenkins retest this please

PR #198, opened 2018-02-23
mass source/sinks in the solid region for poroelasticity

boyceg:


boyceg:
@scott-heath-richardson, @sharpgao --- this is the branch where I am adding support for IBFE source/sinks.

sharpgao:
Hi Boyce, I started to test the IBFE-div-source branch, there is an error (Segmentation Violation) when spreading the fluid source ot the Eulerian grid.  I used the debug version trying to find out where it is, but no further informaiton provided, but I manage to find out at which point when the error triggered, it possiblely related to 

IBExplicitHierarchyIntegrator::integrateHierarchy()
{ ...
d_ib_method_ops->spreadFluidSource(
            d_q_idx, d_p_phys_bdry_op, getProlongRefineSchedules(d_object_name + "::q"), half_time);
...
}
If setting d_p_phys_bdry_op a NULL pointer, it can run, but not with d_p_phys_bdry_op when the FEDataManager tries to call accumulateFromPhysicalBoundaryData(). 

Maybe you can have a quick solution to this?

Bellow is the log file 
[IBFE.log](https://github.com/IBAMR/IBAMR/files/1815225/IBFE.log)

Another question, I am trying to output the source variable into VIsit, by adding OUTPUT_Q = TRUE, seems the IBIntegrator does not pick up this value. 

I can share the example code I am using, through GitLab will be ok?

Hao

boyceg:
@sharpgao I think this is now fixed in the branch.

(Note that I just pushed a fix and rebased to clean up the history, so you will need to run "git reset origin/boyceg/IBFE-div-source" after pulling to synch up with the updated code.)

I plan to merge once the Jenkins tests complete.

Issue #199, opened 2018-02-23
Restarts broken for non-IB examples


nishantnangia:
If you try restarting a simulation from an example that does not contain an immersed structure (i.e. navier_stokes/ex3), you will get the following error message:

_In the future, it will be necessary to include "Silo" in the list of data writer in order to output Silo data.
.
P=00000:Program abort called in file ``../../../IBAMR/ibtk/lib/../src/lagrangian/LSiloDataWriter.cpp'' at line 2404
P=00000:ERROR MESSAGE: 
P=00000:Restart database corresponding to LSiloDataWriter not found in restart file._

The error can be traced back to a call in AppInitializer:
_0x000000000081d83a in IBTK::AppInitializer::AppInitializer (this=0x10a1a80, argc=<optimized out>, argv=<optimized out>, 
    default_log_file_name=...) at ../../../IBAMR/ibtk/lib/../src/utilities/AppInitializer.cpp:263_ 

i.e. this line of code, which looks like it is intended to be deprecated anyway:
https://github.com/IBAMR/IBAMR/blob/master/ibtk/src/utilities/AppInitializer.cpp#L263

==

I understand the purpose of this -- if one doesn't register the Silo writer for an IB case, then this will take care of it. However given that it breaks restarts for purely Eulerian cases, do you think it's safe to remove it at this point in time?

boyceg:
> On Feb 23, 2018, at 12:15 PM, Nishant Nangia <notifications@github.com> wrote:
> 
> If you try restarting a simulation from an example that does not contain an immersed structure (i.e. navier_stokes/ex3), you will get the following error message:
> 
> In the future, it will be necessary to include "Silo" in the list of data writer in order to output Silo data.
> .
> P=00000:Program abort called in file ``../../../IBAMR/ibtk/lib/../src/lagrangian/LSiloDataWriter.cpp'' at line 2404
> P=00000:ERROR MESSAGE:
> P=00000:Restart database corresponding to LSiloDataWriter not found in restart file.
> 
> The error can be traced back to a call in AppInitializer:
> 0x000000000081d83a in IBTK::AppInitializer::AppInitializer (this=0x10a1a80, argc=, argv=,
> default_log_file_name=...) at ../../../IBAMR/ibtk/lib/../src/utilities/AppInitializer.cpp:263
> 
> i.e. this line of code, which looks like it is intended to be deprecated anyway:
[...]

PR #200, opened 2018-02-24
Removing deprecated Silo calls in AppInitializer and ensuring that IB examples are updated

nishantnangia:
Follow up to issue https://github.com/IBAMR/IBAMR/issues/199


boyceg:
👍

PR #211, opened 2018-04-11
fixed a small bug in IBFEMethod.cpp

sach1:
Velocity initialization functions were using coordinate mapping function variables. Fixed this issue.

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -2659,7 +2659,7 @@ IBFEMethod::initializeVelocity(const unsigned int part)
     const unsigned int U_sys_num = U_system.number();
     NumericVector<double>& U_vec = *U_system.solution;
     VectorValue<double> U;
-    if (!d_coordinate_mapping_fcn_data[part].fcn)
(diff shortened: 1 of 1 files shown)


boyceg:
Jenkins please test this.

boyceg:
@deleeke Jenkins seems to be going a little haywire here. Can you take a look?

deleeke:
I think it is doing this because the committer is not a member of the IBAMR organization, so it does not want to run the code unless one of the admin's verifies the patch. There are 3 PR jobs so it makes sense why there are three posts. I can see if it is some how possible to make it only comment once.

deleeke:
So, the way it is set up now with 3 jobs using GitHub Pull Request builder, I think you are going to get a message for each job when a non whitelisted person makes a PR. To add the person to the whitelist from GitHub, comment 
```
add to whitelist
```
To just test this one PR comment:
```
ok to test
```

To ask that tests be run again, comment:
```
test this please
```

boyceg:
OK, got it! I was worried that Jenkins was going crazy again and did not wait for it to run to close the PR...

PR #213, opened 2018-05-03
include normal vector in surface function API

boyceg:


boyceg:
@cpuelz @rossisimone this PR adds normal vectors to the surface force API. Note that the normal vector is in the reference configuration. If you need the vector in the current configuration, you will need to transform it using Nanson's relation. (Maybe there should be a helper function to do this?)

cpuelz:
if its easiest enough, then it could be nice to have both reference and current vectors to be consistent with the current API for positions...

boyceg:
@cpuelz check this out --- it now includes both reference and current normal vectors. I put them before the deformation gradient FF since somehow they seemed related. What do you think of the interface?

cpuelz:
looks good to me!

PR #220, opened 2018-05-18
added a missing sumReduction 

cpuelz:
this addition is needed to check if we are correctly
counting the quadrature points on the meter meshes.


boyceg:
👍

PR #222, opened 2018-05-20
adding simplified version of IBFEMethod for surface meshes

boyceg:


amneetb:
@boyceg : Is this the ALE approach (option 3)?

boyceg:
Right now, it is just a stripped-down version of IBFEMethod that only handles surface meshes. But it could be a framework for setting up those kinds of methods.

PR #224, opened 2018-05-24
constraint forces between overlapping structures

boyceg:


boyceg:
@cpuelz This branch will provide a way to specify tie constraints between pairs of parts. You just need to call `IBFEMethod::constrainPartOverlap(part_id_1, part_id_2, kappa)` after setting up all of the FE meshes and it should work automatically.

The initial commit doesn't work in parallel (yet!), but should work in serial. See if you can set up an example that has two overlapping, non-conforming meshes, and add forces to each mesh that act to pull them apart. Assuming no bugs (fingers crossed!), you should see that the meshes stick together in the overlap region (as long as the penalty spring constant kappa is big enough).

boyceg:
(I should say that I haven't actually tested the code yet --- there are probably some dopey bugs!)

cpuelz:
@boyceg I'll have a look!

PR #226, opened 2018-05-24
Direct forcing IBFEMethod

amneetb:


amneetb:
@boyceg : Take a look at the API. I will start testing the implementation after the interface design is sorted out.

boyceg:
@amneetb, you may need to rebase off master to get the Jenkins tests to work. We had to change the order of boost and libMesh configure tests (see #228).

amneetb:
@boyceg : rebased

amneetb:
@boyceg : Can we zero-out d_F_half_vecs[part] here?
https://github.com/IBAMR/IBAMR/blob/e4c249b4192b8a6d94693a22a70b9a2ffc0c2fd6/src/IB/IBFEMethod.cpp#L731

This will allow direct forcing object to add to d_F_half_vecs[part], instead of overwriting it. 
https://github.com/IBAMR/IBAMR/blob/e4c249b4192b8a6d94693a22a70b9a2ffc0c2fd6/src/IB/IBFEMethod.cpp#L1031
making it work in both scenarios where there is or there is not any other Lag function registered for a part.

boyceg:
> On May 31, 2018, at 11:22 AM, Amneet Bhalla <notifications@github.com> wrote:
> 
> @boyceg <https://github.com/boyceg> : Can we zero-out d_F_half_vecs[part] here?
> https://github.com/IBAMR/IBAMR/blob/e4c249b4192b8a6d94693a22a70b9a2ffc0c2fd6/src/IB/IBFEMethod.cpp#L731 <https://github.com/IBAMR/IBAMR/blob/e4c249b4192b8a6d94693a22a70b9a2ffc0c2fd6/src/IB/IBFEMethod.cpp#L731>
> This will allow direct forcing object to add to d_F_half_vecs[part], instead of overwriting it.
> https://github.com/IBAMR/IBAMR/blob/e4c249b4192b8a6d94693a22a70b9a2ffc0c2fd6/src/IB/IBFEMethod.cpp#L1031 <https://github.com/IBAMR/IBAMR/blob/e4c249b4192b8a6d94693a22a70b9a2ffc0c2fd6/src/IB/IBFEMethod.cpp#L1031>
> making it work in both scenarios where there is or there is not any other Lag function registered for a part.
> 

What about zeroing out when actually computing the force, to allow all callers to accumulate without worrying about whether the thing is initialized or not?

amneetb:
Yes, that works too. So something like VecZeroEnteries(d_F_half_vecs[part]->vec()) before this line?
https://github.com/IBAMR/IBAMR/blob/e4c249b4192b8a6d94693a22a70b9a2ffc0c2fd6/src/IB/IBFEMethod.cpp#L1028

amneetb:
@boyceg : I think I am mostly done with adding examples for this branch. You can review it and send me your comments.

src/IB/IBFEMethod.cpp
```diff
@@ -980,6 +1048,12 @@ IBFEMethod::forwardEulerStep(const double current_time, const double new_time)
         IBTK_CHKERRQ(ierr);
         d_X_new_vecs[part]->close();
         d_X_half_vecs[part]->close();
+
+        if (d_direct_forcing_kinematics_data[part])
```
boyceg: Do we want to do the normal update followed by the direct-forcing update, or should it be an if statement --- if there are specified kinematics, use those kinematics; otherwise, use the interpolated velocity?

src/IB/IBFEMethod.cpp
```diff
@@ -1005,6 +1079,11 @@ IBFEMethod::midpointStep(const double current_time, const double new_time)
         IBTK_CHKERRQ(ierr);
         d_X_new_vecs[part]->close();
         d_X_half_vecs[part]->close();
+        if (d_direct_forcing_kinematics_data[part])
```
boyceg: Same comment as above...

src/IB/IBFEMethod.cpp
```diff
@@ -1033,6 +1112,11 @@ IBFEMethod::trapezoidalStep(const double current_time, const double new_time)
...
+        {
+            d_direct_forcing_kinematics_data[part]->trapezoidalStep(
```
boyceg: Same comment as above...

src/IB/IBFEMethod.cpp
```diff
@@ -1054,7 +1138,14 @@ IBFEMethod::computeLagrangianForce(const double data_time)
         {
             computeStressNormalization(*d_Phi_half_vecs[part], *d_X_half_vecs[part], data_time, part);
         }
+        // Zero-out force vector
+        d_F_half_vecs[part]->zero();
         computeInteriorForceDensity(*d_F_half_vecs[part], *d_X_half_vecs[part], d_Phi_half_vecs[part], data_time, part);
+        if (d_direct_forcing_kinematics_data[part])
+        {
+            d_direct_forcing_kinematics_data[part]->computeLagrangianForce(
```
boyceg: Should these forces should go into a zero_clone() of the force vector, and then get added to the force vector owned by this class. That way implementations of the kinematics class don't have to remember to accumulate.

src/IB/IBFEMethod.cpp
```diff
@@ -1054,7 +1138,14 @@ IBFEMethod::computeLagrangianForce(const double data_time)
         {
             computeStressNormalization(*d_Phi_half_vecs[part], *d_X_half_vecs[part], data_time, part);
         }
+        // Zero-out force vector
+        d_F_half_vecs[part]->zero();
```
boyceg: Does this need to be zero-ed out here?

boyceg:
Please drop the binary file (in a rebase) --- I would like to try to avoid adding new binary files to the repo.

src/IB/IBFEMethod.cpp
amneetb: I think former is more flexible in the sense that implementation of direct-forcing can skip the position update if they want and just use what IBFEMethod does. But I am fine with the other way too. Let me know.

src/IB/IBFEMethod.cpp
amneetb: That also sounds good. I can change it. Let me know.

src/IB/IBFEMethod.cpp
amneetb: That seemed to me the local place to make this change. Other place is in the preprocessIntegrateHierarchy(), but that seemed non-local spot for force calculation. What do you think?

amneetb:
I added .xda file which is ASCII file. I think .xdr is the binary format. I can also remove .xda file if you want. Let me know.

amneetb:
@boyceg : I have removed the xda file and have merged master into it. If I rebase with master, there are plenty of conflicts in the history because both branches had their own examples/IBFE/explicit/ex9. I have moved ex9 to ex10 in this PR before merging. I also merged the PR with master locally and it looks clean. Would plain merge with master not work?

src/IB/IBFEMethod.cpp
boyceg: Yes please do make this change.

src/IB/IBFEMethod.cpp
boyceg: I think that computeInteriorForceDensity is already overwriting the F vector; in that case, the call to zero() is not needed.

src/IB/IBFEMethod.cpp
boyceg: OK with me.

PR #231, opened 2018-05-28
fixing ghosting for overlap constraint forces

boyceg:


boyceg:
@cpuelz, this should fix the problems you were running into with the heart model with overlap constraints enabled.

PR #232, opened 2018-05-28
reorganizing interpolation and spreading to attempt to improve parallel performance

boyceg:


boyceg:
@cpuelz I think you will see some performance improvements with models with lots of parts with this branch --- can you give it at try?

cpuelz:
@boyceg  sounds great... I'll fire off a simulation.

PR #233, opened 2018-05-28
sort the node DOF ids before setting up meter meshes

boyceg:


boyceg:
@cpuelz, this sorts the node IDs before setting up the meters. Please take a look. I haven't added anything to skip renumbering the nodes --- if it is important to avoid renumbering nodes, then please do add the needed function calls.

boyceg:
Also, it looks like you could save a node by repeating the first node in the ring. Can you make that change? It might be easiest to put the centroid last, and to index the nodes around the ring using constructs like `(idx + 1) % num_ring_nodes`.

cpuelz:
@boyceg this looks good.  Do you get an assertions thrown with this code?

I am not sure what you mean by repeating the first node in the ring.  I am actually doing that I think, and in this version of the meter code, I don't actually use the centroid as part of the meter mesh construction.
I'll do another PR to switch to a prettier looking meter mesh with a centroid node.

boyceg:
No assertions with the new heart model.

I think I am getting confused by the indexing in setting up the meters. I think there should be be the number of nodes in the nodeset + one for the centroid?

cpuelz:
These meshes are the "sea shell" like meshes, so they have no interior nodes, just the nodes that are given by the nodesets from the heart model parts.  in particular these meshes do not have the centroid node.

so in this case, the number of nodes = size of the nodeset, and the number of elements = number of nodes - 1.

cpuelz:
sorry, the number of elements in this version should be num nodes - 2.

PR #238, opened 2018-05-30
track maximum displacements in overlap regions

boyceg:


boyceg:
@cpuelz, this code (now in master) prints out some information about the maximum displacement between constrained parts. Give it a try and see if it tells you anything useful about the overlap constraints for the new heart model.

PR #240, opened 2018-05-31
add surface velocity jumps

boyceg:


boyceg:
@AminKolah, this is still WIP code. I am replacing separate U, V, and W data structures with fixed-sized arrays of size NDIM along with loops over dimension.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -150,6 +151,13 @@ libmesh_restart_file_name(const std::string& restart_dump_dirname,
                      << std::setfill('0') << std::right << time_step_number << "." << extension;
     return file_name_prefix.str();
 }
+
+inline int
```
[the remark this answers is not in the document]
AminKolah: sounds like oost::math::iround does the same thing

AminKolah:
@boyceg I noticed when imposing the pressure jump you don't do it in two steps anymore i.e. 1) collecting the intersection points 2) looping over the points to apply at the stencil...Is there a big reason why this has been changed or in another word should I strictly follow the same procedure for the velocity?

boyceg:
Yes there is a reason not to do it in two steps. There appeared to be almost no benefit to collecting the intersection points and then imposing the jump conditions, and doing both at the same time simplified the implementation substantially.

(The original implementation was based on similar code for spreading forces using regularized delta functions, and in that case, it is actually simpler to collect all of the quadrature point positions and forces, and then to spread the forces all at the same time.)

boyceg:
Notice that in the current implementation, what happens is that the code does keep track of previous intersections. However, it does so only to detect duplicated intersections (e.g. intersections along shared edges of elements). The code tries to ensure that duplicated intersections are skipped if they have the same orientation (here "orientation" means "dot product of the intersection axis and the surface normal"). The only case where duplicated intersections are both used is when the orientation changes --- this corresponds to a tangential intersection with a surface, in which case there should be no jump. By including both of the duplicated jumps in this case, the code actually ensures that the jump at the tangential intersection is zero.

AminKolah:
@boyceg This is not so crucial but I recommend using d_P_j_systems instead of d_DP_systems as the jump in the pressure currently account for the value of P itself. Similarly for the jump in the first derivative of the velocity we use d_DU_j_systems which is consistent with the above nomenclature

boyceg:
The other thing the code now does is to perturb the positions of the nodes of the elements in cases where the nodes are close to locations that could be "cut" by a ray that we are using to search for intersections. This avoids having to figure out what to do about intersections that are detected at the nodes, which are tricky to deal with in 3D (since you could have an arbitrary number of elements joining at a common node).

boyceg:
@AminKolah are you able to work some more on updating the code to account for the changes that I've made in the names/types of the jump data structures? I am probably not going to be able to work on this today.

AminKolah:
@boyceg Sure I'll make those changes today along with adding the velocity jump.

AminKolah:
I'm also hoping at the end of the day the collection of the these improvement resolve some of the issues we had with parallel (MPI) cases.

boyceg:
The other thing that would be good to do would be to avoid having essentially duplicated code for the u/v/w jump computations. I am pretty sure that many (or even all) of those can be replaced by loops over the axis index, using the % operator to get the other axes. E.g. if `axis` is the axis aligned with the velocity, the tangent direction axes can be computed as `(axis + 1) % NDIM` and `(axis + 2) % NDIM`.

The goal here is to reduce redundant code.

boyceg:
While you are at it, if you want to update the names of the systems to P_j and DU_j, that is OK with me.

boyceg:
BTW, the pressure jump code seems to work correctly in parallel, so just try to follow their implementation.

amneetb:
@boyceg : Is the intersection code general enough to compute distances of cell centers to surface elements? I need to initialize LS data using boundary mesh.

boyceg:
> @boyceg <https://github.com/boyceg> : Is the intersection code general enough to compute distances of cell centers to surface elements? I need to initialize LS data using boundary mesh.
> 
The intersection code is specialized to looking for intersections between rays that are in the coordinate directions and elements. It allows for "fuzzy" detection of intersections (to try to avoid missing elements in cases where the ray cuts through an element boundary).

Also, in 3D, intersections are currently only implemented for TRI3 and QUAD4 elements. The 2D implementation supports EDGE2 and EDGE3 elements.

boyceg:
@AminKolah, you will want to merge or rebase off IBAMR master to get the fixed triangle intersection code.

To rebase, do: `git fetch`, `get checkout master`, `git pull`, `git checkout boyceg/add-surface-velocity-jumps`, `git pull`, `git rebase master`.

boyceg:
I am removing all of the mask_var stuff; it is not being used as a "mask" variable.

AminKolah:
That's fine with me.I,  however,  would probably have to re-intrduce them later for calculation of traction.

boyceg:
> That's fine with me.I, however, would probably have to re-intrduce them later for calculation of traction.
> 
There should be no need. You can just use the velocity data supplied to the function.

AminKolah:
What about the pressure? ..and does the velocity profile have enough number of ghost cells to calculate derivatives

boyceg:
> On Jun 11, 2018, at 4:01 PM, Ebrahim Kolahdouz <notifications@github.com> wrote:
> 
> What about the pressure? ..and does the velocity profile have enough number of ghost cells to calculate derivatives
> 

If we need the pressure then we can update the interface.

You can control the number of ghost cells through the ghost cell width member variable.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -640,48 +779,389 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+
+                std::vector<int> local_indices;
+                local_indices.clear();
+                const unsigned int upper_bound = n_qp_patch;
+                if (upper_bound == 0) return;
+                local_indices.reserve(upper_bound);
+
+                for (unsigned int k = 0; k < upper_bound; ++k)
+                {
+                    const double* const XX = &x_qp[NDIM * k];
```
boyceg: Please don't mix uppercase X and lower case x --- they have different meanings. (X indicates reference coordinates, x indicates current coordinates.)

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -640,48 +779,389 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+            // Note: Values are interpolated only to those quadrature points that
+            // are within the patch interior
+            const Box<NDIM>& patch_box = patch->getBox();
+            Pointer<PatchData<NDIM> > u_data = patch->getPatchData(side_mask_scratch_idx);
             Pointer<CellData<NDIM, double> > u_cc_data = u_data;
             if (u_cc_data)
             {
                 LEInteractor::interpolate(
-                    U_qp, NDIM, x_qp, NDIM, u_cc_data, patch, interp_box, d_default_interp_spec.kernel_fcn);
+                    U_qp, NDIM, x_qp, NDIM, u_cc_data, patch, patch_box, d_default_interp_spec.kernel_fcn);
```
boyceg: Shouldn't this case result in an error message if velocity jumps are imposed in interpolation?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -640,48 +779,389 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+                    n = (dx_dxi[0].cross(dx_dxi[1])).unit();
+
+                    for (unsigned int i = 0; i < NDIM; ++i)
+                    {
+                        for (unsigned int k = 0; k < n_node; ++k)
                         {
-                            x_qp[NDIM * (qp_offset + qp) + d] += x_node[k][d] * p;
+                            x_qp[NDIM * (qp_offset + qp) + i] += x_node[k][i] * phi[k][qp];
                         }
+                        N_qp[NDIM * (qp_offset + qp) + i] = n(i);
```
boyceg: Please don't mix uppercase N and lower case n --- they have different meanings. (N indicates reference coordinates, n indicates current coordinates.)

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -640,48 +779,389 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
                     }
+
+                    for (unsigned int d = 0; d < NDIM; ++d)
```
boyceg: Let's use `axis` instead of `d` for these kinds of loops.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -640,48 +779,389 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+                                           [range(0, u_depth)]),
+                            boost::fortran_storage_order());
+
+                        for (unsigned int k = 0; k < nindices; ++k)
+                        {
+                            const int s = local_indices[k];
+
+                            for (int d = 0; d < NDIM; ++d)
+                            {
+                                X_shifted[d] = x_qp[d + s * NDIM] + periodic_shifts[d + k * NDIM];
```
boyceg: Please don't mix "X" and "x".

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -640,48 +779,389 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+
+                            boost::multi_array<double, NDIM + 1> Ujump(
+                                boost::extents[range(ic_trimmed_lower[0], ic_trimmed_upper[0] + 1)]
+                                              [range(ic_trimmed_lower[1], ic_trimmed_upper[1] + 1)]
+#if (NDIM == 3)
+                                              [range(ic_trimmed_lower[2], ic_trimmed_upper[2] + 1)]
+#endif
+                                              [range(0, NDIM)]);
+
+#if (NDIM == 2)
```
boyceg: This needs some comments.

Also, could you avoid separate LL/LLL data structures by indexing on NodeIndex?

src/IB/IBFESurfaceMethod.cpp
AminKolah: Currently the velocity jump conditions are only setup for staggered grid. The jumps for cell-centered (cc) velocity need to be modified a little bit and can be added later. The current LEInteractor for velocity interpolation ignore the kink but shouldn't result in an error message, or should it?

src/IB/IBFESurfaceMethod.cpp
AminKolah: Okay I see what you mean. I'll fix this

src/IB/IBFESurfaceMethod.cpp
AminKolah: I knew you're gonna say that:D ...okay I'll make it one index

boyceg:
Jenkins retest this please.

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -110,7 +110,16 @@ class IBFESurfaceMethod : public IBStrategy
...
     static const std::string TANGENTIAL_VELOCITY_SYSTEM_NAME;
+    static const boost::array<std::string, NDIM> VELOCITY_JUMP_SYSTEM_NAME;
     static const std::string VELOCITY_SYSTEM_NAME;
+    static const std::string WSS_O_SYSTEM_NAME;
+    static const std::string WSS_I_SYSTEM_NAME;
+    static const std::string P_O_SYSTEM_NAME;
+    static const std::string P_I_SYSTEM_NAME;
+    static const std::string TAU_SYSTEM_NAME;
+
+    SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > mask_var;
```
boyceg: Can this masking variable be removed from `IBFESurfaceMethod`?

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -300,6 +309,16 @@ class IBFESurfaceMethod : public IBStrategy
         const std::vector<SAMRAI::tbox::Pointer<SAMRAI::xfer::CoarsenSchedule<NDIM> > >& u_synch_scheds,
         const std::vector<SAMRAI::tbox::Pointer<SAMRAI::xfer::RefineSchedule<NDIM> > >& u_ghost_fill_scheds,
         double data_time);
+    /*!
+     * Compute the exterior fluid traction for when using the
```
boyceg: Should we make it easy to do either interior or exterior tractions?

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -300,6 +309,16 @@ class IBFESurfaceMethod : public IBStrategy
...
+     * jump conditions
+     */
+    void computeFluidTraction(double current_time, unsigned int part = 0);
+    /*!
+     * Compute the exterior pressure used in the calculation
```
boyceg: Would it be better to interpolate the pressure in one function and the viscous stress in a separate functions? Sometimes we want only one or the other, right?

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -521,13 +547,23 @@ class IBFESurfaceMethod : public IBStrategy
     const unsigned int d_num_parts;
     std::vector<IBTK::FEDataManager*> d_fe_data_managers;
     SAMRAI::hier::IntVector<NDIM> d_ghosts;
-    std::vector<libMesh::System*> d_X_systems, d_U_systems, d_U_n_systems, d_U_t_systems, d_F_systems, d_DP_systems;
-    std::vector<libMesh::PetscVector<double>*> d_X_current_vecs, d_X_new_vecs, d_X_half_vecs, d_X0_vecs, d_X_IB_ghost_vecs;
+    std::vector<libMesh::System*> d_X_systems, d_U_systems, d_U_n_systems, d_U_t_systems, d_F_systems, d_WSS_o_systems,
```
boyceg: Can you add some comments about these different systems?

@drwells should we make some data structure for packaging together `System`s and `PetscVector`s. This way of doing things is my fault, but it seems silly that if we want to include a new variable here, we repeat the name 2 or 3 times (once for the system, and then once for each of the DOF vectors that we explicitly track).

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -159,7 +159,21 @@ const std::string IBFESurfaceMethod::COORD_MAPPING_SYSTEM_NAME = "IB coordinate
 const std::string IBFESurfaceMethod::FORCE_SYSTEM_NAME = "IB force system";
 const std::string IBFESurfaceMethod::NORMAL_VELOCITY_SYSTEM_NAME = "IB normal velocity system";
 const std::string IBFESurfaceMethod::PRESSURE_JUMP_SYSTEM_NAME = "IB [[p]] system";
+const std::string IBFESurfaceMethod::WSS_O_SYSTEM_NAME = "One sided exterior shear stress system";
+const std::string IBFESurfaceMethod::WSS_I_SYSTEM_NAME = "One sided interior shear stress system";
+const std::string IBFESurfaceMethod::P_O_SYSTEM_NAME = "One sided exterior pressure system";
+const std::string IBFESurfaceMethod::P_I_SYSTEM_NAME = "One sided interior pressure system";
+const std::string IBFESurfaceMethod::TAU_SYSTEM_NAME = "traction system";
 const std::string IBFESurfaceMethod::TANGENTIAL_VELOCITY_SYSTEM_NAME = "IB tangential velocity system";
+const boost::array<std::string, NDIM> IBFESurfaceMethod::VELOCITY_JUMP_SYSTEM_NAME = {
```
boyceg: 👍

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -327,9 +341,33 @@ IBFESurfaceMethod::preprocessIntegrateData(double current_time, double new_time,
     d_F_half_vecs.resize(d_num_parts);
     d_F_IB_ghost_vecs.resize(d_num_parts);
 
-    d_DP_systems.resize(d_num_parts);
-    d_DP_half_vecs.resize(d_num_parts);
-    d_DP_IB_ghost_vecs.resize(d_num_parts);
+    d_P_j_systems.resize(d_num_parts);
```
boyceg: What does `j` stand for? (I understand `i` and `o`.)

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -370,13 +408,60 @@ IBFESurfaceMethod::preprocessIntegrateData(double current_time, double new_time,
         d_F_IB_ghost_vecs[part] = dynamic_cast<PetscVector<double>*>(
             d_fe_data_managers[part]->buildGhostedSolutionVector(FORCE_SYSTEM_NAME, /*localize_data*/ false));
 
-        if (d_use_jump_conditions)
+        if (d_use_pressure_jump_conditions)
         {
-            d_DP_systems[part] = &d_equation_systems[part]->get_system(PRESSURE_JUMP_SYSTEM_NAME);
-            d_DP_half_vecs[part] = dynamic_cast<PetscVector<double>*>(d_DP_systems[part]->current_local_solution.get());
-            d_DP_IB_ghost_vecs[part] =
+            d_P_j_systems[part] = &d_equation_systems[part]->get_system(PRESSURE_JUMP_SYSTEM_NAME);
```
boyceg: @drwells, if we had a data structure that packed together the `System` and `NumericVector`s, maybe we could avoid some of this boilerplate. What do you think?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -404,10 +489,29 @@ IBFESurfaceMethod::preprocessIntegrateData(double current_time, double new_time,
         d_F_systems[part]->solution->close();
         d_F_systems[part]->solution->localize(*d_F_half_vecs[part]);
 
-        if (d_use_jump_conditions)
+        if (d_use_pressure_jump_conditions)
+        {
+            d_P_j_systems[part]->solution->close();
```
boyceg: And also here...

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -672,19 +1037,274 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+                        xo, x_lower_ghost, x_upper_ghost, patch_geom->getDx(), ghost_box.lower(), ghost_box.upper());
+
+                    // Some kind of assertation can be applied here using the indices of the cells away from the interfce
+
+                }
+                if (local_indices.empty()) continue;
+                Index<NDIM> ic_lower, ic_upper, ic_center;
+                boost::array<boost::array<double, 2>, NDIM> w, wr;
+#if (NDIM == 2)
+                boost::array<double, NDIM> LL, LU, UL, UU;
```
boyceg: It might clean up some code lower down to combine these into a single `boost::multi_array` that is indexed for L/U, e.g. using L=`0` and U=`1`.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -672,19 +1037,274 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+                            boost::extents[range(ic_lower[0], ic_upper[0] + 1)][range(ic_lower[1], ic_upper[1] + 1)]
+#if (NDIM == 3)
+                                          [range(ic_lower[2], ic_upper[2] + 1)]
+#endif
+                                          [range(0, NDIM)]);
+
+#if (NDIM == 2)
+                        for (int d = 0; d < NDIM; ++d)
+                        {
+                            LL[d] = (n_qp[s * NDIM] * wr[0][1] + n_qp[s * NDIM + 1] * wr[1][1]) * n_qp[s * NDIM + d];
```
boyceg: If you used a `multi_array` here instead of an `array`, I think that you could replace all of these nearly duplicated statements with loops over sides (L/U) along with the modulo operator.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -672,19 +1037,274 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+                            LU[d] = (n_qp[s * NDIM] * wr[0][1] + n_qp[s * NDIM + 1] * wr[1][0]) * n_qp[s * NDIM + d];
+                            UU[d] = (n_qp[s * NDIM] * wr[0][0] + n_qp[s * NDIM + 1] * wr[1][0]) * n_qp[s * NDIM + d];
+                            UL[d] = (n_qp[s * NDIM] * wr[0][0] + n_qp[s * NDIM + 1] * wr[1][1]) * n_qp[s * NDIM + d];
+                        }
+                        for (int d = 0; d < NDIM; ++d)
+                        {
+                            Ujump[ic_lower[0]][ic_lower[1]][d] =
```
boyceg: Similar story here, I think.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -672,19 +1037,274 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+                            UUU[d] = (n_qp[s * NDIM] * wr[0][0] + n_qp[s * NDIM + 1] * wr[1][0] +
+                                      n_qp[s * NDIM + 2] * wr[2][0]) *
+                                     n_qp[s * NDIM + d];
+                            LUU[d] = (n_qp[s * NDIM] * wr[0][1] + n_qp[s * NDIM + 1] * wr[1][0] +
+                                      n_qp[s * NDIM + 2] * wr[2][0]) *
+                                     n_qp[s * NDIM + d];
+                        }
+                        for (int d = 0; d < NDIM; ++d)
+                        {
+                            Ujump[ic_lower[0]][ic_lower[1]][ic_lower[2]][d] =
```
boyceg: If possible, I would like these very complicated looking statements to be transformed into simpler loops. This kind of code can be tough to maintain, and I don't think we need the performance that can come with manually unrolling as done here.

boyceg:
@drwells, @abarret, this is a branch that has some important new functionality that we would like to get merged into `master`. I think that most of the important changes are in `IBFESurfaceMethod`. Amin may need some help in getting this done, because a lot of the changes predate the C++11 modernization.

include/ibamr/IBFESurfaceMethod.h
AminKolah: I think so. I used in the past but it's not needed any more in the current code.

include/ibamr/IBFESurfaceMethod.h
AminKolah: I actually used to only compute the exterior traction but depends on which side the interesting stuff is happening either the interior or exterior calculation becomes important. Also at least for the sake of convergence study it turns out that numerically du/dn^- is not exactly equal to -([[du/dn]] - du/dn^+) does that make sense?....however maybe we can just keep one variable and a flag to to switch the normal to determine which side needs to be calculated for a specific problem?

include/ibamr/IBFESurfaceMethod.h
boyceg: OK, let's remove it.

include/ibamr/IBFESurfaceMethod.h
AminKolah: So currently if the pressure jump is on, then the pressure on the structure is automatically calculated via interpolatePressureForTraction. Similarly if velocity jump is on, the WSS is computed automatically too (in the interpolateVelocity). There is a separate flag in the input file that asks to whether to compute the traction or not via computeFluidTraction

include/ibamr/IBFESurfaceMethod.h
boyceg: I think that sometimes we will want one or the other?

I _think_ it makes sense that interpolating from one side is not exactly the same as interpolating from the other side and then adding the jump. That is why I think it would be good to be able to do either. :-)

AminKolah: That's right. There were some code exploring the options for taking the mean value of the two or other ideas, but I removed them. The example I used to explore this was the falling sphere (to see which option results in smoother traction on the surface). However I think I can come up with a simpler RBD example to check this maybe, like a 2D disk which runs much faster. The main goal is to have both smoother (help with stability) and more accurate traction values.

src/IB/IBFESurfaceMethod.cpp
AminKolah: so j is the jump. i and o are interior and exterior? If it's not intuitive enough we can change it to something else?

boyceg: Let me think. `j` makes me think of an index.

src/IB/IBFESurfaceMethod.cpp
AminKolah: Okay I can do this.

src/IB/IBFESurfaceMethod.cpp
AminKolah: I see what you mean. It can be a lot more compact that way...I'll take a look.

src/IB/IBFESurfaceMethod.cpp
drwells: That refactoring has been on my todo list for awhile now. I'll take a look once we get some of the IB ghost data changed (another upcoming PR).

drwells:
I am looking at this now.

I have a small suggestion regarding git: I recommend updating branches with `master` via rebasing, instead of merging, so that commits on this branch are on top of `master` and not interleaved with it. Doing this makes all commits from this branch contiguous in the history, which helps bisecting (and makes the branch visualization a lot easier to understand).

src/IB/IBFESurfaceMethod.cpp
drwells: Yes. If the loop bounds are given to the compiler then it will unroll when appropriate.

boyceg: This is not about efficiency --- these should not be dominate computational costs, because they are happening on an `NDIM-1` surface in an `NDIM` computation. What I want to try to accomplish is to enhance the readability of the implementation.

Towards that goal, @AminKolah, it would be helpful if you are doing things in "normal" and "tangential" directions (or looking at "normal" and "tangential" indices, etc.) that you use variable names that include that information --- e.g. i_norm and j_tan.

For indices, it would be fairly natural to do something like:
```
    int i_norm = ... ;
    vector<int> j_tan = (NDIM == 2 ? { (i_norm + 1) % NDIM } : { (i_norm + 1) % NDIM , (i_norm + 2) % NDIM});
```

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -513,6 +717,7 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
         NumericVector<double>* U_t_vec = NULL;
         NumericVector<double>* X_vec = NULL;
         NumericVector<double>* X_ghost_vec = d_X_IB_ghost_vecs[part];
+        boost::array<PetscVector<double>*, NDIM> DU_j_ghost_vec = d_DU_j_IB_ghost_vecs[part];
```
drwells: Here and elsewhere please use `std::array`: its the same thing, but now that we have C++11 we can use the 'real' class.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -557,32 +765,76 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
         UniquePtr<FEBase> fe = FEBase::build(dim, fe_type);
         const std::vector<double>& JxW = fe->get_JxW();
         const std::vector<std::vector<double> >& phi = fe->get_phi();
         boost::array<const std::vector<std::vector<double> >*, NDIM - 1> dphi_dxi;
         dphi_dxi[0] = &fe->get_dphidxi();
         if (NDIM > 2) dphi_dxi[1] = &fe->get_dphideta();
+        
+        
+        FEType fe2_type = DU_j_fe_type;
+        UniquePtr<FEBase> fe2 = FEBase::build(dim, fe2_type);
```
drwells: Also cosmetic: we should use `std::unique_ptr` now.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -769,6 +1416,7 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
     return;
 } // interpolateVelocity
```
drwells: It was already a problem, but in `IBFEMethod` we moved the vector communication and projection outside of the loop over the parts. Its necessary to do this in phases so that we don't have severe load imbalances when running with multiple parts and multiple MPI processes.

We will have to fix this in a future PR.

src/IB/IBFESurfaceMethod.cpp
AminKolah: Okay I just started working on turning these into simple/more compact loops in the interpolateVelocity function. Would you also recommend using things like inner_product instead of component by component multiplication?

boyceg: Definitely!

If you use either libMesh or Eigen vectors, you should be able to do an inner product using `*`. My goal in requesting these changes is to get the expressions to look more like "math". :-)

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1388,162 +2289,1758 @@ void IBFESurfaceMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*h
...
+    d_X_new_vecs[part]->close();
+
+    d_P_o_half_vecs[part]->close();
+    d_P_i_half_vecs[part]->close();
+    d_P_j_IB_ghost_vecs[part]->close();
+    d_X_IB_ghost_vecs[part]->close();
+
+    return;
+
+} // interpolatePressureForTraction
```
drwells: This function has the same logic everywhere for interior and exterior tractions. Could we convert some of this code to loops over interior and exterior data, instead of writing nearly the same thing for each traction?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1388,162 +2289,1758 @@ void IBFESurfaceMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*h
...
+    DU_j_ghost_vec = d_DU_j_IB_ghost_vecs[part];
+    DU_j_vec = d_DU_j_half_vecs[part];
+    for (unsigned int d = 0; d < NDIM; ++d)
+    {
+        DU_j_vec[d]->localize(*DU_j_ghost_vec[d]);
+    }
+    for (unsigned int d = 0; d < NDIM; ++d)
+    {
+        DU_j_vec[d]->localize(*DU_j_ghost_vec[d]);
+    }
```
drwells: Like elsewhere, localization should not be necessary. Explicitly update the ghost region.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1388,162 +2289,1758 @@ void IBFESurfaceMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*h
...
+#if (NDIM == 3)
+                        intersect_line_with_face(intersectionsSide[l], static_cast<Face*>(elem), rs[l], q, tolerance);
+#endif
+                    }
+
+                    if (d_use_pressure_jump_conditions)
+                    {
+                        for (unsigned int k = 0; k < intersections.size(); ++k)
+                        {
+                            const libMesh::Point x = r + intersections[k].first * q;
```
drwells: I think we could move this intersection calculation into an exterior function.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1648,127 +4170,609 @@ IBFESurfaceMethod::imposeJumpConditions(const int f_data_idx,
...
+                                // Make sure we haven't already found this
+                                // intersection.
+                                //
+                                // (Because we are doing this in floating point
+                                // arithmetic, we can't even count on the
+                                // intersection being assigned to the same index!)
+                                bool found_same_intersection_point = false;
+                                for (int shift = -1; shift <= 1; ++shift)
+                                {
+                                    SideIndex<NDIM> i_s_prime = i_s;
```
drwells: I think this is nearly the same, or the same, intersection code as above. It should be in a separate function if possible.

include/ibamr/IBFESurfaceMethod.h
AminKolah: I just realized I am actually using  mask_scratch_idx to synchronize the ghost data for the pressure at the IB.

https://github.com/IBAMR/IBAMR/blob/4d8bcdb9a61d8b84ecd8fd3d85331dbbd40e261a/src/IB/IBFESurfaceMethod.cpp#L533-L551

boyceg: OK, well make this an internal class variable and give it a name that corresponds to its use.

boyceg: (I should have said `private` rather than internal.)

src/IB/IBFESurfaceMethod.cpp
drwells: (well, almost everywhere!)

include/ibamr/IBFESurfaceMethod.h
AminKolah: sounds good.

src/IB/IBFESurfaceMethod.cpp
AminKolah: That's true. I thought something can be done here or whenever we compute quantities on both sides of the IB. Mostly it seems all the computations are the same except for the sign of the normal vector that switches between the two....I'll work on fixing this. Thanks for the suggestion David.

src/IB/IBFESurfaceMethod.cpp
AminKolah: I might actually decide to remove imposePointWiseJumpConditions all together. It was just for comparing with the weak version of the jump quantities. @boyceg What do you think? 

Also I don't think anything has changed in imposeWeakJumpConditions since the last merge into master, but we can still reduce the number lines or create separate functions like the one you've mentioned here.

boyceg: I would remove any un-needed code.

I also would move intersection related code into helper functions, because we will be wanting to find intersections in other places, too.

boyceg: A good rule of thumb (not always followed by me!) is: if you have a nontrivial piece of code and use it more than once, put it in a helper function.

drwells: well, if it works, I see no reason to get rid of it.

I would still like to move the intersection algorithm out since its probably useful in other parts of IBAMR.

src/IB/IBFESurfaceMethod.cpp
abarret: There's a perl script in the scripts directory that will replace UniquePtr with unique_ptr.

AminKolah:
@boyceg  and @drwells I spent sometime on Sunday and today morning and tried to address all (or most?) of the stuff we discussed here. I also tested the code for a couple of examples and things seem to be working as expected. Do you want me create another pull request to discuss the current code or you would like to continue the merge process through the current issue?

AminKolah:
@abarret if you get a chance, it'd be nice if you could take a look too, especially the SAMRAI (index) related changes in interpolateVelocity function

AminKolah:
A few comments on new changes:

- Removed all the calls for closing/localizing of solution vectors., also manually added a call to `batch_vec_ghost_update` in `postprocessIntegrateData` similar to what's done in _IBFEMethod.cpp_ on `master` branch. I suppose this function performs the ghost update for all parts? 

- Created loops by introducing some boost `array`/`multi_array`  and cut down the code in `interpolateVelocity` by huge amount, also tried to use inner product whenever possible. I'm hoping it looks more neat and  simplified now and easier to decipher.  

- Tried to add more annotations to the code for instance when introducing new variables in the .h file

- Got rid of some redundant code related to computation of both and inner an outer pressure/wss. Currently the user determines if the values from the interior side needs to be calculated by the flag `d_calculate_interior_side`. By default pressure and wall shear stress are computed on the exterior side as it's the case for all external flow simulations. The reason I initially had both of them was because I wanted to play around with some mean values including the jump itself for best computation of the fluid traction.

- Removed the point `imposePointWisejumpconditions` since the code was giving basically the same result as simply using L2_Lagrange family for the calculation of the jump quantities in the L2 projection step.

- Folded some of the repeated parts in the` imposeWeakJumpConditions `into a new helper function `checkDoubleCountingIntersection`. Please note that some parts of  imposeWeakJumpConditions ` were pretty tedious to code up, and I would rather not play around too much with what's currently working. Although some lines seem to be similar  ...I believe this function is now relatively compact by introducing loops over normal and tangents ID as Boyce recommended. Let me know if you want to make further change in the way things are written.

Thanks,

Amin

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2854,39 +2854,22 @@ FEDataManager::collectGhostDOFIndices(std::vector<unsigned int>& ghost_dofs,
...
-                if (node->n_dofs(sys_num, var_num) > 0)
-                {
-                    const unsigned int dof_index = node->dof_number(sys_num, var_num, 0);
-                    if (dof_index < first_local_dof || dof_index >= end_local_dof)
-                    {
-                        ghost_dof_set.insert(dof_index);
-                    }
-                }
-            }
-        }
```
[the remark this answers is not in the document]
boyceg: The code for recording ghost DOFs was broken for non-`LAGRANGE` variables. This probably should go into master in a separate PR.

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -1,6 +1,6 @@
 // Filename: IBFESurfaceMethod.h
 // Created on 19 May 2018 by Boyce Griffith
-//
+// Modified on 13  April 2019 by Amin Kolahdouz
```
boyceg:
```suggestion
```

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -466,14 +495,29 @@ class IBFESurfaceMethod : public IBStrategy
...
-     */
-    void imposeJumpConditions(const int f_data_idx,
-                              libMesh::PetscVector<double>& DP_ghost_vec,
-                              libMesh::PetscVector<double>& X_ghost_vec,
-                              const double data_time,
-                              const unsigned int part);
-
+     * Impose weak jump conditions.
+     */
+    void imposeWeakJumpConditions(const int f_data_idx,
```
boyceg:
```suggestion
    void imposeJumpConditions(const int f_data_idx,
```

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -466,14 +495,29 @@ class IBFESurfaceMethod : public IBStrategy
...
+                                  const double data_time,
+                                  const unsigned int part);
+    /*!
+     * \brief Helper function for checking possible double-counting
+     *  intesection points
+     */
+    void checkDoubleCountingIntersection(int axis,
+                                         const double* dx,
+                                         libMesh::VectorValue<double> n,
+                                         const libMesh::Point x,                  // = r + intersections[k].first * q;
```
boyceg: What are these comments?

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -547,6 +613,10 @@ class IBFESurfaceMethod : public IBStrategy
     std::vector<libMesh::Order> d_default_quad_order;
     bool d_use_consistent_mass_matrix;
     bool d_use_direct_forcing;
+    double d_mu;
+    double d_wss_calc_width, d_p_calc_width;
+    double d_traction_activation_time;
+    bool d_use_l2_lagrange_family;
```
boyceg: It would be more generalizable to just provide the `FEFamily` and `FEOrder` values instead of using booleans.

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -595,6 +665,7 @@ class IBFESurfaceMethod : public IBStrategy
     std::string d_libmesh_restart_file_extension;
 
 private:
+
```
boyceg:
```suggestion
```

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1,6 +1,6 @@
 // Filename: IBFESurfaceMethod.cpp
 // Created on 19 May 2018 by Boyce Griffith
-//
+// Modified on 13  April 2019 by Amin Kolahdouz
```
boyceg:
```suggestion
```

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -416,48 +485,102 @@ IBFESurfaceMethod::preprocessIntegrateData(double current_time, double new_time,
 void
 IBFESurfaceMethod::postprocessIntegrateData(double /*current_time*/, double /*new_time*/, int /*num_cycles*/)
 {
+    batch_vec_ghost_update(
+        { d_X_new_vecs, d_U_new_vecs, d_U_n_new_vecs, d_U_t_new_vecs, d_F_half_vecs }, INSERT_VALUES, SCATTER_FORWARD);
+    batch_vec_ghost_update(
+        { d_WSS_half_vecs, d_P_half_vecs, d_P_jump_half_vecs, d_TAU_half_vecs }, INSERT_VALUES, SCATTER_FORWARD);
+    //~ for (unsigned int d = 0; d < NDIM; ++d)
```
boyceg: Drop the dead code here?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -558,6 +700,41 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+            {
+                DU_jump_system[i] = &equation_systems->get_system(VELOCITY_JUMP_SYSTEM_NAME[i]);
+                DU_jump_dof_map[i] = &DU_jump_system[i]->get_dof_map();
+                DU_jump_dof_map_cache[i] = d_fe_data_managers[part]->getDofMapCache(VELOCITY_JUMP_SYSTEM_NAME[i]);
+                DU_jump_fe_type = DU_jump_dof_map[i]->variable_type(0);
+                for (unsigned int d = 0; d < NDIM; ++d)
+                {
+                    TBOX_ASSERT(DU_jump_dof_map[i]->variable_type(d) == DU_jump_fe_type);
+                }
+                // TBOX_ASSERT(U_fe_type == DU_jump_fe_type);
```
boyceg:
```suggestion
```
???

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -558,6 +700,41 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
...
+            WSS_dof_map = &WSS_system->get_dof_map();
+            WSS_dof_map_cache = d_fe_data_managers[part]->getDofMapCache(WSS_SYSTEM_NAME);
+            FEType WSS_fe_type = WSS_dof_map->variable_type(0);
+            for (unsigned int d = 0; d < NDIM; ++d)
+            {
+                TBOX_ASSERT(WSS_dof_map->variable_type(d) == WSS_fe_type);
+            }
+        }
         FEType fe_type = U_fe_type;
         UniquePtr<FEBase> fe = FEBase::build(dim, fe_type);
```
boyceg:
```suggestion
        UniquePtr<FEBase> fe_U = FEBase::build(dim, U_fe_type);
```

src/IB/IBFESurfaceMethod.cpp
boyceg:
```suggestion
        std::unique_ptr<FEBase> fe_DU_jump = FEBase::build(dim, DU_jump_fe_type);
```

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -566,23 +743,18 @@ IBFESurfaceMethod::interpolateVelocity(const int u_data_idx,
         dphi_dxi[0] = &fe->get_dphidxi();
         if (NDIM > 2) dphi_dxi[1] = &fe->get_dphideta();
 
+        FEType fe2_type = DU_jump_fe_type;
+        UniquePtr<FEBase> fe2 = FEBase::build(dim, fe2_type);
+        const std::vector<double>& JxW2 = fe2->get_JxW();
```
boyceg: You don't need this quadrature weight if you are using the same integration rule for both FE families (which I think you are).

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1216,37 +1843,82 @@ IBFESurfaceMethod::initializeFEEquationSystems()
...
+                boost::array<System*, NDIM> DU_jump_system;
+                for (unsigned int d = 0; d < NDIM; ++d)
+                {
+                    DU_jump_system[d] = &equation_systems->add_system<System>(VELOCITY_JUMP_SYSTEM_NAME[d]);
+
+                    for (unsigned int i = 0; i < NDIM; ++i)
+                    {
+                        std::ostringstream os;
+                        os << "DU_jump_" << d << "_" << i;
+                        if (d_use_l2_lagrange_family)
```
boyceg: I would suggest to keep track of the `FEFamily`s used for different types of quantities (displacements/velocities, forces, stresses/jump conditions, etc) instead of using this kind of an `if`-`else` construction.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1406,111 +2110,944 @@ IBFESurfaceMethod::initializeLevelData(Pointer<BasePatchHierarchy<NDIM> > hierar
...
+        d_equation_systems[part]->write(file_name, xdr_mode, write_mode, /*partition_agnostic*/ true);
+    }
+    return;
+}
+
+/////////////////////////////// PROTECTED ////////////////////////////////////
+
+namespace
+{
+struct IndexOrder : std::binary_function<SAMRAI::hier::Index<NDIM>, SAMRAI::hier::Index<NDIM>, bool>
```
boyceg: Isn't this already provided by `IBTK::IndexUtilities`?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1406,111 +2110,944 @@ IBFESurfaceMethod::initializeLevelData(Pointer<BasePatchHierarchy<NDIM> > hierar
...
+                    for (unsigned int k = 0; k < n_node; ++k)
+                    {
+                        const double& p_X = phi[k][qp];
+                        x_io_qp[NDIM * (qp_offset + qp) + i] += x_node[k][i] * p_X;
+                        x_qp[NDIM * (qp_offset + qp) + i] += x_node[k][i] * p_X;
+                    }
+                    N_qp[NDIM * (qp_offset + qp) + i] = n(i);
+                    // Note that here we calculate the pressure on one side as the jump plus the pressure on the other
+                    // side
+                    x_io_qp[NDIM * (qp_offset + qp) + i] += (d_calculate_interior_side ? 1.0 : -1.0) * n(i) * dh;
```
boyceg: Suggest making the flag about interior/exterior side an argument to the function instead of a class member variable.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1406,111 +2110,944 @@ IBFESurfaceMethod::initializeLevelData(Pointer<BasePatchHierarchy<NDIM> > hierar
...
+    d_P_half_vecs[part]->close();
+    d_P_jump_IB_ghost_vecs[part]->close();
+    d_X_IB_ghost_vecs[part]->close();
+
+    return;
+
+} // interpolatePressureForTraction
+
+void
+IBFESurfaceMethod::computeFluidTraction(const double data_time, unsigned int part)
```
boyceg: Same comment as above: Suggest making the flag about interior/exterior side an argument to the function instead of a class member variable.

boyceg:
Looks like there has been some good progress here.

For interpolation, it looks like you made a C++ version of the FORTRAN interpolation routines. FORTRAN 77 makes it difficult to make dimension-independent implementations, but in C++, you should be able to make interpolation routines that are much less tied to the particular value of `NDIM`, e.g. by indexing using `Index` and looping over a `Box`. You could store the data in an `ArrayData` object, which is designed for this kind of thing. It will probably be less performant than the current implementation, but it will be much more compact. I would be surprised if IIM interpolation is a bottleneck.

boyceg:
(Basically, in these implementation files, I am going to point out any code that is explicitly indexing the `0`/`1`/`2` components instead of using a loop!)

boyceg:
Another thing that should be done here is to restore the default behaviors of the examples. Instead of modifying the examples and changing the input file, I would suggest either adding a new IIM input file, or making a whole new example (or even suite of examples) for IIM.

boyceg:
I think that this is getting close to ready. I would like to try to clean up the formatting, which should help with finalizing.

@AminKolah, do you have any pending changes that you need to push to the branch? If not, I can run Aaron's C++11 modernization script and `clang-format` on the `IBFESurfaceMethod` in the branch along with `clang-format` on the corresponding files in `master` to facilitate finalizing.

AminKolah:
Let me work on finalizing this now based on your new suggestions. I'll get back to you shortly. A couple of things:

- I have a feeling the code is running a bit slower after the changes in interpolateVelocity but I wanna calculate the clock time now to see if what I'm guessing is correct. 

- I'm running the code for a longer 3D simulation to make sure it's working as expected (so far there hasn't been an issue.

In any case, I'll go ahead and make the changes you just pointed out if you want to use Aaron's script and then merge. I'll continue working on this branch so we can always review things again and merge new changes to the master. 

Btw, I have made sure I'm doing a  `clang format` before each commit although I can't tell what it really does to the format.

AminKolah:
There are only a few instances of `UniquePtr` in this branch but more instances of  `boost::array. So I think it'll be useful to run Aaron's script

boyceg:
> Btw, I have made sure I'm doing a `clang format` before each commit although I can't tell what it really does to the format.

Sometimes different versions of `clang-format` produce different formatting. I think it could reduce the number of diffs if we to run `clang-format` on `IBFESurfaceMethod.{cpp,h}` in both `master` and this branch.

AminKolah:
@boyceg please take a look whenever you get a chance. Thanks.

AminKolah:
@boyceg I just finished testing the code for a couple of 2D and 3D examples. Things working as expected. Please go ahead with Aaron's script and clang-format, then the merge:)

boyceg:
If you haven’t already, please try to address the review comments.

Also, I think we will want to squash a lot of these commits before merging.

AminKolah:
I believe I addressed all of them in my last commit but please feel free to take a look. Also making a whole new example (or even suite of examples) for IIM sounds like a good idea. Would like to that at this step? Currently I have been only testing the IIM with ex5 and ex9 of IBFE.

drwells:
@boyceg @AminKolah I really want to start using this code. When can we get this started again?

boyceg:
This is another branch where I think we need to throw away the history (e.g. using the git diff-based strategy).

I will not have time to work on this until after a couple of grant applications go in in early June.

drwells:
Will this be done in the next few days? If not, we should move to 0.6.

drwells:
> files changed: 823

I think something unintended may have happened in the merge :)

boyceg:
This is not the final version. This should be tagged WIP.

boyceg:
This is superseded by #637.

PR #242, opened 2018-05-31
support both velocity and force constraints between overlapping structures

boyceg:


boyceg:
@cpuelz, give this a try. Use `registerOverlappingVelocityConstraint(part1, part2)` to say that `part1` follows `part2` in the overlapping region. This only resets the nodal velocities and positions of `part1` where those nodes overlap `part2` in the reference configuration.

PR #246, opened 2018-06-05
Update Jenkins configure to properly treat compiler flags as strings.

abarret:


abarret:
@boyceg This should fix the Jenkins issue we were having.

PR #248, opened 2018-06-27
specialized nodal spreading and interpolation

boyceg:
For nodal quadrature rules (QTRAP for FIRST order elements, QSIMPSON for SECOND order elements), Lagrangian-Eulerian interaction can be performed directly at the nodes, without looping over the elements.

### DIFF
--- a/examples/IBFE/explicit/ex0/example.cpp
+++ b/examples/IBFE/explicit/ex0/example.cpp
@@ -271,7 +271,8 @@ bool run_example(int argc, char** argv, std::vector<double>& u_err, std::vector<
                                                     FIFTH,
                                                     /*use_adaptive_quadrature*/ false,
                                                     /*point_density*/ 2.0,
-                                                    /*use_consistent_mass_matrix*/ true);
+                                                    /*use_consistent_mass_matrix*/ true,
+                                                    /*use_nodal_quadrature*/ false);
             ib_post_processor->registerInterpolatedScalarEulerianVariable(
                 "p_f", LAGRANGE, FIRST, p_var, p_current_ctx, p_ghostfill, p_interp_spec);
         }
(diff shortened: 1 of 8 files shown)


boyceg:
@cpuelz this PR is adding support for direct spreading/interpolation from the FE mesh nodes. So far it only handles spreading. If you get a chance, can you test it out a bit? I would use `FIRST` order elements along with `QTRAP` for the `IB_quad_type`.

boyceg:
OK, I have updated the input files so that nodal interaction is specified via:
```
IBFEMethod {
   // ... other settings ...
   IB_use_nodal_quadrature = TRUE
   IB_use_consistent_mass_matrix = FALSE  // ... recommended but not required ...
}
```
This change makes it easy to use nodal interaction for models that have mixed-order parts (e.g. a combination of `FIRST` and `SECOND` order parts).

It also lets us test the code in comparison to element-based quadrature using equivalent quadrature rules (e.g. `QTRAP` for `FIRST` order meshes, `QSIMPSON` for `SECOND` order meshes).

boyceg:
Note in particular that with the current implementation, if you specify:
```
IBFEMethod {
   // ... other settings ...
   IB_quad_rule = "QTRAP"
   IB_quad_order = "FIRST"
   IB_use_adaptive_quadrature = FALSE
   IB_use_consistent_mass_matrix = FALSE
}
```
the code will loop over the elements and apply `QTRAP` to each element. This is expected to be slower than looping over the nodes, because most nodes will be visited multiple times when the loop is over the elements.

In particular, in the current implementation, to get purely nodal quadrature, you have to specify
```
IBFEMethod {
   // ... other settings ...
   IB_use_nodal_quadrature = TRUE
   IB_use_consistent_mass_matrix = FALSE  // ... recommended but not required ...
}
```

PR #252, opened 2018-07-13
Fix fortran warnings

drwells:
After some issues with AFS I finally have things set up. Here is a little PR to get started!

1. gfortran rightfully complains that the integral types should be declared before the real ones since they are used to define the real ones.
2. I removed unused local variables or used the technique applied elsewhere (`x = x`) to suppress warnings.


drwells:
I think I made @jenkinsbot1024 unhappy :(

boyceg:
It will be happy once you two are properly introduced.

boyceg:
add to whitelist

PR #253, opened 2018-07-18
Fix some unused variable warnings.

drwells:
This uses the classic `(void)x;` trick or places things in a scope surrounded by `#if/#endif`.


boyceg:
Instead of `(void)foo;` I usually follow the SAMRAI convention and use `NULL_USE(foo);` (which is actually the same thing).

drwells:
This is good to know; I'll use that idiom in the future.

I pushed an amended version.

boyceg:
Looks good.

I believe that in SAMRAI 3, TBOX_ASSERT is automatically turned on/off by NDEBUG, and so one of the things to do when we finally update to SAMRAI 3 is to remove a lot of these `#if` guards.

PR #254, opened 2018-07-18
Disambiguate SAMRAI::hier::Index and Eigen::Index.

drwells:
I cannot compile the examples without the namespace qualification.

### DIFF
--- a/examples/adv_diff/ex0/QInit.cpp
+++ b/examples/adv_diff/ex0/QInit.cpp
@@ -106,7 +106,7 @@ QInit::setDataOnPatch(const int data_idx,
     TBOX_ASSERT(Q_data);
 #endif
     const Box<NDIM>& patch_box = patch->getBox();
-    const Index<NDIM>& patch_lower = patch_box.lower();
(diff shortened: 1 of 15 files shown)


boyceg:
I think you should be able to switch from SAMRAI::hier::Index to the slightly shorter hier::Index and have everything work. Can you give it a shot?

drwells:
Yes, that works.

drwells:
I pushed an ammended commit that just writes `hier::Index`.

boyceg:
Are there any unqualified uses of Index left in the examples? If so, should we namespace all of them?

boyceg:
> On Jul 19, 2018, at 11:50 AM, David Wells <notifications@github.com> wrote:
> 
> There were two more unqualified uses of Index (I did not notice them before because GCC was happy with their namespaces). I prepended hier:: to those so now all of them in examples/ are qualified.
> 
Great; if you haven't already do so, can you also check in ibtk/examples? I am always surprised about what will compile and what won't.

drwells:
I found a few there as well; they are now qualified!

boyceg:
👍

PR #255, opened 2018-07-19
Fix some uninitialized variable warnings.

drwells:
The first block just sets up `U_wall_coef` and `h_wall_coef`, so we can move the assignments up and merge the `if` statement. This fixes an uninitialized variable warning for `U_wall_coef` and `h_wall_coef`.

### DIFF
--- a/src/level_set/fortran/levelsetops3d.f.m4
+++ b/src/level_set/fortran/levelsetops3d.f.m4
@@ -415,6 +415,20 @@ c
       REAL    U_wall_coef
       REAL    h_wall_coef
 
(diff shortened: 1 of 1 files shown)


amneetb:
@boyceg: The fix will most likely lead to some side effect. I have discussed with @nishantnangia and we will fix and test it with an example soon.    

@drwells : Thanks for bringing it our attention!

nishantnangia:
I have created a separate PR with this fix: https://github.com/IBAMR/IBAMR/pull/256

drwells:
Superceded by #256.

PR #256, opened 2018-07-19
Fixing uninitialized fortran warnings in level set fortran code

nishantnangia:
@boyceg @drwells @amneetb This should fix the uninitialized variable warnings in the level set Fortran code without unintended consequences. I have tested with level_set/ex0.


amneetb:
@nishantnangia: Looks good 👍

drwells:
@nishantnangia Thanks for the quick fix! Merging those trees of `if`s is definitely a better idea.

PR #258, opened 2018-07-20
c++11 updates

abarret:
This branch contains the condensed updates for c++11 support. It currently requires c++11 compilers and libMesh to be built without the flag --disable-unique-ptr. This branch currently compiles with libMesh master branch. Some c++11 changes will have to wait until we update to SAMRAI 3.

Things left to be done:
1) Replace empty destructors with = default construct.
2) Update example routines.

@boyceg Should I add in the script that replaces libMesh::UniquePtr and libMesh::AutoPtr with std::unique_ptr?


boyceg:
Yes, please replace empty destructors with `= default` and update the examples. In addition:

1. Update all `libMesh::UniquePtr<>` and `libMesh::AutoPtr<>` by `std::unique_ptr<>`.
2. Add script(s) that document the usage of `clang-tidy` and `clang-format` to do these updates. (So that we can easily modernize other branches and/or other applications, and so that we can "easily" do further updates at some later point in time.)

boyceg:
@drwells, can you take a look at the changes (I know there are a lot...) to see if you spot any obvious issues. These updates were mostly done automatically using `clang-tidy` along with some extra scripts and manual work by @abarret.

boyceg:
@abarret, is it feasible to create a script that folks can run on their application codes to do these updates?

drwells:
I'm on it!

abarret:
Write a script to update libMesh's smart pointers? This is already done. I will add it to the repository.

Write a script to run clang-tidy? I think it's possible, but I think it would take a lot of work to make the script usable without a significant amount of user interaction. The big issue being the generation of the compilation database (and it using relative paths instead of absolute paths). I could try to adapt the run-clang-tidy.py script that is provided to work specifically with IBAMR by reading the Makefile for compilation flags and using git to find changed files.

boyceg:
> Write a script to update libMesh's smart pointers? This is already done. I will add it to the repository.

Great.
> Write a script to run clang-tidy? I think it's possible, but I think it would take a lot of work to make the script usable without a significant amount of user interaction. The big issue being the generation of the compilation database (and it using relative paths instead of absolute paths). I could try to adapt the run-clang-tidy.py script that is provided to work specifically with IBAMR by reading the Makefile for compilation flags and using git to find changed files.
> 

That sounds like a reasonable solution to me. You also could require that the code be built "in place" instead of using a VPATH build for the purposes of the script. (I am not sure if that helps or not.)

ibtk/src/utilities/HierarchyIntegrator.cpp
```diff
@@ -185,19 +185,19 @@ HierarchyIntegrator::~HierarchyIntegrator()
         d_registered_for_restart = false;
     }
 
-    for (RefinePatchStrategyMap::iterator it = d_ghostfill_strategies.begin(); it != d_ghostfill_strategies.end(); ++it)
+    for (auto& d_ghostfill_strategie : d_ghostfill_strategies)
```
drwells: This is one of my favorite bugs in clang-modernize.

We should fix these before merging.

src/IB/IBStrategy.cpp
```diff
@@ -91,7 +91,7 @@ namespace IBAMR
 
 /////////////////////////////// PUBLIC ///////////////////////////////////////
 
-IBStrategy::IBStrategy() : d_ib_solver(NULL), d_use_fixed_coupling_ops(false)
+IBStrategy::IBStrategy()
```
drwells: This doesn't seem related to other C++11 stuff; is it?

src/advect/AdvectorPredictorCorrectorHyperbolicPatchOps.cpp
```diff
@@ -605,11 +599,8 @@ AdvectorPredictorCorrectorHyperbolicPatchOps::initializeDataOnPatch(Patch<NDIM>&
     if (initial_time)
     {
         VariableDatabase<NDIM>* var_db = VariableDatabase<NDIM>::getDatabase();
-        for (std::set<Pointer<FaceVariable<NDIM, double> > >::const_iterator cit = d_u_var.begin();
-             cit != d_u_var.end();
-             ++cit)
+        for (auto u_var : d_u_var)
         {
-            Pointer<FaceVariable<NDIM, double> > u_var = *cit;
```
drwells: Nice!

drwells:
I am ambivalent about the `= default` syntax on the destructors: since these classes inherit from classes with virtual destructors (most of the time!) we can just use the implicit destructor and delete the implementation and declaration.

boyceg:
@drwells, this monolithic commit is my fault --- @abarret originally had each of these changes in separate commits, but had not applied formatting updates using `git clang-format` along the way. I suggested that he capture all of the diffs, apply them to a new branch, and then reformat before committing.

ibtk/include/ibtk/CartSideDoubleQuadraticCFInterpolation.h
```diff
@@ -241,7 +241,7 @@ class CartSideDoubleQuadraticCFInterpolation : public CoarseFineBoundaryRefinePa
      * Boolean value indicating whether we are enforcing a consistent
      * interpolation scheme at "Type 2" coarse-fine interface ghost cells.
      */
-    bool d_consistent_type_2_bdry;
+    bool d_consistent_type_2_bdry = false;
```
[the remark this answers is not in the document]
boyceg: I am pretty ambivalent about these sorts of changes.

ibtk/include/ibtk/FaceDataSynchronization.h
```diff
@@ -81,8 +82,8 @@ class FaceDataSynchronization : public SAMRAI::tbox::DescribedClass
         /*!
          * \brief Default constructor.
          */
-        inline SynchronizationTransactionComponent(int data_idx = -1, const std::string& coarsen_op_name = "NONE")
-            : d_data_idx(data_idx), d_coarsen_op_name(coarsen_op_name)
+        inline SynchronizationTransactionComponent(int data_idx = -1, std::string coarsen_op_name = "NONE")
+            : d_data_idx(data_idx), d_coarsen_op_name(std::move(coarsen_op_name))
```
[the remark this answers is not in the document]
boyceg: Although I agree that it will likely not make much of an impact, I think I would like to go ahead and start using `std::move` where appropriate.

ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -447,8 +447,8 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
                              bool can_be_refined,
                              bool initial_time,
                              SAMRAI::tbox::Pointer<SAMRAI::hier::BasePatchLevel<NDIM> > old_level =
-                                 SAMRAI::tbox::Pointer<SAMRAI::hier::BasePatchLevel<NDIM> >(NULL),
-                             bool allocate_data = true);
+                                 SAMRAI::tbox::Pointer<SAMRAI::hier::BasePatchLevel<NDIM> >(nullptr),
+                             bool allocate_data = true) override final;
```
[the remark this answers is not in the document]
boyceg: Yes, this is actually what we want to do here with the current design. (I am happy to revise the design.) There is a separate method, `initializeLevelDataSpecialized`, that can be overridden in subclasses.

boyceg:
@drwells, if it is easy to identify which destructors can be safely dropped, I am happy to remove them (in place of using `= delete`).

abarret:
There are incremental changes in the branch abarret/c++11-updates which I have not deleted yet. Although there are some annoying formatting issues (like extra spaces after auto)

ibtk/include/ibtk/CartSideDoubleQuadraticCFInterpolation.h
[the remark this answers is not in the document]
abarret: We could remove dozens of lines from many .cpp files if we initialized with the declaration with all the variables, not just fundamental types.

ibtk/src/utilities/HierarchyIntegrator.cpp
abarret: I thought I fixed these but I must have missed some (maybe a lot?)! I'll see if I can find the rest.

src/IB/IBStrategy.cpp
abarret: Clang-tidy moved these initializations to the declaration.

ibtk/include/ibtk/CartSideDoubleQuadraticCFInterpolation.h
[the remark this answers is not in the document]
boyceg: I have been looking into this some more, and I think that I am generally in favor of "initialize with the declaration" syntax.

drwells:
Bah, we already have merge conflicts :(

@abarret Thanks for making those changes.

I believe all of the testers are failing since the build system does not check for C++11 compatibility.  I suggest that we proceed as follows: I will put up a PR that checks that compilers support enough of C++11 for our needs and adds `-std=c++11` to the C++ flags. Subsequently, we can then merge either this PR or the individual PRs: since the fixes are individual commits it should be easy to cherry-pick them, but if you think this is going to take too long then we can go ahead with the monolithic version.

@abarret Do you, per chance, have such a PR already written? Otherwise I will get started later this afternoon.

abarret:
I think the c++11 testers are failing specifically because libMesh wasn't built with support for `std::unique_ptr`.

I haven't made any changes to the build system. That still needs to be done. I also need to update the examples. Perhaps this can also be done in a separate PR?

@boyceg @drwells Should I continue with the "initialize with the declaration" changes for non-fundamental types?

drwells:
Doing "initialize with the declaration" does have some advantages and @boyceg is in favor, so lets do it for all types.

Out of curiousity: do you have an example of a non-default initialization for a non-fundamental type? I would like to see what that looks like.

drwells:
@abarret We can always update the examples later: I'll be happy as long as things still compile and work.

ibtk/include/ibtk/muParserRobinBcCoefs.h
```diff
@@ -194,28 +194,28 @@ class muParserRobinBcCoefs : public SAMRAI::solv::RobinBcCoefStrategy<NDIM>
...
 
     /*!
      * The mu::Parser objects which evaluate the data-setting functions.
      */
-    std::vector<mu::Parser> d_acoef_parsers;
-    std::vector<mu::Parser> d_bcoef_parsers;
-    std::vector<mu::Parser> d_gcoef_parsers;
+    std::vector<mu::Parser> d_acoef_parsers = std::vector<mu::Parser>(2 * NDIM);
+    std::vector<mu::Parser> d_bcoef_parsers = std::vector<mu::Parser>(2 * NDIM);
+    std::vector<mu::Parser> d_gcoef_parsers = std::vector<mu::Parser>(2 * NDIM);
```
abarret: @drwells This is an example of how I've done an `std::vector<>` initialization to match the other initializations. Although it would be equally valid (and cleaner) to do:
`std::vector<mu::Parser> d_acoef_parsers(2 * NDIM);`

abarret:
@boyceg I forgot a change to to the IBAMR library. It should be ready for review now while I document changes.

examples/CIBFE/ex0/example.cpp
```diff
@@ -344,7 +344,7 @@ run_example(int argc, char* argv[])
         {
             time_integrator->registerVisItDataWriter(visit_data_writer);
         }
-        libMesh::UniquePtr<ExodusII_IO> exodus_io(uses_exodus ? new ExodusII_IO(mesh) : NULL);
+        std::unique_ptr<ExodusII_IO> exodus_io(uses_exodus ? new ExodusII_IO(mesh) : NULL);
```
boyceg: NULL ---> nullptr?

boyceg: ...oh wait, you did not run modernize on the examples, did you?

abarret: The examples have not been updated to C++11 constructs except for std::unique_ptr.

ibtk/include/ibtk/AppInitializer.h
```diff
@@ -219,46 +219,46 @@ class AppInitializer : public SAMRAI::tbox::DescribedClass
...
-    SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> d_input_db;
+    SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> d_input_db = nullptr;
 
     /*!
      * Restart settings.
      */
-    std::string d_restart_read_dirname;
-    int d_restart_restore_num;
-    bool d_is_from_restart;
+    std::string d_restart_read_dirname = "";
```
boyceg: 👍

boyceg:
> The examples have not been updated to C++11 constructs except for std::unique_ptr.

OK, actually, that is great — it is a pretty good indication that this set of updates will be fairly benign for application codes.

ibtk/include/ibtk/CCPoissonHypreLevelSolver.h
```diff
@@ -300,40 +301,40 @@ class CCPoissonHypreLevelSolver : public LinearSolver, public PoissonSolver
...
 
-    std::string d_solver_type, d_precond_type;
-    int d_rel_change;
-    int d_num_pre_relax_steps, d_num_post_relax_steps;
-    int d_memory_use;
+    std::string d_solver_type = "PFMG", d_precond_type = "none";
+    int d_rel_change = 0;
+    int d_num_pre_relax_steps = 1, d_num_post_relax_steps = 1;
+    int d_memory_use = 0;
     int d_rap_type;
```
boyceg: Do these guys not get default values?

ibtk/include/ibtk/CCPoissonLevelRelaxationFACOperator.h
```diff
@@ -254,27 +254,27 @@ class CCPoissonLevelRelaxationFACOperator : public PoissonFACPreconditionerStrat
...
      */
-    std::string d_level_solver_type, d_level_solver_default_options_prefix;
-    double d_level_solver_abs_residual_tol, d_level_solver_rel_residual_tol;
-    int d_level_solver_max_iterations;
-    std::vector<SAMRAI::tbox::Pointer<PoissonSolver> > d_level_solvers;
-    SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> d_level_solver_db;
+    std::string d_level_solver_type = "", d_level_solver_default_options_prefix = "";
+    double d_level_solver_abs_residual_tol = 1.0e-50, d_level_solver_rel_residual_tol = 1.0e-5;
+    int d_level_solver_max_iterations = 1;
+    std::vector<SAMRAI::tbox::Pointer<PoissonSolver> > d_level_solvers = {};
```
boyceg: Another C++11-ism that we could go ahead and adopt while we are making 1000's of changes would be to convert "> >" to ">>". @drwells, what do you think?

ibtk/include/ibtk/CopyToRootSchedule.h
```diff
@@ -121,15 +121,15 @@ class CopyToRootSchedule
      *
      * \return A reference to this object.
      */
-    CopyToRootSchedule& operator=(const CopyToRootSchedule& that);
+    CopyToRootSchedule& operator=(const CopyToRootSchedule& that) = delete;
 
     void commonClassCtor();
 
     const int d_root_proc;
```
boyceg: Is there a reason why some of these get default values but not others?

boyceg: I guess one reason could be that these were set up by a script, and there were not default values in the original implementation...?

ibtk/include/ibtk/CCPoissonHypreLevelSolver.h
abarret: I only added default values to members who got default values on initialization. For example, here d_rap_type was initialized with an input to the constructor. Others might not have been initialized at all.

I can go back through and initialize on declaration all member variables default values.

boyceg:
> I can go back through and initialize on declaration all member variables default values.

That seems unnecessary. Possibly the only reason maybe to do it now is that with the current diff, it is really easy to spot values that are not initialized in the headers. Perhaps there also is some compiler flag that will catch them, so that we can do it another time?

boyceg:
>> I can go back through and initialize on declaration all member variables default values.
> 
> That seems unnecessary. Possibly the only reason maybe to do it now is that with the current diff, it is really easy to spot values that are not initialized in the headers. Perhaps there also is some compiler flag that will catch them, so that we can do it another time?

This answer from SO seems like the right approach: https://stackoverflow.com/a/11594963 <https://stackoverflow.com/a/11594963>.

Basically, if a member variable is always initialized with the same default value, put that in the header. Otherwise, the member should be initialized in the constructor and not in the header.

ibtk/include/ibtk/HierarchyIntegrator.h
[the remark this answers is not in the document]
boyceg: Let me add just a little more context here. `initializeLevelData()` is a SAMRAI interface. In this class, we are saying: "we know how we want to initialize level data _in general_, and that is what is implemented here; however, there are still some implementation specific details that we want to delegate to subclasses --- those go in `initializeLevelDataSpecialized()`".

abarret:
> Perhaps there also is some compiler flag that will catch them, so that we can do it another time?

The compiler flag `-Weffc++` will detect uninitialized values.
> Basically, if a member variable is always initialized with the same default value, put that in the header. Otherwise, the member should be initialized in the constructor and not in the header.

If I remember correctly, I think this is essentially what I've done, except for initializing values that were originally uninitialized.

ibtk/include/ibtk/HierarchyIntegrator.h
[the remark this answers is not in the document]
boyceg: The same pattern is used in other places. There are SAMRAI interfaces that we want to provide a basic implementation while still delegating some details to subclasses. The delegated member function names are usually named something like `XXXSpecialized()`, where `XXX()` is the function from the SAMRAI interface.

ibtk/include/ibtk/muParserRobinBcCoefs.h
```diff
@@ -194,28 +194,28 @@ class muParserRobinBcCoefs : public SAMRAI::solv::RobinBcCoefStrategy<NDIM>
...
 
     /*!
      * Time and position variables.
      */
-    double* d_parser_time;
-    Point* d_parser_posn;
+    double* d_parser_time = new double;
```
boyceg: Could these be changed to `std::unique_ptr`?

ibtk/include/ibtk/muParserRobinBcCoefs.h
```diff
@@ -194,28 +194,28 @@ class muParserRobinBcCoefs : public SAMRAI::solv::RobinBcCoefStrategy<NDIM>
...
+    std::vector<std::string> d_bcoef_function_strings = {};
+    std::vector<std::string> d_gcoef_function_strings = {};
 
     /*!
      * The mu::Parser objects which evaluate the data-setting functions.
      */
-    std::vector<mu::Parser> d_acoef_parsers;
-    std::vector<mu::Parser> d_bcoef_parsers;
-    std::vector<mu::Parser> d_gcoef_parsers;
+    std::vector<mu::Parser> d_acoef_parsers = std::vector<mu::Parser>(2 * NDIM);
```
boyceg: Does `auto` work here?

ibtk/include/ibtk/private/LDataManager-inl.h
```diff
@@ -165,7 +165,7 @@ LDataManager::getLagrangianStructureNames(const int level_number) const
     TBOX_ASSERT(d_coarsest_ln <= level_number && d_finest_ln >= level_number);
 #endif
     std::vector<std::string> ret_val;
-    for (std::map<int, std::string>::const_iterator cit(d_strct_id_to_strct_name_map[level_number].begin());
+    for (auto cit(d_strct_id_to_strct_name_map[level_number].begin());
```
boyceg: Can this be a range-based for?

ibtk/include/ibtk/private/LDataManager-inl.h
```diff
@@ -181,7 +181,7 @@ LDataManager::getLagrangianStructureIDs(const int level_number) const
     TBOX_ASSERT(d_coarsest_ln <= level_number && d_finest_ln >= level_number);
 #endif
     std::vector<int> ret_val;
-    for (std::map<std::string, int>::const_iterator cit(d_strct_name_to_strct_id_map[level_number].begin());
+    for (auto cit(d_strct_name_to_strct_id_map[level_number].begin());
```
boyceg: Can this be a range-based for?

ibtk/include/ibtk/private/LNode-inl.h
```diff
@@ -102,8 +103,7 @@ inline void
 LNode::registerPeriodicShift(const SAMRAI::hier::IntVector<NDIM>& offset, const Vector& displacement)
 {
     LNodeIndex::registerPeriodicShift(offset, displacement);
-    for (std::vector<SAMRAI::tbox::Pointer<Streamable> >::iterator it = d_node_data.begin(); it != d_node_data.end();
-         ++it)
+    for (auto it = d_node_data.begin(); it != d_node_data.end(); ++it)
```
boyceg: Can this be a range-based for?

ibtk/include/ibtk/private/LNode-inl.h
```diff
@@ -139,8 +139,7 @@ LNode::appendNodeDataItem(const SAMRAI::tbox::Pointer<Streamable>& node_data_ite
 inline void
 LNode::removeNodeDataItem(const SAMRAI::tbox::Pointer<Streamable>& node_data_item)
 {
-    std::vector<SAMRAI::tbox::Pointer<Streamable> >::iterator it =
-        std::find(d_node_data.begin(), d_node_data.end(), node_data_item);
+    auto it = std::find(d_node_data.begin(), d_node_data.end(), node_data_item);
```
boyceg: Can this be a range-based for?

ibtk/include/ibtk/private/PETScSAMRAIVectorReal-inl.h
```diff
@@ -64,7 +64,7 @@ PETScSAMRAIVectorReal::destroyPETScVector(Vec petsc_vec)
 {
     if (petsc_vec)
     {
-        PETScSAMRAIVectorReal* psv = static_cast<PETScSAMRAIVectorReal*>(petsc_vec->data);
+        auto psv = static_cast<PETScSAMRAIVectorReal*>(petsc_vec->data);
```
boyceg: 🥇

ibtk/src/boundary/HierarchyGhostCellInterpolation.cpp
```diff
@@ -112,19 +112,7 @@ static Timer* t_fill_data_set_physical_bcs;
...
-      d_grid_geom(NULL),
-      d_coarsest_ln(-1),
-      d_finest_ln(-1),
-      d_coarsen_alg(NULL),
-      d_coarsen_strategy(NULL),
-      d_coarsen_scheds(),
-      d_refine_alg(NULL),
-      d_refine_strategy(NULL),
-      d_refine_scheds(),
+    : d_transaction_comps(),
```
boyceg: 🥇

ibtk/src/boundary/HierarchyGhostCellInterpolation.cpp
```diff
@@ -209,12 +197,12 @@ HierarchyGhostCellInterpolation::initializeOperatorState(
     VariableDatabase<NDIM>* var_db = VariableDatabase<NDIM>::getDatabase();
     bool registered_coarsen_op = false;
     d_coarsen_alg = new CoarsenAlgorithm<NDIM>();
-    for (unsigned int comp_idx = 0; comp_idx < d_transaction_comps.size(); ++comp_idx)
+    for (auto& d_transaction_comp : d_transaction_comps)
```
boyceg: This is nit-picky, but `d_XXX` variable names are reserved for non-static class member variables (with `s_XXX` being reserved for static class member variables).

boyceg: So, `d_transaction_comp` ---> `transaction_comp`.

boyceg: I think it should be easy to use a regex to convert `for (auto* d_` to `for (auto* ` and then clean up the resulting mangled variable names.

ibtk/src/boundary/HierarchyGhostCellInterpolation.cpp
```diff
@@ -335,11 +323,9 @@ HierarchyGhostCellInterpolation::initializeOperatorState(
 
         const std::vector<RobinBcCoefStrategy<NDIM>*>& robin_bc_coefs = d_transaction_comps[comp_idx].d_robin_bc_coefs;
         bool null_bc_coefs = true;
-        for (std::vector<RobinBcCoefStrategy<NDIM>*>::const_iterator cit = robin_bc_coefs.begin();
-             cit != robin_bc_coefs.end();
-             ++cit)
+        for (auto robin_bc_coef : robin_bc_coefs)
```
boyceg: 💯

ibtk/src/boundary/cf_interface/CartCellDoubleLinearCFInterpolation.cpp
```diff
@@ -151,9 +139,8 @@ CartCellDoubleLinearCFInterpolation::postprocessRefine(Patch<NDIM>& fine,
                                                        const Box<NDIM>& fine_box,
                                                        const IntVector<NDIM>& ratio)
 {
-    for (std::set<int>::const_iterator cit = d_patch_data_indices.begin(); cit != d_patch_data_indices.end(); ++cit)
+    for (int patch_data_index : d_patch_data_indices)
```
boyceg: Should we prefer `auto` here?

boyceg: Of course, I should have said `auto&`.

ibtk/src/boundary/cf_interface/CoarseFineBoundaryRefinePatchStrategy.cpp
```diff
@@ -43,18 +43,6 @@ namespace IBTK
 
 /////////////////////////////// PUBLIC ///////////////////////////////////////
 
-CoarseFineBoundaryRefinePatchStrategy::CoarseFineBoundaryRefinePatchStrategy()
```
boyceg: 🥇

ibtk/src/boundary/physical_boundary/CartExtrapPhysBdryOp.cpp
```diff
@@ -375,13 +359,11 @@ CartExtrapPhysBdryOp::setPhysicalBoundaryConditions_cell(
         const Box<NDIM>& ghost_box = patch_data->getGhostBox();
 
         // Loop over the boundary fill boxes and extrapolate the data.
-        for (std::vector<std::pair<Box<NDIM>, std::pair<int, int> > >::const_iterator it = bdry_fill_boxes.begin();
-             it != bdry_fill_boxes.end();
-             ++it)
+        for (const auto& bdry_fill_box_it : bdry_fill_boxes)
```
boyceg: I am not sure how I feel about the use of `XXX_it` for variables that are no longer iterators...

ibtk/src/lagrangian/FEDataInterpolation.cpp
```diff
@@ -76,9 +61,7 @@ FEDataInterpolation::registerSystem(const System& system,
 {
     TBOX_ASSERT(!d_initialized && (!phi_vars.empty() || !dphi_vars.empty()));
     const unsigned int sys_num = system.number();
-    for (std::vector<const System *>::iterator it = d_noninterp_systems.begin(), it_end = d_noninterp_systems.end();
-         it != it_end;
-         ++it)
+    for (auto it = d_noninterp_systems.begin(), it_end = d_noninterp_systems.end(); it != it_end; ++it)
```
boyceg: Can this be a range-based for?

ibtk/src/lagrangian/FEDataInterpolation.cpp
```diff
@@ -128,7 +111,7 @@ FEDataInterpolation::registerInterpolatedSystem(const System& system,
 {
     TBOX_ASSERT(!d_initialized && (!vars.empty() || !grad_vars.empty()));
     const unsigned int sys_num = system.number();
-    for (std::vector<const System *>::iterator it = d_systems.begin(), it_end = d_systems.end(); it != it_end; ++it)
+    for (auto it = d_systems.begin(), it_end = d_systems.end(); it != it_end; ++it)
```
boyceg: Can this be a range-based for?

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -187,7 +187,7 @@ inline void
 collect_unique_elems(std::vector<Elem*>& elems, const ContainerOfContainers& elem_patch_map)
 {
     std::set<Elem*, ElemComp> elem_set;
-    for (typename ContainerOfContainers::const_iterator it = elem_patch_map.begin(); it != elem_patch_map.end(); ++it)
+    for (auto it = elem_patch_map.begin(); it != elem_patch_map.end(); ++it)
```
boyceg: Can this be a range-based for?

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2695,7 +2683,7 @@ FEDataManager::collectActivePatchElements(std::vector<std::vector<Elem*> >& acti
             const double* const patch_dx = patch_geom->getDx();
             const double patch_dx_min = *std::min_element(patch_dx, patch_dx + NDIM);
 
-            std::set<Elem*>::const_iterator el_it = frontier_elems.begin();
+            auto el_it = frontier_elems.begin();
             const std::set<Elem*>::const_iterator el_end = frontier_elems.end();
             for (; el_it != el_end; ++el_it)
```
boyceg: I picked up this kind of construct from the libMesh examples. It seems like an ideal use case for range-based for. The tools did not do the conversion since the ranges are set outside of the `for` loop. :-(

ibtk/include/ibtk/HierarchyIntegrator.h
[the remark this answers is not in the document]
drwells: Got it.

ibtk/include/ibtk/CCPoissonLevelRelaxationFACOperator.h
drwells: Yes, we should do that.

ibtk/include/ibtk/muParserRobinBcCoefs.h
drwells: Why are these even pointers to begin with? I'll take a look at this.

ibtk/include/ibtk/muParserRobinBcCoefs.h
drwells: This is a strange declaration: it should suffice to just write
```cpp
std::vector<mu::Parser> d_acoef_parsers(2 * NDIM);
```

abarret: @drwells I initialized like this to match the other initializations I did using `= <value>`. I can go back and update these kinds of initializations though to make them cleaner.

ibtk/src/boundary/HierarchyGhostCellInterpolation.cpp
abarret: I left it as `d_transaction_comp` since it is a reference to an element of a container variable (kind of like a member variable?), but I can go back and change them.

boyceg:
> On Aug 1, 2018, at 2:44 PM, Aaron Barrett <notifications@github.com> wrote:
> 
> I left it as d_transaction_comp since it is a reference to an element of a container variable (kind of like a member variable?), but I can go back and change them.
> 
I understand the rationale, but I think I want to stick with the convention that d_XXX indicates a class member variable — although this is a reference to a (part of a) member variable, it is not in fact a member variable.

ibtk/include/ibtk/CCPoissonLevelRelaxationFACOperator.h
abarret: @boyceg clang-format is correcting the `>>` back to `> >`. Should I go ahead and adjust this in the .clang-format file?

boyceg:
> @boyceg <https://github.com/boyceg> clang-format is correcting the >> back to > >. Should I go ahead and adjust this in the .clang-format file?
> 
Yes please.

ibtk/include/ibtk/muParserRobinBcCoefs.h
drwells: Oh, I didn't realize that this was in a header. We might be stuck with this.

abarret:
This must be recent. The examples compiled for me a few commits ago, but I haven't checked since. I'll take a look. Also Jenkins isn't happy so I must have done something...

drwells:
I meant to write "deleted destructors" above (which makes infinitely more sense, I was just tired when I wrote that...)

abarret:
@drwells Try the tests now. I think I fixed the issues.

abarret:
@drwells Whoops... made the same mistake on the commit message...

drwells:
Aside from the fixes that are not in this branch that were provided by #254 all of the examples compile.

abarret:
@boyceg I believe I have addressed all the issues for the IBTK library.

drwells:
I was able to get `clang-tidy` working with IBAMR by using the [build ear](https://github.com/rizsotto/Bear) tool to automatically generate `compilation_commands.json`. I am not sure how y'all managed to generate the first one but that might be a better way to do it in the future.

ibtk/include/ibtk/CCPoissonHypreLevelSolver.h
boyceg: I think that it is fine to omit things that do not always get default values.

ibtk/include/ibtk/CartCellDoubleQuadraticCFInterpolation.h
```diff
@@ -127,7 +126,7 @@ class CartCellDoubleQuadraticCFInterpolation : public CoarseFineBoundaryRefinePa
      * Presently, the implementation does nothing.
      *
      * \param fine      Fine patch containing destination data.
-     * \param coarse    Coarse patch containing source data.
+     :* \param coarse    Coarse patch containing source data.
```
boyceg: Looks like a spurious ":" here.

ibtk/include/ibtk/BJacobiPreconditioner.h
```diff
@@ -248,7 +249,7 @@ class BJacobiPreconditioner : public LinearSolver
      *
      * \param from The value to copy to this object.
      */
-    BJacobiPreconditioner(const BJacobiPreconditioner& from);
+    BJacobiPreconditioner(const BJacobiPreconditioner& from) = delete;
 
     /*!
      * ]brief Assignment operator.
```
boyceg: Looks like an old typo --- should be `\brief`.

ibtk/include/ibtk/CartCellDoubleQuadraticCFInterpolation.h
```diff
@@ -244,7 +243,7 @@ class CartCellDoubleQuadraticCFInterpolation : public CoarseFineBoundaryRefinePa
                                      const SAMRAI::hier::IntVector<NDIM>& ratio);
 
     /*!
-     * \brief Implementations of computeNormalExtension().
+     :* \brief Implementations of computeNormalExtension().
```
boyceg: Here is another random `:`.

ibtk/include/ibtk/LinearSolver.h
```diff
@@ -76,10 +76,10 @@ class LinearSolver : public virtual GeneralSolver
      * but should not assume the basis vectors to be orthonormal.  If the basis
      * vectors are not orthonormal, the solver may normalize them in place.
      */
-    virtual void setNullspace(bool nullspace_contains_constant_vec,
-                              const std::vector<SAMRAI::tbox::Pointer<SAMRAI::solv::SAMRAIVectorReal<NDIM, double> > >&
-                                  nullspace_basis_vecs = std::vector<
-                                      SAMRAI::tbox::Pointer<SAMRAI::solv::SAMRAIVectorReal<NDIM, double> > >());
+    virtual void setNullspace(
+        bool nullspace_contains_constant_vec,
+        const std::vector<SAMRAI::tbox::Pointer<SAMRAI::solv::SAMRAIVectorReal<NDIM, double>>>& nullspace_basis_vecs =
```
boyceg: Is there a more compact way to write this in C++11?

ibtk/include/ibtk/muParserRobinBcCoefs.h
boyceg: This is resolved in master.

ibtk/include/ibtk/private/LNode-inl.h
```diff
@@ -245,10 +244,10 @@ LNode::assignThatToThis(const LNode& that)
 inline void
 LNode::setupNodeDataTypeArray()
 {
-    std::fill(d_node_data_type_arr, d_node_data_type_arr + MAX_SIZE, static_cast<Streamable*>(NULL));
+    std::fill(d_node_data_type_arr, d_node_data_type_arr + MAX_SIZE, static_cast<Streamable*>(nullptr));
     Streamable* it_val;
     int class_id;
-    for (std::vector<SAMRAI::tbox::Pointer<Streamable> >::const_iterator cit = d_node_data.begin();
+    for (std::vector<SAMRAI::tbox::Pointer<Streamable>>::const_iterator cit = d_node_data.begin();
```
boyceg: range-based for?

ibtk/include/ibtk/private/StreamableManager-inl.h
```diff
@@ -111,7 +111,7 @@ StreamableManager::unpackStream(SAMRAI::tbox::AbstractStream& stream,
     {
         data_items[k] = unpackStream(stream, offset);
     }
-    std::vector<SAMRAI::tbox::Pointer<Streamable> >(data_items).swap(data_items); // trim-to-fit
+    data_items.shrink_to_fit(); // trim-to-fit
```
boyceg: Probably can eliminate this comment now that the function name says what it does.

ibtk/src/boundary/HierarchyGhostCellInterpolation.cpp
```diff
@@ -229,15 +217,15 @@ HierarchyGhostCellInterpolation::initializeOperatorState(
...
+    d_coarsen_strategy = nullptr;
 
     d_coarsen_scheds.resize(d_finest_ln + 1);
     if (registered_coarsen_op)
     {
         for (int src_ln = std::max(1, d_coarsest_ln); src_ln <= d_finest_ln; ++src_ln)
         {
-            Pointer<PatchLevel<NDIM> > level = d_hierarchy->getPatchLevel(src_ln);
-            Pointer<PatchLevel<NDIM> > coarser_level = d_hierarchy->getPatchLevel(src_ln - 1);
+            Pointer<PatchLevel<NDIM>> level = d_hierarchy->getPatchLevel(src_ln);
```
boyceg: I propose making this `auto`. I think that it should be clear that the result is a `PatchLevel`.

But, perhaps those kinds of changes can go into a follow-up PR.

boyceg: In general, I would suggest trying to convert `Pointer<T>` to `auto` wherever it makes sense, because it will be that much less code to change when moving from SAMRAI 2 to SAMRAI 3.

ibtk/src/boundary/HierarchyGhostCellInterpolation.cpp
```diff
@@ -517,11 +503,9 @@ HierarchyGhostCellInterpolation::resetTransactionComponents(
         const std::vector<RobinBcCoefStrategy<NDIM>*>& robin_bc_coefs = d_transaction_comps[comp_idx].d_robin_bc_coefs;
 #if !defined(NDEBUG)
         bool null_bc_coefs = true;
-        for (std::vector<RobinBcCoefStrategy<NDIM>*>::const_iterator cit = robin_bc_coefs.begin();
-             cit != robin_bc_coefs.end();
-             ++cit)
+        for (const auto robin_bc_coef : robin_bc_coefs)
```
boyceg: In range-based for loops, shouldn't this always be either `const auto&&` (if we are not changing the contents) or `auto&&` (if we are)?

ibtk/src/boundary/HierarchyGhostCellInterpolation.cpp
```diff
@@ -628,11 +612,11 @@ HierarchyGhostCellInterpolation::fillData(double fill_time)
     for (int dst_ln = d_coarsest_ln; dst_ln <= d_finest_ln; ++dst_ln)
     {
         if (d_refine_scheds[dst_ln]) d_refine_scheds[dst_ln]->fillData(fill_time);
-        Pointer<PatchLevel<NDIM> > level = d_hierarchy->getPatchLevel(dst_ln);
+        Pointer<PatchLevel<NDIM>> level = d_hierarchy->getPatchLevel(dst_ln);
         const IntVector<NDIM>& ratio = level->getRatioToCoarserLevel();
         for (PatchLevel<NDIM>::Iterator p(level); p; p++)
         {
-            Pointer<Patch<NDIM> > patch = level->getPatch(p());
+            Pointer<Patch<NDIM>> patch = level->getPatch(p());
```
boyceg: Also seems like a good place to use `auto`.

ibtk/src/boundary/cf_interface/CartCellDoubleQuadraticCFInterpolation.cpp
```diff
@@ -355,14 +340,14 @@ CartCellDoubleQuadraticCFInterpolation::setPatchHierarchy(Pointer<PatchHierarchy
         d_cf_boundary[ln] = new CoarseFineBoundary<NDIM>(*d_hierarchy, ln, max_ghost_width);
     }
 
-    Pointer<GridGeometry<NDIM> > grid_geom = d_hierarchy->getGridGeometry();
+    Pointer<GridGeometry<NDIM>> grid_geom = d_hierarchy->getGridGeometry();
```
boyceg: We could use `auto` here too, for instance.

ibtk/src/boundary/cf_interface/CartCellDoubleQuadraticCFInterpolation.cpp
```diff
@@ -481,13 +465,13 @@ CartCellDoubleQuadraticCFInterpolation::postprocessRefine_expensive(Patch<NDIM>&
 
         const Box<NDIM>& patch_box_fine = fine.getBox();
         const Index<NDIM>& patch_lower_fine = patch_box_fine.lower();
-        Pointer<CartesianPatchGeometry<NDIM> > pgeom_fine = fine.getPatchGeometry();
+        Pointer<CartesianPatchGeometry<NDIM>> pgeom_fine = fine.getPatchGeometry();
```
boyceg: But we cannot use it here without changing the rhs, because we are relying on `Pointer`'s ability to automatically downcast from a base class to a derived class.

ibtk/src/boundary/physical_boundary/CartCellRobinPhysBdryOp.cpp
```diff
@@ -459,13 +445,13 @@ CartCellRobinPhysBdryOp::fillGhostCellValuesCodim1(const int patch_data_idx,
         const BoundaryBox<NDIM> trimmed_bdry_box(
             bdry_box.getBox() * bc_fill_box, bdry_box.getBoundaryType(), bdry_box.getLocationIndex());
         const Box<NDIM> bc_coef_box = PhysicalBoundaryUtilities::makeSideBoundaryCodim1Box(trimmed_bdry_box);
-        Pointer<ArrayData<NDIM, double> > acoef_data = new ArrayData<NDIM, double>(bc_coef_box, 1);
-        Pointer<ArrayData<NDIM, double> > bcoef_data = new ArrayData<NDIM, double>(bc_coef_box, 1);
-        Pointer<ArrayData<NDIM, double> > gcoef_data = new ArrayData<NDIM, double>(bc_coef_box, 1);
+        Pointer<ArrayData<NDIM, double>> acoef_data = new ArrayData<NDIM, double>(bc_coef_box, 1);
```
boyceg: `auto`

ibtk/src/boundary/physical_boundary/CartExtrapPhysBdryOp.cpp
boyceg: This could be a future PR, but need to change `bdry_fill_box_it` to something else, since the type is not an interator type.

ibtk/src/lagrangian/FEDataInterpolation.cpp
boyceg: range-based for?

ibtk/src/lagrangian/FEDataInterpolation.cpp
boyceg: range-based for?

boyceg: Maybe this could be a future PR, where we update the for loop syntax.

boyceg: ...or is there a more aggressive setting in modernize for for loops?

ibtk/src/lagrangian/FEDataInterpolation.cpp
```diff
@@ -238,9 +221,9 @@ FEDataInterpolation::init(const bool use_IB_ghosted_vecs)
         const System& system = *d_noninterp_systems[system_idx];
         const DofMap& system_dof_map = system.get_dof_map();
         const std::vector<int>& all_vars = d_noninterp_system_all_vars[system_idx];
-        for (unsigned int k = 0; k < all_vars.size(); ++k)
+        for (int k : all_vars)
```
boyceg: Another place where we should update the variable name.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -130,6 +129,7 @@
 #include "tbox/Timer.h"
 #include "tbox/TimerManager.h"
 #include "tbox/Utilities.h"
+#include <memory>
```
boyceg: This include of `memory` should go up above with other system includes.

ibtk/src/solvers/interfaces/LinearSolver.cpp
```diff
@@ -51,23 +51,16 @@ namespace IBTK
 /////////////////////////////// PUBLIC ///////////////////////////////////////
 
 LinearSolver::LinearSolver()
-    : d_initial_guess_nonzero(true), d_nullspace_contains_constant_vec(false), d_nullspace_basis_vecs()
 {
     d_max_iterations = 10000;
```
boyceg: Can these go into the header?

ibtk/src/solvers/interfaces/NewtonKrylovSolver.cpp
```diff
@@ -54,28 +54,13 @@ namespace IBTK
...
-      d_J(NULL),
-      d_krylov_solver(NULL),
-      d_x(NULL),
-      d_b(NULL),
-      d_r(NULL),
-      d_max_evaluations(10000),
-      d_solution_tol(1.0e-8),
-      d_current_linear_iterations(0)
 {
     d_max_iterations = 50;
```
boyceg: Can these go into the header?

boyceg:
@abarret, @drwells, given all of the conflicts, and that most of the changes are automatic, would it be faster / better to apply the modernize changes in a series of PRs using a new branch off master?

drwells:
I think that would make the review process a lot easier. IIRC a lot of the conflicts are difficult to understand too.

abarret:
I don't think it will be faster, unless @drwells (or anyone else) knows how to fix the issues I was having with clang-tidy. I do agree it will probably be easier. I mentioned the issues in an email, but I'll summarize them here:

1) Bear (the tool to generate the compilation database) generates multiple copies of rules to compile files in the database. I think 

2) The run-clang-tidy script provided makes multiple changes in the header files. I feel like this is caused by the issue above, although I don't know for sure.

3) Clang-tidy must be run on each library separately (IBTK and IBAMR) because it uses relative directory paths for the database.

These aren't deal breakers, it just makes some clang-tidy fixes harder to deal with, more error prone, and take longer to apply.

boyceg:
Shall I nuke this PR and branch?

abarret:
Yeah I don't think we need this anymore. The only thing that is in this commit but not master is changing `< <` to `<<`.

drwells:
Surely you mean changing `> >` to `>` :smile:

boyceg:
OK, this is going away, along with the related branches.

If we ever need to restore the branches and cannot find them easily, their commit IDs are 20b8e761e7f29e60bd4af145849ebe1777bd2594 (for abarret/c++11) and 3a483c789d807b1ad69d9989f0af8c4f6858631f (for abarret/c++11-updates).

PR #262, opened 2018-07-24
Nishant/vc examples

amneetb:
1) Important improvements in order of accuracy for corner cases of INSVC integrator.
2) More examples of INSVC integrator.


amneetb:
@nishantnangia : INSVCStaggeredConservativeConvectiveOperator::applyConvectiveOperator(const int U_idx, const int N_idx)

We should get rid of U_idx and associated U_data, U_adv_data for this routine, before we forget. We are working with V_adv.

boyceg:
@amneetb is this ready to merge?

nishantnangia:
@boyceg This isn't ready yet -- I can finish it up now and let you know.

boyceg:
Jenkins is happy. Is @amneetb?

amneetb:
@nishantnangia : Looks good, but did you test it with order of accuracy case? Could you do that? Also do you want commit more examples later to this branch?

boyceg:
Other examples can be added in another branch. (This is more than just examples, since it adds in the CUI convective operator, which I want to test out, and some MG updates.)

boyceg:
However, it would be good if the code passes order-of-accuracy tests!!!

include/ibamr/INSVCStaggeredConservativeConvectiveOperator.h
```diff
@@ -291,15 +322,18 @@ class INSVCStaggeredConservativeConvectiveOperator : public ConvectiveOperator
     // Whether or not the current density field has been set.
     bool d_rho_is_set;
 
+    // Whether or not the current velocity field has been set.
```
boyceg: Do you need a state variable to track this?

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
```diff
@@ -366,7 +366,7 @@ class INSVCStaggeredHierarchyIntegrator : public INSHierarchyIntegrator
     /*!
      * \brief Get the scaling factor used for A, p and u_rhs
      */
-    inline double getPressureScalingFactor() const
+    inline SAMRAI::tbox::Array<double> getPressureScalingFactor() const
```
boyceg: Seems like this is not just a pressure scaling factor?

src/IB/ConstraintIBMethod.cpp
```diff
@@ -635,7 +636,7 @@ ConstraintIBMethod::initializeHierarchyOperatorsandData()
     const bool from_restart = RestartManager::getManager()->isFromRestart();
     if (!from_restart) calculateVolumeElement();
     setInitialLagrangianVelocity();
-
+
```
boyceg: 👎 Are you clang-formatting? :-D

nishantnangia:
@amneetb @boyceg I tested it with an order of accuracy case before committing and everything looked good.

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
nishantnangia: I will change to getScalingFactor()

src/IB/ConstraintIBMethod.cpp
nishantnangia: I am, but maybe this got committed earlier without clang formatting. Let me force clang-format on this file.

nishantnangia:
Makes sense. I think I had added them while I was debugging stuff. I have removed them in the latest commit and re-tested my manufactured solution case again.

boyceg:
@amneetb, @nishantnangia, is this ready to merge?

include/ibamr/INSVCStaggeredConservativeConvectiveOperator.h
```diff
@@ -196,6 +197,29 @@ class INSVCStaggeredConservativeConvectiveOperator : public ConvectiveOperator
...
+     * If V_old_idx or V_new_idx are not set, then they will degenerate to V_current automatically, for the very first
+     * simulation time step and cases where an INS cycle has not been executed, respectively.
+     */
+    void setFluidVelocityPatchDataIndices(int V_old_idx, int V_current_idx, int V_new_idx);
+
+    /*!
+     * \brief Set the cycle number currently being executed by the INS integrator. This will determine the rho advection
+     * velocity.
+     */
+    void setCycleNumber(int cycle_num);
```
boyceg: Can this be removed in lieu of the approach used in the other convective operators?

include/ibamr/INSVCStaggeredConservativeConvectiveOperator.h
```diff
@@ -196,6 +197,29 @@ class INSVCStaggeredConservativeConvectiveOperator : public ConvectiveOperator
...
+
+    /*!
+     * \brief Set the previous time step value between times n - 1 (old) and n (current). This is used during velocity
+     * extrapolation.
+     */
+    void setPreviousTimeStepSize(double dt_prev);
```
boyceg: Velocity extrapolation should not be done in this operator.

include/ibamr/INSVCStaggeredConservativeConvectiveOperator.h
```diff
@@ -275,31 +299,35 @@ class INSVCStaggeredConservativeConvectiveOperator : public ConvectiveOperator
         const double& dt,
         const double* const dx);
 
+    /*!
+     * \brief Enforce divergence free condition at the coarse-fine interface to ensure conservation of mass.
+     */
+    void enforceDivergenceFreeConditionAtCoarseFineInterface(const int U_idx);
```
boyceg: Is this something that also should be done in the constant-coefficient Navier-Stokes solver?

include/ibamr/INSVCStaggeredConservativeHierarchyIntegrator.h
```diff
@@ -169,6 +169,11 @@ class INSVCStaggeredConservativeHierarchyIntegrator : public INSVCStaggeredHiera
      */
     void registerMassDensitySourceTerm(SAMRAI::tbox::Pointer<IBTK::CartGridFunction> S_fcn);
 
+    /*!
+     * Returns the number of cycles to perform for the present time step.
+     */
+    int getNumberOfCycles() const;
```
boyceg: This is already implemented in the parent class. Why does it need to be reimplemented here?

include/ibamr/StaggeredStokesPhysicalBoundaryHelper.h
```diff
@@ -92,16 +92,26 @@ class StaggeredStokesPhysicalBoundaryHelper : public IBTK::StaggeredPhysicalBoun
...
      * the physical boundary.
+     *
+     * \note If enforce_at_all_phy_bdrys = true, then the discrete divergence-free condition will be enforced
+     * at all boundaries.
      */
-    void enforceDivergenceFreeConditionAtBoundary(int u_data_idx, int coarsest_ln = -1, int finest_ln = -1) const;
+    void enforceDivergenceFreeConditionAtBoundary(int u_data_idx,
+                                                  int coarsest_ln = -1,
+                                                  int finest_ln = -1,
+                                                  bool enforce_at_all_phy_bdrys = false) const;
```
boyceg: This last flag seems redundant. Isn't that what this function is supposed to do anyway?

include/ibamr/StaggeredStokesPhysicalBoundaryHelper.h
```diff
@@ -92,16 +92,26 @@ class StaggeredStokesPhysicalBoundaryHelper : public IBTK::StaggeredPhysicalBoun
...
      * enforce the discrete divergence-free condition in the ghost cell abutting
      * the physical boundary.
+     *
+     * \note If enforce_at_all_phy_bdrys = true, then the discrete divergence-free condition will be enforced
+     * at all boundaries.
      */
     void enforceDivergenceFreeConditionAtBoundary(SAMRAI::tbox::Pointer<SAMRAI::pdat::SideData<NDIM, double> > u_data,
-                                                  SAMRAI::tbox::Pointer<SAMRAI::hier::Patch<NDIM> > patch) const;
+                                                  SAMRAI::tbox::Pointer<SAMRAI::hier::Patch<NDIM> > patch,
+                                                  bool enforce_at_all_phy_bdrys = false) const;
```
boyceg: Same comment as above...

src/navier_stokes/INSVCStaggeredHierarchyIntegrator.cpp
```diff
@@ -438,14 +438,30 @@ INSVCStaggeredHierarchyIntegrator::INSVCStaggeredHierarchyIntegrator(const std::
                                  << "  valid choices are: VC_HARMONIC_INTERP, VC_AVERAGE_INTERP\n");
     }
 
-    // Get the scaling coefficient
-    d_A_scale = 1.0;
-    if (input_db->keyExists("condition_no_scaling")) d_A_scale = input_db->getDouble("condition_no_scaling");
-    if (d_A_scale <= 0.0)
+    // Get the scaling coefficients
+    if (input_db->keyExists("condition_no_scaling"))
```
boyceg: Does this actually make a difference in practice?

boyceg:
Actually, I just took another look, and I still have a few questions about the interface.

src/navier_stokes/INSVCStaggeredHierarchyIntegrator.cpp
amneetb: Yes, in extreme cases of viscosity and density ratios

include/ibamr/StaggeredStokesPhysicalBoundaryHelper.h
amneetb: The flag enforces div free condition in the ghost cells. The original IBAMR routine does it for just traction BC. We find it necessary for all BCs to get 2nd order convergence and stability for multiphase flows.

include/ibamr/StaggeredStokesPhysicalBoundaryHelper.h
amneetb: The flag enforces div free condition in the ghost cells. The original IBAMR routine does it for just traction BC. We find it necessary for all BCs to get 2nd order convergence and stability for multiphase flows.

include/ibamr/INSVCStaggeredConservativeHierarchyIntegrator.h
amneetb: The conservative convective class does not use standard IBAMR time stepping MIDPOINT, EULER etc. The parent class checks that the convective scheme adheres to one of the above and faults when not. Hence, we override.

include/ibamr/INSVCStaggeredConservativeConvectiveOperator.h
amneetb: Maybe.

include/ibamr/INSVCStaggeredConservativeConvectiveOperator.h
amneetb: We need velocity extrapolation for the RK scheme for density update. There is a comment in this routine which says:

// Compute the convective derivative with the penultimate density and velocity

We compute div. (rho u u) from an intermediate/penultimate stage  of mass equation, i.e, intermediate value is more important than the actual density update completion. We need this for stability. Hence, we are doing it simultaneously in the same routine.

amneetb:
Avoiding design stalemate and moving on.

amneetb:
@nishantnangia : Can you run the order of accuracy test for my recent commit to see if I inadvertently
 screwed anything?

@boyceg : Made the changes.

include/ibamr/INSStaggeredCUIConvectiveOperator.h
```diff
@@ -0,0 +1,218 @@
+// Filename: INSStaggeredConvectiveOperator.h
```
boyceg: `INSStaggeredCUIConvectiveOperator.h` ?

include/ibamr/INSVCStaggeredConservativeHierarchyIntegrator.h
```diff
@@ -59,14 +60,16 @@ namespace IBAMR
  * for the incompressible Navier-Stokes equations on an AMR grid hierarchy, with variable
  * coefficients.
  *
- * This class always uses a conservative discretization of the form of the momentum equation
+ * This class integrates the conservative form of the momentum equation
```
boyceg: 👍

include/ibamr/INSVCStaggeredConservativeHierarchyIntegrator.h
```diff
@@ -169,6 +172,20 @@ class INSVCStaggeredConservativeHierarchyIntegrator : public INSVCStaggeredHiera
...
+    int getNumberOfCycles() const;
+
+    /*!
+     * Get the convective operator being used by the integrator class.
+     *
+     * \note The class employs INSVCStaggeredConservativeMassMomentum integrator
+     * to compute convective derivative. Therefore, ConvectiveOperator is a NULL
+     * object.
+     */
+    SAMRAI::tbox::Pointer<ConvectiveOperator> getConvectiveOperator();
```
boyceg: I am a little confused by the comment here; if this is always NULL, do we need this method?

include/ibamr/INSVCStaggeredConservativeHierarchyIntegrator.h
```diff
@@ -169,6 +172,20 @@ class INSVCStaggeredConservativeHierarchyIntegrator : public INSVCStaggeredHiera
...
 
+    /*!
+     * Returns the number of cycles to perform for the present time step.
+     */
+    int getNumberOfCycles() const;
+
+    /*!
+     * Get the convective operator being used by the integrator class.
+     *
+     * \note The class employs INSVCStaggeredConservativeMassMomentum integrator
```
boyceg: Please make sure the class name (`INSVCStaggeredConservativeMassMomentumIntegrator`?) is spelled out completely to get linkage in Doxygen.

include/ibamr/INSVCStaggeredConservativeConvectiveOperator.h
amneetb: It prevents density fluctuations and keeps it bounded. Same is true at physical BCs. Hence, we do it for all BCs.

include/ibamr/INSStaggeredCUIConvectiveOperator.h
amneetb: uh-huh 🤓

include/ibamr/StaggeredStokesPhysicalBoundaryHelper.h
boyceg: Can you update the comment for the method not to start with "At open boundaries...". Maybe it could say something like:

"Set normal velocity ghost cell values to enforce ... . The default behavior is to set these values only in cells adjacent to boundary locations where normal traction conditions are imposed."

Finally, is there any possibility that someone may want only to do this at locations where normal velocity boundary conditions are imposed? Seems unlikely, but if so, maybe it would be better to use a bitmask that controls where these values are set that defaults to a value named `NORMAL_TRACTION` but that could be or'ed with another value named `NORMAL_VELOCITY`? (It seems preferable to me to have named quantities instead of boolean values, to make it more clear what is being controlled, but the boolean also is OK here.)

include/ibamr/INSVCStaggeredConservativeHierarchyIntegrator.h
amneetb: Yep😎. Since we are not using any convective op now, we set it to NULL. No point in having a default convective op hanging around and waste initialization compute time.

src/navier_stokes/INSVCStaggeredHierarchyIntegrator.cpp
boyceg: Interesting.

Can we change the key name to something like `operator_scale_factors`? I think that `condition_no_scaling` is a little bit opaque. If you want to use condition number in the name, please switch it to `num` instead of `no` --- `condition_no_scaling` makes it seem like "no scaling" is being done, when it is in fact the opposite.

include/ibamr/INSVCStaggeredConservativeHierarchyIntegrator.h
amneetb: 🤫I will fix it...

include/ibamr/INSVCStaggeredConservativeHierarchyIntegrator.h
boyceg: Maybe this method eventually needs to be pushed down into a constant-coefficient subclass of the parent class...

include/ibamr/INSVCStaggeredConservativeConvectiveOperator.h
boyceg: I guess my only question is whether these values could be accessed through a pointer to the parent `HierarchyIntegrator`, instead of having an interface to set the values. If you can get them from the parent integrator, it seems easier just to get as needed, rather than requiring explicit passing from the parent class.

src/navier_stokes/INSVCStaggeredHierarchyIntegrator.cpp
amneetb: Makes sense. `operator_scale_factors` makes it more clear. @nishantnangia : make this change when you get a chance. I could miss a spot or two.

boyceg:
Updates look good. I am good to merge once you guys are done with any final changes.

include/ibamr/StaggeredStokesPhysicalBoundaryHelper.h
amneetb: Hmm, I can update the comment. Do you want to take care of this named variables and bit mask thing? I don't have a good understanding of how bit mask thing works😬

include/ibamr/INSVCStaggeredConservativeHierarchyIntegrator.h
amneetb: We actually use a convective op for non-conservative form. Density is multiplied to (u. grad u) term obtained from convective op.

include/ibamr/INSVCStaggeredConservativeConvectiveOperator.h
amneetb: Hmm, the only way I see is via friendship. This has to be done down the lane as friendship of base class (future TimeIntegrator base class) won't make the derived class a friend of XXXHierarchyIntegrator. At that point, one may just specialize the XXXTimeIntegrator to obtain patch indices of XXXHierarchyIntegrator.

include/ibamr/StaggeredStokesPhysicalBoundaryHelper.h
boyceg: I just pushed some code that is meant to do this. Take a look, and feel free to change the tag names.

amneetb:
why is `examples/multiphase_flow/ex1/input2d.test` causing a conflict whereas `examples/multiphase_flow/ex1/input2d.cyl` or `examples/multiphase_flow/ex1/input2d_free_fall.cyl` not for identical change?

src/navier_stokes/INSVCStaggeredHierarchyIntegrator.cpp
nishantnangia: Will do.

nishantnangia:
I will test out the order of accuracy and deal with the conflicts later this afternoon or evening. Once confirmed, I'll let you guys know.

drwells:
@amneetb commit 4d90d765a9dafa4ae046e627c2efc6ea8d86a023 fixed that test recently.

boyceg:
@drwells, I think we need to have developer documentation (a wiki page?) talking about how to update test input files, and what other files need to be updated.

include/ibamr/StaggeredStokesPhysicalBoundaryHelper.h
amneetb: Thanks! Looks good to me. Nishant will run suite-of-tests with latest commit to confirm order of accuracy and stability for high density ratio flows. Once passed you can merge it.

boyceg:
Looks good. Let me know if you are ready to merge.

amneetb:
@boyceg We have a script for the the order of accuracy test to extract slopes for various variables. I think we should add it in the example directory for a record. What do you think?

boyceg:
Fine with me. It could go in this PR or a separate one.

nishantnangia:
Just added a very simple error extraction script, which takes the errors from the log file and puts them into appropriate txt files. The user could put it in a loop that runs each example with a different grid spacing to carry out a convergence test.

boyceg:
Great.

Good to go on the merge?

amneetb:
Running multiphase_flows/ex2 for finite elements on dogwood for optimizing smeared width. Will let you know when completed.

amneetb:
@boyceg : Good to merge.

PR #270, opened 2018-08-01
Convert some pointers to mutable variables.

drwells:
The present implementation works since modifying a variable through a pointer is permissible in a const function.

Since we write to these variables and then read them through Eval(), it is reasonable to mark them as mutable since the value set at by the end of the function won't be read again after the function returns.

Mentioned in #258.


drwells:
Jenkins retest this please

boyceg:
Jenkins retest this please

boyceg:
Jenkins retest this please

PR #271, opened 2018-08-02
Fixing ConstraintIB, level set, and vc gtests

nishantnangia:
Fixing some of the gtests mentioned in https://github.com/IBAMR/IBAMR/issues/269


drwells:
That was quick! Nice!

drwells:
I don't think that Travis actually runs the tests now (otherwise we would have noticed the failures). I'll run these on my workstation and report back once I can confirm that gtest is happy.

drwells:
Yup, these all work!

PR #273, opened 2018-08-06
Restore compatibility with libMesh 1.0.0.

drwells:
Commit 6abf9f7fc35 added calls to a libmesh function not present in 1.0.0, which gets the indices of the degrees of freedom corresponding to nodal value functionals.

This commit adds a wrapper function that either calls (if the detected version of libMesh is new enough) the libMesh function or, as a fallback, uses our own implementation of it.


drwells:
AFAICT there is no tutorial program and/or example program that uses this nodal interpolation yet. @boyceg is one in the works?

boyceg:
Jenkins retest this please.

PR #276, opened 2018-08-08
Fix compatibility with libMesh 1.3.0.

drwells:
The latest version of libMesh uses forward declarations of some enumerations, so we need to explicitly include the header files.

This is the only patch I needed to compile with the latest release of libMesh (i.e., 1.3.0).

These headers exist in libMesh 1.0.0 and 1.2.1 (I did not check 1.1.0).


boyceg:
Yeah the old builds are out of control.

We could reduce file sizes by switching to shared libraries. I have a strong aversion to libtool shared libraries — I’ve had miserable experiences with them — and haven’t gotten around to finding a robust way to support both shared and static linking.

drwells:
Adding support for shared libraries is on my post-0.3 checklist. We should chat at some point in a month about the strange things libtool can do.

I am not too familiar with Jenkins, but the CI system for deal.II only saves the log file and deletes everything else IIRC. How hard would it be to get Jenkins to do that?

boyceg:
I think I now have Jenkins set up to delete old builds, but I think it is confused about the state of the build directory. I don’t currently have permissions to nuke those directories, which I think is needed at this point.

boyceg:
Looks like a weird compiler bug caused GCC Toolchain checks to fail. Jenkins retest this please.

drwells:
Is it possible that we just ran out of disk space again?

boyceg:
I think that the server is overloaded. I am going to try restarting it at a smaller number of executors on the VM.

boyceg:
Jenkins retest this please.

PR #277, opened 2018-08-09
Fix the adv_diff ex0 3D test.

drwells:
The test itself records that it compares new error calculations to those measured in October 2016: however, the input file is no longer the same (`DT` and `END_TIME`, among other things, are now different) as it was  then. This commit fixes the input file to match the one used to generate the data in 2016 (i.e., near commit cb7b841f1a9).

The test passes with this fix.


drwells:
That was quick!

boyceg:
No need to wait on Jenkins here...

PR #278, opened 2018-08-09
Initialize IBStandardInitializer in its ctor.

drwells:
This fixes an issue related to the current failure of the `CIB/ex0` test: if, in that test, output is disabled, then `init()` is never run, which leads to failed assertions later on. Since `init()` does not rely on any data not set in the constructor (i.e., the class members are set by `getFromInput`) we can run it there.

### DIFF
--- a/examples/CIB/ex0/input2d.test
+++ b/examples/CIB/ex0/input2d.test
@@ -271,7 +271,10 @@ Main {
 
 // visualization dump parameters
    viz_writer                  = "VisIt","Silo"
-   viz_dump_interval           = int(END_TIME/(3*DT))
+   // Do not write any output (previous versions of IBStandardInitializer had a
+   // bug where setting this to zero triggered an abort, so its set to zero now
+   // to ensure that that bug is fixed)
+   viz_dump_interval           = 0
    viz_dump_dirname            = "viz_cylinder2d"
    visit_number_procs_per_file = 1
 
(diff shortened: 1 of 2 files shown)


boyceg:
This seems fine for `IBStandardInitializer`, but it might be worth taking a quick look at `IBRedundantInitializer` to make sure that the logic seems reasonable there for initializing things.

@abarret, I will wait to hear from you before merging.

abarret:
This fix won't work with `IBRedundantInitializer` because functions to initialize the data are registered after the object is constructed. 

The issue seems to be that there are calls to `getLevelHasLagrangianData` before `init` is called. However, we can't call `init` in `getLevelHasLagrangianData` because it's marked as `const`. If we remove this restriction in `LInitStrategy`, we can fix the issue. Is there any reason this function needs to be labeled as `const`?

boyceg:
The original implementation had all data read in by the constructor, so a lot of the functions were just looking up existing data --- hence `const`. I clearly wasn't thinking about dynamically setting up data. Take a look at `LInitStrategy` and see which methods are currently `const` that perhaps shouldn't be, and we can update accordingly.

boyceg:
I believe that this PR is superseded by #279; I am going to close it.

drwells:
It would still be good to set up a test that doesn't write any output so that we know this worked. I'll put up a new PR at some point tomorrow (I don't have the testing infrastructure set up at home).

PR #279, opened 2018-08-10
Update const function in LInitStrategy and add init call to IBRedundantInitializer

abarret:
This fixes the issues addressed in #278 for `IBRedundantInitializer`. It removes `const` from a member function in `LInitStrategy` and adds an extra `init` call to `IBRedundantInitializer`.

This change should fix the issue for `IBStandardInitializer` as well.

### DIFF
--- a/ibtk/include/ibtk/LInitStrategy.h
+++ b/ibtk/include/ibtk/LInitStrategy.h
@@ -112,6 +112,13 @@ public:
                                       bool initial_time) = 0;
 
     /*!
+     * \brief Initialize structure specific configurations.
+     *
+     * \note A default empty implementation is provided.
+     */
+    virtual void init() {};
+
+    /*!
      * \brief Initialize the structure indexing information on the patch level.
      *
      * \note A default empty implementation is provided.
--- a/ibtk/include/ibtk/LDataManager.h
+++ b/ibtk/include/ibtk/LDataManager.h
@@ -476,6 +476,9 @@ public:
     /*!
      * Register a concrete strategy object with the integrator that specifies
      * the initial configuration of the curvilinear mesh nodes.
+     *
+     * \note This function calls LInitStrategy::init(). All preprocessing should be completed before
+     * registering a LInitStrategy object.
      */
     void registerLInitStrategy(SAMRAI::tbox::Pointer<LInitStrategy> lag_init);
 
--- a/ibtk/src/lagrangian/LDataManager.cpp
+++ b/ibtk/src/lagrangian/LDataManager.cpp
@@ -845,6 +845,7 @@ LDataManager::registerLInitStrategy(Pointer<LInitStrategy> lag_init)
     TBOX_ASSERT(lag_init);
 #endif
     d_lag_init = lag_init;
+    d_lag_init->init();
     return;
 } // registerLInitStrategy
 
--- a/include/ibamr/IBRedundantInitializer.h
+++ b/include/ibamr/IBRedundantInitializer.h
@@ -483,6 +483,13 @@ public:
      */
     void setStructureNamesOnLevel(const int& level_num, const std::vector<std::string>& strct_names);
 
+    /*!
+     * \brief Initialize structure specific configurations.
+     *
+     * \note All functions should be registered with the object before init is called.
+     */
+    virtual void init();
+
 protected:
     /*!
      * \brief Default constructor.
@@ -510,11 +517,6 @@ protected:
      * \return A reference to this object.
      */
     IBRedundantInitializer& operator=(const IBRedundantInitializer& that);
-
-    /*!
-     * \brief Initialize structure specific configurations.
-     */
-    virtual void init();
 
     /*!
      * \brief Configure the Lagrangian Silo data writer to plot the data
--- a/include/ibamr/IBStandardInitializer.h
+++ b/include/ibamr/IBStandardInitializer.h
@@ -436,13 +436,13 @@ public:
      */
     ~IBStandardInitializer();
 
-protected:
-private:
     /*!
      * \brief Initialize structure specific configurations.
      */
     void init();
 
+protected:
+private:
     /*!
      * \brief Default constructor.
      *
@@ -545,11 +545,6 @@ private:
      * and for error reporting purposes.
      */
     std::string d_object_name;
-
-    /*!
-     * Checking if user defined data has been processed./
-     */
-    bool d_data_processed;
 
     /*
      * The boolean value determines whether file read batons are employed to
--- a/src/IB/IBRedundantInitializer.cpp
+++ b/src/IB/IBRedundantInitializer.cpp
@@ -176,8 +176,9 @@ IBRedundantInitializer::registerLSiloDataWriter(Pointer<LSiloDataWriter> silo_wr
     // restart file.
     if (!is_from_restart)
     {
-        // Check if data has been processed.
-        init();
+#if !defined(NDEBUG)
+        TBOX_ASSERT(d_data_processed);
+#endif
 
         for (int ln = 0; ln < d_max_levels; ++ln)
         {
@@ -193,6 +194,10 @@ IBRedundantInitializer::registerLSiloDataWriter(Pointer<LSiloDataWriter> silo_wr
 bool
 IBRedundantInitializer::getLevelHasLagrangianData(const int level_number, const bool /*can_be_refined*/) const
 {
+#if !defined(NDEBUG)
+    TBOX_ASSERT(d_data_processed);
+#endif
+
     return !d_num_vertex[level_number].empty();
 } // getLevelHasLagrangianData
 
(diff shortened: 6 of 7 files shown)


boyceg:
@drwells could this supersede #278?

drwells:
I don't think that this is enough to replace the other PR: with this patch (and not #278) I still get the same error when I run `CIB/ex0` with `viz_dump_interval = 0`. Regardless, though, the main point is good: we should address the problem here first before moving on to #278.

This seems like only a partial fix for the problem (will all functions ultimately have to call `init()`?), and it might break user codes if they call a formerly `const` function inside their own `const` function. Why don't we add checks similar to
```cpp
if (!d_data_processed)
  TBOX_ASSERT("init() must be called before calling this function.");
```

which is what we do in, e.g., `initializeStructurePosition`, to ensure that `init()` (which, as things presently stand, can only be called once anyway) is called before we need it?

drwells:
To follow up: one alternative fix would be to change the examples to do `ib_initializer->init()` before, e.g., `ib_initializer->registerLSiloDataWriter(...)`

boyceg:
@drwells, my preference would be not to require users to add calls to `init()`. Is that preference compatible with not changing the `const`-ness of any member functions in the interface? If not, I think my preference is to change `const`-ness.

boyceg:
Or, if we really want to avoid changing the interface, could the explicit call to `init()` be done in `IBMethod`? I want to try to avoid changes in user code.

drwells:
Hmm. I agree that we should go with the solution that doesn't require changes to user code. I don't mind having it called implicitly by the attach functions (which is what currently happens); I really just want either an explicit error message explaining what goes wrong in the no-output case (i.e., where `init` never gets called) or for `init` to be called at the soonest possible instance.

I will spend some time reading over the implementations of these two classes so that I understand the entry points better. I will post again once I know where things go.

drwells:
I spent some time looking over this. I am certainly not an expert in these two classes so some of the details may be wrong, but the general picture should be right.
1. We should definitely not call `init` in the constructor of `IBRedundantInitializer` since `IBRedundantInitializer::init` uses a bunch of function pointers, passed in after the constructor, to do initialization. (We should think about adding another constructor that sets up everything at once, but lets worry about that later).
2. `IBStandardInitializer` never calls any of the `IBRedundantInitializer::d_init_*_on_level_fcn` functions. In particular, since `IBStandardInitializer` never calls `IBRedundantInitializer::init()`, none of its structural information gets set up (at least in the way `IBRedundantInitializer` would do it).
3. `IBStandardInitializer` makes very little use of functions offered by `IBRedundantInitializer` at all, since it does not use its `init` function: AFAICT inheritance just gets us the constructor (which is mostly duplicated in its own constructor anyway), `registerLSiloDataWriter`, and all of its protected member variables (which are used, just set up differently).
4. There are no public functions of `IBStandardInitializer` that modify its state prior to when `init` should be called; i.e., there is no analog to the `d_init_*_on_level_fcn` functions (and those are ignored by `IBStandardInitializer::init`).

Hence: the original approach for `IBStandardInitializer` seems reasonable to me: there does not appear to be anything set after construction which could possibly alter what this function does so we may as well call it immediately. As for `IBRedundantInitializer`: calling any method aside from `getLevelHasLagrangianData`, the registration functions, and things called only by `init` will call `init`. Hence: by adding a check in `getLevelHasLagrangianData`, keeping the status quo re: calls to init, *and* by making `init` public we can guarantee that the class is properly initialized before we do anything. We could make `IBStandardInitializer::init` also public (and document that it doesn't do anything) to make this consistent.

In summary, I think we should:
1. keep #278 
2. make both `IBStandardInitializer::init` and `IBRedundantInitializer::init` public
3. add `TBOX_ASSERT`s or equivalents to any functions that don't call `init` on their own to clarify the situation to errant users (such as myself)

This should be completely backwards compatible (including possible const correctness issues). I doubt most users will encounter this since the objects automatically set themselves up as soon as needed (with the exception of `getLevelHasLagrangianData`). The `TBOX_ASSERT`s would only guard code that would subsequently crash or return garbage anyway, so they should not introduce any breakage.

boyceg:
> On Aug 13, 2018, at 4:29 PM, David Wells <notifications@github.com> wrote:
> 
> I spent some time looking over this. I am certainly not an expert in these two classes so some of the details may be wrong, but the general picture should be right.
> 
> We should definitely not call init in the constructor of IBRedundantInitializer since IBRedundantInitializer::init uses a bunch of function pointers, passed in after the constructor, to do initialization. (We should think about adding another constructor that sets up everything at once, but lets worry about that later).
Yes.
> IBStandardInitializer never calls any of the IBRedundantInitializer::d_init_*_on_level_fcn functions. In particular, since IBStandardInitializer never calls IBRedundantInitializer::init(), none of its structural information gets set up (at least in the way IBRedundantInitializer would do it).
Yes, `IBStandardInitializer` reads in all data from disk in its constructor.
> IBStandardInitializer makes very little use of functions offered by IBRedundantInitializer at all, since it does not use its init function: AFAICT inheritance just gets us the constructor (which is mostly duplicated in its own constructor anyway), registerLSiloDataWriter, and all of its protected member variables (which are used, just set up differently).
Redundant here is intended to indicate that initialization data are redundantly provided on all processors. This has implications on the scalability of initialization, but has the benefit of allowing a relatively straightforward implementation.

I think that the rationale for deriving `IBStandardInitializer` from `IBRedundantInitializer` is to indicate that the “standard” initializer provides “redundant” data.

I suppose that both could just derive directly from the `LInitStrategy` base class — although, in that case, we may want to rename `IBRedundantInitializer`?
> There are no public functions of IBStandardInitializer that modify its state prior to when init should be called; i.e., there is no analog to the d_init_*_on_level_fcn functions (and those are ignored by IBStandardInitializer::init).
> Hence: the original approach for IBStandardInitializer seems reasonable to me: there does not appear to be anything set after construction which could possibly alter what this function does so we may as well call it immediately.
> 
Sounds fine.
> As for IBRedundantInitializer: calling any method aside from getLevelHasLagrangianData, the registration functions, and things called only by init will call init. Hence: by adding a check in getLevelHasLagrangianData, keeping the status quo re: calls to init, and by making init public we can guarantee that the class is properly initialized before we do anything. We could make IBStandardInitializer::init also public (and document that it doesn't do anything) to make this consistent.
> 
> In summary, I think we should:
> 
> keep #278 <https://github.com/IBAMR/IBAMR/pull/278>
> make both IBStandardInitializer::init and IBRedundantInitializer::init public
> add TBOX_ASSERTs or equivalents to any functions that don't call init on their own to clarify the situation to errant users (such as myself)
Sounds good. @abarret, can you finish this up?

abarret:
If I'm understanding the suggestion, this fix will still require explicit calls to `init` in the main routine if `viz_dump_interval` is set to zero (for the examples at least), although there will be a descriptive error message. Is this correct?

boyceg:
Yes, that is right, with `IBRedundantInitializer`, it will be necessary to call `init()` before calling `getLevelHasLagrangianData()`.

However, you may be able to make that call in `LDataManager` (or possibly in `IBMethod` and `IMPMethod`). For instance, we could have `LDataManager` call `init()` in `LDataManager::registerLInitStrategy()`. This would just require that the initialization object be set up before it is registered, which does not seem like an onerous requirement.

boyceg:
By the way, if `LDataManager` calls `init()` in `registerLinitStrategy()`, then we could remove all of the calls to `init()` in `IBXXXInitializer` and just have `TBOX_ASSERT`ions to assert that the data have been set up any place that they are needed.

abarret:
@boyceg @drwells I moved the `init` call to `LDataManager` and added `TBOX_ASSERT` to the implementations of `LInitStrategy`. I tested with IB/ex0 and CIB/ex0 and they both run. Take a look and let me know if this looks ok.

ibtk/include/ibtk/LInitStrategy.h
```diff
@@ -111,6 +111,11 @@ class LInitStrategy : public virtual SAMRAI::tbox::DescribedClass
                                       bool can_be_refined,
                                       bool initial_time) = 0;
 
+    /*!
+     * \brief Initialize structure specific configurations.
+     */
+    virtual void init() = 0;
```
boyceg: Shall we provide a default empty implementation?

boyceg:
I think this looks great. Let's add a default implementation of `init()` and call it a day.

ibtk/src/lagrangian/LInitStrategy.cpp
```diff
@@ -139,6 +139,14 @@ LInitStrategy::tagCellsForInitialRefinement(const Pointer<PatchHierarchy<NDIM> >
     return;
 } // tagCellsForInitialRefinement
 
+void
+LInitStrategy::init()
+{
+    TBOX_WARNING("LInitStrategy::init()\n"
```
boyceg: I think there is no need to warn here --- just leave it empty, and note in the comments in the header that a default empty implementation is provided.

ibtk/include/ibtk/LInitStrategy.h
boyceg: With the default implementation, this method should not be purely virtual (i.e., please remove ` = 0`).

boyceg:
@abarret, couple small changes, then squash and I will merge.

abarret:
Made those changes and squashed.

PR #281, opened 2018-08-12
Removing exodus io from IBFEInstrumentPanel class

cpuelz:
removing exodus io and replacing probably with output to STLs.

### DIFF
--- a/include/ibamr/IBFEInstrumentPanel.h
+++ b/include/ibamr/IBFEInstrumentPanel.h
@@ -126,6 +126,11 @@ private:
      * \brief write out meshes and equation systems in Exodus file.
      */
     void outputExodus(IBAMR::IBFEMethod* ib_method_ops, int timestep, double loop_time);
+    
+    /*!
+     * \brief write out meshes in STL file.
+     */
+    void outputSTL(IBAMR::IBFEMethod* ib_method_ops, int timestep, double loop_time);
 
     /*!
      * \brief write out nodes.
--- a/include/ibamr/IBFEInstrumentPanel.h
+++ b/include/ibamr/IBFEInstrumentPanel.h
@@ -114,7 +114,7 @@ public:
     const libMesh::MeshBase& getMeterMesh(const unsigned int jj) const;
     
     /*!
-     * \return A reference to the EquationSystem object for jjth meter mesh
+     * \return A reference to the EquationSystems object for jjth meter mesh
      */
     const libMesh::EquationSystems& getMeterMeshEquationSystems(const unsigned int jj) const;
     
--- a/include/ibamr/IBFEInstrumentPanel.h
+++ b/include/ibamr/IBFEInstrumentPanel.h
@@ -99,24 +99,39 @@ public:
                             double data_time);
 
     /*!
-     * \return get the instrument dump interval.
+     * \return get the instrument dump interval
      */
     int getInstrumentDumpInterval() const;
     
     /*!
+     * \return get the number of meter meshes
+     */
+    int getNumberOfMeterMeshes() const;
+    
+    /*!
      * \return get a pointer to the jjth meter mesh
      */
-    libMesh::SerialMesh* getMeterMesh(const unsigned int jj) const;
+    libMesh::MeshBase& getMeterMesh(const unsigned int jj) const;
     
     /*!
      * \return get a pointer to the EquationSystem object for jjth meter mesh
      */
-    libMesh::EquationSystems* getMeterMeshSystem(const unsigned int jj) const;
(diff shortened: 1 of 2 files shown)


boyceg:
Instead of STL, can you just give access to the meshes themselves, so that users can choose any output format?

cpuelz:
Oh okay that sounds good. Like have a getMeterMeshes function?

Get Outlook for iOS<https://aka.ms/o0ukef>

cpuelz:
I went back to look at our previous comments in #272 .  I think I know what to do.

cpuelz:
@boyceg @drwells I added some functions to get mesh and system members for the meter objects, and removed things having to do with exodus.  what do you think?

drwells:
Looks good to me.

I think it would be preferable, though, to return references instead of pointers so that no one is tempted to double-delete the libmesh data structures.

include/ibamr/IBFEInstrumentPanel.h
```diff
@@ -100,15 +99,25 @@ class IBFEInstrumentPanel
...
-
+    int getInstrumentDumpInterval() const;
+    
     /*!
-     * \return return the instrument dump interval.
+     * \return get a pointer to the jjth meter mesh
      */
-    int getInstrumentDumpInterval() const;
-
+    libMesh::SerialMesh* getMeterMesh(const unsigned int jj) const;
```
boyceg: It looks like the meshes are created only once, and so it should be safe to replace this return type by `libMesh::MeshBase&`.

include/ibamr/IBFEInstrumentPanel.h
```diff
@@ -100,15 +99,25 @@ class IBFEInstrumentPanel
...
+    
+    /*!
+     * \return get a pointer to the EquationSystem object for jjth meter mesh
+     */
+    libMesh::EquationSystems* getMeterMeshSystem(const unsigned int jj) const;
```
boyceg: As above, let's change the return type to `libMesh::EquationSystems&`.

Also, let's go ahead and change the method name to `getMeterMeshEquationSystems`, since `System` is a separate concept from `EquationSystems` in libMesh.

src/IB/IBFEInstrumentPanel.cpp
```diff
@@ -313,11 +312,10 @@ IBFEInstrumentPanel::IBFEInstrumentPanel(SAMRAI::tbox::Pointer<SAMRAI::tbox::Dat
       d_nodes(),
       d_node_dof_IDs(),
       d_meter_systems(),
-      d_exodus_io(),
       d_meter_meshes(),
       d_meter_mesh_names(),
       d_nodeset_IDs_for_meters(),
-      d_instrument_dump_interval(),
+      d_instrument_dump_interval(),
```
boyceg: Let's remove this whitespace-only change.

src/IB/IBFEInstrumentPanel.cpp
```diff
@@ -563,6 +559,10 @@ IBFEInstrumentPanel::initializeHierarchyIndependentData(IBFEMethod* ib_method_op
             d_U_dof_idx[jj].push_back(U_dof_index);
         }
     }
+    
+    // output the nodes of each meter mesh into a .dat file
+    outputNodes();
```
boyceg: If the point is to handle output outside of the class, should this call to `outputNodes()` still be in the implementation?

include/ibamr/IBFEInstrumentPanel.h
boyceg: By the way, since this is indexed by the meter number, there needs to be a method for finding out the number of meters (e.g., `getNumberOfMeterMeshes()`). It doesn't look like such a function currently exists in the interface.

include/ibamr/IBFEInstrumentPanel.h
```diff
@@ -100,15 +99,40 @@ class IBFEInstrumentPanel
...
     /*!
-     * \return return the instrument dump interval.
+     * \return get the number of meter meshes
      */
-    int getInstrumentDumpInterval() const;
-
+    int getNumberOfMeterMeshes() const;
+    
+    /*!
+     * \return get a pointer to the jjth meter mesh
```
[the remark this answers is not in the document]
cpuelz: I also returned const references for the Mesh and EquationSystems objects.  Is that okay?

cpuelz:
I also returned const references for the Mesh and EquationSystems objects. Is that okay?

boyceg:
Unfortunately, the interface for the libMesh class `ExodusII_IO` requires a non-`const` reference to a `MeshBase` --- so you would need to return mutable (non-`const`) references to `MeshBase`. I would also return a non-`const` reference to the `EquationSystems` object.

cpuelz:
okay... made those changes.

PR #286, opened 2018-08-15
Fixing latex

amneetb:


boyceg:
I think it maybe is this: "\f$ dt \sim dx$ and \f$ 1/dx = N $”. Look out for mis-paired \f$ and $.

amneetb:
😬Check now (squashed and pushed)

boyceg:
Let's see if this makes Travis happy.

Issue #290, opened 2018-08-16
Some tests fail with abort


drwells:
I previously located failing tests by searching for gtest output. However, gtest doesn't print anything if we `abort` from inside IBAMR itself, so I did not catch these before.

- [x] `adv_diff/ex1/test2d` (`DT` is not defined in the database)
- [x] `adv_diff/ex1/test3d` (`DT` is not defined in the database)
- [x] `IBFE/explicit/ex5/test2d` (`C1_S` is not defined in the database)
- [x] `IBFE/explicit/ex6/test2d` (`C1_S` is not defined in the database)

PR #293, opened 2018-08-20
Fix aborting tests in adv_diff.

drwells:
The issue with these tests is that viz_dump_interval depends on DT, which is never defined. We can get around this by simply never writing output.

Part of #290.


boyceg:
...and not writing output is probably what most of the tests should be doing in the first place!

PR #294, opened 2018-08-21
Fix aborting IBFE tests

drwells:
It looks like we defined these tests and examples without adding all physical parameters to the input files. I added them and also cleaned up the implementation a bit: the relevant IBFE objects have nice support for pointers to context objects, so I moved the state there from global statically allocated variables.

This is enough of a fix to make the tests pass (and avoid basic errors, e.g., looking for variables in the database that were not defined in the input file) but the examples still fail with the default values. I will open separate issues for those.


drwells:
Partial fix for #290.

drwells:
Since #293 is merged this now closes #290.

PR #296, opened 2018-08-22
Run one test at stricter tolerances.

drwells:
Commit e6aeeb564ca changed the way we do a particular L2 projection in `IBFE/explicit/ex0`: instead of using a given initial guess for the solution we now zero the vector. This caused the computed solution to be slightly different from the known value.

This commit changes the test to run at a much stricter numerical tolerance, and compares it to values computed (with the same tolerances) before e6aeeb564ca.

We should ultimately find a better way to jointly handle PETSc and SAMRAI options with the input file or command line parameters, but I think we should delay that fix until after 0.3 is done.


drwells:
Fixes #269.

boyceg:
👍 

We should add prefixes to those libMesh KSPs, but let's do that after 0.3.

PR #301, opened 2018-08-29
Add parentheses suggested by -Wparentheses.

drwells:
While the present code is correct, the order of operations (`==`, `&&`, then `||`) is a bit esoteric, so GCC reasonably suggests adding parentheses.


boyceg:
👍

PR #307, opened 2018-09-04
Add unordered map headers

drwells:
We use `boost::unordered_map` in IBTK but the bundled copy of boost does not include the necessary headers. I added enough headers to get this working.

While I was here I deleted two unused boost `#include`s.

This is an issue on macs which usually do not have a copy of boost available in a system directory.

Part of #266.

### DIFF
--- a/ibtk/contrib/boost/boost/core/no_exceptions_support.hpp
+++ b/ibtk/contrib/boost/boost/core/no_exceptions_support.hpp
@@ -1,0 +1,44 @@
+#ifndef BOOST_CORE_NO_EXCEPTIONS_SUPPORT_HPP
+#define BOOST_CORE_NO_EXCEPTIONS_SUPPORT_HPP
+
+#if defined(_MSC_VER)
+#  pragma once
+#endif
+
+//----------------------------------------------------------------------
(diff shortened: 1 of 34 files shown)


drwells:
I am not sure how serious this is, but is very insiduous. One has to build libmesh without boost *and* not have a system copy of boost installed to find this problem.

drwells:
Thanks again to @masmath for help in finding all of these obscure bugs :)

boyceg:
I tried to setup `configure` to complain if we are using `libMesh` and *not* using an external Boost. The reason was to avoid version clashes between the bundled Boost libs from IBAMR and `libMesh`.

boyceg:
(When I set this up, I came to the conclusion that IBAMR could not simply use `libMesh`'s bundled Boost. This conclusion may have been reached in error.)

drwells:
We presently use (taking into account the present patch)

```
#include "boost/array.hpp"
#include "boost/math/special_functions/round.hpp"
#include "boost/multi_array.hpp"
#include <boost/unordered_map.hpp>
#include <boost/version.hpp>
```

It looks like libmesh 1.0.0 does not include `boost/array.hpp` or `boost/multi_array.hpp` so you were correct to bundle our own.

PR #311, opened 2018-09-04
Fix links in top level README.

abarret:
Part of #266.  This updates a few links to correctly direct to the `github.io` pages. Note that we claim in the README that there are visualization examples on the `github.io` pages (or we will once this is merged), but there are no such visualizations at this time.


boyceg:
I have updated the README.

drwells:
I clicked on all the links; they work!

boyceg:
I changed "here" to names of the links --- I prefer names of links since it is easy to gloss over "here".

boyceg:
Should this also go into master?

PR #313, opened 2018-09-05
Check for an incompatible libMesh setting.

drwells:
We cannot use libMesh when it is compiled with its own implementation of `boost::unique_ptr` due to conflicts with our own boost usage.

This change causes `configure` to print a useful message and quit early on instead of compilation crashing with difficult to understand (this took me most of the afternoon to figure out) boost template conflict errors.

### DIFF
--- a/configure
+++ b/configure
@@ -23452,6 +23452,56 @@ fi
   LIBMESH_FCFLAGS="`$LIBMESH_CONFIG --fflags`"
   LIBMESH_LIBS="`$LIBMESH_CONFIG --libs`"
 
+  # libMesh implements UniquePtr by including parts of boost in
+  # libmesh/unique_ptr.hpp (or in contrib/unique_ptr/unique_ptr.hpp). This
+  # causes compilation errors when one #includes both that file and the original
+  # file from boost, since the preprocessor guards on the libMesh-provided
+  # unique_ptr.hpp do not match the preprocessor guards on the boost headers (so
+  # we get multiple definition errors). To ensure that this does not happen,
+  # check that libMesh will not include unique_ptr.hpp in its own auto_ptr.hpp
+  # file (which we use). libMesh uses the preprocessor guard symbol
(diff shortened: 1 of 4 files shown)


drwells:
Note that we can be a bit more permissive, with boost 1.1 or newer, than just compiling with `--disable-unique-ptr`: as long as a copy of boost is available, newer versions of libMesh will use the complete boost implementation instead of duplicating parts of boost in their own file.

boyceg:
This only affects libMesh 1.0.0. More recent releases don’t seem to exhibit the same behavior.

drwells:
This is true on my machines, but it looks like libMesh 1.2.1 picks up my system copy of boost even when it is compiled with `--with-boost=no` (i.e., it defines `LIBMESH_HAVE_BOOST_MOVELIB_UNIQUE_PTR` but does not define `LIBMESH_HAVE_BOOST`), which may cause conflicts (though I have not yet observed them). I suspect (though I do not have mac or other machine without a system copy of boost available to test this on) that it is still possible to include the conflicting header when we compile libMesh 1.0, 1.1, or 1.2.1 without C++11 and without boost.

boyceg:
I think that I tried to set up the configuration stuff for libMesh to insist on using an external Boost for exactly this reason --- that _seemed_ like the easiest way to ensure that everything is using the same Boost lib.

boyceg:
It looks like this test is more robust, since it captures the actual problematic header.

drwells:
> I think that I tried to set up the configuration stuff for libMesh to insist on using an external Boost for exactly this reason --- that seemed like the easiest way to ensure that everything is using the same Boost lib.

Yes, this was a good choice.

> It looks like this test is more robust, since it captures the actual problematic header.

I think so. There are a lot of permutations in the boost configuration, but our troubles spring from this header regardless of what gets brought in to libmesh.

drwells:
Yes.

PR #316, opened 2018-09-10
Fixing out-of-bounds error for vectors

nishantnangia:
For certain compilers, one will hit the following assertion in debug mode when trying to access or pass `&some_vector[0]` when some_vector is empty:

```
/usr/include/c++/4.8.3/debug/vector:346:error: attempt to subscript 
    container with out-of-bounds index 0, but container only holds 0     
    elements.
```

I found this because some of the examples will crash in debug mode during the first time step. The changes in this PR fixes that.

For IBStandardForceGen, I think it's safer to include the return statement. This is already done for beams and springs, but it was missing for target points.

For  IBFEDirectForcingKinematics, nodal_X_values should be resized before it gets passed here:
https://github.com/IBAMR/IBAMR/blob/nishant/fix-vector-dbg-assert/src/IB/IBFEDirectForcingKinematics.cpp#L747

@boyceg @amneetb Thoughts?

This should probably go in 0.3 as well.

### DIFF
--- a/src/IB/IBStandardForceGen.cpp
+++ b/src/IB/IBStandardForceGen.cpp
@@ -1267,6 +1267,7 @@ IBStandardForceGen::computeLagrangianTargetPointForce(Pointer<LData> F_data,
     double max_displacement = 0.0;
 
     const int num_target_points = static_cast<int>(d_target_point_data[level_number].petsc_node_idxs.size());
+    if (num_target_points == 0) return;
     const int* const petsc_node_idxs = &d_target_point_data[level_number].petsc_node_idxs[0];
     const double** const kappa = &d_target_point_data[level_number].kappa[0];
     const double** const eta = &d_target_point_data[level_number].eta[0];
--- a/src/IB/IBStandardForceGen.cpp
+++ b/src/IB/IBStandardForceGen.cpp
@@ -1267,11 +1267,12 @@ IBStandardForceGen::computeLagrangianTargetPointForce(Pointer<LData> F_data,
     double max_displacement = 0.0;
 
     const int num_target_points = static_cast<int>(d_target_point_data[level_number].petsc_node_idxs.size());
-    if (num_target_points == 0) return;
-    const int* const petsc_node_idxs = &d_target_point_data[level_number].petsc_node_idxs[0];
-    const double** const kappa = &d_target_point_data[level_number].kappa[0];
-    const double** const eta = &d_target_point_data[level_number].eta[0];
-    const Point** const X0 = &d_target_point_data[level_number].X0[0];
+    const bool uses_target_points = (num_target_points > 0);
+    const int* const petsc_node_idxs =
+        uses_target_points ? &d_target_point_data[level_number].petsc_node_idxs[0] : NULL;
+    const double** const kappa = uses_target_points ? &d_target_point_data[level_number].kappa[0] : NULL;
+    const double** const eta = uses_target_points ? &d_target_point_data[level_number].eta[0] : NULL;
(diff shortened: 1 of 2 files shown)


drwells:
You're right: we should be using `resize` in `IBFEDirectForcingKinematics.cpp` for both arrays to fix this.

nishantnangia:
Using `resize` for the first array won't work as intended, since it is being filled via `push_back` later:
https://github.com/IBAMR/IBAMR/blob/nishant/fix-vector-dbg-assert/src/IB/IBFEDirectForcingKinematics.cpp#L740

I guess the `reserve` for the first array isn't actually necessary.

I didn't run the full test suite, so it is possible this is an issue in other places as well? I came across it while working on my own applications.

src/IB/IBStandardForceGen.cpp
```diff
@@ -1267,6 +1267,7 @@ IBStandardForceGen::computeLagrangianTargetPointForce(Pointer<LData> F_data,
     double max_displacement = 0.0;
 
     const int num_target_points = static_cast<int>(d_target_point_data[level_number].petsc_node_idxs.size());
+    if (num_target_points == 0) return;
```
boyceg: I would be careful here --- I think there are some collective operations that happen in this routine, and so the routine really should be rewritten to work when `num_target_points` is `0`.

boyceg:
`reserve` for `nodal_indices` is not necessary, but it doesn't hurt.

`resize` for `nodal_X_values` is needed for this to work as written. It also should be possible to get a pointer to the underlying data in `X_petsc` and avoid directly copying those values.

drwells:
I think it would be worthwhile to run the full test suite with the debug version of GLIBCXX (i.e., with `-D_GLIBCXX_DEBUG`). I will try to get to that today or tomorrow.

boyceg:
This should happen automatically with and build configured with a libMesh this is using the debug method.

nishantnangia:
Ah now it makes sense why I never saw this error before -- I just started building with libMesh :-)

src/IB/IBStandardForceGen.cpp
nishantnangia: I took a look and it seems like this routine (before my changes) doesn't do anything if `num_target_points` is `0`. Both the loops will exit without any iterations. Could you confirm when you get a chance; maybe I'm missing something?

boyceg:
There is a collective that computes the max over all processors using `SAMRAI_MPI::maxReduction()`, and that will hang if some callers return early.

boyceg:
The same issues show up in the other force computation routines. I am surprised that we haven't hit this before.

An alternative would be to use the original containers (e.g. `std::vector<T>`) and to get rid of the prefetch mumbo jumbo. I really don't know how much of a performance benefit that provides. I added it because we noticed that on some models, computing these types of forces was taking forever. One would hope that the intervening 10+ years of compiler development would make this manually unrolling crap unnecessary.

boyceg:
(Of course, the least invasive change would be just to set the pointers to `NULL` if there are no points.)

boyceg:
@nishantnangia, while you are at it, can you make the same change to all of the force computation routines?

boyceg:
Yes I think that is correct.

boyceg:
👍

boyceg:
I will merge once Jenkins is happy.

Although it looks like Jenkins may not be running currently?

boyceg:
Jenkins retest this please.

boyceg:
Jenkins retest this please.

boyceg:
Looks like Jenkins ran out of space (again!). I cleared out all workspaces and reduced the number of old builds Jenkins keeps around (from 10 to 5).

boyceg:
@drwells, this should go into 0.3.

PR #327, opened 2018-09-12
Allow versions of PETSc newer than 3.9.

drwells:
PETSc 3.10 will be out soon and we are compatible with it. However, the m4 scripts currently limit us to using 3.9 or older; this PR deletes the upper bound.

The PETSc interfaces we use are quite stable and are unlikely to change in the near future, so there is no reason to specify a maximum version: we should simply document what works and let users use new versions if they like.

### DIFF
--- a/configure
+++ b/configure
@@ -30268,8 +30268,8 @@ fi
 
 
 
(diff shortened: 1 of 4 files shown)


boyceg:
👍
Will merge once Jenkins is happy.

boyceg:
Jenkins retest this please.

PR #330, opened 2018-09-14
conditional compilation in BuildSystem makefiles

### LINKED ISSUE #323: fix conditional compilation in BuildSystem builds

boyceg:
This PR resolves #323.


boyceg:
For the hypre solvers, it is not enough just to conditionally compile, there also need to be compile guards in the source code, because hypre solvers are included in various places.

boyceg:
Components that use libMesh are currently self-contained, and so there is no need to conditionally compile in the source code (other than to maintain consistency within the codebase).

boyceg:
I guess one thing to complain about is that this change effectively requires hypre to be provided by PETSc, since it uses `PETSC_HAVE_HYPRE` in the source code.

drwells:
Ah. We could just do our own check, i.e., conditionally define `IBAMR_PETSC_WITH_HYPRE`. That is what we do in deal.II. Do you have another fix in mind?

boyceg:
Yep. There is already such a check in the autotools-based configure that I am updating now.

PR #331, opened 2018-09-14
make configure.new work for environments with both python2 and python3

boyceg:
This PR resolves #331.


drwells:
I think the more usual way to do this is to check `sys.version_info.major`.

Any machine with python should have a `python2` interpreter, so the rest looks good.

boyceg:
I am simply following the approach used in PETSc `configure` to check python versions. :-)

boyceg:
It appears that `sys.version_info` is not a tuple for python < 2.7. I think that the check in `configure.new` avoids that problem.

drwells:
What they do will work; I think the intent behind the original was python 2.6 compatibility (which is why we have the tuple check).

configure.new
```diff
@@ -1,10 +1,14 @@
 #!/usr/bin/env python
 
-import sys
+import sys, os
 if not type(sys.version_info) is tuple and sys.version_info.major > 2:
-  print('Configure does not support Python 3 yet, please run as')
+  if os.path.basename(sys.executable) == 'python2':
+    # Exit to prevent an infinite loop in case the environment is corrupted such
+    # that "python2" in PATH is actually a Python 3 interpreter.
+    print('Executable not a valid Python 2 interpreter: ' + sys.executable)
```
drwells: I am a bit frightened to ask whether or not this actually happened.

boyceg:
I am going to merge these even through Jenkins is not happy because Jenkins' unhappiness is not related to the changes in this PR.

PR #336, opened 2018-09-17
adding support to BuildSystem to specify the spatial dimension

boyceg:


boyceg:
@knepley, does this looks like the right approach to setting the spatial dimension?

Also, is there any other way that would be useful to set/record this variable other than through
```
    self.addMakeMacro('DIM', self.framework.argDB['DIM'])
```
in `IBAMR/Configure.py`?

boyceg:
@knepley, with this change, I am now getting the following queries when running `configure.new`:
```
TESTING: alternateConfigureLibrary from config.packages.amanzi(/Users/boyceg/sfw/petsc/petsc-maint/config/BuildSystem/coPlease enter value for download-samrai-commit: 
TESTING: configureLibrary from IBAMR.packages.SAMRAI(/Users/boyceg/sfw/petsc/petsc-maint/config/BuildSystem/config/packaPlease enter value for download-samrai-configure-arguments: 
```

config/IBAMR/packages/SAMRAI.py
```diff
@@ -6,11 +6,18 @@ def __init__(self, framework):
     self.download  = ['http://ftp.mcs.anl.gov/pub/petsc/externalpackages/SAMRAI-v2.4.4p.tar.gz', 'https://computation-rnd.llnl.gov/SAMRAI/download/SAMRAI-v2.4.4.tar.gz']
     self.functions = []
     self.includes  = ['Box.h']
-    self.liblist   = [['libSAMRAI.a', 'libSAMRAI2d_algs.a', 'libSAMRAI2d_appu.a', 'libSAMRAI2d_geom.a', 'libSAMRAI2d_hier.a', 'libSAMRAI2d_math_std.a', 'libSAMRAI2d_mesh.a', 'libSAMRAI2d_pdat_std.a', 'libSAMRAI2d_solv.a', 'libSAMRAI2d_xfer.a']]
+    if framework.argDB['DIM'] == 2:
```
knepley: I do not think querying the argDB in __init__ works. Configure creates all the nodes, then it calls setupHelp() on all of them, then it calls setup(), and then it calls configure(). I think you need to check this in setup() or configure(). I would lean towards the latter.

knepley:
> @knepley, does this looks like the right approach to setting the spatial dimension?
> 
> Also, is there any other way that would be useful to set/record this variable other than through
> 
> ```
>     self.addMakeMacro('DIM', self.framework.argDB['DIM'])
> ```
> in `IBAMR/Configure.py`?

You could also add a C define if you need that

```
    self.addDefine('DIM', self.framework.argDB['DIM'])
```

in the same place. It will go into the configure header.

drwells:
Superceded by #1020.

Issue #250, opened 2018-06-30
INS integrators never call regridHierarchy() in IB simulations


nishantnangia:
During regridding, IBHierarchyIntegrator::regridHierarchy() calls HierarchyIntegrator::regridHierarchy():

https://github.com/IBAMR/IBAMR/blob/master/src/IB/IBHierarchyIntegrator.cpp#L480

I think this means that regridHierarchy routines implemented by derived INS integrators will *never* be called during regridding (i.e. the ones implemented by INSStaggeredHierarchyIntegrator, etc.). Is this done intentionally, or is this a bug?

==

As an aside, we had added calls to initialize composite grid data to HierarchyIntegrator, so that they would be called during IBHierarchyIntegrators' regrid routine:
https://github.com/IBAMR/IBAMR/blob/master/ibtk/src/utilities/HierarchyIntegrator.cpp#L496

Perhaps these calls need to be moved to the derived integrators' regridHierarchy() routines?

amneetb:
Perhaps these calls need to be moved to the derived integrators' regridHierarchy() routines?

@nishantnangia : not moved but derived classes should also make a call to these routines if they override HierarchyIntegrator::regridHierarchy().

nishantnangia:
@amneetb Right, that seems sensible.

boyceg:
> On Jun 29, 2018, at 8:07 PM, Nishant Nangia <notifications@github.com> wrote:
> 
> During regridding, IBHierarchyIntegrator::regridHierarchy() calls HierarchyIntegrator::regridHierarchy():
> 
> https://github.com/IBAMR/IBAMR/blob/master/src/IB/IBHierarchyIntegrator.cpp#L480 <https://github.com/IBAMR/IBAMR/blob/master/src/IB/IBHierarchyIntegrator.cpp#L480>
> I think this means that regridHierarchy routines implemented by derived INS integrators will never be called during regridding (i.e. the ones implemented by INSStaggeredHierarchyIntegrator, etc.). Is this done intentionally, or is this a bug?
> 
I am not sure about the intention, but I do know that we don't want to call both HierarchyIntegrator::regridHierarchy() and also INSHierarchyIntegrator::regridHierarchy(), because both call GriddingAlgorithm::regridAllFinerLevels().

The right thing to do might be to call INSHierarchyIntegrator::regridHierarchy() instead of HierarchyIntegrator::regridHierarchy().

[...]

PR #343, opened 2018-09-26
Make call to INS regridHierarchy in IB integrator

nishantnangia:
In reference to https://github.com/IBAMR/IBAMR/issues/250

We should call the INS specific regridHierarchy() functions to ensure that their regridProjection() routines are called. Also adding calls to initializeCompositeHierarchyData() in the overridden calls.

### DIFF
--- a/src/IB/IBHierarchyIntegrator.cpp
+++ b/src/IB/IBHierarchyIntegrator.cpp
@@ -477,7 +477,7 @@ IBHierarchyIntegrator::regridHierarchy()
 
     // Use the INSHierarchyIntegrator to handle Eulerian data management.
     if (d_enable_logging) plog << d_object_name << "::regridHierarchy(): regridding the patch hierarchy\n";
-    HierarchyIntegrator::regridHierarchy();
+    d_ins_hier_integrator->regridHierarchy();
 
     // After regridding, finish Lagrangian data movement.
     if (d_enable_logging) plog << d_object_name << "::regridHierarchy(): finishing Lagrangian data movement\n";
--- a/src/navier_stokes/INSCollocatedHierarchyIntegrator.cpp
+++ b/src/navier_stokes/INSCollocatedHierarchyIntegrator.cpp
@@ -1563,6 +1563,10 @@ INSCollocatedHierarchyIntegrator::regridHierarchy()
     // Project the interpolated velocity.
     regridProjection();
 
+    // Reinitialize composite grid data.
+    const bool initial_time = false;
+    initializeCompositeHierarchyData(d_integrator_time, initial_time);
+
     // Synchronize the state data on the patch hierarchy.
     synchronizeHierarchyData(CURRENT_DATA);
     return;
--- a/src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
+++ b/src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
@@ -1491,6 +1491,10 @@ INSStaggeredHierarchyIntegrator::regridHierarchy()
         regridProjection();
     }
 
+    // Reinitialize composite grid data.
+    const bool initial_time = false;
+    initializeCompositeHierarchyData(d_integrator_time, initial_time);
+
     // Synchronize the state data on the patch hierarchy.
     synchronizeHierarchyData(CURRENT_DATA);
     return;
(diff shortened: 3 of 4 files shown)


drwells:
Jenkins retest this please

amneetb:
We need to merge this with NWT. There is an obvious bug in IBHierarchyIntegrator for not calling INSHI's regridHierarchy().

boyceg:
I will take a look this week. If you need it before then, I would suggest making a temporary integration branch — but please do not merge into master or the other feature branch.

amneetb:
Ok, thanks. 

Right now all IB cases do have regridProjection() method called after regridding. This is a severe bug for INSVC because initial velocity needs to projected onto div-free space using density-weighted op.

boyceg:
I am imagining a case where you have specialized reinitialization that happens in multiple places, e.g. you may need to rebuild the level set function on the new finest level as well as reproject the velocity. I think that this is what I had in mind when I moved regridding into `HierarchyIntegrator`, but I clearly did not quite finish the job...

Would it make sense for `HierarchyIntegrator::regridHierarchy()` to be marked `final`, and to make it responsible for actually calling `GriddingAlgorithm::regridAllFinerLevels()`, and then for the parent integrator to call, for itself and for each of the nested integrators:

1. a new `HierarchyIntegrator::regridHierarchySpecialized()` method (default empty implementation, would do `regridProjection()` in the Navier-Stokes solvers)
2. `HierarchyIntegrator::initializeCompositeHierarchyData()` and 
3. `HierarchyIntegrator::synchronizeHierarchyData()`

For the N-S solvers, `INSHierarchyIntegrator::regridHierarchySpecialized()` could call a virtual `INSHierarchyIntegrator::regridProjection()` method.

This ensures that `regridAllFinerLevels()` is called "in the right place" and eliminates some code duplication. On the other hand, it is not straightforward to specify any dependencies between implementations of `regridHierarchySpecialized()`.

An alternative would be almost the same as above: `HierarchyIntegrator::regridHierarchy()` is responsible for actually calling `GriddingAlgorithm::regridAllFinerLevels()`, and it calls `HierarchyIntegrator::regridHierarchySpecialized()` for only the parent `HierarchyIntegrator`, and then have the parent integrator call that method for the other nested integrators. This seems a little brittle, however.

amneetb:
I like the first approach. I also agree it is not straightforward to know the dependencies between regridHierarchySpecialized().

boyceg:
> I would prefer avoiding boolean, but if it makes sense to you I am OK with that.
> 
> I definitely like the idea of renaming it to HierarchyIntegrator::initializeDataAfterRegrid() as it makes the intent clear.
> 

The basic question is whether we want to allow a "parent" HierarchyIntegrator to call initializeDataAfterRegrid() for other integrators that it manages in the parent's implementation of initializeDataAfterRegrid(), to coordinate the reinitialization, or if we just want an implementation in the base HierarchyIntegrator class to coordinate this.

If we don't care about allowing for customization of the order in which these methods are called, there is no need to keep extra state.

If we want to allow for customization, we also need to ensure that initializeDataAfterRegrid() is called for all of the hierarchy integrator instances. (I don't think that we want to require implementations to know about all of the associated integrators and to call initializeDataAfterRegrid() manually for all of them.)

Let me see if I can cook up a proposed implementation...

drwells:
This if the first commit from #418: figuring this out is necessary for figuring out how to proceed there.

@boyceg did you ever cook up a proposed implementation?

drwells:
If not, I will start working on one :)

boyceg:
I had some thoughts about this but would need to reconstruct them. If you want to give this some thought, that would be good with me. :-)

drwells:
I spent some time looking over this this afternoon.

I found two incompatibilities with this patch:
1. If we run any of the IBFE codes then we never regrid after the first time step.
2. Some of the structure positions (reported by `estimated upper bound on IB point displacement`) are now off by a factor of 100 at the end of the first time step. I don't know why this happens.

We do need to fix this but this patch needs some more TLC before we can merge it.

I think @boyceg's [outline](https://github.com/IBAMR/IBAMR/pull/343#issuecomment-448102301) above is the reasonable way to fix the problem on NWT and also preserve compatibility: we should only ever call `regridHierarchy` on the parent integrator (we should add an assertion to enforce this, i.e., `TBOX_ASSERT(d_parent_integrator == nullptr)`) and that function should call `regridHierarchySpecialized`/`initializeDataAfterRegrid`/whatever on all child integrators after moving data around to inform them that they need to do something. 

It might be worthwhile to handle this in a more general way with `regridHierarchySpecializedBegin` and `regridHierarchySpecializedEnd` (or whatever name we end up using), but since SAMRAI handles the data movement I don't think this is absolutely necessary right now.

drwells:
Superceded by #524, which solves the problem with projection in a more permanent way.

PR #344, opened 2018-09-27
Remove boost array

### LINKED ISSUE #341: Keep track of major changes
One thing that we do in deal.II that is quite handy is associating all non-trivial (e.g., things larger than a typo fix) changes with a changelog entry, see, e.g.,

https://github.com/dealii/dealii/tree/master/doc/news/changes

We should adopt this format, or something similar, so that we can write release summaries that say things like "we fixed major bug X, added feature Y, and added new tutorial program Z" where authors can get credit for their contributions.

This would also make it a lot easier to write release summaries (which is exactly why I am thinking about this now).

drwells:
Fixes #341 (at least until its time to release).

This is a breaking patch that will make sure we really do have C++11 enabled: I replaced `boost::array` with the identical C++11 class `std::array`. This does add two more conflicting lines to #258 (sorry Aaron).

One issue (noted in #341) with the 0.3.0 release is that it is hard to summarize the changes by just reading back over git logs; it would be a lot easier to give people credit for their contributions if we kept track of things at a higher level. The first commit in this PR adds some infrastructure for doing this: I added three folders for minor, major, and incompatible changes and also a file describing this change.


boyceg:
I would be happy to remove all dependency on boost.

boyceg:
...but I don't know a good replacement for `boost::multi_array`.

drwells:
I do not know of one either.

boyceg:
Since we are now requiring C++11 for `configure` to work, let's use some C++11 language features!

boyceg:
@abarret, if we are making our own modernize-like script(s), perhaps one that converts `boost::array` to `std::array` is one thing that should go in.

abarret:
I can add that change to the UniquePtr -> unique_ptr script.

PR #347, opened 2018-10-01
Use the C++11 std::unordered_map.

drwells:
It looks like we do not use `boost.m4` anywhere but we should keep it up to date.

This is a followup to #307, which was only on the 0.3.0 branch.

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -42,6 +42,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <unordered_map>
 
 #include "BasePatchLevel.h"
 #include "CellVariable.h"
@@ -53,7 +54,6 @@
 #include "StandardTagAndInitStrategy.h"
 #include "VariableContext.h"
 #include "boost/multi_array.hpp"
-#include "boost/unordered_map.hpp"
 #include "ibtk/ibtk_utilities.h"
 #include "libmesh/dof_map.h"
 #include "libmesh/elem.h"
@@ -136,7 +136,7 @@ public:
 
     private:
         libMesh::DofMap& d_dof_map;
-        boost::unordered_map<libMesh::dof_id_type, std::vector<std::vector<unsigned int> > > d_dof_cache;
+        std::unordered_map<libMesh::dof_id_type, std::vector<std::vector<unsigned int> > > d_dof_cache;
     };
 
     /*!


boyceg:
I think we use `boost.m4` to configure boost?

Note that `boost.m4` also should be updated in the `m4` subdirectory. Basically all of the m4 files should be duplicated between `m4` and `ibtk/m4`.

drwells:
You're right; I fixed `m4/boost.m4`.

I should clarify: we do use `ibtk/m4/configure_boost.m4`, but we never use anything from `ibtk/m4/boost.m4`.

boyceg:
```
$ grep BOOST_REQUIRE configure.ac 
BOOST_REQUIRE([1.57.0],[
$ grep BOOST_REQUIRE */*.m4
m4/boost.m4:# The first macro is BOOST_REQUIRE.  It will simply check if it's possible to
m4/boost.m4:# BOOST_REQUIRE([VERSION], [ACTION-IF-NOT-FOUND])
m4/boost.m4:AC_DEFUN_ONCE([BOOST_REQUIRE],
m4/boost.m4:])# BOOST_REQUIRE
m4/boost.m4:[AC_REQUIRE([BOOST_REQUIRE])dnl
m4/boost.m4:# (eg, >= 1.33) you *must* invoke BOOST_REQUIRE before this macro.
m4/boost.m4:[AC_REQUIRE([BOOST_REQUIRE])dnl
m4/configure_boost.m4:#BOOST_REQUIRE([1.57.0],[AC_MSG_WARN([could not find system Boost library, using bundled Boost library])])
```

boyceg:
`BOOST_REQUIRE` is called from `configure.ac` and provided by `boost.m4`.

drwells:
You're right. This is weird: deleting the header check from the m4 script made no difference to `configure`, but it is clearly using something else in the m4 script. I will update #348 correspondingly.

boyceg:
@drwells can you revert the `boost.m4` changes? If you want to update `boost.m4`, let's just synch with upstream.

drwells:
Fixed.

boyceg:
Is there a good replacement for `boost::math::iround`? (I personally don't like to use casting for rounding.) After this merge, that looks like the only use of boost that is left other than `multi_array` and friends (which, as discussed elsewhere, do not appear to have good alternatives).

amneetb:
Please do not eliminate boost from IBAMR. I am using it for ODE integrators and special functions like erf() and erfc() in my applications.

drwells:
It looks like C++11 has its own `std::round` function. I'll play around with it to make sure it has the same rounding mode as the boost one.

PR #349, opened 2018-10-02
Modernize headers

drwells:
@abarret and I corresponded regarding the current state of #258 and we agreed that it would be better to chop up the first commit to avoid merge conflicts. I also promised to help with the effort: this PR uses the `modernize-deprecated-headers` option to replace C headers with C++ equivalents (e.g., `string.h` is `cstring`).

I followed up on the changes and put everything that I could find that should be in `std::` in `std::`, e.g., `std::snprintf` instead of `snprintf`.


include/ibamr/IBMethod.h
```diff
@@ -542,7 +542,7 @@ class IBMethod : public IBImplicitStrategy
      * List of local indices of local anchor points.
      *
      * NOTE: IB points are automatically considered to be anchored if they are
-     * within 2.0*sqrt(epsilon_mach) of the physical boundary.
+     * within 2.0*std::sqrt(epsilon_mach) of the physical boundary.
```
boyceg: 2.0*sqrt here?

boyceg:
Instead of namespacing all of these math operations, could/should we use `using` declarations?

I also can live with the fully namespaced functions.

Otherwise looks good to me.

include/ibamr/IBMethod.h
drwells: Fixed.

boyceg:
👍

Issue #350, opened 2018-10-02
Split up #258's first commit


drwells:
@abarret and I corresponded by email about this and we agreed (as long as I do some of the work) that we should apply the modernization changes in separate PRs. I believe this will be faster than working through the merge conflicts present in #258 (though, of course, we should still use all commits on that branch besides the first once these are through).

- [x] modernize-deprecated-headers (@drwells, #349)
- [x] modernize-loop-convert (risky didn't make an incorrect change so it's fine to use) (@abarret, #361)
- [x] modernize-pass-by-value (#363)
- [x] modernize-shrink-to-fit (no results)
- [x] modernize-use-auto (MinTypeNameLength=0, RemoveStars=1) (@abarret #356)
- [x] modernize-use-bool-literals (this only effects `eigen`)
- [x] modernize-use-default-member-init (@abarret, #380 )
- [x] modernize-use-equals-default
- [x] modernize-use-equals-delete (@abarret #381)
- [x] modernize-use-nullptr (@drwells #351)
- [x] modernize-use-override (@drwells #355)
- [x] modernize-use-using (@abarret #354)

drwells:
The `nullptr` modernization patch is done but conflicts on one line with #349, so I won't post it until that is merged.

boyceg:
I think we are all done?

boyceg:
Oh, wait, have we run modernize-use-equals-default? I only see one `= default` in the codebase; is that the only one modernize found?

abarret:
I don't know how that one got in there. I'm working on the ` = default` now.

boyceg:
OK, so now this really is all done. 👍

PR #351, opened 2018-10-03
Replace 0 and NULL with nullptr.

drwells:
Part of #350.

### DIFF
--- a/src/utilities/RNG.cpp
+++ b/src/utilities/RNG.cpp
@@ -256,7 +256,7 @@ RNG::parallel_seed(int global_seed)
     {
         if (seed == 0)
         {
-            seed = static_cast<int>(std::time(0));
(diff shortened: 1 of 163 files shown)


src/utilities/RNG.cpp
```diff
@@ -256,7 +256,7 @@ RNG::parallel_seed(int global_seed)
     {
         if (seed == 0)
         {
-            seed = static_cast<int>(std::time(0));
+            seed = static_cast<int>(std::time(nullptr));
```
boyceg: Looks wrong.

boyceg: Good catch.

boyceg:
👍

drwells:
I got bear to cooperate by deleting all the 2D build instructions, which caused clang-tidy to miss a bunch of `NULL`s. I am updating this now.

PR #352, opened 2018-10-04
Remove boost::math::{iround,round}.

drwells:
We now have std::round so this is not needed.

Followup from #347.

### DIFF
--- a/ibtk/src/lagrangian/LDataManager.cpp
+++ b/ibtk/src/lagrangian/LDataManager.cpp
@@ -84,7 +84,6 @@
 #include "VariableContext.h"
 #include "VariableDatabase.h"
 #include "VisItDataWriter.h"
-#include "boost/math/special_functions/round.hpp"
 #include "boost/multi_array.hpp"
 #include "ibtk/IBTK_CHKERRQ.h"
 #include "ibtk/IndexUtilities.h"
@@ -1448,7 +1447,7 @@ LDataManager::beginDataRedistribution(const int coarsest_ln_in, const int finest
             IntVector<NDIM> periodic_offset;
             for (int d = 0; d < NDIM; ++d)
             {
-                periodic_offset[d] = boost::math::round(periodic_displacement[d] / level_dx[d]);
+                periodic_offset[d] = int(std::round(periodic_displacement[d] / level_dx[d]));
             }
             if (periodic_offset != IntVector<NDIM>(0))
             {
--- a/src/IB/IBFESurfaceMethod.cpp
+++ b/src/IB/IBFESurfaceMethod.cpp
@@ -61,7 +61,6 @@
 #include "SideIndex.h"
 #include "Variable.h"
 #include "VariableDatabase.h"
-#include "boost/math/special_functions/round.hpp"
 #include "boost/multi_array.hpp"
 #include "ibamr/IBFESurfaceMethod.h"
 #include "ibamr/IBHierarchyIntegrator.h"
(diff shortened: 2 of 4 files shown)


src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1669,7 +1668,8 @@ IBFESurfaceMethod::imposeJumpConditions(const int f_data_idx,
                         const libMesh::Point x = r + intersections[k].first * q;
                         const libMesh::Point& xi = intersections[k].second;
                         SideIndex<NDIM> i_s(i_c, axis, 0);
-                        i_s(axis) = boost::math::iround((x(axis) - x_lower[axis]) / dx[axis]) + patch_lower[axis];
+                        i_s(axis) = int(std::round((x(axis) - x_lower[axis]) / dx[axis]))
```
boyceg: `static_cast<int>`? And, if it needs a cast anyway, should we just omit the call to `std::round` and let the cast do the rounding for us?

ibtk/src/lagrangian/LDataManager.cpp
```diff
@@ -1448,7 +1447,7 @@ LDataManager::beginDataRedistribution(const int coarsest_ln_in, const int finest
             IntVector<NDIM> periodic_offset;
             for (int d = 0; d < NDIM; ++d)
             {
-                periodic_offset[d] = boost::math::round(periodic_displacement[d] / level_dx[d]);
+              periodic_offset[d] = int(std::round(periodic_displacement[d] / level_dx[d]));
```
boyceg: `static_cast<int>`? And, if it needs a cast anyway, should we just omit the call to `std::round` and let the cast do the rounding for us?

src/IB/IBRedundantInitializer.cpp
```diff
@@ -1046,7 +1045,7 @@ IBRedundantInitializer::initializeDataOnPatchLevel(const int lag_node_index_idx,
             IntVector<NDIM> periodic_offset;
             for (int d = 0; d < NDIM; ++d)
             {
-                periodic_offset[d] = boost::math::round(periodic_displacement[d] / patch_dx[d]);
+                periodic_offset[d] = int(std::round(periodic_displacement[d] / patch_dx[d]));
```
boyceg: `static_cast<int>`? And, if it needs a cast anyway, should we just omit the call to `std::round` and let the cast do the rounding for us?

src/IB/IMPMethod.cpp
```diff
@@ -127,7 +126,7 @@ kernel(const double X,
        boost::multi_array<double, 1>& dphi)
 {
     const double X_o_dx = (X - patch_x_lower) / dx;
-    stencil_box_lower = boost::math::round(X_o_dx) + patch_box_lower - kernel_width;
+    stencil_box_lower = int(std::round(X_o_dx)) + patch_box_lower - kernel_width;
```
boyceg: `static_cast<int>`? And, if it needs a cast anyway, should we just omit the call to `std::round` and let the cast do the rounding for us?

drwells:
I am not sure that I understand the difference between `static_cast` and the converting constructor. To avoid overflow `std::round` returns `double` instead of `int`, so we need to do one or the other, and I think that for fundamental types (e.g., `int`) the two approaches are equivalent (but we need to use one or the other to avoid warnings about truncation).

The difference between the cast and round is that round will find the nearest integer, which I assume is the desired behavior (instead of chopping). In at least one place we are using the rounding function to compute an array index so I suspect this would give different results.

amneetb:
I would not go for changes in this PR unless I am 100% sure about the behavior of these functions. There needs to be testing with different configuration of IB points in parallel to see the before and after effects.

drwells:
@amneetb has a decent point; we cannot really verify a lot of these changes without a bigger test suite.

boyceg:
I would like to just use `std::round` everywhere, but let's close this for now.

PR #354, opened 2018-10-04
Prefer using syntax over typedef.

abarret:
This replaces all typedef's with the new using syntax.


ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -167,35 +167,35 @@ level_can_be_refined(int level_number, int max_levels)
...
+
+using MatrixNd = Eigen::Matrix<double, NDIM, NDIM>;
+using VectorNd = Eigen::Matrix<double, NDIM, 1>;
+using ColumnVectorNd = Eigen::Matrix<double, NDIM, 1>;
+using RowVectorNd = Eigen::Matrix<double, 1, NDIM>;
+
+using MatrixXd = Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic>;
+using VectorXd = Eigen::Matrix<double, Eigen::Dynamic, 1>;
+using ColumnVectorXd = Eigen::Matrix<double, Eigen::Dynamic, 1>;
+using RowVectorXd = Eigen::Matrix<double, 1, Eigen::Dynamic>;
```
abarret: @boyceg Eigen already has most of these typedefs, e.g.
`typedef Eigen::Matrix<double, 3, 3> Eigen::Matrix3d;`
Would it be worth it to get rid of these in favor of Eigen's typedefs?

drwells:
:+1:

ibtk/include/ibtk/ibtk_utilities.h
abarret: As far as I can tell, the only place these typedefs are used are in `GeneralizedIBMethod.cpp` and `IBKirchoffRodForceGen.cpp`. All other uses of Matrix/Vector2/3d utilize the Eigen typedef.

boyceg: Hmmm, why did I add in those `typedef`s?

The reason is that I used to do this stuff with Blitz++ classes, but then Blitz++ development slowed down, and I switched to Eigen. I put in the typedefs to make the IBAMR/IBTK interface more agnostic to the linear algebra backend.

I am sure that @drwells will chime in about whether this makes sense or not. :-)

boyceg: (Also it is certainly possible that I put the `typedef`s with the _intention_ of "hiding" the backend but then forgot to follow through on it...)

abarret: Yes we could go the other way and replace all uses of Eigen's `typedef`s with these.

boyceg: Let's see what @drwells says...

drwells: I don't think we should worry about being agnostic w.r.t. the dense linear algebra background.

It looks like `Eigen` already offers `Eigen::VectorXd` (and we use it in `IBFEPatchRecoveryPostProcessor.cpp`). Since there is some overlap between what we use as `Eigen::VectorXd` and `IBTK::Vector` I think it would be better to just stick to the eigen-provided using declarations.

Either way: we can fix this in a follow-up PR since we do use `IBTK::Vector` a lot. I suggest we use `[[deprecated]]` or similar to mark these as deprecated and switch all of our stuff to use, e.g., `Eigen::VectorXd`.

boyceg:
👍

PR #355, opened 2018-10-05
Mark overriding functions as such.

drwells:
I couldn't get clang-tidy's modernize-override to work so I instead used the warnings generated by clang (`-Winconsistent-missing-override`) and GCC (`-Wsuggest-override`) to find all of these.


boyceg:
Ouch. Do you have a script, or did you have to go by hand?

drwells:
I used an emacs macro, so I hit `C-x e` 1006 times. It wasn't that bad.

PR #360, opened 2018-10-11
eliminating the use of localize in IBFEMethod

boyceg:


drwells:
You beat me to it!

boyceg:
@drwells can you see what this does to the performance numbers? I have eliminated all uses of `NumericVector<T>::localize()` in `IBFEMethod` but not yet in `FEDataManager`.

boyceg:
Whoops, noticed a dopey bug. Just rebased a fix.

drwells:
Sure.

I did some more profiling and I did not notice this showing up in the four chamber model, just ex4. I'll see what this does.

boyceg:
If there is no performance impact maybe we should just leave it as is.

drwells:
I get errors of the following sort:
```
[4]PETSC ERROR: Incompatible vector local lengths parameter # 1 local size 57804 != parameter # 2 local size 67473
```

drwells:
(that was on examples/IBFE/explicit/ex4)

boyceg:
Uh oh.

boyceg:
This could point me me misunderstanding the parallel data layout for `NumericVector`s with ghosted DOFs.

boyceg:
Oh, I see, libMesh is doing a local form copy.

boyceg:
Give this a whirl.

boyceg:
(I must admit that I am being a little lazy here by having `copy_and_synch` take in a `NumericVector` when it relies on it being a `PetscVector`... Can fix this up if it looks like this is worthwhile.)

drwells:
I did not check wall times (I will do that next) but with this branch `preprocessIntegrateHierarchy` requires about 0.5% of the total run time; on `master` (with the same input file) this function requires 14%.

Its worth mentioning that this function only really becomes a bottleneck when there is a lot of Lagrangian data; it doesn't show up in the default settings for example 4.

drwells:
It looks like the wall times are essentially equal: everything we gained here was gobbled up by polling elsewhere. Thats depressing.

drwells:
(for example: the polling on rank 0 jumped from 56% to 70%)

boyceg:
So is this just revealing load balance issues elsewhere?

drwells:
I think so, but its curious that this barely changes the wall time.

boyceg:
Although, `PetscVector<T>::localize()` appears ultimately to be nonscalable, right? So should we just do this anyway?

drwells:
Yes, but it looks like the main slowdown was caused by our already poor load balancing.

boyceg:
OK, I just pushed some code to check (in `DEBUG` mode) that the input vector is actually a `PetscVector<T>`. (OTOH, we assume `PetscVector<T>` practically everywhere currently.)

Anyway, even if this is not a big performance win, I think it is good to try to avoid extra `VecScatter`s and just do ghost updates in most places where we previously had been using `NumericVector<T>::localize()`.

drwells:
Before we merge, would you mind adding a changelog entry? If not I can do it.

boyceg:
Where do I do it?

cpuelz:
I wonder if this fixes parallel issues I’ve been having with prolongData.

Get Outlook for iOS<https://aka.ms/o0ukef>

drwells:
Add an entry like the one in `doc/news/changes/incompatibilities/20180927DavidWells` to the folder in `doc/new/changes/minor/` that says something like "Improved: rather than using the `localize` function to copy data, we now only copy the necessary ghost data."

doc/news/changes/minor/20181012BoyceGriffith
```diff
@@ -0,0 +1,7 @@
+Improved: Replaced the use of <code>NumericVector<T>::localize()</code> with a new helper function, <code>copy_and_synch</code>, that is optimized only to communicate ghost data.  Changes have been made in:
+<ul>
+  <li><code>IBFEMethod</code>,</li>
+  <li><code>FEDataManager</code>,</li>
+  <li><code>FEDataInterpolator</code>
+</ul>
+(Boyce Griffith, 2018/10/12)
```
drwells: :+1:

drwells:
@cpuelz The main difference this change makes is that we now no longer propagate all ghost data to all vectors. I have not seen any evidence so far to imply there was a problem with correctness, just performance. Could you open an issue for whatever is happening with prolongation if things don't work?

cpuelz:
@drwells I was thinking about issues I was having here:
https://github.com/IBAMR/IBAMR/pull/208#discussion_r221096892

Could have been user error :-).

drwells:
I'm not sure. We need a better test suite...

drwells:
This does pass the full test suite, but I am concerned that we exit this function while leaving the libMesh vector in an inconsistent state: we update ghost values, but do not set `this->_is_closed = true`. While it is less efficient (since we will end up with more calls to `VecAssemblyBegin` and `VecAssemblyEnd`) I think it would be better to just use `libMesh::PetscVector::operator=` here instead. We will still come out massively ahead since we avoid the all-to-all scatter.

I'll rerun my torturous version of ex4 with this and see what happens.

boyceg:
I agree with you in principle.

However, in practice, I am sort of ambivalent about the state of the `NumericVector`, because any access to the vector that uses `PetscVector<T>::vec()` potentially leaves the libMesh vector in an inconsistent state --- and we already do this elsewhere in `IBFEMethod` to get access to some specific PETSc `Vec` functions.

Further, calls to methods like `NumericVector<double>::set()` are not collective (which is a good thing). But this means that we cannot rely on `NumericVector<double>::closed()` to tell us anything useful about the state of the vector unless we wrap it in something like a max-reduction, which I don't think we want to do.

My conclusion is that `NumericVector<t>::closed()` is not a useful way to keep track of the state of the vector.

On the other hand, I also agree that it probably is not a good thing to go out of our way to _break_ the state, which is what the current code potentially does.

drwells:
My main worry is that libMesh will break in some unfathomable way if we do this, but it sounds like this is not a concern in practice. I think the worst that could happen is libMesh could just call `close()` again later on anyway.

boyceg:
In that case, I am OK with going back to just using `close()` here, as originally implemented.

boyceg:
@knepley, if a `Vec` is already fully assembled, do `VecAssemblyBegin`/`VecAssemblyEnd` do anything?

knepley:
> @knepley, if a `Vec` is already fully assembled, do `VecAssemblyBegin`/`VecAssemblyEnd` do anything?

A little. They must check if anyone has anything to contribute. This is a synchronization, although it can be done in a nice way due to Hoefler that Jed coded up. I would be surprised if it hurts at the scale you care about.

boyceg:
@drwells, I don't think there is any need to switch to `PetscVector<T>::operator=` here, but I think it is OK to switch back do using `close()` in `copy_and_synch` instead of manually communicating ghost values. Can you make this change and, if your tests run as expected, go ahead and merge?

drwells:
I spent a long time considering whether or not we should call `VecDuplicate` here: I believe the answer is 'no' since PETSc will complain if we copy and the local sizes are different. I am running the test suite now.

boyceg:
@drwells, this now uses `NumericVector<T>::close()` to synch ghost data for the receiving vector in `copy_and_synch()`. Once you are satisfied that this works as expected, let's merge and move on to other improved load balancing tasks.

drwells:
I made precisely the same change locally but did not push. I will merge once the test suite is finished and passes.

drwells:
This passes the tests and everything can be built on the CI server so lets merge.

PR #361, opened 2018-10-12
Prefer range based for loops.

abarret:
This is a run of modernize-loop-convert with clang-tidy.

I'm still working on removing unused `typedef`'s, but I went ahead and opened this so changes can be reviewed.

### DIFF
--- a/ibtk/include/ibtk/private/LDataManager-inl.h
+++ b/ibtk/include/ibtk/private/LDataManager-inl.h
@@ -165,11 +165,9 @@ LDataManager::getLagrangianStructureNames(const int level_number) const
     TBOX_ASSERT(d_coarsest_ln <= level_number && d_finest_ln >= level_number);
 #endif
     std::vector<std::string> ret_val;
-    for (std::map<int, std::string>::const_iterator cit(d_strct_id_to_strct_name_map[level_number].begin());
-         cit != d_strct_id_to_strct_name_map[level_number].end();
-         ++cit)
+    for (const auto& map : d_strct_id_to_strct_name_map[level_number])
     {
-        ret_val.push_back(cit->second);
+        ret_val.push_back(map.second);
     }
     return ret_val;
 } // getLagrangianStructureNames
@@ -181,11 +179,9 @@ LDataManager::getLagrangianStructureIDs(const int level_number) const
     TBOX_ASSERT(d_coarsest_ln <= level_number && d_finest_ln >= level_number);
 #endif
     std::vector<int> ret_val;
-    for (std::map<std::string, int>::const_iterator cit(d_strct_name_to_strct_id_map[level_number].begin());
-         cit != d_strct_name_to_strct_id_map[level_number].end();
-         ++cit)
+    for (const auto& map : d_strct_name_to_strct_id_map[level_number])
     {
-        ret_val.push_back(cit->second);
+        ret_val.push_back(map.second);
     }
     return ret_val;
 } // getLagrangianStructureIDs
--- a/ibtk/src/boundary/physical_boundary/muParserRobinBcCoefs.cpp
+++ b/ibtk/src/boundary/physical_boundary/muParserRobinBcCoefs.cpp
@@ -244,12 +244,12 @@ muParserRobinBcCoefs::muParserRobinBcCoefs(const std::string& object_name,
     const double pi = 3.1415926535897932384626433832795;
     const double* const xLower = grid_geom->getXLower();
     const double* const xUpper = grid_geom->getXUpper();
-    for (auto cit = all_parsers.begin(); cit != all_parsers.end(); ++cit)
+    for (const auto& parser : all_parsers)
     {
         // Various names for pi.
-        (*cit)->DefineConst("pi", pi);
-        (*cit)->DefineConst("Pi", pi);
-        (*cit)->DefineConst("PI", pi);
+        parser->DefineConst("pi", pi);
+        parser->DefineConst("Pi", pi);
+        parser->DefineConst("PI", pi);
 
         // The extents of the domain.
         for (unsigned int d = 0; d < NDIM; ++d)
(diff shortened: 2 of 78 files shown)


boyceg:
@abarret is this ready for review?

ibtk/include/ibtk/private/LDataManager-inl.h
```diff
@@ -165,11 +165,9 @@ LDataManager::getLagrangianStructureNames(const int level_number) const
     TBOX_ASSERT(d_coarsest_ln <= level_number && d_finest_ln >= level_number);
 #endif
     std::vector<std::string> ret_val;
-    for (std::map<int, std::string>::const_iterator cit(d_strct_id_to_strct_name_map[level_number].begin());
-         cit != d_strct_id_to_strct_name_map[level_number].end();
-         ++cit)
+    for (const auto& map : d_strct_id_to_strct_name_map[level_number])
```
boyceg: 👍

ibtk/src/boundary/physical_boundary/muParserRobinBcCoefs.cpp
```diff
@@ -244,12 +244,12 @@ muParserRobinBcCoefs::muParserRobinBcCoefs(const std::string& object_name,
     const double pi = 3.1415926535897932384626433832795;
     const double* const xLower = grid_geom->getXLower();
     const double* const xUpper = grid_geom->getXUpper();
-    for (auto cit = all_parsers.begin(); cit != all_parsers.end(); ++cit)
+    for (auto& all_parser : all_parsers)
```
boyceg: `parser`

ibtk/src/lagrangian/FEDataInterpolation.cpp
```diff
@@ -225,9 +225,9 @@ FEDataInterpolation::init(const bool use_IB_ghosted_vecs)
         const System& system = *d_systems[system_idx];
         const DofMap& system_dof_map = system.get_dof_map();
         const std::vector<int>& all_vars = d_system_all_vars[system_idx];
-        for (unsigned int k = 0; k < all_vars.size(); ++k)
+        for (const auto& all_var : all_vars)
```
boyceg: `var`

drwells: (same here, with all_var)

abarret:
I updated the variable names as well as a few `it` and `cit` names I missed. If there's nothing else, I can squash into a single commit.

boyceg:
@drwells, should we prefer to use `const auto&` in range-based for loops? If so, @abarret, could you switch everything to initialize using `const auto&` and then back off to `auto&` or `auto` where needed? (Perhaps you already have done that.)

Looks great!

drwells:
Yes.

abarret:
Done. I used the compiler to check for correctness on the changes. Is there any case the compiler wouldn't catch for this kind of switch?

boyceg:
@abarret, do you mean, are there cases where `const auto&` is not the right thing to use in the range-based for loop? The only cases are where the loop is actually modifying the objects that are being iterated over.

The following is not OK:
```
vector<T> vec = ... ;
for (const auto& v : vec)
{
    v.non_const_member_function();
}
```
but the following is OK:
```
vector<T*> vec = ... ;
for (const auto& v : vec)
{
    v->non_const_member_function();
}
```
because here, `v` is a constant reference to a `T*`.

On the other hand, this does not work:
```
vector<const T*> vec = ... ;
for (const auto& v : vec)
{
    v->non_const_member_function();
}
```
The reason is that here, `v` is a const reference to a `const T*`, so we cannot call any non-const member functions.

I am 99% sure that if you default to using `const auto&` in the range-based for loops, then basically all you will need to do is fix any cases where the compiler complains about const correctness.

boyceg:
In cases where the compiler does complain about const correctness, you will need to determine whether you want a copy (`auto`) or a reference (`auto&`) or, possibly, an r-value reference (`auto&&`).

abarret:
> I am 99% sure that if you default to using `const auto&` in the range-based for loops, then basically all you will need to do is fix any cases where the compiler complains about const correctness.

This is what I did. I couldn't think of a case where the compiler wouldn't catch an error, but I also can't think of many things, so I wanted to double check.

I went back and checked all the range-based for loops that didn't use `const auto&` and I don't think any of those need to be changed. I'm done making changes unless someone else (@drwells?) suggests changes.

src/IB/IBLagrangianForceStrategySet.cpp
```diff
@@ -61,9 +61,9 @@ IBLagrangianForceStrategySet::~IBLagrangianForceStrategySet()
 void
 IBLagrangianForceStrategySet::setTimeInterval(const double current_time, const double new_time)
 {
-    for (auto cit = d_strategy_set.begin(); cit != d_strategy_set.end(); ++cit)
+    for (const auto& set : d_strategy_set)
```
boyceg:
```suggestion
    for (const auto& strategy : d_strategy_set)
```

src/IB/IBLagrangianForceStrategySet.cpp
```diff
@@ -61,9 +61,9 @@ IBLagrangianForceStrategySet::~IBLagrangianForceStrategySet()
...
     {
-        (*cit)->setTimeInterval(current_time, new_time);
+        set->setTimeInterval(current_time, new_time);
```
boyceg:
```suggestion
        strategy->setTimeInterval(current_time, new_time);
```

abarret:
@boyceg I've fixed the naming issues. Once you give me the thumbs up on the rest of the changes I will squash and fix the conflicts. I'm pretty sure these conflicts are just overlaps with #367.

ibtk/src/lagrangian/LSiloDataWriter.cpp
```diff
@@ -469,9 +468,9 @@ build_local_ucd_mesh(DBfile* dbfile,
 
     // Prune duplicate edges.
     std::set<std::pair<int, int> > local_edge_set;
-    for (auto it = edge_map.begin(); it != edge_map.end(); ++it)
+    for (const auto& map : edge_map)
```
boyceg: Is this a map?

src/IB/IBRedundantInitializer.cpp
```diff
@@ -403,9 +403,9 @@ IBRedundantInitializer::initializeSprings()
 
                 int min_idx = 0;
                 int max_idx = d_num_vertex[ln][j];
-                for (auto it = d_spring_edge_map[ln][j].begin(); it != d_spring_edge_map[ln][j].end(); ++it)
+                for (const auto& edge_map : d_spring_edge_map[ln][j])
```
boyceg: Are these maps? Or edges?

boyceg:
Couple more tiny things that I just commented on. Also, I think that most occurrences of
```
for (auto& k : is_XXX)
```
should be
```
for (auto& is : is_XXX)
```

It might be worth doing a quick check for range-based for loops where the variable name is a common index variable name (`d`, `i`, `j`, `k`, `n` come to mind) and see if any of those now no longer refer to indexes but to actual variables.

After you do that, feel free to squash, and I will merge once Jenkins give the 👍.

ibtk/src/lagrangian/LSiloDataWriter.cpp
abarret: I'm not sure what an appropriate name would be. `edge_map` is of type `std::multimap<int, std::pair<int, int> >`. Perhaps `elem` for an element of the map? This kind of situation happens several times

src/IB/IBRedundantInitializer.cpp
drwells: They are `std::pair<int, Edge>`s. I usually write something like `edge_pair`.

boyceg:
Or just edge (or even just e for edge).

abarret:
@boyceg I fixed a few more names. This branch is now squashed and rebased on top of master

boyceg:
Let's do this.

PR #365, opened 2018-10-19
Adding missing virtual keyword for INSVCStaggeredHierarchyIntegrator destructor

nishantnangia:
This destructor should be virtual -- not sure how I missed this...

### DIFF
--- a/include/ibamr/INSVCStaggeredHierarchyIntegrator.h
+++ b/include/ibamr/INSVCStaggeredHierarchyIntegrator.h
@@ -140,7 +140,7 @@ public:
      * integrator object with the restart manager when the object is so
      * registered.
      */
-    ~INSVCStaggeredHierarchyIntegrator();
+    virtual ~INSVCStaggeredHierarchyIntegrator();
 
     /*!
      * Get the convective operator being used by this solver class.


include/ibamr/INSVCStaggeredHierarchyIntegrator.h
```diff
@@ -140,7 +140,7 @@ class INSVCStaggeredHierarchyIntegrator : public INSHierarchyIntegrator
      * integrator object with the restart manager when the object is so
      * registered.
      */
-    ~INSVCStaggeredHierarchyIntegrator();
+    virtual ~INSVCStaggeredHierarchyIntegrator();
```
amneetb: Does this matter in practice?

nishantnangia: Not really. But in many cases in main.cpp, we have,

`Pointer<INSVCStaggeredHierarchyIntegrator> ins_integrator = new INSVCStaggeredConservativeHierarchyIntegrator(...)`

which is deleted through the base class destructor when the simulation ends. I noticed that sometimes at the end of the simulation (i.e. when main returns), I get a segfault. Making the destructor virtual fixes that, I think. This is the safe thing to do.

amneetb: Yeah, segfault will happen with this usage and without `virtual`keyword.

drwells:
There is no harm in doing this but since the SAMRAI base class has a virtual destructor this should already be virtual: its possible that there is another destruction bug lurking around.

boyceg:
I am going to revert this merge -- I just double-checked, and @drwells is correct; this destructor should already be virtual thanks to the virtual destructor in `SAMRAI::tbox::DescribedClass`.

BTW, @drwells and @abarret, this brings up an issue when we move from SAMRAI 2 to SAMRAI 3 --- SAMRAI 3 removes `SAMRAI::tbox::Pointer<T>` in favor of `std::shared_ptr<T>`, and the `DescribedClass` base class will go away. So, we will need to make sure that virtual destructors are in place wherever needed...

drwells:
Good catch. I'll put together a patch that explicitly marks destructors of things that presently inherit from `DescribedClass` as virtual.

nishantnangia:
I see; didn't notice that.

drwells:
To be totally honest: past `std::complex` and `std::array` I don't really think it matters if classes have virtual destructors. Looking up a virtual function is much cheaper than deallocating memory, so I don't mind making them virtual on everything that is not a POD class.

PR #369, opened 2018-10-22
Revert "Adding missing virtual keyword for INSVCStaggeredHierarchyIntegrator destructor"

boyceg:
Reverts IBAMR/IBAMR#365


drwells:
I don't have strong preferences on this patch either way, but lets at least wait until GitHub is back to 100% functionality so that we do not have to do this PR twice.

Issue #263, opened 2018-07-24
SAMRAI 3 conversion


drwells:
I have discussed this with @abarret and @boyceg both in person and over email; we should summarize everything here.

I talked to @abarret about this yesterday: since ~~SAMRAI 3.x~~ SAMRAI 3.12 requires C++11 (e.g., `Pointer` is replaced by `std::shared_ptr`) we should merge #261 before switching over.

One major change in the new SAMRAI releases is that most classes were de-templatized for a variety of reasons (IIRC, some classes don't even use the template argument). Here is a brief list of what needs to happen:
- [x] come up with a list of incompatibilities between the two versions (@abarret has this)
- [ ] write a script that can automatically switch most of the constructs (I believe @abarret has this too)
- [ ] examine the necessity of the patches @deleeke wrote: these performance problems have *probably* been fixed over the last decade but we should double check (and send the SAMRAI guys a patch if we need to!)
- [ ] 'press the button' and convert IBAMR to require this newer version (update the library usage and `m4/configure_samrai.m4`)

Since the new versions of SAMRAI are de-templatized, some of the motivation for compiling things in 2D and 3D are now gone. However, since we do (IIRC) compile-time switches between calling one FORTRAN kernel or another I believe it would be best to keep the 2D/3D split for now.

@abarret Please edit this comment if I am missing something.

Edit: It is SAMRAI 3.12 that requires C++11. I believe the present intent is to switch to that version (which is the newest).

boyceg:
> One major change in the new SAMRAI releases is that most classes were de-templatized for a variety of reasons (IIRC, some classes don't even use the template argument). Here is a brief list of what needs to happen:
> 
>  come up with a list of incompatibilities between the two versions (@abarret <https://github.com/abarret> has this)
>  write a script that can automatically switch most of the constructs (I believe @abarret <https://github.com/abarret> has this too)
>  examine the necessity of the patches @deleeke <https://github.com/deleeke> wrote: these performance problems have probably been fixed over the last decade but we should double check (and send the SAMRAI guys a patch if we need to!)
I believe that this performance issue is still present in SAMRAI 3. I reported it when we discovered it earlier in the AY.
>  'press the button' and convert IBAMR to require this newer version (update the library usage and m4/configure_samrai.m4)
You are very optimistic! :-)
[...]

PR #370, opened 2018-10-22
Mark more destructors as virtual.

drwells:
From the discussion in #365: lets fix this before we forget.

These destructors are, presently, already virtual because we inherit from `tbox::DescribedClass`. However, this class was removed in SAMRAI 3, so we will need to do explicitly make them virtual when we require SAMRAI 3 in the future.


drwells:
Related to #263.

PR #376, opened 2018-10-26
Remove unused destructors, part 1.

drwells:
We have a lot of empty destructor implementations (hundreds). This commit gets rid of a few.

I checked that, in all cases, we will not end up converting a destructor from virtual to non-virtual; all classes with virtual destructors here already have a base class with a virtual destructor.


boyceg:
👍 

Shall I merge?

PR #377, opened 2018-10-31
setting up restart for IBFEInstrumentPanel

cpuelz:


cpuelz:
@boyceg how does this look?  it occurred to me that we need this for the really long heart simulations.

boyceg:
I think that it is probably better not to explicitly append.

The reason is that sometimes you will want to restart after a "dirty" shutdown (i.e. a shutdown that does not occur just after writing restart files). In such cases, you will need to do some surgery to patch the log files together. Maybe I have just been inured to doing things in complicated ways, but it seems better to me to keep the files separate and then to merge manually as needed. Otherwise, you have to hunt around in the middle of the automatically merged files for the "duplicated" entries.

PR #378, opened 2018-10-31
Add versioning macros.

drwells:
This commit adds basic preprocessor constants and one function for doing version checks. One can now do things like
```cpp
#if IBAMR_VERSION_GTE(1, 0, 0)
    std::vector<MeshBase *> meshes;
#else
    std::vector<Mesh *> meshes;
#endif
```
for compatibility with the changes made in #368.

I really wish I had realized this was missing before the 0.3.0 release; better late than never, I suppose.

### DIFF
--- a/config/IBAMR_config.h.tmp.in
+++ b/config/IBAMR_config.h.tmp.in
@@ -209,3 +209,15 @@
 
 /* Version number of package */
 #undef VERSION
+
+/* Major version of IBAMR. */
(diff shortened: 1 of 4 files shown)


drwells:
autoheader is trying to outsmart me... this doesn't work. I'll patch it in a few minutes.

drwells:
That should work.

Jenkins retest this please

PR #380, opened 2018-10-31
Initialize in header files.

abarret:
Removes initializations in constructors and sets default values in header files.


drwells:
This is going to take awhile to get through.

@abarret and I talked about about style: I somewhat dislike writing
```cpp
     SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> d_input_db = nullptr;
    [...]
    std::string d_restart_read_dirname = "";
```

because, in both cases, the default constructor suffices (in particular `std::shared_ptr` begins life as a `nullptr`). @boyceg do you have a preference for this?

Since I brought it up, I volunteer to write (at least part of) a follow-up patch :)

drwells:
Another concern that I have, that @abarret pointed out, is that this change is nontrivial and huge; we don't really have sufficient test coverage (at this point) to provide a reasonable guarantee that an integer that was once `-1` is not now `0`. We can comb through this all we want but its still a risk.

abarret:
> Since I brought it up, I volunteer to write (at least part of) a follow-up patch :)

@drwells I'll take care of it right now. It shouldn't take long.

abarret:
@drwells Take a look at it now. I removed initializations that used default constructors and `SAMRAI::tbox::Pointer` objects that were initialized with `nullptr`

ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -885,28 +886,30 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
...
      * Time and time step size data read from input or set at initialization.
      */
-    double d_integrator_time, d_start_time, d_end_time;
-    double d_dt_init, d_dt_min, d_dt_max, d_dt_growth_factor;
-    int d_integrator_step, d_max_integrator_steps;
+    double d_integrator_time = std::numeric_limits<double>::quiet_NaN(), d_start_time = 0.0,
+           d_end_time = std::numeric_limits<double>::max();
+    double d_dt_init = std::numeric_limits<double>::max(), d_dt_min = 0.0,
+           d_dt_max = std::numeric_limits<double>::max(), d_dt_growth_factor = 2.0;
+    int d_integrator_step = 0, d_max_integrator_steps = std::numeric_limits<double>::max();
```
drwells: This cannot be the correct value for `d_max_integrator_steps` since it causes an overflow (one is `int` and the other is `double`).

ibtk/src/solvers/impls/CCPoissonLevelRelaxationFACOperator.cpp
```diff
@@ -142,20 +142,12 @@ CCPoissonLevelRelaxationFACOperator::CCPoissonLevelRelaxationFACOperator(const s
           CELLG,
           input_db,
           default_options_prefix),
-      d_level_solver_type(),
-      d_level_solvers(),
-      d_level_solver_db(),
-      d_coarse_solver(nullptr),
-      d_coarse_solver_db()
+      d_level_solver_default_options_prefix(default_options_prefix + "level_"),
```
drwells: We get a new warning based on rearranging things in the header.

ibtk/include/ibtk/HierarchyIntegrator.h
abarret: Yes. It should be `std::numeric_limits<int>::max()`. I'll fix this once you find all of my mistakes...

include/ibamr/CIBSaddlePointSolver.h
```diff
@@ -360,23 +362,23 @@ class CIBSaddlePointSolver : public SAMRAI::tbox::DescribedClass
...
-    std::string d_object_name, d_ksp_type, d_pc_type;
-    bool d_is_initialized;
-    bool d_reinitializing_solver;
+    std::string d_object_name, d_ksp_type = KSPGMRES, d_pc_type = "none";
+    bool d_is_initialized = false;
+    bool d_reinitializing_solver = false;
 
-    Vec d_petsc_x, d_petsc_b;
+    Vec d_petsc_x = nullptr, d_petsc_b = nullptr;
     std::string d_options_prefix;
```
drwells: We also get a reordering warning here (from the source file).

drwells:
its a huge PITA to just copy and paste error descriptions into GitHub when there are more than a few (there are a couple dozen). Let me know if you want me to push a patch or if you can handle it.

abarret:
> its a huge PITA to just copy and paste error descriptions into GitHub when there are more than a few (there are a couple dozen). Let me know if you want me to push a patch or if you can handle it.

I'll take care of it. I remember noticing those but I forgot to go back and fix them.

drwells:
The test suite is not happy:

`*** Error in `./test2d': double free or corruption (!prev): 0x0000000004cd8510 ***`

at least it can catch something obvious :)

drwells:
We still have some failing tests.

drwells:
I am working through this now.

It might help to give `-1` a special name like `ibtk::invalid_index`. What do you think?

drwells:
I am surprised that `SAMRAI::tbox::Array<int> d_tag_buffer = { 0 };` actually works...

drwells:
I want to check that `IntVector<NDIM> d_stencil_width = 1` is the same as `IntVector<NDIM> d_stencil_width(1)`. I thought the first filled with ones but the second just set the first value to 1 (and the rest to zero).

abarret:
I believe I have fixed the failing test. There was a raw pointer that wasn't initialized but still being deleted. Perhaps we should put replacing raw pointers with smart pointers on the to do list?

> I am surprised that `SAMRAI::tbox::Array<int> d_tag_buffer = { 0 };` actually works...

`SAMRAI::tbox::Array` will be replaced with `std::vector` once the SAMRAI 3 upgrade comes along.

ibtk/include/ibtk/PETScNewtonKrylovSolver.h
```diff
@@ -374,18 +374,18 @@ class PETScNewtonKrylovSolver : public NewtonKrylovSolver
...
 
     std::string d_options_prefix;
 
-    MPI_Comm d_petsc_comm;
-    SNES d_petsc_snes;
-    Mat d_petsc_jac;
-    bool d_managing_petsc_snes;
-    bool d_user_provided_function;
-    bool d_user_provided_jacobian;
+    MPI_Comm d_petsc_comm = PETSC_COMM_WORLD;
```
drwells: This is already initialized in the constructor.

ibtk/include/ibtk/FACPreconditioner.h
```diff
@@ -295,10 +295,10 @@ class FACPreconditioner : public LinearSolver
 
     SAMRAI::tbox::Pointer<FACPreconditionerStrategy> d_fac_strategy;
     SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > d_hierarchy;
-    int d_coarsest_ln;
-    int d_finest_ln;
+    int d_coarsest_ln = 0;
+    int d_finest_ln = 0;
     MGCycleType d_cycle_type;
```
drwells: Can we initialize this also to `V_CYCLE`?

boyceg:
> On Nov 2, 2018, at 12:22 PM, David Wells ***@***.***> wrote: I want to check that IntVector<NDIM> d_stencil_width = 1 is the same as IntVector<NDIM> d_stencil_width(1). I thought the first filled with ones but the second just set the first value to 1 (and the rest to zero).
> Hrmmm, I too thought that IntVector<NDIM> d_stencil_width(1) fills the vector with 1's: https://ibamr.github.io/IBAMR-docs/samrai/2.4.4/html/classSAMRAI_1_1hier_1_1IntVector.html#f50d4e58d4a831476542ef2dcd85c471 <https://ibamr.github.io/IBAMR-docs/samrai/2.4.4/html/classSAMRAI_1_1hier_1_1IntVector.html#f50d4e58d4a831476542ef2dcd85c471>

In fact, here is the code from SAMRAI 2.4.4:
```
template<int DIM> inline 
IntVector<DIM>::IntVector(const int value)
{
   for(int i=0; i < DIM; i++)
      d_vector[i] = value;
}
```

Are we sure that `IntVector<NDIM> stencil = { 1 }` isn't just setting the first element to 1 and leaving the others undefined?

boyceg:
One of the things that changes in SAMRAI 3 is the way that classes like `IntVector` and `Index` are initialized --- because these classes are no longer templated on dimension, the dimension information has to come in through the constructor somehow or another.

It might behoove us to go ahead and eliminate usages like `IntVector<NDIM> i = 1` or `Index<NDIM> i = 0` in favor of named variables. I am not crazy about any of these suggestions, but maybe something like `ibtk_zero_vector` or `ibtk_all_zeros` or `ibtk_all_ones`?

ibtk/src/solvers/impls/CCPoissonLevelRelaxationFACOperator.cpp
```diff
@@ -142,20 +142,12 @@ CCPoissonLevelRelaxationFACOperator::CCPoissonLevelRelaxationFACOperator(const s
           CELLG,
           input_db,
           default_options_prefix),
-      d_level_solver_type(),
-      d_level_solvers(),
-      d_level_solver_db(),
-      d_coarse_solver(nullptr),
-      d_coarse_solver_db()
+      d_level_solver_type(CCPoissonSolverManager::PETSC_LEVEL_SOLVER),
+      d_level_solver_default_options_prefix(default_options_prefix + "level_")
```
drwells: This is a change from what we had before; I assume this fixes something?

boyceg:
> It might behoove us to go ahead and eliminate usages like `IntVector<NDIM> i = 1` or `Index<NDIM> i = 0` in favor of named variables. I am not crazy about any of these suggestions, but maybe something like `ibtk_zero_vector` or `ibtk_all_zeros` or `ibtk_all_ones`?

...or to have a helper function that builds an `IntVector` or `Index` with a specified constant value?  `ibtk::constant_vector(x)`?

drwells:
I finished reading through the IBTK changes (I left comments where I had them).

drwells:
@boyceg I am cautiously optimistic that since `tbox::Dimension` cannot be converted to an integer that our current set of constructor calls will mostly work; i.e., `Index(42, 42)` sets up a 2D index while `Index(42, 42, 42)` sets up a 3D index.

drwells:
No, I think I might be wrong for our most common use case; `Index(int)` is no longer a valid constructor; it needs a dimension argument.

drwells:
I'll let @abarret do the hard thinking about this type of vector initialization; I suspect he already knows the fix.

boyceg:
> In ibtk/include/ibtk/FACPreconditioner.h <https://github.com/IBAMR/IBAMR/pull/380#discussion_r230435171>:
> 
> > @@ -295,10 +295,10 @@ class FACPreconditioner : public LinearSolver
>  
>      SAMRAI::tbox::Pointer<FACPreconditionerStrategy> d_fac_strategy;
>      SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > d_hierarchy;
> -    int d_coarsest_ln;
> -    int d_finest_ln;
> +    int d_coarsest_ln = 0;
> +    int d_finest_ln = 0;
>      MGCycleType d_cycle_type;
> Can we initialize this also to V_CYCLE?
> 
Yes please.

abarret:
> Are we sure that `IntVector<NDIM> stencil = { 1 }` isn't just setting the first element to 1 and leaving the others undefined?

I've tested this and both `IntVector<NDIM> stencil = { 1 }` and `IntVector<NDIM> stencil = 1` sets all elements of `stencil` to one. Although this will have to be changed once we move to SAMRAI 3 since this kind of initialization won't work because of the `Dimension` object requirement.


> It might behoove us to go ahead and eliminate usages like `IntVector<NDIM> i = 1` or `Index<NDIM> i = 0` in favor of named variables. I am not crazy about any of these suggestions, but maybe something like `ibtk_zero_vector` or `ibtk_all_zeros` or `ibtk_all_ones`?

SAMRAI 3 is already ahead of you on this one. There are static methods to construct `IntVector`'s of all ones and zeros using notation like `IntVector::getZero(dim)` or `IntVector::getOne(dim)`. I have a script that _should_ address these things.

ibtk/src/solvers/impls/CCPoissonLevelRelaxationFACOperator.cpp
abarret: I moved a line that set a value inside the constructor to the initialization list. This should avoid an extra constructor call, right?

drwells:
I am a bit stumped right now: we have introduced a bug into `CIBMethod::getStructureHandle`. One of the CIB tests now manages to get to the nonsense `return -1;` statement at the end.

abarret:
> I am a bit stumped right now: we have introduced a bug into `CIBMethod::getStructureHandle`. One of the CIB tests now manages to get to the nonsense `return -1;` statement at the end.

`IBStandardInitializer` is reading in the structure input files twice for some reason. I'm not sure why yet.

abarret:
> > I am a bit stumped right now: we have introduced a bug into `CIBMethod::getStructureHandle`. One of the CIB tests now manages to get to the nonsense `return -1;` statement at the end.
> 
> `IBStandardInitializer` is reading in the structure input files twice for some reason. I'm not sure why yet.

I think I found it. I think `IBStandardInitializer` and `IBRedundantInitializer` are both reading in the input filenames. I'm not sure why this is just now causing an error though.

drwells:
I'll rerun the test suite :)

drwells:
Yup; everything passes on master.

abarret:
@boyceg unless @drwells has any suggestions with the IBAMR changes, this branch is ready for review.

ibtk/include/ibtk/HierarchyGhostCellInterpolation.h
```diff
@@ -39,6 +39,11 @@
 #include <string>
 #include <vector>
 
+#include "ibtk/CartCellRobinPhysBdryOp.h"
+#include "ibtk/CartExtrapPhysBdryOp.h"
+#include "ibtk/CartSideRobinPhysBdryOp.h"
+#include "ibtk/CoarseFineBoundaryRefinePatchStrategy.h"
+
```
boyceg: Let's not separate IBAMR/IBTK includes from SAMRAI/libMesh/etc includes.

boyceg: Also, presumably the forward declarations below can be removed if these are now needed.

ibtk/include/ibtk/PETScSAMRAIVectorReal.h
```diff
@@ -269,7 +269,7 @@ class PETScSAMRAIVectorReal
      * PETSc vector object corresponding to this PETScAbstractVectorReal object.
      */
     Vec d_petsc_vector;
-    bool d_vector_created_via_duplicate, d_vector_checked_out_read_write, d_vector_checked_out_read;
+    bool d_vector_created_via_duplicate, d_vector_checked_out_read_write = false, d_vector_checked_out_read = false;
```
boyceg: Should we default `d_vector_created_via_duplicate` to `false`?

ibtk/include/ibtk/muParserRobinBcCoefs.h
```diff
@@ -223,9 +223,9 @@ class muParserRobinBcCoefs : public SAMRAI::solv::RobinBcCoefStrategy<NDIM>
     /*!
      * The mu::Parser objects which evaluate the data-setting functions.
      */
-    std::vector<mu::Parser> d_acoef_parsers;
-    std::vector<mu::Parser> d_bcoef_parsers;
-    std::vector<mu::Parser> d_gcoef_parsers;
+    std::vector<mu::Parser> d_acoef_parsers = std::vector<mu::Parser>(2 * NDIM);
```
boyceg: Is there some way to avoid repeating the typename here?

ibtk/src/boundary/HierarchyGhostCellInterpolation.cpp
```diff
@@ -111,23 +111,6 @@ static Timer* t_fill_data_set_physical_bcs;
 /////////////////////////////// PUBLIC ///////////////////////////////////////
 
 HierarchyGhostCellInterpolation::HierarchyGhostCellInterpolation()
-    : d_is_initialized(false),
```
boyceg: 💯

ibtk/src/boundary/cf_interface/CartCellDoubleLinearCFInterpolation.cpp
```diff
@@ -99,12 +99,7 @@ static const int GHOST_WIDTH_TO_FILL = 1;
 /////////////////////////////// PUBLIC ///////////////////////////////////////
 
 CartCellDoubleLinearCFInterpolation::CartCellDoubleLinearCFInterpolation()
-    : d_patch_data_indices(),
-      d_consistent_type_2_bdry(false),
-      d_refine_op(new CellDoubleConstantRefine<NDIM>()),
-      d_hierarchy(nullptr),
-      d_cf_boundary(),
-      d_periodic_shift()
+    : d_refine_op(new CellDoubleConstantRefine<NDIM>())
```
boyceg: Can this go in the header too?

ibtk/src/boundary/cf_interface/CartCellDoubleQuadraticCFInterpolation.cpp
```diff
@@ -214,12 +214,7 @@ is_corner_point(const Index<NDIM>& i,
 /////////////////////////////// PUBLIC ///////////////////////////////////////
 
 CartCellDoubleQuadraticCFInterpolation::CartCellDoubleQuadraticCFInterpolation()
-    : d_patch_data_indices(),
-      d_consistent_type_2_bdry(false),
-      d_refine_op(new CartesianCellDoubleLinearRefine<NDIM>()),
-      d_hierarchy(nullptr),
-      d_cf_boundary(),
-      d_periodic_shift()
+    : d_refine_op(new CartesianCellDoubleLinearRefine<NDIM>())
```
boyceg: Same as above.

ibtk/src/boundary/cf_interface/CartSideDoubleQuadraticCFInterpolation.cpp
```diff
@@ -163,11 +163,7 @@ static const int GHOST_WIDTH_TO_FILL = 1;
 /////////////////////////////// PUBLIC ///////////////////////////////////////
 
 CartSideDoubleQuadraticCFInterpolation::CartSideDoubleQuadraticCFInterpolation()
-    : d_patch_data_indices(),
-      d_consistent_type_2_bdry(false),
-      d_refine_op(new CartesianSideDoubleConservativeLinearRefine<NDIM>()),
-      d_hierarchy(nullptr),
-      d_cf_boundary(),
+    : d_refine_op(new CartesianSideDoubleConservativeLinearRefine<NDIM>()),
```
boyceg: Looks like clang-tidy isn't moving things that initialize `Pointer<T>` via `new T`?

ibtk/src/solvers/impls/CCPoissonLevelRelaxationFACOperator.cpp
```diff
@@ -142,20 +142,12 @@ CCPoissonLevelRelaxationFACOperator::CCPoissonLevelRelaxationFACOperator(const s
           CELLG,
           input_db,
           default_options_prefix),
-      d_level_solver_type(),
-      d_level_solvers(),
-      d_level_solver_db(),
-      d_coarse_solver(nullptr),
-      d_coarse_solver_db()
+      d_level_solver_type(CCPoissonSolverManager::PETSC_LEVEL_SOLVER),
```
boyceg: Can this go in the header?

ibtk/src/solvers/impls/CCPoissonLevelRelaxationFACOperator.cpp
```diff
@@ -142,20 +142,12 @@ CCPoissonLevelRelaxationFACOperator::CCPoissonLevelRelaxationFACOperator(const s
...
+      d_level_solver_default_options_prefix(default_options_prefix + "level_")
 {
     // Set some default values.
     d_prolongation_method = "LINEAR_REFINE";
```
boyceg: Header for these inits?

ibtk/src/solvers/impls/PETScLevelSolver.cpp
```diff
@@ -127,23 +127,11 @@ generate_petsc_is_from_std_is(std::vector<std::set<int> >& overlap_std,
...
-      d_shell_pc_type(""),
-      d_options_prefix(""),
-      d_petsc_ksp(nullptr),
-      d_petsc_mat(nullptr),
-      d_petsc_pc(nullptr),
-      d_petsc_x(nullptr),
-      d_petsc_b(nullptr)
 {
     // Setup default options.
     d_max_iterations = 10000;
```
boyceg: Header?

ibtk/src/solvers/impls/PoissonSolver.cpp
```diff
@@ -52,7 +52,7 @@ namespace IBTK
 
 /////////////////////////////// PUBLIC ///////////////////////////////////////
 
-PoissonSolver::PoissonSolver() : d_poisson_spec(""), d_default_bc_coef(nullptr), d_bc_coefs()
+PoissonSolver::PoissonSolver() : d_poisson_spec("")
```
boyceg: `d_poisson_spec` looks like something that can be initialized in the header?

ibtk/src/solvers/impls/SCPoissonHypreLevelSolver.cpp
```diff
@@ -94,25 +94,7 @@ enum HypreSStructRelaxType
...
-      d_solver_type("Split"),
-      d_precond_type("none"),
-      d_split_solver_type("PFMG"),
-      d_rel_change(0),
-      d_num_pre_relax_steps(1),
-      d_num_post_relax_steps(1),
-      d_relax_type(RELAX_TYPE_WEIGHTED_JACOBI),
-      d_skip_relax(1),
-      d_two_norm(1)
+    : d_relax_type(RELAX_TYPE_JACOBI)
```
boyceg: ?

ibtk/src/boundary/cf_interface/CartSideDoubleQuadraticCFInterpolation.cpp
abarret: It didn't move most of these things :). I think it only moves built in types.

I didn't change these because I was getting an error that I thought was due to typecasting that couldn't be done in header files, but now I realize that it's just due to namespaces. I'll update these.

ibtk/src/solvers/impls/CCPoissonLevelRelaxationFACOperator.cpp
abarret: These were members that were inherited from a parent class. Some of these override default values like
`d_prolongation_method = "LINEAR_REFINE";`
overrides the default value of
`"DEFAULT"` provided in `PoissonFACPreconditionerStrategy.h`
but others are the same and could be deleted such as
`d_coarse_solver_rel_residual_tol = 1.0e-5;`

I can go back and check all of these variables.

ibtk/include/ibtk/muParserRobinBcCoefs.h
abarret: We could instead use the brace initialization
`std::vector<mu::Parser> d_acoef_parsers{ 2*NDIM };`
but I chose to use consistent notation for initialization instead of compactness. Which would you prefer?

ibtk/include/ibtk/HierarchyGhostCellInterpolation.h
abarret: Actually these headers are no longer needed since the default constructor for the `std::vector` objects have been deleted.

boyceg:
> @boyceg there are two questions left,
> 
> 1. Should I add in default values to variables that look like they should have default values.
>    Pro: It's probably a good thing to have default values.
>    Con: We don't have a good way to check if the changes will have side effects.

Hmm. I think it is a good thing to have this done. Thoughts:

1. Is there any way to automate this? If so, then I think we should just do it.
2. Perhaps these changes could go into a single commit, so that if it turns out that you and David are right that we shouldn't have done it, we can easily undo it.

> 2. In some situations should we change how initializations are done for compactness, i.e.
>    `std::vector<mu::Parser> d_acoef_parsers = std::vector<mu::Parser>(2 * NDIM);`
>    vs
>    `std::vector<mu::Parser> d_acoef_parsers{ 2 * NDIM };`

I think it is desirable to avoid duplicating the typename. However, in cases like this, can we simplify things further by replacing the existing code with a fixed-size `std::array`?

boyceg:
> > @boyceg there are two questions left,
> > 
> > 1. Should I add in default values to variables that look like they should have default values.
> >    Pro: It's probably a good thing to have default values.
> >    Con: We don't have a good way to check if the changes will have side effects.
> 
> Hmm. I think it is a good thing to have this done. Thoughts:
> 
> 1. Is there any way to automate this? If so, then I think we should just do it.
> 2. Perhaps these changes could go into a single commit, so that if it turns out that you and David are right that we shouldn't have done it, we can easily undo it.

The reason why I think it is a good thing to do is because it provides clear documentation, in one place, of the default values.

On the other hand, these default values are not captured in the documentation, and so the impact of doing this is primarily in reducing LOCs.

Maybe this is something to do _only_ if it can be automated...

¯\_(ツ)_/¯

abarret:
@boyceg @drwells Any last suggested changes?

boyceg:
Once this gets David’s +1, lets merge.

drwells:
Bah, now I have to read the whole thing :)

drwells:
I am still reading through this. I'll keep posting as I find things.

It looks like `CIBFEMethod::commonConstructor` still does some initializations that could be in the header.

drwells:
We converted some `>>`s to `> >`s in `IBFEInstrumentPanel.h`.

drwells:
In `struct BeamSpec` in `IBRedundantInitializer.h`: we should set the `curvature` value at the declaration. I think we can do the same for the pair with
`std::pair<int, int> neighbor_idxs = {0, 0}`.

drwells:
Quick comment (nothing needs to be changed): we removed the member variable `d_object_name` from `IBInitializer` since the base class `IBRedundantInitializer` already has exactly that member variable.

drwells:
We don't seem to use `d_dt` in INSVCStaggeredConservativeMassMomentumIntegrator. I think we should still initialize it to `quiet_NaN` just to be sure.

drwells:
I am not sure if it matters (we overwrite it later) but we previously set `d_num_steps` to `1` and not `-1` in INSVCStaggeredConservativeMassMomentumIntegrator.

drwells:
Okay, I have reached the end of the changes in IBAMR. We should be set aside from the items pointed out above: 

- [ ] `CIBFEMethod::commonConstructor` still does some initializations that could be in the header.
- [ ] In IBFESurfaceMethod.cpp: we should set `d_use_consistent_mass_matrix` to `true`.
- [ ] In struct BeamSpec in IBRedundantInitializer.h: we should set the curvature value at the declaration. I think we can do the same for the pair with `std::pair<int, int> neighbor_idxs = {0, 0}`.
- [ ] We don't seem to use d_dt in INSVCStaggeredConservativeMassMomentumIntegrator. I think we should still initialize it to quiet_NaN just to be sure.
- [ ] We previously set `d_num_steps` to 1 and not -1 in INSVCStaggeredConservativeMassMomentumIntegrator.

abarret:
@boyceg Rebased and squashed into 2 commits. This is ready to merge once Jenkins gives the ok.

PR #381, opened 2018-11-01
Use delete on private special member functions.

abarret:
This is a run of clang-tidy with modernize-use-equals-delete

### DIFF
--- a/ibtk/include/ibtk/AppInitializer.h
+++ b/ibtk/include/ibtk/AppInitializer.h
@@ -208,7 +208,7 @@ private:
      *
      * \param from The value to copy to this object.
      */
-    AppInitializer(const AppInitializer& from);
+    AppInitializer(const AppInitializer& from) = delete;
 
     /*!
      * \brief Assignment operator.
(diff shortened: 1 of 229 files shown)


drwells:
I did not look very carefully but the lines changed count doesn't seem right: we should be able to delete all implementations but we are at +609 -609. Does this compile with the implementations?

boyceg:
This is not worth doing by hand, but if you can cobble together a regex to wipe out the comments that are attached to these deleted functions, that would be awesome.

drwells:
I'll see if I can throw something together.

drwells:
Actually, it looks like a lot of these things were just never defined anyway (its not like we could use them so this never caused linking errors).

drwells:
I could not find a single case of an implemented private copy constructor (or copy operator) so I think we are okay. clang-tidy did miss a few `= delete`s, though; I'll push an update to this branch.

drwells:
Automatically rewriting all the comments will take a bit of work, but I think I know how to script it.

drwells:
Yes. Lets get merge so that we can get on with the SAMRAI 3 migration. I'll open a separate issue for rewording the documentation. I will get on that in a day or two when I need something less exciting to do :)

PR #383, opened 2018-11-02
Add some checks for valid structure ids.

drwells:
Caught while debugging something in #380.

I opted to add the check at the call site instead of the function itself since, conceivably, someone else already has such a check (and an invalid structure id may not mean the program cannot continue).


drwells:
This won't fix the bug introduced in #380 but it will cause the code to abort instead of silently corrupting the heap.

PR #386, opened 2018-11-07
Add code for partitioning the Lagrangian mesh according to Eulerian data.

drwells:
Super work in progress right now. Obviously I'll clean up the history, add more formal tests, and fix bugs as I get this working.

As I mentioned via email: I had some trouble figuring out data dependencies so I completely rewrote `IBFE/explicit/ex4/` to make them more obvious. This example step is now, presently, broken (more stuff to fix) but I have a much better understanding now of how we start an IBFE simulation (that is, how we set up Eulerian data, Lagrangian data, and the couplings between them).


drwells:
I am still working on this. I rewrote ex4 in a way that makes the data interdependencies more clear (at this point mostly for my own benefit).

I am still trying to figure out some of the order in which we initialize certain objects; in particular the postprocessor adds links between the Eulerian and Lagrangian data. I believe we need to set up the postprocessor before setting up the time integrator (at least the whole program crashes in libMesh code if I set it up after the NSE integrator).

I talked about this with @boyceg over email: the tentative conclusion is that we should set up the partitioning in the libMesh mesh near the end when setting up the NSE integrator (in particular, in `IBFEMethod::beginDataRedistribution()` which is called by `IBHierarchyIntegrator::initializePatchHierarchy`). This is where we do it in the IB code (not IBFE).

 think life would be easier if we did the partitioning early on (before setting up discretizations). All we need are the patches to set up the libMesh `Mesh`.

boyceg:
> I am still working on this. I rewrote ex4 in a way that makes the data interdependencies more clear (at this point mostly for my own benefit).
> 
> I am still trying to figure out some of the order in which we initialize certain objects; in particular the postprocessor adds links between the Eulerian and Lagrangian data. I believe we need to set up the postprocessor before setting up the time integrator (at least the whole program crashes in libMesh code if I set it up after the NSE integrator).
> 
> I talked about this with @boyceg over email: the tentative conclusion is that we should set up the partitioning in the libMesh mesh near the end when setting up the NSE integrator (in particular, in `IBFEMethod::beginDataRedistribution()` which is called by `IBHierarchyIntegrator::initializePatchHierarchy`). This is where we do it in the IB code (not IBFE).

Not quite. What happens is that the AMR grid is regenerated between the calls to `IBFEMethod::beginDataRedistribution()` and `IBFEMethod::endDataRedistribution()`. So you want to use the patch hierarchy configuration to determine the parallel distribution of the `libMesh::Mesh` in `IBFEMethod::endDataRedistribution()`. If you do it in `IBFEMethod::beginDataRedistribution()`, then 

> think life would be easier if we did the partitioning early on (before setting up discretizations). All we need are the patches to set up the libMesh `Mesh`.

If I understand this correctly, this cannot work, because the AMR grid data structures are dynamically repartitioned throughout the computation.

boyceg:
> > I am still trying to figure out some of the order in which we initialize certain objects; in particular the postprocessor adds links between the Eulerian and Lagrangian data. I believe we need to set up the postprocessor before setting up the time integrator (at least the whole program crashes in libMesh code if I set it up after the NSE integrator).
> > I talked about this with @boyceg over email: the tentative conclusion is that we should set up the partitioning in the libMesh mesh near the end when setting up the NSE integrator (in particular, in `IBFEMethod::beginDataRedistribution()` which is called by `IBHierarchyIntegrator::initializePatchHierarchy`). This is where we do it in the IB code (not IBFE).
> 
> Not quite. What happens is that the AMR grid is regenerated between the calls to `IBFEMethod::beginDataRedistribution()` and `IBFEMethod::endDataRedistribution()`. So you want to use the patch hierarchy configuration that is set when `IBFEMethod::endDataRedistribution()` is called to determine the parallel distribution of the `libMesh::Mesh`. If you do it in `IBFEMethod::beginDataRedistribution()`, then you will be repartitioning the `Mesh` using the AMR patch hierarchy that is about to be torn down.

The idea with `IBStrategy::beginDataRedistribtuion()` is to package up data that is currently in the "old" distribution before that distribution gets torn down, and then to reassemble things in `IBStrategy::endDataRedistribution()`. In between those calls, the AMR grid is regenerated and data living on the AMR grid are mapped from the old grid to the new one.

By the way, with nonuniform load balancing, the same thing can happen in even a uniform grid run, since the patches will not generally be uniform in such cases.

drwells:
See #396 for a simplified branch where we ignore the new partitioner and focus on data.

PR #387, opened 2018-11-08
Replace empty special member functions with = default.

abarret:
Last c++11 update.

Note that there were four files that have been deleted since they are now empty.


boyceg:
@abarret is the reigning champion of removing LOC!

boyceg:
This is good to merge once @drwells and Jenkins give the 👍.

abarret:
My goal is to forever have a net negative impact on LOC.

drwells:
If you believe https://github.com/IBAMR/IBAMR/graphs/contributors then @abarret has removed 319,000 more lines than he has added, and with this PR his lead is only widening :)

We actually removed some files this time; not bad.

drwells:
It looks like we need to add
```cpp
#include "ibamr/INSVCStaggeredConservativeMassMomentumIntegrator.h"
```

to `INSVCStaggeredConservativeHierarchyIntegrator.cpp`.

boyceg:
You wanna squash before I merge?

Issue #358, opened 2018-10-08
post #356: get rid of `new double`


drwells:
We should grep through the source and remove any manual allocations of `double` arrays. In all cases
```cpp
auto x = new double[n];
// ...
delete[] x;
```
is just a less convenient version of
```cpp
std::vector<double> x(n);
```

fuhuifang:
That makes sense. I'll make those changes.

boyceg:
Addressed by #393.

PR #389, opened 2018-11-09
get rid of new double in CIBMethod.cpp

fuhuifang:


src/IB/CIBMethod.cpp
```diff
@@ -1402,8 +1402,9 @@ CIBMethod::constructMobilityMatrix(const std::string& /*mat_name*/,
     }
 
     // Get the position data
-    double* XW = nullptr;
-    if (rank == managing_rank) XW = new double[size];
+    std::vector<double> XW;
+    if (rank == managing_rank) XW.resize(size);
+    XW.data();
```
drwells: This statement won't have any effect alone; we should just have a blank line.

src/IB/CIBMethod.cpp
```diff
@@ -1462,7 +1463,6 @@ CIBMethod::constructMobilityMatrix(const std::string& /*mat_name*/,
                 }
             }
         }
-        delete[] XW;
```
drwells: :)

fuhuifang: Just pushed the changes.

boyceg:
This looks good to me. @fuhuifang can you squash into one or two commits (see, e.g., [here](https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit)), and then I will merge.

boyceg:
@amneetb, can you take a look before we merge this?

amneetb:
@fuhuifang: If the CIB examples are building and running (serial and parallel both), then it should be fine. The changes look OK to me.

boyceg:
This is not something that would have been introduced by this PR; it sounds like something that was already broken. @amneetb, can you take a look? @drwells, are we running parallel cases in the tests at this point?

drwells:
We do not have any parallel test cases yet (i.e., nothing is invoked with `mpirun`/`mpiexec`).

This is most likely fallout from #380; I'll take a look.

boyceg:
May be worth running IB/explicit/ex0 in parallel to see if it is CIB or anything IB-like.

boyceg:
I am going to go ahead and merge because I think these changes are all fine.

drwells:
@fuhuifang BTW: if you write 'closes #358' in a comment GitHub will automatically close the issue once this is merged.

boyceg:
I just checked, and this PR does not fully close #358 -- there are still some uses of `new double` lurking; just noting in `CIBMethod.cpp`.

Fuhui, do you want to make a new PR that tries to finish up the rest? Do
```
$ grep "new double" `find . -name *.cpp`
```
You will see that all of this comes from stuff under the auspices of @amneetb. ;-)

boyceg:
BTW, getting rid of `new T` is more difficult, because we initialize `SAMRAI::tbox::Pointer<T>` using `new`.

Because SAMRAI 3 now uses `std::shared_ptr<T>`, we can go back and remove any remaining uses of `new` after that update.

drwells:
Ah, I thought the warning was printed to the screen. I see it in the log file too :)

boyceg:
I am about to push a fix. It looks like what happened is that the member `LDataManager::d_needs_synch` was being initialized incorrectly _before_ we moved initialization to the header.

drwells:
Yes. This issue predates #380.

Issue #359, opened 2018-10-10
static_cast<unsigned> instead of (unsigned)


drwells:
We should use `static_cast<T>` instead of `(T)` so that the casting semantics are clear. In particular, we do this with `(unsigned)`; see the discussion in #356.

drwells:
We should do the same thing for the fundamental type conversion operators, e.g., write `static_cast<float>(x)` instead of `float(x)`.

boyceg:
I think this is closed by #391.

boyceg:
Now we can close this I think.

PR #391, opened 2018-11-11
fix casting in #359

fuhuifang:
Changed the casting semantics to static_cast<T>.

### DIFF
--- a/examples/ConstraintIB/eel3d/eelgenerator3d.cpp
+++ b/examples/ConstraintIB/eel3d/eelgenerator3d.cpp
@@ -226,7 +226,7 @@ main()
     }
     std::fstream eelstream;
     eelstream.open("eel3d.vertex", std::fstream::out);
-    assert(static_cast<unsigned>(total_lag_pts) == shape_new[0].size());
+    assert(static_cast<std::size_t>(total_lag_pts) == shape_new[0].size());
     eelstream << total_lag_pts << "\n";
 
     for (int k = 1; k <= total_lag_pts; ++k)
--- a/examples/ConstraintIB/eel3d/eelgenerator3d.cpp
+++ b/examples/ConstraintIB/eel3d/eelgenerator3d.cpp
@@ -127,8 +127,8 @@ main()
     interp_coefs[3] = a3;
 
     // No. of points on the backbone and till head.
-    const int headNs = int(ceil(LENGTH_TILLHEAD / dx));
-    const int tailNs = int(ceil((LENGTH_FISH - LENGTH_TILLHEAD) / dx));
+    const int headNs = static_cast<int>(ceil(LENGTH_TILLHEAD / dx));
+    const int tailNs = static_cast<int>(ceil((LENGTH_FISH - LENGTH_TILLHEAD) / dx));
     const int bodyNs = headNs + tailNs + 1;
 
     std::vector<std::pair<int, int> > immersedBodyData(bodyNs);
@@ -139,8 +139,8 @@ main()
         const double s = (i - 1) * dx;
         const double section = sqrt(2 * WIDTH_HEAD * s - s * s);
         const double height = MINOR_AXIS * std::sqrt(1 - pow((s - MAJOR_AXIS) / MAJOR_AXIS, 2));
-        const int numPtsInSection = int(ceil(section / dy));
-        const int numPtsInHeight = int(ceil(height / dz));
+        const int numPtsInSection = static_cast<int>(ceil(section / dy));
+        const int numPtsInHeight = static_cast<int>(ceil(height / dz));
         immersedBodyData[i - 1] = std::make_pair(numPtsInSection, numPtsInHeight);
         immersedBodyWidthHeight[i - 1] = std::make_pair(section, height);
     }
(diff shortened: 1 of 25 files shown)


boyceg:
Looks good.

I am not sure if the intention here is to remove all C-style casts, but it looks like there are many that remain. Try running, e.g.,
```
$ for T in int double float unsigned; do grep "($T)" `find . -name *.cpp` | grep -v "sizeof($T)" | grep -v contrib; done
```
and
```
$ for T in int double float unsigned; do grep " $T(" `find . -name *.cpp` | grep -v contrib; done
```
I am not sure that there is an easy way to automate replacing these.

boyceg:
Looks good to me.

@amneetb can you also take a look at this one?

@drwells will probably tell us that many of the casts are actually not needed because of type conversion and promotion rules. :-)

drwells:
Technically, yes, but I very much agree with Stroustrup (D&E, page 333):
> The new-style casts are part of a continuing effort to eliminate holes in the C++ type system. The aim is to minimize and localize unsafe and error-prone programming practices.

It is always good to mark narrowing casts explicitly and `static_cast` has much clearer semantics than C-style casts. I believe `int(x)` is the same as `static_cast<int>(x)` unless `x` is some bizarre user-defined class (i.e., not relevant here) so I am happy to use the more explicit notation.

examples/ConstraintIB/eel3d/eelgenerator3d.cpp
```diff
@@ -226,7 +226,7 @@ main()
     }
     std::fstream eelstream;
     eelstream.open("eel3d.vertex", std::fstream::out);
-    assert((unsigned)total_lag_pts == shape_new[0].size());
+    assert(static_cast<unsigned>(total_lag_pts) == shape_new[0].size());
```
drwells: This is still a bit weird (but not incorrect) since on the right we have `std::size_t`. Could you use that instead of `unsigned` here?

examples/ConstraintIB/falling_sphere/sphereGen3d.cpp
```diff
@@ -62,7 +62,7 @@ main()
 
     std::fstream sphere_coord_stream;
     sphere_coord_stream.open("sphere3d.vertex", std::fstream::out);
-    assert((unsigned)total_lag_pts == sphere_coords[0].size());
+    assert(static_cast<unsigned>(total_lag_pts) == sphere_coords[0].size());
```
drwells: (same)

src/IB/CIBMethod.cpp
```diff
@@ -788,7 +788,7 @@ CIBMethod::midpointStep(double current_time, double new_time)
 
         // Get structures on this level.
         const std::vector<int> structIDs = d_l_data_manager->getLagrangianStructureIDs(ln);
-        const auto structs_on_this_ln = (unsigned)structIDs.size();
+        const auto structs_on_this_ln = static_cast<unsigned>(structIDs.size());
```
drwells: I'd just get rid of this cast completely: its already a larger unsigned type.

src/IB/CIBSaddlePointSolver.cpp
```diff
@@ -1102,8 +1102,8 @@ CIBSaddlePointSolver::monitorKSP(KSP ksp, int it, PetscReal rnorm, void* /*mctx*
     }
 
     std::streamsize old_precision = tbox::plog.precision(16);
-    tbox::plog << std::scientific << it << " SP_KSP " << print_normtype << " resid norm " << (double)rnorm
-               << " true resid norm " << (double)truenorm << " ||r(i)||/||b|| " << (double)(truenorm / bnorm)
+    tbox::plog << std::scientific << it << " SP_KSP " << print_normtype << " resid norm " << static_cast<double>(rnorm)
+               << " true resid norm " << static_cast<double>(truenorm) << " ||r(i)||/||b|| " << static_cast<double>(truenorm / bnorm)
```
drwells: I don't think there is any reason to cast these: `PetscReal` is either `float` or `double` anyway.

src/IB/KrylovMobilitySolver.cpp
```diff
@@ -994,8 +994,8 @@ KrylovMobilitySolver::monitorKSP(KSP ksp, int it, PetscReal rnorm, void* /*mctx*
     }
 
     std::streamsize old_precision = tbox::plog.precision(16);
-    tbox::plog << std::scientific << it << " KMInv_KSP " << print_normtype << " resid norm " << (double)rnorm
-               << " true resid norm " << (double)truenorm << " ||r(i)||/||b|| " << (double)(truenorm / bnorm)
+    tbox::plog << std::scientific << it << " KMInv_KSP " << print_normtype << " resid norm " << static_cast<double>(rnorm)
+               << " true resid norm " << static_cast<double>(truenorm) << " ||r(i)||/||b|| " << static_cast<double>(truenorm / bnorm)
```
drwells: same here (don't cast PETSc types)

src/IB/KrylovFreeBodyMobilitySolver.cpp
```diff
@@ -508,8 +508,8 @@ KrylovFreeBodyMobilitySolver::monitorKSP(KSP ksp, int it, PetscReal rnorm, void*
     }
 
     std::streamsize old_precision = tbox::plog.precision(16);
-    tbox::plog << std::scientific << it << " KFBMInv_KSP " << print_normtype << " resid norm " << (double)rnorm
-               << " true resid norm " << (double)truenorm << " ||r(i)||/||b|| " << (double)(truenorm / bnorm)
+    tbox::plog << std::scientific << it << " KFBMInv_KSP " << print_normtype << " resid norm " << static_cast<double>(rnorm)
+               << " true resid norm " << static_cast<double>(truenorm) << " ||r(i)||/||b|| " << static_cast<double>(truenorm / bnorm)
```
drwells: same here (don't cast PETSc types)

boyceg:
(What I meant is that I think that in many cases, we do not need any cast at all. I put a lot of this stuff in originally because I wanted to make it clear when computations were being performed in, e.g., double precision or integer arithmetic for folks who may not know the C/C++ type conversion and promotion rules by heart.)

drwells:
This is also true: we can do `1.0 / NDIM` and get the correct value. However, since its easy to write `1/NDIM` I still think writing `static_cast` is a good idea.

boyceg:
> This is also true: we can do `1.0 / NDIM` and get the correct value. However, since its easy to write `1/NDIM` I still think writing `static_cast` is a good idea.

This is a good point that I did not think of --- even though the explicit cast is not necessary for `1.0 / NDIM`, it makes it impossible to make the extremely difficult-to-find bug of `1 / NDIM`.

drwells:
Looks good to me. I read it through two more times just to check.

PR #392, opened 2018-11-12
fixing initialization of LDataManager::d_needs_synch

boyceg:
Prior to #380, `std::vector<bool> LDataManager::d_needs_synch` had been initialized via an initialization list in `LDataManager.cpp` via
```
    ...
    d_needs_synch(true),
    ...
```
I think that what was happening here is that `true` was being silently converted to the integer value `1`, and then the constructor was populating `d_needs_synch` with `false`.

This member actually does not need to be initialized to anything, because its values will be set correctly in `LDataManager::setPatchLevels()` when setting up the Lagrangian data structures.

### DIFF
--- a/ibtk/include/ibtk/LDataManager.h
+++ b/ibtk/include/ibtk/LDataManager.h
@@ -1165,7 +1165,7 @@ private:
     /*!
      * Indicates whether the LData is in synch with the LNodeData.
      */
-    std::vector<bool> d_needs_synch = { true };
+    std::vector<bool> d_needs_synch;
 
     /*!
      * PETSc AO objects provide mappings between the fixed global Lagrangian


boyceg:
This PR should address the warning message mentioned by @fuhuifang in #389.

drwells:
Whoa, this is tricky. Sometimes C++ makes me sad. I think `bool` predates `explicit`, which permits this bit of conversion insanity.

Proof:
```cpp
#include <vector>
#include <iostream>

int main()
{
    std::cout << "Test 1: possibly erroneous initialization\n";
    std::cout << std::boolalpha;
    {
        std::vector<bool> booleans(true);
        std::cout << "value: " << booleans[0] << '\n';
    }

    std::cout << "Test 2: initializer list initialization\n";
    {
        std::vector<bool> booleans {true};
        std::cout << "value: " << booleans[0] << '\n';
    }
}
```
generates
```
[drwells@auricle int-bool-test]$ g++ -std=c++11 test.cc
[drwells@auricle int-bool-test]$ ./a.out 
Test 1: possibly erroneous initialization
value: false
Test 2: initializer list initialization
value: true
```

drwells:
Is it true that we want to initialize to `false` anyway, i.e., the object does not need synchronization right after it is constructed?

boyceg:
Synchronization is not needed until later on. It is the job of the initializer to make sure that stuff is initialized in a consistent state.

PR #398, opened 2018-11-20
Rearrange ibfe ex4

drwells:
I'm not sure if we want to go for this, but I figured I would put it up before deleting the commits on #396.


drwells:
The structure volume and logfiles are identical on this branch and master.

boyceg:
I have to admit that I am not crazy about the arrangement here, although it provides one model for how to get rid of a lot of the boilerplate code that shows up in all of these examples/mini applications.

drwells:
It is completely different from the rest of the examples (which is not great). We might want to put this in another place to show how data dependencies work. I'll sort it out once the partitioning is ready.

drwells:
Lets close this for now since it makes this example completely different from all the other ones. I'll keep the branch on GitHub in case we ever want to look at it again.

PR #399, opened 2018-11-25
Fixing CIB and its examples

amneetb:


boyceg:
I got rid of those calls to `std::make_pair` --- looks cleaner to me, but YMMV.

amneetb:
I have learnt all my internet acronyms from you ;-) . 

Looks like C++ is becoming more like MATLAB, cryptic yet sleek. Will follow this initialization style moving forward. There should be some doc page on ibamr.github.io on new C++ style to ensure uniformity among all developers.

boyceg:
Could you give that a whirl?

amneetb:
OK, this made it work. As you pointed out earlier, it is imperative to scrutinize #380 under a microscope.

PR #401, opened 2018-11-27
Update a few default initialization values.

abarret:
Follow up to #380. This fixes a spelling error and an incorrect deletion.


drwells:
Since no one has further comment and this clearly fixes a bug lets merge.

PR #402, opened 2018-11-29
Document SystemDofMapCache.

drwells:
I was looking at this while working on #396 and added some comments.


boyceg:
I have to be honest that I am not 100% sure that this class is necessary. :-D

I did some timings at some point that seemed to show that in some IBFE applications, we were getting killed by DOF indexing, and so I slapped this class together. It is probably worth revisiting, especially once we are repartitioning the libMesh data structures.

PR #403, opened 2018-11-29
Fedatamanager unique ptrs

drwells:
Related to #396: we need to recalculate all members of `FEDataManager` after we repartition the mesh. Its getting a bit complicated to call `delete` in a dozen places, so I replaced all owning raw pointers with `std::unique_ptr`s.

This is a lot easier if we make the (minor) change of really requiring that `libMesh::UniquePtr` is `std::unique_ptr`. I updated the configuration scripts to check this.

### DIFF
--- a/configure
+++ b/configure
@@ -24225,15 +24225,9 @@ fi
   LIBMESH_FCFLAGS="`$LIBMESH_CONFIG --fflags`"
   LIBMESH_LIBS="`$LIBMESH_CONFIG --libs`"
 
-  # libMesh implements UniquePtr by including parts of boost in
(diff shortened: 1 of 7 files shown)


abarret:
I have a script that replaces all instances of `LibMesh::UniquePtr` with `std::unique_ptr` if that would make things easier. I wasn't sure if we wanted to do this.

boyceg:
Lets just require libMesh uses the standard unique pointer class.

drwells:
@abarret this patch requires that those two be identical so we might as well run that script after this is done.

boyceg:
@abarret time to run that script.

abarret:
I'll run it after #405 is merged since there's some overlap.

PR #404, opened 2018-12-03
Reformat loops

drwells:
These commits are cosmetic improvements taken from #396 that are not directly related to partitioning.

I think we should always put `TBOX_ASSERT` in curly brackets since it is a macro; weird things can happen if half of the macro is in one scope and the other half is in another.


boyceg:
Perhaps we should make an inline function `ibtk_assert` instead of relying on the `TBOX_ASSERT` macro.

boyceg:
This looks fine --- will merge once Jenkins is satisfied.

drwells:
I don't think it is possible to avoid macros when writing a good `assert` function, since we need the line number and stringified version of the line itself. It would be nice to have our own IBTK_ASSERT with some more features, though :)

PR #405, opened 2018-12-03
Use unique_ptr and references in IBFEMethod.

drwells:
Some of the IBFE examples (i.e., IBFE/explicit/ex10) leak memory at the end of the calculation: in particular, libMesh claims that we do not delete all `EquationSystem` objects. This PR gets rid of manual memory management of these objects: it doesn't completely fix the problem but it does allow us to get rid of a few tricky `delete` statements.

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -358,10 +358,6 @@ IBFEMethod::IBFEMethod(const std::string& object_name,
 
 IBFEMethod::~IBFEMethod()
 {
-    for (unsigned int part = 0; part < d_num_parts; ++part)
-    {
-        delete d_equation_systems[part];
-    }
     if (d_registered_for_restart)
     {
         RestartManager::getManager()->unregisterRestartItem(d_object_name);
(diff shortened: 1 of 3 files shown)


src/IB/IBFEMethod.cpp
```diff
@@ -1129,24 +1125,26 @@ IBFEMethod::computeLagrangianForce(const double data_time)
...
         for (unsigned int k = 0; k < d_X_half_vecs.size(); ++k)
         {
             if (d_is_overlap_force_part[k])
             {
                 PetscVector<double>* X_half_vec = d_X_half_vecs[k];
                 std::vector<numeric_index_type> ghost_idxs(d_overlap_force_part_ghost_idxs[k].begin(),
                                                            d_overlap_force_part_ghost_idxs[k].end());
-                X_half_ghost_vecs[k] = new PetscVector<double>(
-                    X_half_vec->comm(), X_half_vec->size(), X_half_vec->local_size(), ghost_idxs);
+                X_half_ghost_vecs[k] = std::unique_ptr<libMesh::PetscVector<double>>(
```
boyceg: Can we avoid a `new` here?

boyceg: Maybe not because of the ghosting.

PR #406, opened 2018-12-04
Replaced UniquePtr with unique_ptr

abarret:
Follow up to #403. Replace all instances of `LibMesh::UniquePtr` with `std::unique_ptr`. A perl script is provided in the scripts directory.

If we want to keep using the Libmesh alias we can close this. I won't be offended :)

### DIFF
--- a/examples/CIBFE/ex0/example.cpp
+++ b/examples/CIBFE/ex0/example.cpp
@@ -353,7 +353,7 @@ run_example(int argc, char* argv[])
         {
             time_integrator->registerVisItDataWriter(visit_data_writer);
         }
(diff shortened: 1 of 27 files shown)


boyceg:
Looks good to me; let's merge once Jenkins is happy.

Although this PR gives you a net increase in LOC. ;-)

abarret:
Sure. How about `replaceLibmeshPointers.pl`? `libMesh::AutoPtr` is not used anywhere in IBAMR, but I thought I would include it for completeness. I could also just remove the `AutoPtr` replacement.

> Although this PR gives you a net increase in LOC. ;-)

I figured I could add a few and still be under the 331,000 lines deleted...

boyceg:
Sounds good.

abarret:
Done and rebased.

PR #412, opened 2018-12-13
cache libMesh RHS vectors

boyceg:


boyceg:
@drwells can you see if this makes any difference in performance?

boyceg:
Also @cpuelz.

drwells:
IIRC this patch does help slightly but we won't realize any substantial performance improvements until we figure out the load balancing (i.e., why some processors are ending up with zero libMesh elements).

I think that this caching strategy will save us from recomputing the communication pattern over and over again. I want to convince myself that this is what happens when we clone vectors before proceeding, though.

boyceg:
There appears to be something wrong with the way that I am using either `System::add_vector()` or `System::retrieve_vector()` in handling the "new" and "half" time step variables in the most recent commit.

boyceg:
@drwells can you see if you can run, say, IBFE ex1?

drwells:
I'll take a look.

drwells:
Some of the vectors are not set up correctly: we get a segmentation fault on

```cpp
800        d_X_new_vecs[part] = dynamic_cast<PetscVector<double>*>(d_X_systems[part]->request_vector("new"));
```

in IBFEMethod.cpp since that vector does not exist.

boyceg:
If I switch it out with
```
d_X_new_vecs[part] = dynamic_cast<PetscVector<double>*>(&d_X_systems[part]->add_vector("new", true, GHOSTED));
```
I get the same seg fault.

boyceg:
Ah yes, I am an idiot. The cached vectors ... don't need to be deleted.

However, there is still something screwy. I get a different answer in IBFE/ex1 with the branch than I do in master. Investigating...

boyceg:
OK, I think this is sorted out.

@drwells, can you take a look?

boyceg:
Jenkins retest this please

boyceg:
OK, I think I have done everything that I am planning on this PR unless @drwells has some comments.

The only tricky thing here is that there are some places where we used to use `zero_clone()`; need to make sure that the vectors are properly zeroed out if needed. I think I caught everything.

drwells:
We could simplify this a lot by putting all of the arrays of pointers for each component into a `struct`. I can do that after we merge.

This looks good to me; it passes the full test suite. I want to read through it a few more times to ensure I understand what it changes.

boyceg:
Please simplify! :-)

include/ibamr/IBFEMethod.h
```diff
@@ -768,12 +769,13 @@ class IBFEMethod : public IBStrategy
     const unsigned int d_num_parts = 1;
     std::vector<IBTK::FEDataManager*> d_fe_data_managers;
     SAMRAI::hier::IntVector<NDIM> d_ghosts = 0;
-    std::vector<libMesh::System*> d_X_systems, d_U_systems, d_F_systems, d_Q_systems, d_Phi_systems;
-    std::vector<libMesh::PetscVector<double> *> d_X_current_vecs, d_X_new_vecs, d_X_half_vecs, d_X_IB_ghost_vecs;
-    std::vector<libMesh::PetscVector<double> *> d_U_current_vecs, d_U_new_vecs, d_U_half_vecs;
-    std::vector<libMesh::PetscVector<double> *> d_F_half_vecs, d_F_IB_ghost_vecs;
-    std::vector<libMesh::PetscVector<double>*> d_Q_half_vecs, d_Q_IB_ghost_vecs;
-    std::vector<libMesh::PetscVector<double>*> d_Phi_half_vecs;
+    std::vector<libMesh::ExplicitSystem*> d_X_systems, d_U_systems, d_F_systems, d_Q_systems, d_Phi_systems;
```
drwells: Is there a reason we switched to `ExplicitSystem`?

src/IB/IBFEMethod.cpp
```diff
@@ -767,62 +767,70 @@ IBFEMethod::preprocessIntegrateData(double current_time, double new_time, int nu
...
-            d_U_current_vecs[part]->clone().release()); // WARNING: must be manually deleted
-        d_U_half_vecs[part] = dynamic_cast<PetscVector<double>*>(
-            d_U_current_vecs[part]->clone().release()); // WARNING: must be manually deleted
+        d_U_rhs_vecs[part] = dynamic_cast<PetscVector<double>*>(d_U_systems[part]->rhs);
+        d_U_new_vecs[part] = dynamic_cast<PetscVector<double>*>(d_U_systems[part]->request_vector("new"));
+        d_U_half_vecs[part] = dynamic_cast<PetscVector<double>*>(d_U_systems[part]->request_vector("half"));
 
-        d_F_systems[part] = &d_equation_systems[part]->get_system(FORCE_SYSTEM_NAME);
+        d_F_systems[part] = &d_equation_systems[part]->get_system<ExplicitSystem>(FORCE_SYSTEM_NAME);
         d_F_half_vecs[part] = dynamic_cast<PetscVector<double>*>(d_F_systems[part]->current_local_solution.get());
```
drwells: This seems to work but it doesn't match what we do for all of the other vectors, which is suspicious.

src/IB/IBFEMethod.cpp
```diff
@@ -767,62 +767,70 @@ IBFEMethod::preprocessIntegrateData(double current_time, double new_time, int nu
...
+        d_F_rhs_vecs[part] = dynamic_cast<PetscVector<double>*>(d_F_systems[part]->rhs);
+        d_F_tmp_vecs[part] = dynamic_cast<PetscVector<double>*>(d_F_systems[part]->request_vector("tmp"));
         d_F_IB_ghost_vecs[part] = dynamic_cast<PetscVector<double>*>(
             d_fe_data_managers[part]->buildGhostedSolutionVector(FORCE_SYSTEM_NAME, /*localize_data*/ false));
 
         if (d_lag_body_source_part[part])
         {
-            d_Q_systems[part] = &d_equation_systems[part]->get_system(SOURCE_SYSTEM_NAME);
+            d_Q_systems[part] = &d_equation_systems[part]->get_system<ExplicitSystem>(SOURCE_SYSTEM_NAME);
             d_Q_half_vecs[part] = dynamic_cast<PetscVector<double>*>(d_Q_systems[part]->current_local_solution.get());
```
drwells: Same thing: I guess this weirdness was already here.

boyceg:
I do not think that System has a RHS vector.

We could stick with System and add a RHS vector.

drwells:
Good point. I guess I never noticed that `System` is missing `rhs`.

boyceg:
F and Q are forcing terms that are not state variables and are only evaluated at the “half” time step, so we don’t have “new” in the current time stepping.

drwells:
I double checked the vector zeroing and it looks good to me.

I spent some time reading over how PETSc does parallel partitioning: AFAICT its not possible to actually copy the vector partitioner, which is what we spend a ton of time recomputing. Caching all of these vectors should help a lot (particularly once we improve the load balancing) since the communication pattern only changes when we regrid.

drwells:
Lets merge this now. I'll write a followup patch that tidies things up a bit later :)

PR #414, opened 2018-12-14
Add lib mesh partitioning visualization

drwells:
Another spinoff from #396: this PR adds two new functions to visualize the libMesh partitioning (both `Elem`-wise and `Node`-wise).

### DIFF
--- a/ibtk/include/ibtk/libmesh_utilities.h
+++ b/ibtk/include/ibtk/libmesh_utilities.h
@@ -873,6 +873,35 @@ get_nodal_dof_indices(const libMesh::DofMap &dof_map,
 #endif
 }
 
+/*!
+ * Save, in a plain text file, the libMesh partitioning, with the format
+ *
+ *     x,y,z,rank
+ *
+ * where x, y, and z are the center of an Elem and rank is the current MPI
+ * rank.
+ *
+ * @note this function collates the output from all MPI processors in the
+ * communicator assigned to @p position_system, so it is an inherently serial
+ * function.
+ */
+inline void
+write_elem_partitioning(const std::string& file_name, const libMesh::System& position_system);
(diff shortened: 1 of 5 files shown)


boyceg:
@drwells let's leave the header file name `libmesh_utilities.h`. The unwritten convention that I think that I've used is that headers for "collections of functions" are all lowercase with words separated by underscores. The same convention is used for static functions. "Camel case" is used for class names and associated headers. In any case, I know that we include `libmesh_utilities.h` in some application codes, and so this change breaks those codes.

boyceg:
(See, e.g., `ibamr_utilities.h` and `ibtk_utilities.h`.)

drwells:
I can remove the rename: I did this since we have, e.g., `IndexUtilities.h` in IBTK.

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -35,845 +35,9 @@
...
-            nodal_indices.push_back(node->dof_number(system_n, variable_n, component_n));
-        }
-    }
-#else
-    dof_map.dof_indices(node, nodal_indices, variable_n);
-#endif
-}
-
-} // namespace IBTK
+#warning "This header is deprecated: use ibtk/LibMeshUtilities.h instead."
```
drwells: Its not too hard to rename headers in a compatible way if we leave a warning, though.

boyceg:
That would be fine too.

drwells:
I think it would be worthwhile to just consistently name all utility headers to make this clearer in the future. How about we move the rest to `index_utilities.h` etc and add stubs for compatibility? We can remove the stubs after a few releases.

drwells:
Fixed!

PR #415, opened 2018-12-14
Rearrange load balancer

drwells:
I spent a large part of today reviewing how we (and SAMRAI) do load balancing. This little PR is the first step in utilizing SAMRAI's very nice infrastructure for doing nonuniform load balancing: it just moves some code around so that we can get a load balancer in INSHierarchyIntegrator.

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -813,7 +813,7 @@ private:
      * balancing.
      */
     SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > d_workload_var;
-    int d_workload_idx;
+    int d_workload_idx = IBTK::invalid_index;
 
     /*
      * The default kernel functions and quadrature rule used to mediate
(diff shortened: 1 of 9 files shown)


boyceg:
What is the benefit of each of these classes having a `worload_var` and `workload_idx`? (I ask this question fully realizing that I am the one that started this particular practice.) I wonder if this should just be owned by the parent `HierarchyIntegrator` and then passed around as needed to fill in the workload estimates.

drwells:
By these classes, do you mean the Hierarchy classes, or things like FEDataManager?

drwells:
Hm. The only things we do with the load balancer are to call `load_balancer->setWorkloadPatchDataIndex` and `load_balancer->setUniformWorkload`. Neither of those are really the concern of the HierarchyIntegrator but it means we will have to set it up somewhere else.

boyceg:
Yeah, one of these classes needs to manage allocation for the workload index data. I usually put that into the integrator, but agree that there is probably a better place.

We probably do not need to handle non uniform load balancing except for applications using IB, so it could possibly just go into IBStrategy, and no where else.

boyceg:
This is fine for now; I will go ahead and merge.

drwells:
I'm still thinking about how to do this: i.e., `FEDataManager` only needs the workload index.

boyceg:
> I'm still thinking about how to do this: i.e., FEDataManager only needs the workload index.
> 
LDataManager could use it too in an IBMethod-based simulation.

PR #416, opened 2018-12-18
Add more assertions to FEDataManager.

drwells:
The load balancing code (which is not currently used from this class, but is implemented and largely correct) does not, presently, reinitialize all arrays to be the correct length in this code. This PR just adds a bunch of array length and array index checks.


drwells:
Surprisingly this causes some failures. I am investigating.

drwells:
Fixed.

boyceg:
I prefer to rely on `-D_GLIBCXX_DEBUG` to ensure that these vector access are in bounds, although IIRC @drwells does _not_ like to use it...

drwells:
Thats a good point: I tend not to use it because I *think* its slow (and I have to recompile everything, though thats easy with spack) but I have never actually measured it. Lets hold off on merging this (since these are purely checks that help now while I am debugging) until I actually measure what it costs to compile everything with that flag.

PR #419, opened 2018-12-19
use VecGhostScatterBegin/VecGhostScatterEnd in IBFE interpolation

boyceg:


boyceg:
Hmmm, what boneheaded git thing did I do to get 4f5fde3 showing up in this PR?

boyceg:
I just reorganized the communication in `IBFEMethod::interpolateVelocity` to group together all of the calls to `VecGhostScatterBegin`, and all of the calls to `VecGhostScatterEnd`. I do not have time right now to test for correctness -- @drwells can you give this a try?

drwells:
Nice. I'll give it a whirl.

drwells:
This does seem to work, but I want to stare at it some more before approving it for merging.

boyceg:
Does it make any performance difference?

drwells:
I am working on that now: I am comparing it to the rest of the load balancing code.

boyceg:
I just added some helper functions for batching calls to `VecCopy` and to `VecGhostUpdateBegin`/`VecGhostUpdateEnd`, and switched to using `VecGhostUpdateBegin`/`VecGhostUpdateEnd` in `IBFEMethod::spreadForce()`.

boyceg:
Do you mean 6 procs or 16?

Note that this PR does a lot more index lookups to get the local indices for accumulating the ghost values — that could be slowing things down more than using the ghosted communication routines speeds things up. It should be fairly easy to cache this stuff if it is indeed the problem.

boyceg:
Should we include some simple diagnostics to report min/mean/max cells/elems?

You may also need to play around with SAMRAI’s grid generation algorithm to get more reasonable looking patches.

boyceg:
@knepley, there is an accumulation step that we could perform using the local form of a ghosted vector, or we can just use `VecSet`. (In `master` we are currently just using `VecSet` followed by `VecAssemblyXXX`. This PR switches to accumulating into the local form of the ghosted vectors followed by `VecGhostUpdateXXX`.) Do you have an opinion about whether we should be doing one vs. the other?

boyceg:
This is superseded by #430.

boyceg:
@drwells, this is the PR I was babbling about this afternoon. It uses the local form in `FEDataManager::interpWeighted()` to do the local part of the accumulation followed by a scatter to do parallel communication.

I will see if I can pull out the relevant stuff a bit later this evening.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1387,6 +1436,17 @@ FEDataManager::interpWeighted(const int f_data_idx,
...
+                    // TODO: This probably should be cached.
+                    if (ghosted_vec)
+                    {
+                        F_local_dof_indices[i].resize(F_dof_indices[i].size());
+                        std::transform(F_dof_indices[i].begin(),
+                                       F_dof_indices[i].end(),
+                                       F_local_dof_indices[i].begin(),
+                                       [&](unsigned int idx) -> unsigned int {
+                                           return F_petsc_vec->map_global_to_local_index(idx);
+                                       });
```
drwells: This is, AFAIK, exactly what we need to do right now (until we get #513 finished)

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1270,17 +1302,34 @@ FEDataManager::interpWeighted(const int f_data_idx,
...
+        double* F_local_soln;
+        if (ghosted_vec)
+        {
+            ierr = VecGhostGetLocalForm(F_global_vec, &F_local_vec);
+            IBTK_CHKERRQ(ierr);
+            ierr = VecZeroEntries(F_local_vec);
+            IBTK_CHKERRQ(ierr);
+            ierr = VecGetArray(F_local_vec, &F_local_soln);
+            IBTK_CHKERRQ(ierr);
+        }
```
drwells: I think that this and the indexing are really all we need.

boyceg: I agree. I can do this later tonight (I think!).

drwells: @boyceg can you do this or shall I take a look?

boyceg:
By all means --- I did not make it to this last night!

drwells:
This should be a much smaller change now that the vectors are ghosted. Let me give it a spin.

PR #421, opened 2018-12-20
Require libMesh 1.1.0 or newer.

drwells:
We use ReplicatedMesh now, which was new with libMesh 1.1.0.

### DIFF
--- a/configure
+++ b/configure
@@ -24075,19 +24075,17 @@ else
 fi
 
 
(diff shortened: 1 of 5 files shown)


drwells:
We can compile IBAMR with libMesh 1.0.0, but since we should be using ReplicatedMesh in IBFE we should just require a slightly newer version for IBAMR 1.0.0.

PR #423, opened 2019-01-08
Detect the same C++ version as libMesh.

### LINKED ISSUE #411: Improve C++11/C++14 detection
I just ran into an issue with selecting the correct version flag and libMesh: since newer versions of g++ default to `-std=gnu++14` it is possible, without user intervention, that libMesh gets compiled with C++14 support. The problem with this is we strictly enforce, at the moment, the flag `-std=c++11`.

We should try to detect if libMesh was compiled with C++14 support: if so, and if the compiler supports it, we should use `-std=c++14` too.

drwells:
Presently, new GNU compliers (e.g., g++ 6.0 and newer) default to C++14. This causes libMesh to be compiled with C++14 support. However, the present version macros force the usage of C++11 in IBAMR: this causes compilation errors with libMesh headers.

Get around this by always using the current (1.3.1) libMesh C++ version detection macro logic: this guarantees that we will always have C++11 available *and* match whatever libMesh finds.

This patch is a little weird in that we are formally downgrading the m4 macro we used to an older version, but it makes sense: if we do exactly what libMesh does we will never have a version flag conflict.

Fixes #411. I think. I need to reinstall a copy of GCC 4.7 to triple check.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -89,7 +89,7 @@ build_triplet = @build@
 host_triplet = @host@
 subdir = .
 ACLOCAL_M4 = $(top_srcdir)/aclocal.m4
-am__aclocal_m4_deps = $(top_srcdir)/m4/ax_cxx_compile_stdcxx.m4 \
+am__aclocal_m4_deps = $(top_srcdir)/m4/ax_cxx_compile_stdcxx_11.m4 \
 	$(top_srcdir)/m4/ax_prefix_config_h.m4 \
 	$(top_srcdir)/m4/ax_prog_cc_mpi.m4 \
 	$(top_srcdir)/m4/ax_prog_cxx_mpi.m4 $(top_srcdir)/m4/boost.m4 \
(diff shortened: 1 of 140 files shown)


boyceg:
Looks good. Any more testing before merging?

drwells:
I checked with GCC 4.4 and the macro correctly detects no C++11 support is available: since we cover all this cases I believe this is good to go.

roystgnr:
I stumbled upon this ticket while trying to get Google to find https://github.com/libMesh/libmesh/pull/2842 - and as long as I'm here, I guess I ought to warn you guys to take a look at that too?  We upgraded our C++ detection macro, and added some autoconf code around it, and so now it's entirely possible that you'll find that libMesh is turning on C++17 while you're using a macro that doesn't know that C++17 exists.

drwells:
Thanks for the heads up. I'll take a look!

PR #425, opened 2019-01-09
Sprinkle/rigid fhd

brennansprinkle:
added two new classes CIBStochasticMethod.cpp and CIBStochasticStokesSolver.cpp to account for thermal fluctuations in rigid body motion.


drwells:
There is a lot of autotools churn in this PR: in particular, it looks like you are using an older copy of autotools. Would it be possible for you to get autotools 1.16.1 installed and regenerate everything? Alternatively, would you mind if I rewrote (via force pushing) your commits to use the same autotools version we have on master?

boyceg:
At some point in the past couple of years, I tried to setup configure to get consistent autotools. I can't remember if this is still in the library or not.

brennansprinkle:
I'm using Ubuntu 18 and after some cursory googling it isn't obvious to me how to update my version of autotools (can't use aptitude and doing it from source looks hard but maybe it isn't). Is it easy to to the force push thing? If not, I can absolutely keep trying to update my systems autotools.

amneetb:
@brennansprinkle : You can configure, make and make install autools (automake and autoconfigure) in your home directory and add the binary in front of your path in .bashrc. This is what I do on clusters, which use older version of autotools.

amneetb:
Are you using git-clang format or you just clang-formatted all the files? You will also need to update clang-format and git clang-format. Again these can be configured and installed in your home directory.

I am not sure what version of clang-format is used by other developers but on my system I have 
`$ clang-format --version
clang-format version 8.0.0 (tags/google/stable/2018-10-04) `

amneetb:
If these are only commits you have, then perhaps @drwells can fix them using his version of autotools and clang-format.

If you have not, can you add a stochastic rigid body example in examples/CIB directory. @drwells can fix formatting and autotools after everything is done from your side.

amneetb:
Yes, adding a well tested example (possibly from paper) in much more important than gtest mumbo-jumbo.

brennansprinkle:
Sorry for the delay, I wanted to make sure a few things were consistent in the example I'm using. I just pushed a commit with an added ex5 in the /examples/CIB directory

drwells:
I think you forgot a file: Jenkins errors out with the message
```
config.status: error: cannot find input file: `examples/CIB/ex5/Makefile.in'
```

brennansprinkle:
I added one but I'm not sure how to tell if it worked?

abarret:
I think you may have forgotten to add the whole ex5 directory: all the necessary .cpp files, the input files, Makefile.am, and any other files you might use with the example.

brennansprinkle:
ok, all the files are there this time at least

boyceg:
@brennansprinkle, sorry this has taken a while to get back to.

There are a large number of changes that are related to using a different version of Automake (1.16 instead of 1.16.1), which is resulting in merge conflicts as well as lots of diffs. Tar files for the current versions of autotools that we are using are actually available in `IBAMR/ibtk/contrib/autotools`.

If you are having trouble setting those up, an alternative could be to make a new branch that just commits the changed code and input files without the autotools stuff, and then one of us can setup autotools for you.

drwells:
We have gotten stuck for too long on this autotools issue that is easy for us to deal with and less easy for novices. I will put up a new branch.

drwells:
Closed in favor of #523: same relevant changes but no autotools issues.

PR #426, opened 2019-01-09
Updated packaged automake files.

abarret:
This commit updates the packaged automake files to 1.16.1 which is the current version used by the library. This should help reduce the autotool churn produced by i.e. #425.


boyceg:
Do these packaged autotools actually get used properly?

boyceg:
While we are at it, it might be worth checking the logic that attempts to decide when to use the packaged autotools.

abarret:
Yes. I actually use these because there's no libtool on the coeur machines and I'm too lazy to update my path. 

The only issue with the logic that I can see is that if one tool is missing it rebuilds _all_ the tools. But the actual building process only takes a couple of minutes so it's not a big deal when added onto to rest of the process.

PR #418, opened 2018-12-18
amneetb/numerical wave tank

amneetb:
New tools to do:

1) Distance calculation from FE surface mesh.
2) Transport of level set in Lagrangian manner.
3) Partitioned FSI for rigid body dynamics. 
4) Wave generation and wave damping capabilities.

PR #428, opened 2019-01-14
Fix alignment of Eigen objects.

drwells:
Eigen will automatically use AVX512 and other vectorized instruction sets if they are available. This causes segmentation faults (on AMD64, at least) when objects are not aligned to 128-bit (or better) addresses. This commit gets around this by specifying alignment requirements (or, for std::vector [1], a custom allocator) for objects with Eigen members.

[1] http://eigen.tuxfamily.org/dox-devel/group__TopicStlContainers.html

### DIFF
--- a/ibtk/include/ibtk/ibtk_utilities.h
+++ b/ibtk/include/ibtk/ibtk_utilities.h
@@ -39,6 +39,7 @@
 #include <algorithm>
 
 #include "Eigen/Core" // IWYU pragma: export
+#include "Eigen/StdVector"
 #include "tbox/PIO.h"
 #include "tbox/Utilities.h"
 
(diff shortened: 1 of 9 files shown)


drwells:
This is only evident if one compiles with support for the available instruction set (e.g., `-march=native` in GCC).

amneetb:
These are some horrible looking constructors ;-). Can we mask them with some typedefs?

drwells:
Hm. We could do something like

```cpp
template <typename T>
using EigenAlignedVector = std::vector<T, Eigen::aligned_allocator<T>>;
```
which would save some typing. I would prefer to keep writing out the (admittedly horrible) name so that its clear we still have a `std::vector` (with a custom allocator). @boyceg do you have a style preference?

drwells:
We should also check for the IBTK::Point and friends typedefs of the Eigen objects before fixing this.

drwells:
@amneetb do you like the typedef?

amneetb:
Yes. Can you also make the changes to amneetb/numerical_wave_tank branch where I have additional Eigen::Vector3d and Eigen::Quaterniond

drwells:
I'll take a look once this is merged.

drwells:
I'll merge this later today unless there are further comments since this is a minor fix.

drwells:
@amneetb I can apply the same changes to the wave tank branch (a.k.a. #418). Do you mind if I rebase and force push so that it includes these changes?

amneetb:
Yes, please. If you want to merge master into NWT please go ahead with that as well. We will be committing quite a few examples to this branch in the coming days and would want to have the latest updates from master. I can run one or two examples today to make sure that nothing is broken with merging master.

PR #430, opened 2019-01-16
eliminating some calls to NumericVector<T>::close()

boyceg:

PR #433, opened 2019-01-23
eliminate some calls to NumericVector<T>::close()

boyceg:
This PR supersedes #430, which had a couple of bugs that changed the computed solution in parallel.

### DIFF
--- a/ibtk/include/ibtk/libmesh_utilities.h
+++ b/ibtk/include/ibtk/libmesh_utilities.h
@@ -169,6 +169,118 @@ copy_and_synch(libMesh::NumericVector<double>& v_in,
     if (close_v_out) v_out.close();
 }
 
+inline void
+batch_vec_copy(const std::vector<libMesh::PetscVector<double>*>& x_vecs,
+               const std::vector<libMesh::PetscVector<double>*>& y_vecs)
+{
+#if defined(NDEBUG)
+    TBOX_ASSERT(x_vecs.size() == y_vecs.size());
+#endif
+    for (unsigned int k = 0; k < x_vecs.size(); ++k)
+    {
+        if (!x_vecs[k] || !y_vecs[k]) continue;
+        int ierr = VecCopy(x_vecs[k]->vec(), y_vecs[k]->vec());
+        IBTK_CHKERRQ(ierr);
+    }
+}
+
+inline void
+batch_vec_copy(const std::vector<std::vector<libMesh::PetscVector<double>*> >& x_vecs,
+               const std::vector<std::vector<libMesh::PetscVector<double>*> >& y_vecs)
+{
+#if defined(NDEBUG)
+    TBOX_ASSERT(x_vecs.size() == y_vecs.size());
+#endif
+    for (unsigned int n = 0; n < x_vecs.size(); ++n)
+    {
+        batch_vec_copy(x_vecs[n], y_vecs[n]);
+    }
+}
+
+inline void
+batch_vec_assembly(const std::vector<libMesh::PetscVector<double>*>& vecs)
+{
+    for (const auto& v : vecs)
+    {
+        if (!v) continue;
+        int ierr = VecAssemblyBegin(v->vec());
+        IBTK_CHKERRQ(ierr);
+    }
+    for (const auto& v : vecs)
+    {
+        if (!v) continue;
+        int ierr = VecAssemblyEnd(v->vec());
+        IBTK_CHKERRQ(ierr);
+    }
+}
+
+inline void
+batch_vec_assembly(const std::vector<std::vector<libMesh::PetscVector<double>*> >& vecs)
+{
+    for (unsigned int n = 0; n < vecs.size(); ++n)
+    {
+        for (const auto& v : vecs[n])
+        {
+            if (!v) continue;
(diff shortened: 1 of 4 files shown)


boyceg:
@drwells can we merge this?

drwells:
Probably: let me read it through once more then I will merge.

boyceg:
I am sure there is a better way to set up the helper function for arrays of arrays of `Vec`s.

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -169,6 +169,126 @@ copy_and_synch(libMesh::NumericVector<double>& v_in,
...
+#if defined(NDEBUG)
+        TBOX_ASSERT(x_vecs[n].size() == y_vecs[n].size());
+#endif
+        for (unsigned int k = 0; k < x_vecs[n].size(); ++k)
+        {
+            if (!x_vecs[n][k] || !y_vecs[n][k]) continue;
+            int ierr = VecCopy(x_vecs[n][k]->vec(), y_vecs[n][k]->vec());
+            IBTK_CHKERRQ(ierr);
+        }
+    }
```
drwells: We could make this simpler by just calling the previous function.

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -169,6 +169,126 @@ copy_and_synch(libMesh::NumericVector<double>& v_in,
...
+    for (unsigned int n = 0; n < vecs.size(); ++n)
+    {
+        for (const auto& v : vecs[n])
+        {
+            if (!v) continue;
+            int ierr = VecAssemblyEnd(v->vec());
+            IBTK_CHKERRQ(ierr);
+        }
+    }
```
drwells: same

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -169,6 +169,126 @@ copy_and_synch(libMesh::NumericVector<double>& v_in,
...
+    for (unsigned int n = 0; n < vecs.size(); ++n)
+    {
+        for (const auto& v : vecs[n])
+        {
+            if (!v) continue;
+            int ierr = VecGhostUpdateEnd(v->vec(), insert_mode, scatter_mode);
+            IBTK_CHKERRQ(ierr);
+        }
+    }
```
drwells: same

src/IB/IBFEMethod.cpp
```diff
@@ -987,28 +987,19 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
...
-        int ierr = VecAssemblyEnd(d_U_rhs_vecs[part]->vec());
-        IBTK_CHKERRQ(ierr);
-    }
+    batch_vec_assembly(d_U_rhs_vecs);
 
     // Solve for the interpolated data.
     for (unsigned int part = 0; part < d_num_parts; ++part)
     {
-        d_fe_data_managers[part]->computeL2Projection(*U_vecs[part],
+        *d_U_systems[part]->solution = *U_vecs[part]; // TODO: Commenting out this line changes the solution slightly.
```
drwells: Is this still true or is it a leftover from the previous PR?

src/IB/IBFEMethod.cpp
```diff
@@ -1117,14 +1101,29 @@ void
...
-        computeInteriorForceDensity(*d_F_half_vecs[part], *d_X_half_vecs[part], d_Phi_half_vecs[part], data_time, part);
+        d_F_rhs_vecs[part]->zero();
+        assembleInteriorForceDensityRHS(
+            *d_F_rhs_vecs[part], *d_X_half_vecs[part], d_Phi_half_vecs[part], data_time, part);
+    }
+    batch_vec_assembly(d_F_rhs_vecs);
+    for (unsigned part = 0; part < d_num_parts; ++part)
+    {
+        d_F_systems[part]->solution->zero(); // TODO: Commenting out this line changes the solution slightly but
+                                             // probably speeds up this solve slightly too.
```
drwells: If there is any different more than something tiny (as a result of a different initial guess with a Krylov method) then we should open an issue. Otherwise I wouldn't bother with this comment.

ibtk/include/ibtk/libmesh_utilities.h
boyceg: Will do.

ibtk/include/ibtk/libmesh_utilities.h
boyceg: I think we want to do `VecAssemblyBegin` for all `Vec`s before starting `VecAssemblyEnd`.

ibtk/include/ibtk/libmesh_utilities.h
boyceg: Same as assembly.

boyceg:
@drwells, if you want to change those initial guesses, that is OK with me --- feel free to do so and then merge.

For the helper functions that involve communication, I think that we want to do all of the "begins" before starting any of the "ends".

drwells:
My suggestion with combining the assembly functions as not well thought out :( Lets not do that!

I ran a bunch of tests for this on Friday and everything worked (up to solver tolerance). I think PETSc will abort the program if we really screw this up so I think we are okay.

Issue #436, opened 2019-01-26
Avoid calling FE::reinit when possible.


drwells:
The heart solver (with the afterload 10 benchmark) spends about 20% of its time calling `libMesh::FE::reinit`: unfortunately, specifying exactly what needs to be recomputed to libMesh is a bit tricky so we end up doing a lot of work there. In other benchmarks that have thicker structures (e.g., IBFE/explicit/ex4) we spend nearly half of our time calling this function. I'll have to do some more profiling to see where we are with the full heart model.

There is a bunch of stuff we can do to fix this:
- [x] A lot of the calls to `reinit` come from changing quadrature rules. We should just cache multiple `FE` objects (one per quadrature rule) to get around this. This is mostly done (see #437). This speeds up the first part of `interpWeighted` where we only need the function values.
- [x] We can get the `JxW` values by using `FEMap`. We should implement a similar cache strategy here: `FEMap` only depends on details of the finite element space (FE order, FE family) and the quadrature rule, so we can avoid recomputing everything but the Jacobians if we have cache here that uses the same strategy.
- [x] There are a few places where we build a new `Quadrature` object, but we only need the quadrature rule and number of points. We should clean that up too.
- [ ] Add caches to `IBTK::FEDataInterpolation`.

boyceg:
`FEMap` _does_ use the element geometry for evaluating any quantities in "physical" coordinates. It does not depend on only the FE order / FE family. In particular, for `JxW` and derivatives of shape functions, it needs to know the mapping from the reference element to the physical configuration of the element.

drwells:
@boyceg I think you're right but we need to be careful: `FEMap::init_reference_to_physical_map` does something with the FE order and type, and we do call it: however, I am pretty sure the data it computes is not read again.

drwells:
FEDataInterpolation takes up only a small amount of the total run time (it totals at 5%, while something like `interpWeighted` is closer to 50%) and is not a massive source of bottlenecks like `spread` is, so I think we should close this for now: the 'high priority' part of easy to implement performance improvements is done. I'll open a separate issue for it like the `JxW` calculations.

We lowered the total number of cycles executed from 465 billion to 339 billion in the default settings for `IBFE/explicit/ex4/main3d` so I am declaring victory.

drwells:
Quick update, for anyone following this thread: @cpuelz just reported that the heart model is running about 2x faster now that these are in :tada: :tada: :tada:

[...]

PR #437, opened 2019-01-26
Add IBTK::FECache.

drwells:
First part of #436: this lowers run time for my favorite example (IBFE/explicit/ex4) by about 10%. I suspect it will help more with more highly deformed meshes.

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -79,6 +79,7 @@
 #include "VariableContext.h"
 #include "VariableDatabase.h"
 #include "boost/multi_array.hpp"
+#include "ibtk/FECache.h"
 #include "ibtk/FEDataManager.h"
 #include "ibtk/IBTK_CHKERRQ.h"
 #include "ibtk/IndexUtilities.h"
@@ -1154,7 +1155,7 @@ FEDataManager::interpWeighted(const int f_data_idx,
     const unsigned int dim = mesh.mesh_dimension();
     std::unique_ptr<QBase> qrule;
 
-    // Extract the FE systems and DOF maps, and setup the FE object.
+    // Extract the FE systems and DOF maps, and setup the FECache objects.
     System& F_system = d_es->get_system(system_name);
     const unsigned int n_vars = F_system.n_vars();
     const DofMap& F_dof_map = F_system.get_dof_map();
@@ -1178,16 +1179,9 @@ FEDataManager::interpWeighted(const int f_data_idx,
         TBOX_ASSERT(X_dof_map.variable_type(d) == X_fe_type);
         TBOX_ASSERT(X_dof_map.variable_order(d) == X_order);
     }
-    std::unique_ptr<FEBase> F_fe_autoptr(FEBase::build(dim, F_fe_type)), X_fe_autoptr;
-    if (F_fe_type != X_fe_type)
-    {
-        X_fe_autoptr = std::unique_ptr<FEBase>(FEBase::build(dim, X_fe_type));
-    }
-    FEBase* F_fe = F_fe_autoptr.get();
-    FEBase* X_fe = X_fe_autoptr.get() ? X_fe_autoptr.get() : F_fe_autoptr.get();
-    const std::vector<double>& JxW_F = F_fe->get_JxW();
-    const std::vector<std::vector<double> >& phi_F = F_fe->get_phi();
-    const std::vector<std::vector<double> >& phi_X = X_fe->get_phi();
+
+    FECache F_fe_cache(dim, F_fe_type);
+    FECache X_fe_cache(dim, X_fe_type);
 
     // Communicate any unsynchronized ghost data.
     for (const auto& f_refine_sched : f_refine_scheds)
@@ -1200,6 +1194,24 @@ FEDataManager::interpWeighted(const int f_data_idx,
     // Check to see if we are using nodal quadrature.
     const bool use_nodal_quadrature =
         interp_spec.use_nodal_quadrature && (F_fe_type == X_fe_type && F_order == X_order);
+
+    // We only use the FECache objects if we do *not* use nodal quadrature so
+    // only perform sanity checks in that case:
+    if (!use_nodal_quadrature)
+    {
+        // This will break, at some point in the future, if we ever use
+        // nonnodal-interpolating finite elements. TODO: more finite elements
+        // will probably work.
+        std::vector<FEFamily> fe_family_whitelist {LAGRANGE, L2_LAGRANGE};
+        TBOX_ASSERT(std::find(fe_family_whitelist.begin(),
+                              fe_family_whitelist.end(),
+                              F_fe_type.family)
+                    != fe_family_whitelist.end());
+        TBOX_ASSERT(std::find(fe_family_whitelist.begin(),
+                              fe_family_whitelist.end(),
+                              X_fe_type.family)
+                    != fe_family_whitelist.end());
+    }
 
     if (use_nodal_quadrature)
     {
@@ -1360,6 +1372,7 @@ FEDataManager::interpWeighted(const int f_data_idx,
             // quadrature points.
             qrule.reset();
             unsigned int qp_offset = 0;
+            std::set<std::pair<libMesh::QuadratureType, libMesh::Order> > used_X_quadratures;
             for (unsigned int e_idx = 0; e_idx < num_active_patch_elems; ++e_idx)
             {
                 Elem* const elem = patch_elems[e_idx];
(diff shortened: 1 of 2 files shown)


ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1399,21 +1422,34 @@ FEDataManager::interpWeighted(const int f_data_idx,
...
+                    F_fe.get_phi();
+                    F_fe.get_JxW();
+                    used_F_quadratures.insert(key);
                 }
-                F_fe->reinit(elem);
+
+                // Unlike above, we always reinit here since we need JxW
+                F_fe.reinit(elem);
+                const std::vector<std::vector<double> >& phi_F = F_fe.get_phi();
+                const std::vector<double> &JxW_F = F_fe.get_JxW();
```
drwells: I'm not convinced that this is right: we don't really use this mapping, since we also have to take into account the deformation caused by `X`, which should alter the Jacobian (i.e., the contravariant transformation depends on `X`, right?)

boyceg:
JxW depends on the geometry of the element. In IBFE, the mesh geometry defines the Lagrangian reference coordinates. So, JxW is not the same for every element, even if the element type is the same.

X shows up to determine where the quadrature points are in the current configuration.

In the IBFE paper, the libMesh variable X is the deformation mapping chi. I probably should update the code sometime to use consistent notation.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: This routine is a little bit confusing, but what is happening is that we are setting the values for a RHS vector for an L2 projection, but we are not actually solving for the L2 projection. (That is what I mean when I say that it is a "weighted" interpolation.)

The projection is in reference coordinates. We would need to worry about `X` if we did this in current coordinates. Does this make sense?

boyceg:
`FECache` needs to work for cases where `dim != NDIM` and with FE families different than `LAGRANGE`. How much work would it be to do that?

boyceg:
Oops I hit the wrong button.

drwells:
Since we circumvent the `reinit` function this strategy won't work if we use, e.g., RT elements (or anything thats not just a nodal interpolant) since the function values then depend on the cell geometry. The best answer here would be to whitelist `LAGRANGE` elements and call `reinit` in case we have something else.

I think it would be best to de-template the function and just use `FEBase` to avoid the dimension problem (and also support more FEs).

ibtk/src/lagrangian/FEDataManager.cpp
drwells: >  what is happening is that we are setting the values for a RHS vector for an L2 projection, but we are not actually solving for the L2 projection.

I was confused by that and did a lot of googling for 'weighted interpolations' before I figured out what was going on :) I'd vote to rename this to `setupLoadVector` or something like that (since the result is a dual space vector).

I think I am still confused about this: I probably just need to think harder to sort this out. We compute the quadrature points by reading `X` values, so the `J` in `JxW` needs to be computed w.r.t. that same coordinate transformation. Am I misunderstanding where we use the `X` data? It looks like we use it to compute the quadrature point locations in `get_values_for_interpolation`.

boyceg: We are sampling a field using the current coordinates of the quadrature points. `X` is the motion map from reference coordinates to current coordinates. The integration happens in reference coordinates, so `JxW` depends on the fixed shape of the element in the reference configuration, but is independent of the deformation from the reference configuration to the current configuration.

I think of this as happening in two stages. Using `s` as the mesh/reference coordinates and `X(s,t)` as the current (deformed) position of material point `s` at time `t`, first we evaluate `F_IB(s,t) = \sum_{i,j,k} f_{i,j,k} \delta_h(x_{i,j,k} - X(s,t)) h^3`. This is "the usual IB interpolation scheme".

We cannot use `F_IB` directly, however, because it is typically not directly representable in terms of the FE basis functions. So we define its projection `F` via `(F,phi_l)_2 = (F_IB,phi_l)_2 = \int(F_IB(s,t) phi_l(s) ds` for all FE basis functions `phi_l`. This is where we use `JxW` in this routine.

boyceg:
This looks good.

Before merging, should we go ahead and:
1. Avoid reinit'ing the `FE`/`FECache` object just to get quadrature rule point counts.
2. Think about breaking `F_fe` into two different objects, one used for init'ing shape functions and the other for init'ing `JxW`. That way we only recompute `JxW` on each element, and shape functions when the element type/quadrature rule changes.

I am surprised that profiling does not suggest that the same changes should be made in `FEDataManager::spread()`. Are you planning on making the same kinds of changes there?

drwells:
1. I don't understand what you are asking. Do you mean the calls to `updateInterpQuadratureRule`?
2. Unfortunately there is no way to get *only* the `JxW` values due to this bit of logic in `fe_base.C`:
```cpp
template <typename OutputType>
void FEGenericBase<OutputType>::determine_calculations()
{
  this->calculations_started = true;

  // If the user forgot to request anything, we'll be safe and
  // calculate everything:
  if (!this->calculate_phi && !this->calculate_dphi && !this->calculate_curl_phi && !this->calculate_div_phi)
    {
      this->calculate_phi = this->calculate_dphi = this->calculate_dphiref = true;
      if (FEInterface::field_type(fe_type.family) == TYPE_VECTOR)
        {
          this->calculate_curl_phi = true;
          this->calculate_div_phi  = true;
        }
    }
  // ...
}
```
Furthermore, unless the element is just a translation of the previous element we saw, we always recompute the shape values (again from `fe.C`):
```cpp
                for (unsigned int n = 1; n < elem->n_nodes(); ++n)
                  {
                    if (!(elem->point(n) - elem->point(0)).relative_fuzzy_equals
                        ((cached_nodes[n] - cached_nodes[0]), 1e-13))
                      {
                        cached_nodes_still_fit = false;
                        break;
                      }
                  }
// ...
  // Compute the shape functions and the derivatives at all of the
  // quadrature points.
  if (!cached_nodes_still_fit)
    {
      if (pts != nullptr)
        this->compute_shape_functions (elem,*pts);
      else
        this->compute_shape_functions(elem,this->qrule->get_points());
    }
```
I think the libMesh guys picked the safe option here since (as we previously discussed) the shape values can depend on the element geometry. Its unfortunate that they don't have some optimization in place to avoid recomputing shape values but this can get pretty [complicated when done in full generality](https://dealii.org/9.0.0/doxygen/deal.II/group__UpdateFlags.html).

Finally, regarding `spread`: since we need to torture libMesh into recomputing the Jacobians but not the shape values we will need to implement a second class `FEMapCache` to store `FEMap`s instead of relying on the ones owned by the `FE`s: AFAICT its just not possible to get libMesh to avoid recomputing shape values when the element geometry changes.

drwells:
Hence: I would like to tackle the Jacobians in a followup patch since we will need another cache type (`FEMapCache`) to get the results we need.

boyceg:
OK, got it with item 2.

For 1, IIRC, there are some places in `interpWeighted()` where we `reinit` an `FE` or `FECache` just to query the number of quadrature points for the purpose of preallocation. Can we avoid those calls to `reinit`?

(However, I may be getting this mixed up with what happens in `spread()`.)

I think that `spread()` also will be recomputing a lot of shape function values and should benefit from caching in the same way that is done in `interpWeighted()`.

drwells:
We need `JxW` in the only loop in `spread` where we already call `reinit` so I don't think that we can fix that until we separate the Jacobian and shape calculations.

drwells:
I don't see any such `reinit` calls in `interpWeighted`.

I didn't look too closely at it before, but `updateQuadratureRule` is also relatively expensive: we spend about 5% of our *total* time calling this function when all, once the caches are fully configured, we actually need is the number of quadrature points. Would you mind if I left this until after we do the Jacobian caching? I would prefer to fix all occurrences of that function at once instead of some now and some later.

boyceg:
It is fine to break up into multiple PRs.

Are you ready to merge this thing?

boyceg:
In `spread()` you could conceivably use `FECache` for the `X_fe` object.

The most typical use case is for `X_fe` and `F_fe` to be the same, however, and so this will not make any difference in current usage.

drwells:
I would prefer to merge now and start working on caching `JxW` values: that will make the work in `spread` clearer too.

boyceg:
If you haven't already, can you run ex5 with an input file that is set up to use an interfacial (surface) mesh instead of a volume mesh, to test the case where `dim != NDIM`?

If that works, please go ahead and merge.

drwells:
This actually causes some tests to fail in weird ways. I am investigating now.

drwells:
We should categorize quadrature points also by the element type in case we have composite meshes with multiple element types. I'll tackle this in a new PR.

PR #440, opened 2019-01-29
fix F_tmp_vecs setup

### LINKED ISSUE #439: Vector caching makes some examples fail
IBFE/explicit/ex10 and IBFE/explicit/ex11 now crash due to incorrectly set up `libMesh::PETScVector`s. I used bisect and found that this happened after d94e1e0dcd5a106.

This is one of the problems with #372: we don't print out failing tests: the failures are hidden in make messages in a giant log instead of being easy to find.

boyceg:
fixes #439.


boyceg:
Whoops. Copy-and-paste error when setting up vector caching. This works for me on ex10 and ex11, in serial and parallel.

To protect everyone from me, we should try to run more extensive Jenkins tests on the new build server.

PR #442, opened 2019-01-29
Add fe map cache

drwells:
This passes the test suite. I am doing some profiling now.

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -80,6 +80,7 @@
 #include "VariableDatabase.h"
 #include "boost/multi_array.hpp"
 #include "ibtk/FECache.h"
+#include "ibtk/FEMapCache.h"
 #include "ibtk/FEDataManager.h"
 #include "ibtk/IBTK_CHKERRQ.h"
 #include "ibtk/IndexUtilities.h"
@@ -1184,6 +1185,7 @@ FEDataManager::interpWeighted(const int f_data_idx,
     using quad_key_type = std::tuple<libMesh::ElemType, libMesh::QuadratureType, libMesh::Order>;
     FECache F_fe_cache(dim, F_fe_type);
     FECache X_fe_cache(dim, X_fe_type);
+    FEMapCache fe_map_cache(dim);
 
     // Communicate any unsynchronized ghost data.
     for (const auto& f_refine_sched : f_refine_scheds)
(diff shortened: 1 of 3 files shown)


boyceg:
Looks great!

Interested to see if it helps with performance. :-)

drwells:
This lowers the total cycle count (for IBFE/explicit/ex4 in 3D) from 423 billion to 391 billion: about 7.5%.

It also lowers the total number of calls to `reinit` from about 2.5 million to 1.8 million. There is still more to do but between this and #437 we have lowered the total number of cycles from 465 billion to 391 billion: thats pretty good considering how easy it was :)

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1180,8 +1181,11 @@ FEDataManager::interpWeighted(const int f_data_idx,
         TBOX_ASSERT(X_dof_map.variable_order(d) == X_order);
     }
 
+    // convenience alias for the quadrature key type used by FECache and FEMapCache
+    using quad_key_type = std::tuple<libMesh::ElemType, libMesh::QuadratureType, libMesh::Order>;
     FECache F_fe_cache(dim, F_fe_type);
     FECache X_fe_cache(dim, X_fe_type);
+    FEMapCache F_fe_map_cache(dim);
```
boyceg: There is no need to identify the cached `FEMap` as associated with the `F` variable --- it probably is clearer to call it `fe_map_cache`, because it is independent of the FE space used for `F`.

boyceg:
BTW, I was doing some profiling last night, and even without the improvements in this branch, `interpWeighted` and `spread` are taking about the same amount of time. I think it will pay off to make these changes in `spread` too.

drwells:
I am working on spread now.

drwells:
That was straightforward: its the same changes as `interpWeighted` (no surprise there).

drwells:
This also passes the test suite.

boyceg:
Looks great. Merge whenever you are ready.

What is the cycle count down to? :-)

drwells:
367 billion with the changes to spread. Thats a 21% reduction in cycles compared to when we started!

drwells:
libMesh's implementation of the Jacobian calculations is not great (as usual, its *general* and does the right thing for isoparametric elements, but we don't really care): its the fourth most expensive thing now and it really shouldn't be. Here is the callgrind map:

callgrind-femap

boyceg:
BTW, note that we call `interpWeighted` twice per time step but only call `spread` once; this is why `lagrangian_ib_4_interp3d` has ~2x the count of `lagrangian_ib_4_spread3d`.

boyceg:
Also, I am surprised that `VecGetOwnershipRange` and `PetscCheckPointer` are so prominent here.

boyceg:
Should we roll our own Jacobian code? Looks like that is close to 25% of cycles?

boyceg:
`VecGetOwnershipRange` is called by `PetscVector::map_global_to_local_index`. We call `PetscVector::map_global_to_local_index` a _lot_. Another place where we could cache...?

boyceg:
Can you flip the callstack to see if there is a relatively small number of places in IBAMR that result in all of these calls to `VecGetOwnershipRange` and `PetscCheckPointer`?

drwells:
> Should we roll our own Jacobian code? Looks like that is close to 25% of cycles?

This is very tempting. I'll open an issue.

drwells:
The calls to `VecGetOwnershipRange` are mostly in `IBTK::get_values_for_interpolation` and `IBTK::FEDataInterpolation::collectDataForInterpolation`: see #445.

`VecGetOwnershipRange` is the main function calling `PetscCheckPointer` (every PETSc function calls it, but we call `VecGetOwnershipRange` *a lot*).

drwells:
Since this passes the tests and we seem happy with the implementation I will merge.

PR #443, opened 2019-01-30
cache quadrature rules used by FEDataManager::updateQuadratureRule

boyceg:


boyceg:
This makes a modest difference in `FEDataManager::updateQuadratureRule` by caching the quadrature rules. The implementation is not ready for prime time --- it uses a `static` data structure defined in the implementation of `updateQuadratureRule` to keep track of the rules.

@drwells, if this looks like the right approach, would you be able to clean up a bit? I probably am not going to have time to work on this again today.

The other function that seems to take a _lot_ of time in `updateQuadratureRule` is the call to `elem->is_node_on_edge()`. Perhaps these data could also be cached?

boyceg:
@drwells, if you are so inclined, please feel free to rewrite my history on this. :-D

drwells:
@boyceg Uh oh; I planned on completely removing all calls that create a `Quadrature` object and let the caches handle calling `Quadrature::build`. At least in `interpWeighted` we only need the number of quadrature points and never the quadrature object itself: we can get away with just returning `std::tuple<ElemType, QuadratureType, Order>` and the number of points everywhere (we would add a utility function that takes a key and computes the number of quadrature points in the associated rule).

There may be places where we need to actually read the quadrature object instead of just putting it in `FE` or `FEMap` and never thinking about it again (I am definitely biased towards the later since its the deal.II way :laughing:); if so we will need a caching strategy like this.

I'll take a closer look once #442 is done.

boyceg:
Can you easily get the number of points without actually building the quadrature rule?

drwells:
I think so, though we may have to cheat and just hard-code details of the quadrature rules libMesh uses (e.g., Dunavant rules). It might be better to cache these things.

drwells:
There is also some really awkward logic (as a result of our libMesh hijacking) where we need to initialize `FEMap` objects with a quadrature rule and then call them again with the vector of quadrature weights. Weird.

boyceg:
Maybe there should be a quadrature rule oracle that we can ask for quadrature rules as needed.

drwells:
That is more-or-less what I have in mind, just not as a global object (it would be a member variable of `FEDataManager`.

drwells:
There are already some conflicts but there is a lot of stuff in here worth doing. I'll start work on some spinoffs.

boyceg:
I think that this will be superseded by #448, so I will close it out and wipe out the branch.

If we need to resuscitate, the last commit was 76318525029e43f702ea2960025d18679c56bf52.

PR #447, opened 2019-01-30
Unroll some loops to avoid zeroing data.

drwells:
According to callgrind this saves us about 4% of the total cycles used in lagrangian_ib_4_interp3d, and about 0.6% of overall runtime.

This doesn't save us as much as the other optimizations but it was easy to implement.

### DIFF
--- a/ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
+++ b/ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
@@ -280,8 +280,17 @@ c
 c     Interpolate u onto V.
 c
          do d = 0,depth-1
-            V(d,s) = 0.d0
-            do ic1 = ic_trimmed_lower(1),ic_trimmed_upper(1)
+c
+c     Avoid explicitly zeroing the data by doing one outer iteration:
+c
+            do ic1 = ic_trimmed_lower(1),ic_trimmed_lower(1)
+               do ic0 = ic_trimmed_lower(0),ic_trimmed_upper(0)
+                  V(d,s) = w(0,ic0-ic_lower(0))
+     &                 *w(1,ic1-ic_lower(1))
+     &                 *u(ic0,ic1,d)
+               enddo
+            enddo
+            do ic1 = ic_trimmed_lower(1)+1,ic_trimmed_upper(1)
                do ic0 = ic_trimmed_lower(0),ic_trimmed_upper(0)
                   V(d,s) = V(d,s)
      &                 +w(0,ic0-ic_lower(0))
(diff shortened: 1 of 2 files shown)


drwells:
This shows up in a lot of places!

boyceg:
Yeah, that pattern shows up a lot.

The "right" thing to do here would be to generate these subroutines from a template (e.g. using `m4`).

drwells:
I believe it. Most of the FORTRAN in Overture is automatically generated by a macro language Bill Henshaw built out of perl that solves this problem (I don't remember if he bothered with a basic optimization like this, but it does handle similar stuff).

drwells:
At the same time: I have no idea how to handle this in m4 without going crazy since this doesn't apply to all loops, some have different numbers of loops, etc.

boyceg:
I have no particular fondness for m4. I started using it because SAMRAI provides very convenient m4 macros to handle the sizing of different SAMRAI data types.

boyceg:
BTW, I am sorry to tell you this, but I also do the same thing in the Cartesian grid patch operations. E.g., in `flux.f.m4`,
```
               g0(i0,i1,i2) = 0.d0
               do d = 0,NDIM - 1
                  g0(i0,i1,i2) = alpha0(i0,i1,i2,d)*dU_dx(d)
     &                 + g0(i0,i1,i2)
               enddo
```
In this case, NDIM is actually replaced by the correct dimension by m4. I would hope that an optimizing compiler can figure out what to do, but I also have to admit that I have not checked.

drwells:
I'll rewrite those if you want me to, but I don't think its worth our time right now: we spend a lot less time in those kernels so improving them by a few percent is relatively small compared to improving the libMesh caching.

drwells:
Uh oh: this causes some failures. I am investigating.

boyceg:
Is the unrolling correct? I think that you want to visit each `V(d,s)` only once in the first pass. I think the unrolled loop will actually hit `V(d,s)` `n_stencil**(NDIM-1)` times?

drwells:
I think so; its right inside the loop over `d`. I'll keep staring at it.

This only causes 2 2D tests to fail so I think I just have a typo somewhere.

boyceg:
For this to work, it seems like you need to process (in 2D for simplicity) the lower corner, `(ic_lower0,ic_lower1)` first, then the rest of the first row, `(ic_lower0+1:ic_upper0,ic_lower0)`, and then the rest of the data.

boyceg:
Note that here, `d` is indexing the "depth" and `s` is indexing the particular point within the Lagrangian vector. (For IBFE, those are vectors of quadrature points, not actual degrees of freedom.)

The original loops are, e.g.,
```
           V(d,s) = 0.d0
            do ic1 = ic_trimmed_lower(1),ic_trimmed_upper(1)
               do ic0 = ic_trimmed_lower(0),ic_trimmed_upper(0)
                  V(d,s) = V(d,s)
     &                 +w(0,ic0-ic_lower(0))
     &                 *w(1,ic1-ic_lower(1))
     &                 *u(ic0,ic1,d)
               enddo
            enddo
         enddo
```
To avoid initializing to zero, I would think we would need to do:
```
c       Lower corner of the stencil:
           V(d,s) =
     &                +w(0,ic_trimmed_lower(0)-ic_lower(0))
     &                 *w(1,ic_trimmed_lower(1)-ic_lower(1))
     &                 *u(ic_trimmed_lower(0),ic_trimmed_lower(1),d)
c       Rest of the first row of the stencil:
            do ic0 = ic_trimmed_lower(0)+1,ic_trimmed_upper(0)
               V(d,s) = V(d,s)
     &              +w(0,ic0-ic_lower(0))
     &              *w(1,ic_trimmed_lower(1)-ic_lower(1))
     &              *u(ic0,ic_trimmed_lower(1),d)
            enddo
c       Rest of the stencil:
            do ic1 = ic_trimmed_lower(1)+1,ic_trimmed_upper(1)
               do ic0 = ic_trimmed_lower(0),ic_trimmed_upper(0)
                  V(d,s) = V(d,s)
     &                 +w(0,ic0-ic_lower(0))
     &                 *w(1,ic1-ic_lower(1))
     &                 *u(ic0,ic1,d)
               enddo
            enddo
```

boyceg:
I wonder if `ifort` is able to do this automatically. This does not seem like something that we should have to do ourselves.

boyceg:
@drwells, do you want to revisit or close?

drwells:
The improvements are marginal compared to more pressing things; lets close this for now.

PR #448, opened 2019-01-30
Add quadrature cache

drwells:
This is the last caching object that we can use. I am still running callgrind to see how much this saves us.

I want to add one more optimization: store a `quad_key_type` for each `Elem` during the first loop over elements in `interpWeighted` and `spread`. Since `quad_key_type` is just three integers this is not a lot of memory but it will save us a lot of repeated calculations. I'll try to add that later tonight or tomorrow (I have to be somewhere at 7).

### DIFF
--- a/ibtk/include/ibtk/FECache.h
+++ b/ibtk/include/ibtk/FECache.h
@@ -30,6 +30,9 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+#ifndef included_IBTK_FECache
+#define included_IBTK_FECache
+
 /////////////////////////////// INCLUDES /////////////////////////////////////
 
 #include <libmesh/enum_elem_type.h>
@@ -186,3 +189,5 @@ FECache::operator[](const FECache::key_type &quad_key)
 } // namespace IBTK
 
 //////////////////////////////////////////////////////////////////////////////
+
+#endif //#ifndef included_IBTK_FECache
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -53,6 +53,7 @@
 #include "VariableContext.h"
 #include "boost/multi_array.hpp"
 #include "ibtk/ibtk_utilities.h"
+#include "ibtk/QuadratureCache.h"
 #include "libmesh/dof_map.h"
 #include "libmesh/elem.h"
 #include "libmesh/enum_order.h"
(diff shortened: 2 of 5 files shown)


drwells:
This shaves off 6 billion more cycles (367 -> 361), but we can get rid of about 12 billion more cycles if we cache the quadrature keys between calls to `interpWeighted` and `spread`. I'll think about how to do this.

In general, we should have some auxiliary data structures where we store some extra information about `libMesh` objects that we recreate after repartitioning to speed up our calculations. This would be a good thing to compute once and reuse a lot.

drwells:
Okay, I think I figured out what I want to do. This will also play well with partitioning.

We are now up to four cache (`SystemDofMapCache`, `FEMapCache`, `FECache`, and `QuadratureCache`) objects and I want to add a fifth (well, just `std::vector<std::vector<quad_key_type>>` for each patch level): it would make sense to group these into a class that is (partially) cleared whenever we repartition. That would keep all of our libMesh performance hacks (some of which depend on mesh data) in one place.

That being said: I think I would like to merge this first. It would be easier to reach a consensus on this, refactor, and then add the last cache and declare victory on this set of optimizations.

drwells:
Nevermind, that won't work; the individual quadrature rules depend on `X` so that cannot be cached between calls (we can, of course, compute them once instead of 3 times in `interpWeighted` though). Hence we don't really need to implement a caching object when `SystemDofCache` is the only cached object that needs to survive between calls.

boyceg:
> Nevermind, that won't work; the individual quadrature rules depend on `X` so that cannot be cached between calls (we can, of course, compute them once instead of 3 times in `interpWeighted` though).

Yes: the quadrature occurs in the reference coordinate system (i.e. the mesh coordinates), but here the particular rule will depend on the strain state in the current configuration --- and hence depends on `X`.

My thought in #443 was to maintain a library of pre-initialized quadrature rules that are initialized in a "lazy" fashion.

drwells:
Saving the quadrature keys helps a lot: its about 12 billion more instructions saved. For comparison, thats about the total number of instructions spent on the godunov, minmod, etc. kernels.

boyceg:
> Saving the quadrature keys helps a lot: its about 12 billion more instructions saved. For comparison, thats about the total number of instructions spent on the godunov, minmod, etc. kernels.

This is before using the quadrature rule cache in `updateQuadratureRule()`?

drwells:
Its after adding that. The main cost for the quadratures is determining which rule, not necessarily building them.

boyceg:
@drwells is this ready to merge? It looks like `FEDataManager::updateQuadratureRule()` is not currently using a cache.

drwells:
You're right; I did not change what `updateQuadratureRule` does. I doubt that anyone is using it in this way, but since the first argument is a `unique_ptr` we will have to break this function to make it return a cached quadrature rule instead. I kept the old semantics so that the interface doesn't change, albeit we don't really need use this function any more.

How about I add two more commits:

1. Get rid of the remaining calls to `updateInterpQuadratureRule` and then we can just deprecate `updateQuadratureRule`, `updateSpreadQuadratureRule`, and `updateInterpQuadratureRule`
2. Give each `FEDataManager` its own `QuadratureCache` data member so that we don't have to recompute quadrature rules more than once per part.

boyceg:
Sounds reasonable.

drwells:
I added the last two commits.

boyceg:
Looks good to me.

boyceg:
@drwells have you verified that this passes the test suite? One way that the code could run way faster would be if it is now using quadrature rules with (many) fewer quadrature points...

boyceg:
FWIW, on my laptop, I am seeing ~5% performance difference between current master and the current state of this branch for IBFE ex4 in 2D, and a smaller difference for the ex4 3D test.

boyceg:
@drwells is this ready to roll?

PR #449, opened 2019-02-01
Enable a faster code path in index lookups.

### LINKED ISSUE #445: Cut down on calls to VecGetOwnershipRange
This shows up surprisingly often in callgrind profiling: it is about 5% of our total run time. We call it a lot, apparently, in  `IBTK::get_values_for_interpolation` and `IBTK::FEDataInterpolation::collectDataForInterpolation`. We should find a way to cut down on these calls: we call it roughly one million times per time step which is silly.

This function is dead simple: it just [reads two integers](https://www.mcs.anl.gov/petsc/petsc-current/src/vec/vec/interface/vector.c.html#VecGetOwnershipRange). I don't know why we are spending so many cycles on it.

drwells:
If we explicitly localize the vector through the libMesh interface then the local sizes are cached: i.e., we no longer need to call `VecGetOwnershipRange` repeatedly when looking up local indexes since libMesh assumes the value is constant.

Since libMesh uses exactly the same localization strategy that we do its worthwhile to just let it manage this part so that we can use its own optimizations when looking up indices.

Fixes #445: we no longer go down that code path and we don't need to call that function millions of times per time step.

I haven't done any profiling of this yet. I'll post results in the morning when the run is done.

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -722,25 +722,19 @@ FEDataManager::spread(const int f_data_idx,
     }
     else
     {
-        // Extract local form vectors.
+        // Extract local form vectors. PetscVector::get_array_read()
+        // ultimately returns a pointer to a local form of the ghosted vector
+        // (i.e., via VecGhostGetLocalForm and VecGetArrayRead) which is
+        // exactly what we want.
+        //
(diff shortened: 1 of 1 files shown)


boyceg:
We may want to do this anywhere that we use `get_values_for_interpolation`.

drwells:
Superceded by #450, which includes these changes and more.

PR #450, opened 2019-02-01
use PetscVector::get_array_read() to access local form arrays

boyceg:
This includes the changes in #449 but also adds extra calls to `PetscVector::get_array_read()` to access local form arrays.

### DIFF
--- a/ibtk/include/ibtk/libmesh_utilities.h
+++ b/ibtk/include/ibtk/libmesh_utilities.h
@@ -332,14 +332,9 @@ get_values_for_interpolation(MultiArray& U_node,
                              const std::vector<unsigned int>& dof_indices)
 {
     libMesh::PetscVector<double>* U_petsc_vec = dynamic_cast<libMesh::PetscVector<double>*>(&U_vec);
-    Vec U_global_vec = U_petsc_vec->vec();
-    Vec U_local_vec;
-    VecGhostGetLocalForm(U_global_vec, &U_local_vec);
-    double* U_local_soln;
-    VecGetArray(U_local_vec, &U_local_soln);
+    const double* const U_local_soln = U_petsc_vec->get_array_read();
     get_values_for_interpolation(U_node, *U_petsc_vec, U_local_soln, dof_indices);
-    VecRestoreArray(U_local_vec, &U_local_soln);
-    VecGhostRestoreLocalForm(U_global_vec, &U_local_vec);
+    U_petsc_vec->restore_array();
     return;
 } // get_values_for_interpolation
 
@@ -350,14 +345,9 @@ get_values_for_interpolation(MultiArray& U_node,
                              const std::vector<std::vector<unsigned int> >& dof_indices)
 {
     libMesh::PetscVector<double>* U_petsc_vec = dynamic_cast<libMesh::PetscVector<double>*>(&U_vec);
-    Vec U_global_vec = U_petsc_vec->vec();
-    Vec U_local_vec;
-    VecGhostGetLocalForm(U_global_vec, &U_local_vec);
-    double* U_local_soln;
-    VecGetArray(U_local_vec, &U_local_soln);
+    const double* const U_local_soln = U_petsc_vec->get_array_read();
     get_values_for_interpolation(U_node, *U_petsc_vec, U_local_soln, dof_indices);
-    VecRestoreArray(U_local_vec, &U_local_soln);
-    VecGhostRestoreLocalForm(U_global_vec, &U_local_vec);
+    U_petsc_vec->restore_array();
     return;
 } // get_values_for_interpolation
 
--- a/ibtk/include/ibtk/FEDataInterpolation.h
+++ b/ibtk/include/ibtk/FEDataInterpolation.h
@@ -338,6 +338,7 @@ private:
     std::vector<std::vector<int> > d_system_all_vars, d_system_vars, d_system_grad_vars;
     std::vector<std::vector<size_t> > d_system_var_idx, d_system_grad_var_idx;
     std::vector<libMesh::NumericVector<double>*> d_system_data;
+    std::vector<const double*> d_system_local_vecs;
     std::vector<std::vector<size_t> > d_system_var_fe_type_idx, d_system_grad_var_fe_type_idx;
     std::vector<std::vector<std::vector<double> > > d_system_var_data;
     std::vector<std::vector<std::vector<libMesh::VectorValue<double> > > > d_system_grad_var_data;
--- a/ibtk/src/lagrangian/FEDataInterpolation.cpp
+++ b/ibtk/src/lagrangian/FEDataInterpolation.cpp
@@ -51,6 +51,12 @@ FEDataInterpolation::FEDataInterpolation(const unsigned int dim, FEDataManager* 
 
 FEDataInterpolation::~FEDataInterpolation()
 {
+    const size_t num_systems = d_systems.size();
+    for (size_t system_idx = 0; system_idx < num_systems; ++system_idx)
+    {
+        auto system_data = static_cast<PetscVector<double>*>(d_system_data[system_idx]);
+        system_data->restore_array();
+    }
     return;
 }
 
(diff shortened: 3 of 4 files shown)


boyceg:
@drwells, I think this captures all of the important code paths (and some unimportant one) that could benefit from this optimization.

There are a few places in `IBFEMethod` that could benefit, but they are uncommonly used at present. (Those code paths are related to constraining the motion of the overlapping regions of two parts in ways that, AFAICT, do not work.)

boyceg:
This second commit moves the calls to `PetscVector::get_array_read()` in `FEDataInterpolation` into `FEDataInterpolation::init()`. This seems slightly risky because there is not a clean way to call `PetscVector::restore_array()` once we are done with those vectors.

Perhaps there should be a `FEDataInterpolation::finalize()` that is called once we are done with the interpolation? I didn't want to add that just yet, and so I just put the calls to `restore_array()` in `~FEDataInterpolation()`.

boyceg:
FWIW, I just ran some timing tests that seem to indicate that while these changes do improve performance, they do not make much impact.

boyceg:
(I still think we should make some or all of these changes. Every little bit helps! :-)

drwells:
I'm profiling this now.

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -350,14 +345,9 @@ get_values_for_interpolation(MultiArray& U_node,
...
-    VecGhostGetLocalForm(U_global_vec, &U_local_vec);
-    double* U_local_soln;
-    VecGetArray(U_local_vec, &U_local_soln);
+    const double* const U_local_soln = U_petsc_vec->get_array_read();
     get_values_for_interpolation(U_node, *U_petsc_vec, U_local_soln, dof_indices);
-    VecRestoreArray(U_local_vec, &U_local_soln);
-    VecGhostRestoreLocalForm(U_global_vec, &U_local_vec);
+    U_petsc_vec->restore_array();
     return;
 } // get_values_for_interpolation
```
drwells: This was the first approach I considered, but it is a lot less efficient than doing the read operation at the beginning of, e.g., `interpWeighted` and restoring at the end since libMesh will destroy the locally ghosted `Vec` when calling `restore_array` by calling `VecRestoreArrayRead` and `VecGhostRestoreLocalForm`.

Doing it this way is somewhat faster than what we have on master, but I think the right approach is to put this kind of localization outside the loop that calls `get_values_for_interpolation` repeatedly to avoid this. The downside is that we have to modify all such loops to get the full performance improvement.

boyceg:
That is really a convenience function that is not used in any hot code paths (I think!).

boyceg:
Because it is slow, I was thinking about removing or renaming that function. Maybe we should deprecate it?

ibtk/include/ibtk/libmesh_utilities.h
boyceg: Replying by email did not preserve the thread here. To recap:

That is really a convenience function that is not used in any hot code paths (I think!). Because it is slow, I was thinking about deprecating it. I think it is very unlikely to show up in applications, however, and so it also is tempting to just delete it.

I am pretty sure that this particular version of `get_values_for_interpolation` is only used in a couple of places in the code, in code related to computing constraint forces between overlapping structures. Those constraint forces do not work properly, however, and hopefully no one is using them...

boyceg:
> It might be worthwhile to keep this PR as-is; its still a big improvement.

What would you change about the usage of `get_array_read()` / `restore_array()`? I think that those arrays are accessed well outside of inner loops. Do you see places where this is not happening?

> I am getting some nonsense results from my machine right now (for some reason master is 4x slower than this branch). I am investigating.

Did you figure this out?

ibtk/include/ibtk/libmesh_utilities.h
drwells: Thats a good point; we could avoid the problem by just inlining this whole function. I'll take a look.

drwells:
I didn't figure out the weird profiling results: it doesn't make any sense that this patch speeds up the benchmark by a factor of two. I must be doing something dumb.

I'll take a look at this this afternoon; it should be easy to sort out.

ibtk/include/ibtk/libmesh_utilities.h
boyceg: I am just repeating myself here, but note that the version of the function that this comment is referring to _is not_ used in IBAMR except for a couple of places. Most calls to `get_values_for_interpolation` use the version that includes the extracted C-style array (see [here](https://github.com/IBAMR/IBAMR/blob/80cc1551e82308b88fbf64b980b80b915091aef2/ibtk/include/ibtk/libmesh_utilities.h#L286)). I think that all of the calls to that function use that version of the function, and the calls use C arrays that are extracted outside of the element or patch loop.

boyceg:
I only see a ~5% speedup when I ran this the other night on IBFE ex0. But I could imagine cases where there is a bigger impact.

drwells:
I think there is something wrong with how I compiled things; that sounds more reasonable.

boyceg:
@drwells is this ready to roll?

drwells:
@boyceg I have some concerns about the code in FEDataInterpolation: I am pretty sure that it works, but I want to convince myself that we won't run into trouble by accessing the raw pointer to the localized buffer after an `FEDataInterpolation` instance is destroyed. I believe that this is not an issue since we do not support GPUs or other interesting hardware but I want to be sure.

Otherwise this is ready. We could chop those parts out and merge what we have.

drwells:
i.e., I am convinced that the changes made to `FEDataManager` are all good, but I am less sure about what will happen if we do "open, open, close, read, close" at the "read" stage.

boyceg:
> i.e., I am convinced that the changes made to `FEDataManager` are all good, but I am less sure about what will happen if we do "open, open, close, read, close" at the "read" stage.

Is that really what is happening? I don't think there are any closes before getting to reading the data from the raw arrays.

Also, the arrays will be restored when the `FEDataInterpolator` is destroyed.

The paranoid thing to do would be to read on clones of the vectors built in `FEDataInterpolator::init()`?

drwells:
> Is that really what is happening? I don't think there are any closes before getting to reading the data from the raw arrays.

I read through everything and I cannot find a time when this happens; it looks like its just hypothetical.

> The paranoid thing to do would be to read on clones of the vectors built in `FEDataInterpolator::init()`?

I'll take a closer look tomorrow, but I think the best thing to do would be to put all the open/close calls in `FEDataManager` so that no one else is tampering with the state of objects they don't own. This won't work with `IBFEMethod`, though, but I don't remember seeing those functions take up a lot of time in the profiler.

Cloning is not a bad idea either; its definitely cheaper than all the quadrature rule calculations, but I think for now it would be best to be lazy and not add these calls to the interpolation code so that we don't get invalid pointers in unexpected places, even if the invalidation happens at the end of the block.

boyceg:
> > Is that really what is happening? I don't think there are any closes before getting to reading the data from the raw arrays.
> 
> I read through everything and I cannot find a time when this happens; it looks like its just hypothetical.
> 
> > The paranoid thing to do would be to read on clones of the vectors built in `FEDataInterpolator::init()`?
> 
> I'll take a closer look tomorrow, but I think the best thing to do would be to put all the open/close calls in `FEDataManager` so that no one else is tampering with the state of objects they don't own. This won't work with `IBFEMethod`, though, but I don't remember seeing those functions take up a lot of time in the profiler.
> 
> Cloning is not a bad idea either; its definitely cheaper than all the quadrature rule calculations, but I think for now it would be best to be lazy and not add these calls to the interpolation code so that we don't get invalid pointers in unexpected places, even if the invalidation happens at the end of the block.

Note that the constitutive model calculations in the IBAMR examples are quite simple. It might be worthwhile to do some profiling of the heart model.

boyceg:
OTOH, the interface for `FEDataInterpolation` is already pretty gross.

Here is something simpler, although I am not sure it is necessarily any better: Add functions to `FEDataInterpolation` to "open" and "close" the system vectors, and to use the extracted arrays when available, otherwise defaulting to the slower implementation.

boyceg:
I think I am going to close this for now. In any case, whatever we do in `FEDataInterpolation` should be a little different from what is currently in this branch.

boyceg:
If we want to recover this branch, the last commit was 80cc1551e82308b88fbf64b980b80b915091aef2.

Issue #438, opened 2019-01-28
Running bootstrap updates configuration on master


drwells:
Running the updated `bootstrap` script (see #426) still changes things on master. Should we rerun this script on `master` and push the result? Is this deliberate?

@abarret *ping*

abarret:
Hmmm... It looks like most of the changes are in `ibtk/contrib/muparser` or `ibtk/config` and they seem to be mostly benign (or at least to my untrained eye they do). My guess is that whenever we initially started switching to the newer version of automake we never included these updated config files. I'll create a pull request.

drwells:
Fixed by #451.

PR #451, opened 2019-02-01
Rerun of bootstrap on master.

abarret:
This is a rerun of bootstrap on master as addressed in #438 and as a follow up to #426.


drwells:
@boyceg this is the commit I mentioned earlier: we avoid more autotools churn now.

PR #456, opened 2019-02-04
Fix ibfe restart

drwells:
Supercedes #454: since we avoid localization this should always set up ghost data. I am pretty sure that `VecScatterCreate`, when called with complete index sets, will set up the current vector with a global set of ghost data: not what we want.

@xxjaehoxx would it be possible for you to give this a try?

The same ex0 test program passes for me in debug mode, so this should be okay.


xxjaehoxx:
@drwells is this change pushed to boyceg/fix-IBFE-restart branch? I did git pull, but I don't see any changes.

boyceg:
@drwells looks good to me. Are those lambdas useful enough to put into `libmesh_utilities.h`?

@xxjaehoxx this is a different branch. You can get it via:
```
git checkout -b drwells-fix-IBFE-restart master
git pull https://github.com/drwells/IBAMR.git fix-IBFE-restart
```

PR #457, opened 2019-02-06
use PetscVector::get_array_read() to access local form arrays in FEDataManager

boyceg:
This PR includes the parts of #450 that @drwells likes and omits the parts that he doesn't.

### DIFF
--- a/ibtk/include/ibtk/libmesh_utilities.h
+++ b/ibtk/include/ibtk/libmesh_utilities.h
@@ -332,14 +332,9 @@ get_values_for_interpolation(MultiArray& U_node,
                              const std::vector<unsigned int>& dof_indices)
 {
     libMesh::PetscVector<double>* U_petsc_vec = dynamic_cast<libMesh::PetscVector<double>*>(&U_vec);
-    Vec U_global_vec = U_petsc_vec->vec();
-    Vec U_local_vec;
-    VecGhostGetLocalForm(U_global_vec, &U_local_vec);
-    double* U_local_soln;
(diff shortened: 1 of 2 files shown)


boyceg:
Can you run callgrind to see if this addresses the large numbers of calls to `PetscCheckPointer`?

drwells:
The changes in FEDataManager should do this if they were the same as my original patch. I will check.

drwells:
Recompiling this on my laptop is going to take close to an hour so I'll verify this in the morning.

drwells:
We eliminated about 99% of the calls to `VecGetRange`, which in turn eliminated about half the calls to `PetscCheckPointer`. Its not perfect, but it is now much less prominent in the profiling results.

knepley:
@boyceg @drwells Did you guys run with -check_pointer_intensity 0 in order to turn off the pointer checks?

boyceg:
I haven't yet, although @drwells is running most of these checks. They do not show up prominently in the Apple Time Profiler runs that I have done.

drwells:
@knepley I do not see any changes when running with `-check_pointer_intensity 0` as a command line argument: the callgrind results don't change at all when I add that.

knepley:
Do you mean time or number of calls? We see significant time savings on hot routines. That is strange if you see nothing.

drwells:
I saw absolutely no difference in either when executing the test program as

```
valgrind --tool=callgrind ./main3d ./input3d -check_pointer_intensity 0
```

PR #458, opened 2019-02-06
use tsl::robin_map instead of std::unordered_map

boyceg:
In profiling, I am seeing that the DOF index cache is relatively expensive. However, most of those lookups are related to getting data from a map.

This uses [tsl::robin_map](https://github.com/Tessil/robin-map) in place of `std::unordered_map` in an attempt to speed things up.

At least on my Mac, there is very little performance difference.

### DIFF
--- a/ibtk/contrib/robin-map/robin_growth_policy.h
+++ b/ibtk/contrib/robin-map/robin_growth_policy.h
@@ -1,0 +1,324 @@
+/**
+ * MIT License
+ * 
+ * Copyright (c) 2017 Tessil
+ * 
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ * 
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ * 
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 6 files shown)


drwells:
I'll take a look.

boyceg:
On Mac, I think that `std::unordered_map` is from libc++. I have not made any comparisons for GNU libstdc++.

I don't know if look-ups for other caching objects are worth optimizing, but we could try to make a hash function for the `std::tuple` keys.

drwells:
I am yet to see any of the other lookups show up in profiling, but adding a hash would be pretty easy: the keys are, currently, just three integers. In particular, since each integer is bounded by something small (e.g., we only use up to order 43 quadrature rules) we could probably do this by bit shifting their values onto a 64 bit key with something like `union {std::size_t, short int[4]}`.

There has to be a better way to cache these values that doesn't involve a hash table: there are always N elements contiguously. One possibility would be to add even more band-aids and just start indexing libMesh cells with our own ids so we can look stuff up directly in a contiguous buffer.

boyceg:
For the other caching data structures we could just use `boost::multi_array`.

drwells:
Using a non-sparse data structure could result in some very large arrays: e.g., if we have 100 valid element types, 50 valid orders, and 10 valid quadrature rule types we are now allocating room for 50,000 pointers when we only need a few dozen or so :)

boyceg:
Oh yeah, I was thinking there were fewer element types at play.

I guess we should just leave these alone unless they show up in profiling.

drwells:
I think I know another way to improve this. In every case that we look up data with `SystemDofMapCache` in `FEDataManager` we always have `e_idx` around: i.e., we can always look up any element dof info by indexing an array instead of traversing an array. We will have to add some more assertions to guarantee that we are always on the correct patch but that should be straightforward.

boyceg:
Are `e_idx` ranges continuous? If so then it would be better to just use an array.

boyceg:
I can take a look at this while you work on partitioning.

boyceg:
But those ranges change dynamically. (Would need to invalidate the caches, which we can do when resetting the active element lists.)

drwells:
Do they change more frequently than the interval at which we regrid?

boyceg:
Looking more carefully at the profiling data, I think that the expensive part of the DOF cache lookup is actually coping the data out of the cache, not looking it up. I am going to experiment with just returning a pointer to the data and see if it makes a difference.

PR #461, opened 2019-02-06
Refactor data reinitialization in IBFEMethod.

drwells:
First change from #396. Getting this merged will be easier if we do it piecewise.

This doesn't change anything at the moment, but is necessary for using the SAMRAI-based partitioner.


boyceg:
Looks good.

drwells:
I'll take that as an approving review :)

PR #462, opened 2019-02-06
Clear the system dof cache on partitioning.

drwells:
Another spinoff of #396.

libMesh guarantees that each processor owns a contiguous range of dofs. Hence, when we repartition, the new dof numbering on each elem and node is unrelated to the old numbering, so all of this information is out of date.

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -107,6 +107,11 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
 public:
     /*!
      * Class which enables fast lookup of dofs on a given libMesh <code>elem</code>.
+     *
+     * @note The contents of this cache are invalidated when we regrid and the
+     * caches should be reset at that point. Copies of this class should
+     * always be retrieved via FEDataManager::getDofCache() to avoid this
+     * problem.
      */
     class SystemDofMapCache
     {
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -471,6 +471,10 @@ FEDataManager::reinitElementMappings()
 {
     IBTK_TIMER_START(t_reinit_element_mappings);
 
+    // We reinitialize mappings after repartitioning, so clear the cache since
+    // its content is no longer relevant:
+    d_system_dof_map_cache.clear();
+
     // Delete cached hierarchy-dependent data.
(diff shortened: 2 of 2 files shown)


ibtk/include/ibtk/FEDataManager.h
```diff
@@ -118,6 +118,15 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
         {
         }
 
+        /*!
+         * Clear the contents of the cache.
+         */
+        inline void
+        clear()
+        {
+            d_dof_cache.clear();
```
boyceg: I assume that the `DofMap` object itself does not change, but I wonder if it would nonetheless be "safer" to wipe out the `SystemDofMapCache` objects instead, and to repopulate them from scratch as needed. `LDataManager` is already set up to automatically rebuild the cache object if the stored pointer is null.

drwells: That's a better idea: that way we are covered in case the `DofMap` object is destroyed.

drwells: Fixed!

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -471,6 +471,14 @@ FEDataManager::reinitElementMappings()
 {
     IBTK_TIMER_START(t_reinit_element_mappings);
 
+    // We reinitialize mappings after repartitioning, so clear the cache since
```
boyceg: This is fine, but I would suggest simply using `d_system_dof_map_cache->clear()`. As long as we always access via `::getSystemDofMapCache()` --- which we do currently --- then these will get set up automatically there as needed.

boyceg:
Feel free to make the suggested change and merge, or just to merge --- up to you. :-)

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -107,6 +107,9 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
 public:
     /*!
      * Class which enables fast lookup of dofs on a given libMesh <code>elem</code>.
+     *
+     * @note The contents of this cache are invalidated when we regrid and the
```
boyceg: I guess maybe we should say that these are intended to be accessed only via `FEDataManager::getDofCache()`?

drwells:
These are good changes since they allow us to make everything shorter. I did both.

boyceg:
Looks good to me!

PR #463, opened 2019-02-06
Partitioning 2c

drwells:
Another spinoff from #396.

This PR adds the actual partitioner class as well as a 'partitioning box' class. The former inherits from `libMesh::Partitioner` and implements all functionality necessary for integration with libMesh's partitioning code. The later is like a bounding box but uniquely chops up the domain with a tensor product of half-open intervals.

None of this code is actually used in the library at present: I will put up a PR for that once this, #462, and #461 are merged.

### DIFF
--- a/ibtk/include/ibtk/BoxPartitioner.h
+++ b/ibtk/include/ibtk/BoxPartitioner.h
@@ -1,0 +1,117 @@
+// Filename: BoxPartitioner.h
+// Created on 6 Nov 2018 by David Wells
+//
+// Copyright (c) 2018, Boyce Griffith, David Wells
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright notice,
+//      this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//    * Neither the name of The University of North Carolina nor the names of
+//      its contributors may be used to endorse or promote products derived from
+//      this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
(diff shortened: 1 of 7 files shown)


ibtk/src/lagrangian/BoxPartitioner.cpp
```diff
@@ -0,0 +1,332 @@
...
+            local_node_ids.push_back(node_n);
+            TBOX_ASSERT(node_n < node_ids.size());
+            node_ids[node_n] = current_rank + 1;
+            ++n_local_nodes;
+        }
+        ++node_n;
+    }
+
+    // step 1.5: log the current state of affairs:
+    const int n_processes = SAMRAI::tbox::SAMRAI_MPI::getNodes();
```
boyceg: Should we make this optional, and off by default?

drwells: Sure.

(note to self: add `bool d_enable_logging = false;` to the header and initialize it in the constructor.)

drwells: Fixed!

drwells:
I copied the boolean logging logic from the IBTK linear solvers.

ibtk/include/ibtk/BoxPartitioner.h
```diff
@@ -0,0 +1,117 @@
...
+     *
+     * x,y,z,r
+     *
+     * format.
+     */
+    void writePartitioning(const std::string &file_name) const;
+
+    virtual std::unique_ptr<libMesh::Partitioner> clone() const override;
+
+    virtual void _do_partition(libMesh::MeshBase& mesh, const unsigned int n) override;
```
boyceg: Should this be `protected`?

drwells: Good catch: yes.

ibtk/include/ibtk/PartitioningBox.h
```diff
@@ -0,0 +1,146 @@
...
+    /// Return true if the point is inside the set of boxes and false
+    /// otherwise.
+    bool contains(const Point &point) const;
+
+    /// Return a pointer to the first partitioning box.
+    const PartitioningBox * begin() const;
+
+    /// Return a pointer to one past the end of the end of the partitioning
+    /// box array.
+    const PartitioningBox * end() const;
```
boyceg: Worth providing a range for range-based for?

drwells:
I fixed the access specifier.

boyceg:
FWIW, I like to use
```
#include "ibtk/namespaces.h" // IWYU pragma: keep
```
to provide several namespace `using` declarations, which include SAMRAI and libMesh, so that they can be omitted in implementation files.

ibtk/src/utilities/PartitioningBox.cpp
```diff
@@ -0,0 +1,124 @@
...
+                = patch.getPatchGeometry();
+            boxes.emplace_back(*patch_geometry);
+        }
+        return boxes;
+    }
+}
+
+PartitioningBoxes::PartitioningBoxes(const SAMRAI::hier::PatchHierarchy<NDIM> &hierarchy)
+{
+    const auto boxes = get_partitioning_boxes(hierarchy);
```
boyceg: Not sure that this use of `auto` follows the unofficial rules-of-auto.

boyceg:
BTW, did all of this get run through clang-format?

(I guess we are running everything through soon enough.)

There is a script (scripts/setup_clang_format_precommit_hook.sh) that will set up a clang-format git precommit hook.

ibtk/include/ibtk/PartitioningBox.h
drwells: Unless I am mistaken providing the pair there is enough to get range-based for working. I did use one laster.

boyceg: OK great! You know better than me.

drwells:
I haven't been running clang-format; I'll push a commit that does that.

ibtk/src/utilities/PartitioningBox.cpp
drwells: This is easy to redo. I'll fix it.

drwells:
I am not a big fan of the namespace headers since we already have a lot of conflicts between eigen and libmesh or eigen and samrai, but I added it for consistency.

boyceg:
Looks good to me.

boyceg:
> I am not a big fan of the namespace headers since we already have a lot of conflicts between eigen and libmesh or eigen and samrai, but I added it for consistency.

I think all of the namespaces make things hard to read, but don't include these `using` statements if you don't want to!

drwells:
We do it everywhere else so I added it in; I'll live!

drwells:
Since @boyceg approved and this does not actually change anything in the library I will merge.

PR #465, opened 2019-02-07
use (local_patch_num,element_index) to look up cached DOF indices

boyceg:
An alternative approach to that in #458 in trying to speed up retrieval of cached DOF indices.

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -38,9 +38,9 @@
 #include <map>
 #include <ostream>
 #include <string>
+#include <unordered_map>
 #include <utility>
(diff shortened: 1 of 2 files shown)


drwells:
I want to take some time to read over this and try to understand how libMesh implements `dof_object->id()`:  I poked around in a debugger and the results imply that elements are numbered consecutively on each processor (i.e., the patch contained element 3, element 4, element 5, etc), which would make this a lot easier to implement and maintain.

boyceg:
Yes indeed; I had this though myself as I was putting together this terrible looking code.

I will ping libmesh-users.

boyceg:
I am closing this per the discussion in #458.

PR #467, opened 2019-02-07
Partitioning 2d

drwells:
Another spinoff from #396: this implements one of the requested features (user-specified workload weights).

There is some new code in here that is not just cherry-picked from #396, so I want to test this some more (i.e., look at workload estimate pictures) before we merge, but I am happy with the current approach and management of multiple parts.

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -100,6 +100,18 @@ namespace IBTK
  * \brief Class FEDataManager coordinates data required for Lagrangian-Eulerian
  * interaction between a Lagrangian finite element (FE) mesh.
  *
+ * <h3>Parameters effecting workload estimate calculations</h3>
+ * FEDataManager can estimate the amount of work done in IBFE calculations
+ * (such as FEDataManager::spread). Since most calculations use a variable
+ * number of quadrature points on each libMesh element this estimate can vary
+ * quite a bit over different Eulerian cells corresponding to a single
+ * mesh. The current implementation estimates the workload on each cell of the
+ * background Eulerian grid by applying a background value representing the
+ * work on the Eulerian cell itself and a weight times the number of
+ * quadrature points on that cell. These values are set at the time of object
+ * construction through the FEDataManager::WorkloadSpec object, which contains
+ * reasonable defaults.
+ *
  * \note Multiple FEDataManager objects may be instantiated simultaneously.
  */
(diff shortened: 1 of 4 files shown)


ibtk/include/ibtk/FEDataManager.h
```diff
@@ -220,6 +232,26 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
...
+        /// the calculation. This should be true unless some other object is
+        /// also adding values into the workload estimate.
+        bool clear_estimate = true;
+
+        /// The multiplier applied to each quadrature point.
+        double q_point_weight = 2.0;
+
+        /// Multiplier on Eulerian cells. At the present time this is always
+        /// assumed to be 1.
+        static constexpr double eulerian_cell_weight = 1.0;
```
boyceg: Do we need this parameter? I am understanding this to say: `work_estimate(i,j,k) = eulerian_cell_weight + n_q_points(i,j,k) * q_point_weight`. Because the actual value of the workload estimate does not matter, we can always divide through by eulerian_cell_weight and get: `work_estimate(i,j,k) = 1 + n_q_points(i,j,k) * q_point_weight_prime` with `q_point_weight_prime = q_point_weight / eulerian_cell_weight`.

drwells: Technically no, and getting ODR worked out for this is complicated. I'll just document that it is always 1.

drwells:
I fixed a few more bugs that only appear when someone does not ever set up a workload index variable. This should be ready now.

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -220,6 +232,22 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
...
+     */
+    struct WorkloadSpec
+    {
+        /// Whether or not to clear the workload index arrays at the start of
+        /// the calculation. This should be true unless some other object is
+        /// also adding values into the workload estimate.
+        bool clear_estimate = true;
+
+        /// The multiplier applied to each quadrature point.
+        double q_point_weight = 2.0;
```
boyceg: Is there a way for users to set this value?

boyceg: (I mean in input files, not programatically.)

boyceg:
Looks good to me.

ibtk/include/ibtk/FEDataManager.h
drwells: Yes; its in `IBFEMethod`.

PR #471, opened 2019-02-08
Conditionally log FEDataManager's workload data.

drwells:
Each IBFEMethod propagates its settings to the FEDataManagers it owns.

This is another spinoff of #396, and, like the last four PRs, does not do a lot but add functionality that will be used by the partitioning and load balancing functions.


boyceg:
Looks good to me; merge whenever you are ready.

PR #472, opened 2019-02-11
cache Eulerian data used by FEDataManager::spread

boyceg:
This appears to cut about 3-4% off runtime for the heart model running in serial on my mac. It must be some NUMA thing.

One thing that is not great is that currently there is no mechanism to free the cloned patch data indices. However, they are only cloned once per (data centering , data depth) pair, and the actual allocated data get destroyed correctly, so I am not sure this is a show stopper.


drwells:
I looked at my profiling data on Friday and I wasn't able to spot this hotspot on any of the Linux machines, which is odd. I'll take a closer look after #468.

boyceg:
This showed up for me in the Apple Developer Tools Time Profiler, which is a sampling-based profiling tool. I suspect that you will not notice it in callgrind because the actual operation that is being performed is not very computationally intensive, and should be well optimized.

boyceg:
@drwells @abarret, I have cleaned this up a little bit. Could one of you take a look?

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -890,6 +890,8 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
      */
     SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > d_hierarchy;
     int d_coarsest_ln = IBTK::invalid_level_number, d_finest_ln = IBTK::invalid_level_number;
+    using f_data_idx_map_key_type = std::tuple<int, int>;
+    std::map<f_data_idx_map_key_type, int> d_f_data_idx_map;
```
drwells: It would be clearer to add a little enum here for cell-centered vs. face-centered. I don't know if SAMRAI has such a thing but its easier to understand than making one zero and the other one.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -726,11 +726,34 @@ FEDataManager::spread(const int f_data_idx,
...
+    {
+        Pointer<CellDataFactory<NDIM, double> > f_cc_fac = f_cc_var->getPatchDataFactory();
+        depth = f_cc_fac->getDefaultDepth();
+    }
+    if (sc_data)
+    {
+        Pointer<SideDataFactory<NDIM, double> > f_sc_fac = f_sc_var->getPatchDataFactory();
+        depth = f_sc_fac->getDefaultDepth();
+    }
+    const f_data_idx_map_key_type f_key{ cc_data ? 0 : 1, depth };
```
drwells: I am sure that this works, but I had to think really hard about operator precedence between the ternary expression and the comma operator. Could you put the first bit in parentheses?

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -726,11 +726,34 @@ FEDataManager::spread(const int f_data_idx,
...
+    int f_copy_data_idx = -1;
+    auto it = d_f_data_idx_map.find(f_key);
+    if (it == d_f_data_idx_map.end())
```
drwells: we can skip a line if we just do `d_f_data_idx_map.count(f_key) == 0`

ibtk/include/ibtk/FEDataManager.h
drwells: Also, could you add some documentation for this member variable describing exactly why we are storing this cloned data?

boyceg: Sure.

Could something like [`type_index`](https://en.cppreference.com/w/cpp/types/type_index) work without having to introduce an `enum`?

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: Addressing the above comment will address this one as well.

ibtk/include/ibtk/FEDataManager.h
drwells: Yup.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: Isn't `count()` more expensive than just checking if we are at the end of the map?

drwells: I don't think so. `count` for a `std::map` is equivalent to searching since there is only one value per key.

boyceg:
@drwells, please take a look. I switched to using `type_index` for deciding the data centering. Let me know what you think.

boyceg:
@drwells, if you prefer the slightly more verbose:
```
    if (f_type_index == typeinfo(CellVariable<NDIM,double>))
    {
        // ...
    }
    else if (f_type_index == typeinfo(sideVariable<NDIM,double>))
    {
        // ...
    }
```
to
```
    if (is_cc_data)
    {
        // ...
    }
    else if (is_sc_data)
    {
        // ...
    }
```
then I will also make those changes in other parts of this implementation file. I think I like using `typeinfo` here but am not wedded to it.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: I am not following this suggestion. We need to try to find the key, to get the cached index; shouldn't we go ahead and use it to see if we didn't find the key?

drwells: Good point!

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1561,18 +1590,23 @@ FEDataManager::interpWeighted(const int f_data_idx,
...
             Pointer<PatchData<NDIM> > f_data = patch->getPatchData(f_data_idx);
-            if (cc_data)
+            if (f_type_id == typeid(CellVariable<NDIM, double>))
             {
                 Pointer<CellData<NDIM, double> > f_cc_data = f_data;
                 LEInteractor::interpolate(
                     F_node, n_vars, X_node, NDIM, f_cc_data, patch, interp_box, interp_spec.kernel_fcn);
             }
-            if (sc_data)
+            else if (f_type_id == typeid(CellVariable<NDIM, double>))
```
abarret:
```suggestion
            else if (f_type_id == typeid(SideVariable<NDIM, double>))
```

boyceg: Whoops!

boyceg:
@abarret, @drwells, what do you think about using `typeinfo` here?

I think that using this kind of construct will help with SAMRAI 3 migration. Right now, a lot of the RTTI that occurs is performed via implicit downcasting enabled by `SAMRAI::tbox::Pointer`. Is `typeinfo` better or worse than switch to explicit downcasts?

drwells:
I prefer `typeinfo` since it includes more type checking. I think the SAMRAI 3 approved idiom is to use `dynamic_ptr_cast`, e.g.,
```cpp
// b is of type std::shared_ptr<Base>
auto d = std::dynamic_ptr_cast<Derived>(b); TBOX_ASSERT(d);
```

abarret:
I think I prefer explicit downcasting, but I have no good reason, and do not feel strongly either way.

Would it be beneficial to have a helper function? Something like

```
template <typename S, typename T>
bool inline compare_types(const S& obj1, const T& obj2)
{
	std::type_index type = typeid(obj1);
	return type == typeid(obj2);
}
```

drwells:
@abarret I think you just reinvented `std::is_same` :)

drwells:
I wrote that too quickly.

That will work whenever we need to downcast, but it won't help in the present circumstance if we want to see if a type is equal to a known type. We might need to do something like
```cpp
template <typename U, typename T>
bool
is_same_type(const T& t)
{
    return typeid(t) == typeid(std::declval<U>());
}
```

boyceg:
On second thought, it seems better to do a `dynamic_cast`, because I guess we could imagine wanting to be able to support anything that inherits from `CellVariable` or `SideVariable`.
```
template <typename U, typename T>
bool
can_convert_to(T* t)
{
    return dynamic_cast<U*>(t) != nullptr;
}
```

PR #475, opened 2019-02-11
Clear stale data after repartitioning.

drwells:
These data structures contain information referring to the previous partitioning so they need to be rebuilt.

Since everything we clear is recreated on-demand there should not be any need to add more checks elsewhere regarding whether or not something is set up.

This is another spinoff of #396.


boyceg:
Looks good to me.

Is there a more robust way to do this kind of thing? It can be so easy to forget to call clear...

drwells:
I think we could put everything in its own cache class, but we would either need to do something weird (like placement new) or allocate this cache on the heap and destroy it, to get automatic cleanup of all such data and avoid writing `clear` half a dozen times. I don't think its much of an improvement over writing `clear` half a dozen times here (though IBFEMethod could use an improvement like that one).

PR #478, opened 2019-02-13
Rearrange workload estimation

drwells:
A (larger) spinoff from #396, with some more fixes.

This came out of a conversation I had with @boyceg this afternoon: we don't really have a consistent way of setting up workload estimates. However, most of the necessary infrastructure to do this is already there. The new version assumes that the parent hierarchy integrator (i.e., whatever we call `advanceHierarchy` on) 'owns' the workload estimate array. To estimate the work, it sets all cells to one and then asks each child integrator to add in its own contributions. Things like `IBFEMethod`, `LDataManager`, etc. do not need to know anything about load balancers or store the workload index (or even store a pointer to the hierarchy): they just add their contribution when asked via `addWorkloadEstimate`.

I am labeling this as 'do not merge' since I want to do some more testing to guarantee that all the old code paths actually still work (they should; they've just been deprecated) and that this fits with the rest of the load balancing stack.

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -320,6 +320,9 @@ public:
 
     /*!
      * \brief Register a load balancer for non-uniform load balancing.
+     *
+     * @deprecated Since the correct workload index is passed in via
+     * addWorkloadEstimate this function is no longer necessary.
      */
     void registerLoadBalancer(SAMRAI::tbox::Pointer<SAMRAI::mesh::LoadBalancer<NDIM> > load_balancer,
                               int workload_data_idx);
@@ -630,11 +633,12 @@ public:
                                            double dx_min);
 
     /*!
-     * \brief Update the cell workload estimate. Does nothing unless a
-     * workload variable was previously registered via
-     * FEDataManager::registerLoadBalancer.
+     * \brief Update the cell workload estimate by adding a value (see the
+     * main documentation of this class for information on how this is
+     * computed) to the <code>d_workload_idx</code> cell variable.
      */
-    void updateWorkloadEstimates(int coarsest_ln = -1, int finest_ln = -1);
+    void addWorkloadEstimate(SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy,
+                             const int workload_data_idx, const int coarsest_ln = -1, const int finest_ln = -1);
 
     /*!
      * Initialize data on a new level after it is inserted into an AMR patch
@@ -848,6 +852,9 @@ private:
 
     /*
      * We cache a pointer to the load balancer.
+     *
+     * @deprecated This pointer is never used and will be removed in the
+     * future.
      */
     SAMRAI::tbox::Pointer<SAMRAI::mesh::LoadBalancer<NDIM> > d_load_balancer;
 
@@ -880,6 +887,10 @@ private:
      * SAMRAI::hier::Variable pointer and patch data descriptor indices for the
      * cell variable used to determine the workload for nonuniform load
      * balancing.
+     *
+     * @deprecated d_workload_var and d_workload_idx will not be stored in a
+     * future release since the correct workload index will be provided as an
+     * argument to addWorkloadEstimate.
      */
     SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > d_workload_var;
     int d_workload_idx = IBTK::invalid_index;
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -239,11 +239,6 @@ public:
      */
     struct WorkloadSpec
     {
-        /// Whether or not to clear the workload index arrays at the start of
-        /// the calculation. This should be true unless some other object is
-        /// also adding values into the workload estimate.
-        bool clear_estimate = true;
-
         /// The multiplier applied to each quadrature point.
         double q_point_weight = 2.0;
     };
--- a/ibtk/include/ibtk/LDataManager.h
+++ b/ibtk/include/ibtk/LDataManager.h
@@ -501,6 +501,10 @@ public:
 
     /*!
      * \brief Register a load balancer for non-uniform load balancing.
+     *
+     * @deprecated This method is deprecated since the current strategy for
+     * handling non-uniform load balancing does not require that this object
+     * store a pointer to the load balancer.
      */
     void registerLoadBalancer(SAMRAI::tbox::Pointer<SAMRAI::mesh::LoadBalancer<NDIM> > load_balancer,
                               int workload_data_idx);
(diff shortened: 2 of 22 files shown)


ibtk/include/ibtk/FEDataManager.h
```diff
@@ -239,11 +239,6 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
      */
     struct WorkloadSpec
     {
-        /// Whether or not to clear the workload index arrays at the start of
-        /// the calculation. This should be true unless some other object is
-        /// also adding values into the workload estimate.
-        bool clear_estimate = true;
-
```
drwells: Clearing is now done by the parent integrator.

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -324,7 +319,12 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
     static void freeAllManagers();
 
     /*!
-     * \brief Register a load balancer for non-uniform load balancing.
+     * \brief Register a load balancer for non-uniform load balancing. This
+     * function must be called to set up the workload index before
+     * updateWorkloadEstimates is called.
+     *
+     * @deprecated Since the correct workload index is passed in via
+     * addWorkloadEstimate this function is no longer necessary.
```
drwells: We never used the load balancers anywhere (and they should only be used by the parent hierarchy integrator) so all of these are deprecated. In principle, a non-hierarchy doesn't need to even store a pointer to any samrai object: it just needs to know how to write its workload estimate into the given hierarchy and variable index.

drwells:
I rewrote a massive portion of this patch today. I don't think that there is any way to be backwards compatible with the old `updateWorkloadEstimates` code that was partially implemented in several classes so I deleted all of it (I deprecated all of the related code where we attach load balancers to, e.g., FEDataManager).

This is now in good shape and (*crosses fingers*) is the last major change we need before we can get the partitioner changes full merged.

drwells:
All tests pass for me with this. In fact, I don't think any of the examples (or the heart model) use any of this code (yet).

ibtk/include/ibtk/FEDataManager.h
boyceg: You know, I really think it would be OK to just drop this interface --- I would be shocked if anyone was using it.

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -635,11 +633,12 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
...
     /*!
-     * \brief Update the cell workload estimate. Does nothing unless a
-     * workload variable was previously registered via
-     * FEDataManager::registerLoadBalancer.
+     * \brief Update the cell workload estimate by adding a value (see the
+     * main documentation of this class for information on how this is
+     * computed) to the <code>d_workload_idx</code> cell variable.
      */
-    void updateWorkloadEstimates(int coarsest_ln = -1, int finest_ln = -1);
+    void addWorkloadEstimate(SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy,
```
boyceg: "accumulate" or "account" instead of "add" here?

boyceg: Or maybe `addLocalWorkloadEstimate` or `addClassWorkloadEstimate`?

(`addWorkloadEstimate` sounds to me like we are telling the class about the workload estimate, rather than asking the class to account for its workload.)

ibtk/include/ibtk/FEDataManager.h
drwells: The sheer ad-hocness of that approach in FEDataManager (which, I admit, was all my work) was the main motivation to bite the bullet and redo everything here :)

ibtk/include/ibtk/LDataManager.h
```diff
@@ -501,6 +501,10 @@ class LDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::Sta
 
     /*!
      * \brief Register a load balancer for non-uniform load balancing.
+     *
+     * @deprecated This method is deprecated since the current strategy for
```
boyceg: As above, I would be surprised if anyone is using this interface.

ibtk/include/ibtk/LDataManager.h
```diff
@@ -577,6 +581,10 @@ class LDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::Sta
 
     /*!
      * \brief Get the patch data descriptor index for the workload cell data.
+     *
+     * @deprecated This method is deprecated since, in future versions of
```
boyceg: As above, I would be surprised if anyone is using this interface.

include/ibamr/IBFEMethod.h
```diff
@@ -600,23 +593,19 @@ class IBFEMethod : public IBStrategy
...
-     * FEDataManager::updateWorkloadEstimates to compute the estimated
-     * Lagrangian contribution to the total amount of work.
+     * Add the estimated computational work from the current object (i.e., the
+     * work required by the owned Lagrangian objects) per cell into the
+     * specified <code>workload_data_idx</code>.
      */
-    void updateWorkloadEstimates(SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy,
-                                 int workload_data_idx) override;
+    void addWorkloadEstimate(SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy,
+                             const int workload_data_idx) override;
```
boyceg: Try to avoid `const int` (or similar) in the header.

ibtk/include/ibtk/FEDataManager.h
drwells: I like 'accumulate' since it implies mutability.

boyceg:
Looks good to me. Feel free to make any changes or not. I am going to go ahead and approve, so you can merge whenever you are ready.

include/ibamr/IBFEMethod.h
drwells: I am curious; why? I don't know what difference the const really makes.

ibtk/include/ibtk/FEDataManager.h
drwells: I meant to ask you about that but I forgot: I cannot forsee any reason that any of these objects would need the load balancer. It sounds like you are not aware of any either.

ibtk/include/ibtk/FEDataManager.h
drwells: Though, if you are willing to approve it as-is, I like 'add' more since it is shorter :imp:

drwells:
I also cannot find any evidence that anyone is using these deprecated functions. However, I think it would still be best to go through the motions and deprecate them before removing them (if for no other reason than to build good habits).

ibtk/include/ibtk/FEDataManager.h
drwells: I think I will keep it as add. Using 'accumulate' here reminds me of `std::accumulate`, which does something different: in fact, `updateWorkloadEstimates` should really be `accumulateWorkloadEstimates` if we want to be more pedantic.

drwells:
Since @boyceg approved it as-is I will merge since I very much prefer the current names to the suggested ones (sorry).

PR #479, opened 2019-02-14
Partitioning 2g

drwells:
This is the last partitioning change :tada: 

I want to run some more tests before we merge this, so I am marking it as a work in progress for now.

@cpuelz If you want to give this a spin with the heart model, use this branch and attach the load balancer to the time integrator. Look at how I modified example 4 to see how this is done. You can check the logs to see how it repartitions:

```
workload estimate on processor 0  = 136880 (pid is 19028)
workload estimate on processor 1  = 124576 (pid is 19029)
workload estimate on processor 2  = 126832 (pid is 19030)
workload estimate on processor 3  = 156626 (pid is 19031)
workload estimate on processor 4  = 124834 (pid is 19032)
workload estimate on processor 5  = 143008 (pid is 19033)
workload estimate on processor 6  = 133872 (pid is 19034)
workload estimate on processor 7  = 132770 (pid is 19035)
workload estimate on processor 8  = 128098 (pid is 19036)
workload estimate on processor 9  = 139744 (pid is 19037)
workload estimate on processor 10 = 135072 (pid is 19038)
workload estimate on processor 11 = 128354 (pid is 19039)
workload estimate on processor 12 = 141834 (pid is 19040)
workload estimate on processor 13 = 121366 (pid is 19041)
workload estimate on processor 14 = 129468 (pid is 19042)
workload estimate on processor 15 = 130472 (pid is 19043)
local active DoFs on processor 0  = 177674 (pid is 19028)
local active DoFs on processor 1  = 168746 (pid is 19029)
local active DoFs on processor 2  = 168720 (pid is 19030)
local active DoFs on processor 3  = 200760 (pid is 19031)
local active DoFs on processor 4  = 165180 (pid is 19032)
local active DoFs on processor 5  = 189079 (pid is 19033)
local active DoFs on processor 6  = 178133 (pid is 19034)
local active DoFs on processor 7  = 173842 (pid is 19035)
local active DoFs on processor 8  = 168656 (pid is 19036)
local active DoFs on processor 9  = 185534 (pid is 19037)
local active DoFs on processor 10 = 179810 (pid is 19038)
local active DoFs on processor 11 = 170213 (pid is 19039)
local active DoFs on processor 12 = 189811 (pid is 19040)
local active DoFs on processor 13 = 160696 (pid is 19041)
local active DoFs on processor 14 = 165431 (pid is 19042)
local active DoFs on processor 15 = 139953 (pid is 19043)
```
which is a beautiful partitioning. You will also need to adjust `max_workload_factor` (the value 0.25 seems to work well) so that we can split enough patches.

### DIFF
--- a/ibtk/src/utilities/HierarchyIntegrator.cpp
+++ b/ibtk/src/utilities/HierarchyIntegrator.cpp
@@ -554,6 +554,16 @@ HierarchyIntegrator::registerLoadBalancer(Pointer<LoadBalancer<NDIM> > load_bala
     return;
 } // registerLoadBalancer
 
+int
+HierarchyIntegrator::getWorkloadDataIndex() const
+{
+    if (d_parent_integrator)
+    {
+        return d_parent_integrator->getWorkloadDataIndex();
+    }
+    return d_workload_idx;
+}
+
 void
 HierarchyIntegrator::registerVisItDataWriter(Pointer<VisItDataWriter<NDIM> > visit_writer)
 {
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -78,6 +78,7 @@
 #include "ibtk/RobinPhysBdryPatchStrategy.h"
 #include "ibtk/ibtk_utilities.h"
 #include "ibtk/libmesh_utilities.h"
+#include "ibtk/BoxPartitioner.h"
 #include "libmesh/boundary_info.h"
 #include "libmesh/compare_types.h"
 #include "libmesh/dense_vector.h"
@@ -1785,7 +1786,19 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
     // if we are not initialized then there is nothing to do
     if (d_is_initialized)
     {
-        // This is where we will repartition in a future commit
+        // At this point in the code SAMRAI has already redistributed the
+        // patches (usually by taking into account the number of IB points on
+        // each patch). Here is the other half: we inform libMesh of the
+        // updated partitioning so that libMesh Elems and Nodes are on the
+        // same processor as the relevant SAMRAI patch.
+        for (unsigned int part = 0; part < d_num_parts; ++part)
+        {
+            EquationSystems& equation_systems = *d_fe_data_managers[part]->getEquationSystems();
+            MeshBase& mesh = equation_systems.get_mesh();
+            BoxPartitioner partitioner(*d_hierarchy,
+                                       equation_systems.get_system(COORDS_SYSTEM_NAME));
+            partitioner.repartition(mesh);
+        }
 
         reinitializeFEData();
         for (unsigned int part = 0; part < d_num_parts; ++part)
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1689,11 +1689,6 @@ IBFEMethod::registerLoadBalancer(Pointer<LoadBalancer<NDIM> > load_balancer, int
     TBOX_ASSERT(load_balancer);
     d_load_balancer = load_balancer;
     d_workload_idx = workload_data_idx;
-
-    for (unsigned int part = 0; part < d_num_parts; ++part)
-    {
-        d_fe_data_managers[part]->registerLoadBalancer(load_balancer, workload_data_idx);
-    }
     return;
 } // registerLoadBalancer
 
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1781,18 +1781,26 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
     // if we are not initialized then there is nothing to do
     if (d_is_initialized)
     {
+        // Checking the workload index like this breaks incapsulation, but
+        // since this is inside the library and not user code its not so bad
+        const bool workload_is_setup = d_ib_solver
+            ? d_ib_solver->getWorkloadDataIndex() != IBTK::invalid_index : false;
         // At this point in the code SAMRAI has already redistributed the
         // patches (usually by taking into account the number of IB points on
         // each patch). Here is the other half: we inform libMesh of the
         // updated partitioning so that libMesh Elems and Nodes are on the
         // same processor as the relevant SAMRAI patch.
-        for (unsigned int part = 0; part < d_num_parts; ++part)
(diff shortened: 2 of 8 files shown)


drwells:
For comparison: if we comment out the repartitioning and change the estimated work per quadrature point to zero then the workload estimates no longer make sense, but we can see that the parallelization is quite poor by looking at the number of quadrature points per processor:


```
quadrature points on processor 0  = 30048 (pid is 47732)
quadrature points on processor 1  = 94096 (pid is 47733)
quadrature points on processor 2  = 65500 (pid is 47734)
quadrature points on processor 3  = 0 (pid is 47737)
quadrature points on processor 4  = 0 (pid is 47739)
quadrature points on processor 5  = 10835 (pid is 47740)
quadrature points on processor 6  = 114808 (pid is 47741)
quadrature points on processor 7  = 126025 (pid is 47742)
quadrature points on processor 8  = 139401 (pid is 47743)
quadrature points on processor 9  = 137346 (pid is 47744)
quadrature points on processor 10 = 123094 (pid is 47748)
quadrature points on processor 11 = 124629 (pid is 47749)
quadrature points on processor 12 = 48246 (pid is 47750)
quadrature points on processor 13 = 8546 (pid is 47751)
quadrature points on processor 14 = 0 (pid is 47752)
quadrature points on processor 15 = 0 (pid is 47753)
```

i.e., processors 3, 4, 14, and 15 have nothing to do, and most of the work is on processors 7, 8, 9, 10, and 11. We may as well be only running on five processors instead of sixteen.

examples/IBFE/explicit/ex4/input2d
```diff
@@ -4,14 +4,14 @@ RHO = 1.0
...
 // grid spacing parameters
-MAX_LEVELS = 1                                      // maximum number of levels in locally refined grid
-REF_RATIO  = 2                                      // refinement ratio between levels
-N = 64                                              // actual    number of grid cells on coarsest grid level
-NFINEST = (REF_RATIO^(MAX_LEVELS - 1))*N            // effective number of grid cells on finest   grid level
-DX0 = L/N                                           // mesh width on coarsest grid level
-DX  = L/NFINEST                                     // mesh width on finest   grid level
-MFAC = 2.0                                          // ratio of Lagrangian mesh width to Cartesian mesh width
-ELEM_TYPE = "TRI6"                                  // type of element to use for structure discretization
+MAX_LEVELS = 3                                     // maximum number of levels in locally refined grid
```
boyceg: Let's not change the "default" input files just yet --- instead add in a new file, `input2d.partitioner`?

ibtk/src/utilities/HierarchyIntegrator.cpp
```diff
@@ -558,6 +558,11 @@ void
 HierarchyIntegrator::registerVisItDataWriter(Pointer<VisItDataWriter<NDIM> > visit_writer)
 {
     d_visit_writer = visit_writer;
+    if (d_workload_idx != IBTK::invalid_index)
```
boyceg: Let's include a run-time option to turn this on/off. Seems fine to me to default to "on".

src/IB/IBFEMethod.cpp
```diff
@@ -1785,7 +1781,19 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
     // if we are not initialized then there is nothing to do
     if (d_is_initialized)
     {
-        // This is where we will repartition in a future commit
+        // At this point in the code SAMRAI has already redistributed the
```
boyceg: Should we add a runtime option to turn on/off dynamic repartitioning of libMesh data?

drwells: Absolutely. I will move the checklist from #396 to here (which included such items).

examples/IBFE/explicit/ex4/input2d
drwells: I only changed this for testing purposes (e.g., printing workloads). I'll get rid of the changes once that is done.

drwells:
Interestingly, adding more patches improves load balancing but wrecks the performance (at least in 2D) of the fluid solver. Lowering `max_workload_factor` to 0.0625 (from 0.25) doubled the amount of time spent in the fluid solver.

boyceg:
I suspect that this will be less of an issue in 3D, but it would be good to have some tests to find out.

drwells:
The remaining items are:
- [ ] Get rid of the changes to the example step.
- [ ] Clean up the logging
- [ ] Make the (graphical) output of workload estimates conditional
- [ ] Fix the issue where logging info doesn't seem to show up

drwells:
Is getting the patch data index out of the variable database actually possible? I am stuck on the current problem:
1. I want to register a variable with the visit data writer in user code by looking it up in the variable database.
2. the visit writer needs the variable name and patch data index.
3. The method that returns the patch data index, `getVariableId`, is protected.

I am going to solve the original problem by adding code to return the workload patch data index from the parent hierarchy integrator, but it seems odd that registering data for output requires access to protected data.

boyceg:
To get the patch data index, you need to have the Variable / VariableContext pair. Your solution sounds like the easiest one to me.

drwells:
I think this is now done, but I have run out of time today to test the new 'when to use the partitioner' logic. I'll verify that everything works at some point over the next few days.

drwells:
I'm now convinced that the database settings are correctly propagated, so this is done.

src/IB/IBFEMethod.cpp
```diff
@@ -1861,6 +1877,7 @@ IBFEMethod::putToDatabase(Pointer<Database> db)
     db->putBool("d_split_tangential_force", d_split_tangential_force);
     db->putBool("d_use_jump_conditions", d_use_jump_conditions);
     db->putBool("d_use_consistent_mass_matrix", d_use_consistent_mass_matrix);
+    db->putString("d_libmesh_partitioner_type", enum_to_string<LibmeshPartitionerType>(d_libmesh_partitioner_type));
```
drwells: I forgot about force pushing the change that added this. This is the only change I made since Friday.

boyceg:
Shall we merge?

drwells:
I'm doing some last-minute checks. I'll merge when they are done.

drwells:
This changes some results at the level of roundoff, but I think this is within the bounds of the linear solver tolerance. Lets merge!

boyceg:
What causes the changes? There should be no effect unless the load balance is explicitly enabled?

drwells:
I checked with the load balancer explicitly enabled vs. not enabled (i.e., master).

PR #481, opened 2019-02-17
Improve the performance of SystemDofMapCache.

drwells:
Alternative to #468. 

This improves performance with a few changes:
1. Combining the calls to the cache so that only one lookup is done for each system on each cell (instead of each variable).
2. Avoid copying the dofs unless we need to apply constraints.

This lowers the amount of time we spend looking up dof indices by about 75%. Its essentially the same technique as #468 but just returns all variables at once, which we should do anyway; we always need, e.g., all NDIM sets of X dofs.

I left the old lookup function in place so any user code that uses the lookup by variable and copying should still work.

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -146,18 +146,28 @@ public:
                     std::vector<unsigned int>& dof_indices,
                     const unsigned int var = 0)
         {
-            const libMesh::dof_id_type elem_id = elem->id();
-            std::vector<std::vector<unsigned int> >& elem_dof_indices = d_dof_cache[elem_id];
-            if (elem_dof_indices.size() <= var)
-            {
-                elem_dof_indices.resize(var + 1);
-            }
-            if (elem_dof_indices[var].empty())
-            {
-                d_dof_map.dof_indices(elem, elem_dof_indices[var], var);
-            }
-            dof_indices = elem_dof_indices[var];
+            dof_indices = this->dof_indices(elem)[var];
             return;
+        }
+
+        /*!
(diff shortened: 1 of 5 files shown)


drwells:
I am marking as WIP because I still need to read this through again and run the test suite. One nice thing about this patch is that, if we look up the wrong dofs, things go wrong in a very obvious way.

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -160,6 +160,26 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
...
+        /*!
+         * Alternative indexing operation: retrieve all dof indices of all
+         * variables in the given system at once by reference.
+         */
+        inline
+        const std::vector<std::vector<libMesh::dof_id_type>>&
+        dof_indices(const libMesh::Elem* const elem)
+        {
+            std::vector<std::vector<libMesh::dof_id_type>>& elem_dof_indices = d_dof_cache[elem->id()];
+            if (elem_dof_indices.size() == 0)
```
boyceg: Yes, definitely cleaner than my approach.

However, I think that the size can be between 0 and the number of variables thanks to the original interface. I think that case also needs to change to look up all at once for this to work robustly.

boyceg:
Looks good to me!

ibtk/include/ibtk/FEDataManager.h
drwells: Good point: that needs to be fixed for both methods to coexist.

drwells:
I fixed the old method by simply making it call the new one.

This passes the test suite for me locally so lets merge once Jenkins is happy.

PR #483, opened 2019-02-19
Workload fixes

drwells:
I found the bug last night when I started doing some profiling.

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1707,7 +1707,7 @@ IBFEMethod::addWorkloadEstimate(Pointer<PatchHierarchy<NDIM> > hierarchy, const 
 
         std::vector<double> workload_per_processor(n_processes);
         HierarchyCellDataOpsReal<NDIM, double> hier_cc_data_ops(hierarchy);
-        workload_per_processor[current_rank] = hier_cc_data_ops.L1Norm(d_workload_idx, -1, true);
+        workload_per_processor[current_rank] = hier_cc_data_ops.L1Norm(workload_data_idx, IBTK::invalid_index, true);
 
         const auto right_padding = std::size_t(std::log10(n_processes)) + 1;
 
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1870,6 +1870,7 @@ IBFEMethod::putToDatabase(Pointer<Database> db)
     db->putBool("d_use_jump_conditions", d_use_jump_conditions);
     db->putBool("d_use_consistent_mass_matrix", d_use_consistent_mass_matrix);
     db->putString("d_libmesh_partitioner_type", enum_to_string<LibmeshPartitionerType>(d_libmesh_partitioner_type));
+    db->putDouble("workload_quad_point_weight", d_default_workload_spec.q_point_weight);
     return;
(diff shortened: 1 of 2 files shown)


src/IB/IBFEMethod.cpp
```diff
@@ -1870,6 +1870,7 @@ IBFEMethod::putToDatabase(Pointer<Database> db)
     db->putBool("d_use_jump_conditions", d_use_jump_conditions);
     db->putBool("d_use_consistent_mass_matrix", d_use_consistent_mass_matrix);
     db->putString("d_libmesh_partitioner_type", enum_to_string<LibmeshPartitionerType>(d_libmesh_partitioner_type));
+    db->putDouble("workload_quad_point_weight", d_default_workload_spec.q_point_weight);
```
boyceg: Do we get this value from the restart database? If not, we don't need to store it hre.

drwells: This value is read in `getFromInput` so I figured we need it here too. Is this correct? I have to admit that I don't know the difference between the variable and restart databases.

boyceg: We don't really have any system for keeping track of which variables are stateful and should be stashed in the restart database, and which are stateful but should be overwritten by new input at restart. (You could imaging wanting to change some settings when you restart, but that there are other settings for which changes cannot be accommodated.)

boyceg: Any variable whose value is determined from the input database upon class creation does not need to go into the restart database. Variables that describe the fundamental state of the class, or that should not change upon restart, should go into the restart database and be set from the values stored in that database. Does that make sense?

boyceg: (Of course, if you have a variable whose value is set from the input database and then reset from the restart database --- its value needs to be stored in the restart database.)

drwells: > Any variable whose value is determined from the input database upon class creation does not need to go into the restart database.

Sounds like I don't need to use it here, then.

drwells:
I removed the putToDatabase code.

drwells:
@boyceg is this okay? I would like to merge it so that we can get the 0.4 release out sooner.

src/IB/IBFEMethod.cpp
```diff
@@ -1707,7 +1707,7 @@ IBFEMethod::addWorkloadEstimate(Pointer<PatchHierarchy<NDIM> > hierarchy, const
 
         std::vector<double> workload_per_processor(n_processes);
         HierarchyCellDataOpsReal<NDIM, double> hier_cc_data_ops(hierarchy);
-        workload_per_processor[current_rank] = hier_cc_data_ops.L1Norm(d_workload_idx, -1, true);
+        workload_per_processor[current_rank] = hier_cc_data_ops.L1Norm(workload_data_idx, -1, true);
```
boyceg: `IBTK::invalid_index` instead of `-1`?
```suggestion
        workload_per_processor[current_rank] = hier_cc_data_ops.L1Norm(workload_data_idx, invalid_index, true);
```

drwells: Since this is not an IBTK object let me explicitly write `IBTK::invalid_index`, but yes.

drwells: I force pushed a fix.

PR #485, opened 2019-02-21
Clean up the libMesh detection scripts a bit.

### LINKED ISSUE #435: Error out earlier if libMesh was compiled without PETSc
@xxjaehoxx hit this earlier and, while `configure` does eventually fail, the error message we get is not great. We should quit `configure` earlier if `libMesh` is compiled without PETSc.

drwells:
This adds some comments that make it easier to read the m4 and adds another check to verify that libMesh was compiled with support for PETSc. It also fixes a silly error where I state that we require libMesh 1.1 but the code checks for libMesh 1.2.

Fixes #435.

### DIFF
--- a/configure
+++ b/configure
@@ -23958,7 +23958,7 @@ $as_echo_n "checking for libMesh version 1.1.0 or newer... " >&6; }
 
 #include <libmesh/libmesh_config.h>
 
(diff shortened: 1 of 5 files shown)


boyceg:
You don't like my comment-free m4?

drwells:
My main issue was that I cannot get good syntax highlighting for those files, so all the macros blur together in my eyes and I have a lot of trouble reading where one starts and another ends :)

PR #494, opened 2019-03-08
Overhaul test suite

### LINKED ISSUE #372: Overhaul the test suite
I had a nice chat with @knepley earlier today and this is something we discussed. We presently have a couple dozen (roughly) tests that check norms to verify that the code works. These tests rely on the gtest framework. I personally run the test suite frequently and use it to check my own work; even this small test suite is extremely valuable. However, this framework has several shortcomings:
1. We don't have a dedicated test runner program: `make -j32 --keep-going --ignore-errors --output-sync` works but is a bit crude (it cannot display output until it finishes). Tests that fail to compile or fail to execute (i.e., experience an exception or system error like a segfault) don't fail; they just don't print `OK` at the end. The previous make command cannot capture this since we have to ignore *all* errors or *no* errors with make. More succinctly: something else should invoke make on each test and record what happens instead of running it recursively.
2. We do not have a way to write unit tests: everything is presently based around compiling the examples in a certain way. It would be nice to check things like the new MPI class in a simple, isolated context (we don't need to run it inside the Eulerian solver to check the max function).
3. We cannot check things that are not easily expressed without assertions: i.e., we cannot verify graphical output.
4. All results must be expressed in the input file itself, which is inflexible.

A comparison: deal.II has a dedicated `tests/` directory containing about 5,000 distinct tests. One can run any subset (specified with a regular expression) via ctest. We use numdiff to verify output with some rounding tolerance. Importantly, ctest is aware of the build system (it is managed by cmake) and can track dependencies of tests to recompile and relink as needed. Additionally, we can run the same test executable with multiple configurations, which is very handy (example: one test can be run several times, each with a different number of MPI processes, to verify that some parallel algorithms).

I propose that we package up the current tests as-is and move them over to a dedicated `tests/` directory to start. We can continue using `make gtest` for the time being as we migrate to something more sophisticated. How does this sound?

drwells:
I have been working on this off and on this year and finally got it to work, albeit with a bunch of hardcoded values and undocumented dependencies. Fixes #372. Since we have been talking about tests recently I figured that it might be helpful to put this up (even though it is far from done) to show what I intend to do with the test suite.

@abarret This replaces the 'parallelize the test suite' branch that I wrote in the fall, though (of course) this branch isn't that useful since it only has four real tests (and a hello world test).

I would like to do some more serious refactoring while I improve the ghost data communication in IBFE/FEDataManager, but it is difficult to do this sort of work without a test suite for finding my mistakes.

This is the first in a long set of commits that will overhaul the way we test IBAMR. The current test setup has some serious flaws:
1. We have to write GTEST_ASSERT everywhere, which is annoying.
2. If a test fails to compile then we get a silent test failure (nothing is printed to the screen via gtest).
3. If a test crashes then we get another silent test failure (same as above).
4. We cannot run multiple tests concurrently.
5. The current tests are all based on the examples: we should also write unit tests. The current coupling of tests to examples precludes this.
6. Tests cannot utilize MPI.

The contributed test runner, attest, solves all of these problems by simply requiring that the build system compile all tests: the test runner executes the tests and checks their log files against known
output with numdiff. attest can correctly schedule MPI jobs so that the given number of cores is always used: e.g., if we have ten cores available via -j10, attest will run a 7-core test and a 3-core test. Should the 3-core test finish first it will begin, e.g., a 2-core test and a 1-core test.

The test runner is a python script. It is based pretty heavily on how ctest works (particularly how we use it in deal.II), but has the advantage that it can schedule MPI tests more efficiently since it doesn't compile anything. I am not aware of any tool that can handle both MPI tests and do numdiff-based output checking, both of which we really need.

This approach fits how we write examples and applications pretty well: a test is simply a source file, an input file, and an output file.


drwells:
Regarding the discussion in #372: the main contribution of this patch is setting up a `tests/` directory of things that need to be compiled by the build system (i.e., this will work with either autotools or buildsystem) and the actual test runner. Using PETSc-style tests would also work: I have simply included the equivalent information in the file names themselves.

boyceg:
This looks good. One comment is that it looks like you are putting all of the tests for a given "module" in the same directory. I worry a bit about concurrent tests stepping on each others' toes (e.g. viz or log files).

The other thing that I would like to do is to cut down on the huge amount of boilerplate code that shows up in any particular test or example. It might be good to try to do that before setting up the tests. What do you think?

drwells:
> This looks good. One comment is that it looks like you are putting all of the tests for a given "module" in the same directory. I worry a bit about concurrent tests stepping on each others' toes (e.g. viz or log files).

The tests are executed in individually unique temporary directories for exactly this reason: any extra stuff will be deleted once `numdiff` runs.

> The other thing that I would like to do is to cut down on the huge amount of boilerplate code that shows up in any particular test or example. It might be good to try to do that before setting up the tests. What do you think?

It would be nice to have some general object that sets up all the SAMRAI stuff that we use every single time but I haven't really spent much time thinking about it. I did try to delete all of the IO code since I turned that off.

tests/IBFE/interpolate_velocity_01.c.output
```diff
@@ -0,0 +1,12 @@
...
+IBFEMethod: mesh part 0 is using FIRST order LAGRANGE finite elements.
+
+IBHierarchyIntegrator::initializePatchHierarchy(): tag_buffer = 1 3
+FEDataManager::buildGhostedSolutionVector(): building ghosted solution vector for system: IB coordinates system
+FEDataManager::buildGhostedSolutionVector(): constructing ghost DOF index list for system: IB coordinates system
+FEDataManager::buildGhostedSolutionVector(): building ghosted solution vector for system: IB force system
+FEDataManager::buildGhostedSolutionVector(): constructing ghost DOF index list for system: IB force system
+FEDataManager::buildL2ProjectionSolver(): building L2 projection solver for system: IB velocity system
+max vertex distance: 0.00892356
+max norm errors: 2.754841494412296754e-05   3.9301631085009169908e-05
```
drwells: These max norm errors are correct up to ten digits which (while we could improve this) is much more reasonable.

tests/IBFE/interpolate_velocity_01.a.mpirun=4.output
```diff
@@ -0,0 +1,12 @@
...
+IBFEMethod: mesh part 0 is using FIRST order LAGRANGE finite elements.
+
+IBHierarchyIntegrator::initializePatchHierarchy(): tag_buffer = 1 3
+FEDataManager::buildGhostedSolutionVector(): building ghosted solution vector for system: IB coordinates system
+FEDataManager::buildGhostedSolutionVector(): constructing ghost DOF index list for system: IB coordinates system
+FEDataManager::buildGhostedSolutionVector(): building ghosted solution vector for system: IB force system
+FEDataManager::buildGhostedSolutionVector(): constructing ghost DOF index list for system: IB force system
+FEDataManager::buildL2ProjectionSolver(): building L2 projection solver for system: IB velocity system
+max vertex distance: 0.0258831
+max norm errors: 0.0001942783290962424303   0.00030017549307626012478
```
drwells: These values don't match the serial case, though both are second order.

tests/IBFE/spread_01.cpp
```diff
@@ -0,0 +1,336 @@
...
+        auto &equation_systems = *fe_data_manager.getEquationSystems();
+        auto &force_system = equation_systems.get_system(IBFEMethod::FORCE_SYSTEM_NAME);
+        auto &half_f_vector = dynamic_cast<libMesh::PetscVector<double> &>(*force_system.current_local_solution);
+        for (unsigned int i = half_f_vector.first_local_index(); i <= half_f_vector.last_local_index(); ++i)
+        {
+            half_f_vector.set(i, i % 10);
+        }
+        half_f_vector.close();
+
+        ib_method_ops->spreadForce(f_ghost_idx, nullptr, {}, time_integrator->getIntegratorTime() + dt/2);
```
drwells: Its not clear to me how to compute convergence or errors for a spread force, so I just print out the values in this test.

drwells:
This is now ready.

drwells:
The only thing that I have not fully resolved is what we should do with the test programs that previously inhabited `tests/`: for now I moved these into a new folder `tests/unfinished-tests/` until we can merge them with the test suite proper.

boyceg:
Those old “tests” really were programs for testing out functionality. They were not tests in the sense of “unit tests”. Possibly it is more accurate to describe them as “experiments”.

drwells:
Ah. Well, they look good; is there somewhere else they could go? I am not sure what to do with them.

boyceg:
They could become "examples"...?

drwells:
Good idea. Let me open an issue to convert these things to example programs.

boyceg:
> Looks good.
> 
> Should we get rid of the tests in `examples` and merge the source code back into a single `main.cpp` file for each example? I think that basically all that is needed is to change `run_example()` to `main()` and get rid of the extra drivers for examples and tests.

(We could leave the source code for the example in `example.cpp`.)

drwells:
IIRC the majority of gtest tests only verify that the code executes successfully. I think we should skip those in favor of writing more unit tests.

Thats not to say that those tests aren't useful; currently not all of the examples can actually exit successfully with the given input files. However, we should focus on writing tests that verify output or convergence rates.

boyceg:
> IIRC the majority of gtest tests only verify that the code executes successfully. I think we should skip those in favor of writing more unit tests.
> 
> Thats not to say that those tests aren't useful; currently not all of the examples can actually exit successfully with the given input files. However, we should focus on writing tests that verify output or convergence rates.

I agree with you. I am asking if we should just strip out the gtest tests in `examples`. This will make the example code simpler. In the gtests in `examples`, we are trying to re-use the same basic code to provide both "examples" and "tests". (Full disclosure: this was my idea!) This makes the example code somewhat more complicated. More importantly, it does not seem to provide a good foundation for starting a standalone application.

My suggestion is to convert the examples to be just examples (either in this PR or in a separate one). This will result in a lot of changed files, but should be easy to automate.

drwells:
Oh, I see. Yes. I think its okay to have the same code in an example and a test (and not use `#include`) in some place.

drwells:
@boyceg AFAICT you have approved this so I will merge. I just rebased against master and checked one last time that everything is in order.

PR #495, opened 2019-03-15
Cache interpolated positions.

drwells:
We spend about 5% of our total run time in the heart model calling this function. This commit reduces the number of calls per cell from five to two (i.e., instead of calling it three times per cell in `interpWeighted` we call it once and instead of twice in `spread` now we just call it one once).

Even though the percent change is small, since we call this in a place where the load balancer has trouble (i.e., we spend a lot of time waiting for other processors to finish interpolating data) so it should also slightly decrease the amount of time we spend waiting for other processors to finish those two functions.


drwells:
I have been staring at profiling results all day and I figured I should at least fix something easy.

Issue #424, opened 2019-01-09
The 0.4 Release


drwells:
The current situation with C++ version support causes confusing compilation in 0.3.0 that were just fixed for 1.0 (see #423). I believe that we can solve this with the same approach by making C++11 support optional, but the default: i.e., let libMesh detect C++11 support if it is available (and do not instruct users to provide any flags like `--disable-cxx11` or `--disable-unique-ptr`). This should be compatible with versions of libMesh compiled with those flags anyway: we never assume (on 0.3.0) that `libMesh::UniquePtr` is `std::unique_ptr`, so compiling IBAMR with a newer language standard should work (I'll check this).

Can anyone else think of fixes that should go into this release?

boyceg:
Perhaps all of the performance fixes that we are currently throwing into master? :-D

drwells:
Honestly, I think thats the right thing to do: we aren't close to pulling the trigger on the SAMRAI 3 transition and we have made a lot of progress on improving IBAMR since the 0.3 release. It would make sense to do another major release that requires C++11 with what we have (and the rest of #436) and any other imminent fixes. We should aim to get that out in February.

drwells:
Given that #427 exists I will try to find time to put out the new release this week once the partitioner code is merged in.

drwells:
Lets get this done this week since people keep having C++14 issues. Here is the checklist:
- [ ] Finish partitioning (i.e., review and merge #479)
- [ ] Use FECache in FEDataInterpolation (see #482)
- [ ] Improve the performance of SystemDofMapCache (i.e., review and merge #481)

drwells:
@jabrown893 Will you be around this afternoon? It would help me out if you could download the release candidate (which is not yet released) and check that it compiles and runs on a mac.

jabrown893:
@drwells I will be around kind of intermittently. I have an appt. to give blood at 2:45.

drwells:
@xxjaehoxx To the best of my knowledge we were not able to get things working with commits predating 0.3, so I think we should go ahead with the release without this patch. We can always release 0.6 later if we figure it out.

xxjaehoxx:
This sounds good to me!

drwells:
There is still a lot to do with #418, so lets wrap things up as-is.

drwells:
I created the branch: see https://github.com/IBAMR/IBAMR/tree/IBAMR-0.4

drwells:
[...]

PR #500, opened 2019-03-22
Update deprecated lib mesh

drwells:
We use a lot of deprecated libMesh functions. I went through and got rid of everything that is deprecated as of 1.4.0 where possible: in one case I had to use the preprocessor to keep compatibility with 1.1.0. This stops a lot of annoying warnings.

Since we require 1.1.0 or newer this shouldn't break anything.

Candidate for inclusion in 0.4; see #424.

### DIFF
--- a/src/IB/IBFEInstrumentPanel.cpp
+++ b/src/IB/IBFEInstrumentPanel.cpp
@@ -338,7 +338,17 @@ IBFEInstrumentPanel::initializeHierarchyIndependentData(IBFEMethod* ib_method_op
     std::vector<std::set<dof_id_type> > temp_node_dof_ID_sets;
     std::vector<std::vector<libMesh::Point> > temp_nodes;
     std::vector<libMesh::Point> meter_centroids;
+    // new API in 1.4.0
+#if 1 <= LIBMESH_MAJOR_VERSION && 4 <= LIBMESH_MINOR_VERSION
+    const std::vector<std::tuple<dof_id_type, boundary_id_type> > node_list = boundary_info.build_node_list();
+    for (const std::tuple<dof_id_type, boundary_id_type>& pair : node_list)
+    {
+        nodes.push_back(std::get<0>(pair));
+        bcs.push_back(std::get<1>(pair));
+    }
+#else
     boundary_info.build_node_list(nodes, bcs);
+#endif
 
     // check to make sure there are node sets to work with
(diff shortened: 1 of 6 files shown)


drwells:
I can't reproduce the build failure with any version of libMesh (1.2, 1.3, 1.4, etc). @abarret what version of libMesh is used with the failing build?

abarret:
Looks like 1.3.0

drwells:
Weird. Let me take another look.

drwells:
Okay, I got mixed up when I read multiple versions of this header: the function was deprecated in 1.4, not 1.3, so that should fix it.

drwells:
@AminKolah Would you be willing to review this? Boyce mentioned that you are a libMesh expert :)

AminKolah:
Sure @drwells , I'll review them tonight....btw, I wasn't actually aware of some of these new changes in the latest release of libmesh. I've been compiling my branch with an older 1.2.1 version but I'd like to work on merging with master soon (hopefully this weekend). So I'm sure reviewing this patch will be useful.

src/IB/IBFEInstrumentPanel.cpp
```diff
@@ -338,7 +338,17 @@ IBFEInstrumentPanel::initializeHierarchyIndependentData(IBFEMethod* ib_method_op
     std::vector<std::set<dof_id_type> > temp_node_dof_ID_sets;
     std::vector<std::vector<libMesh::Point> > temp_nodes;
     std::vector<libMesh::Point> meter_centroids;
+    // new API in 1.4.0
```
boyceg: Should we likewise declare our support for libMesh < 1.4 to be deprecated?

boyceg:
Looks fine. I am surprised that these are the only deprecated libMesh functions that we are using!

src/IB/IBFEInstrumentPanel.cpp
drwells: I think its worth supporting a few old versions of libMesh: we only have checks like this in two places at the moment.

drwells: i.e., I don't think we should declare it deprecated in some way right now since its very easy to support.

drwells:
Well, its good that the libMesh guys aren't reinventing the wheel constantly :)

PR #502, opened 2019-03-22
Update deprecated lib mesh 0.4

drwells:
Same as #500, but now for 0.4.


drwells:
Since this is a release branch (and no one should be relying on it) I am going to go ahead and merge this and the other 0.4 patches.

PR #503, opened 2019-03-22
Fix a signed integer comparison warning.

drwells:
Same as #501, but for 0.4.


drwells:
I am merging this for the same reason as #502.

PR #505, opened 2019-03-22
Improve max elem edge length

drwells:
I wrote this patch while testing out some new parts of the test runner. In fact, #494 already includes a test for this change to verify that the new algorithm doesn't change the computed edge length.

This function takes up roughly 4% of our total run time in the heart model: these changes lower, in valgrind, the total number of cycles spent in this function from 262 billion to 30 billion (and total time down to 0.46%).

### DIFF
--- a/ibtk/include/ibtk/libmesh_utilities.h
+++ b/ibtk/include/ibtk/libmesh_utilities.h
@@ -975,6 +975,104 @@ get_nodal_dof_indices(const libMesh::DofMap &dof_map,
 #endif
 }
 
+
+/**
+ * Return the maximum edge length of a given element with mapped nodes.
+ *
+ * @p elem The given libMesh element.
+ * @p X_node Mapped coordinates of the nodes.
+ */
+template <class MultiArray>
+inline double
+get_max_edge_length(const libMesh::Elem* const elem, const MultiArray& X_node)
+{
+#ifndef NDEBUG
+    TBOX_ASSERT(elem->n_nodes() == X_node.shape()[0]);
+    TBOX_ASSERT(elem->dim() == X_node.shape()[1]);
+#endif
+    const libMesh::ElemType elem_type = elem->type();
+
+    double max_edge_length_2 = 0.0;
+    // TODO: implement additional element-specific versions as needed.
+    switch (elem_type)
+    {
+    case libMesh::TET4:
+    {
+        // all nodes are connected to all other nodes
+        constexpr unsigned int edge_pairs[6][2] = { { 0, 1 }, { 0, 2 }, { 0, 3 }, { 1, 2 }, { 1, 3 }, { 2, 3 } };
+        for (const unsigned int(&pair)[2] : edge_pairs)
+        {
+            const unsigned int n1 = pair[0];
+            const unsigned int n2 = pair[1];
+            double diff_sq = 0.0;
+            for (unsigned int d = 0; d < 3; ++d) // TET4 implies 3D
+                diff_sq += (X_node[n1][d] - X_node[n2][d]) * (X_node[n1][d] - X_node[n2][d]);
+            max_edge_length_2 = std::max(max_edge_length_2, diff_sq);
+        }
+        break;
+    }
+    case libMesh::HEX8:
+    {
+        // see the connectivity diagram in cell_hex8.h to see why these are the edges
+        constexpr unsigned int edge_pairs[12][2] = { { 0, 1 }, { 0, 3 }, { 0, 4 }, { 1, 2 }, { 1, 5 }, { 2, 3 },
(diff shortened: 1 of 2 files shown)


drwells:
While this is fine as-is, we may want to rename this function (and change the semantics of the function this replaces) so that things work consistently with higher-order elements. It is not clear to me whether it makes more sense to compute a length scale based on the size of the element (which is done here) or on the maximum distance between finite element nodes. These are, of course, the same for low-order elements but are not for, e.g., `Tet10` or `Hex27`.

drwells:
@boyceg and I just talked about this and came to the conclusion that we, for this particular part of IBAMR, need the length of the edge. We can extend this to higher-order elements by computing the length as the sum of multiple line segments.

drwells:
Clarification: we don't have the higher-order element implementation yet (that will actually change some computed lengths), so the proposed implementation is identical to the original implementation in all cases.

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -975,6 +975,104 @@ get_nodal_dof_indices(const libMesh::DofMap &dof_map,
...
+    const libMesh::ElemType elem_type = elem->type();
+
+    double max_edge_length_2 = 0.0;
+    // TODO: implement additional element-specific versions as needed.
+    switch (elem_type)
+    {
+    case libMesh::TET4:
+    {
+        // all nodes are connected to all other nodes
+        constexpr unsigned int edge_pairs[6][2] = { { 0, 1 }, { 0, 2 }, { 0, 3 }, { 1, 2 }, { 1, 3 }, { 2, 3 } };
```
boyceg: Is there any way to look this up from `libMesh` rather than writing it out by hand?

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -975,6 +975,104 @@ get_nodal_dof_indices(const libMesh::DofMap &dof_map,
...
+            for (unsigned int d = 0; d < 3; ++d) // TET4 implies 3D
+                diff_sq += (X_node[n1][d] - X_node[n2][d]) * (X_node[n1][d] - X_node[n2][d]);
+            max_edge_length_2 = std::max(max_edge_length_2, diff_sq);
+        }
+        break;
+    }
+    case libMesh::HEX8:
+    {
+        // see the connectivity diagram in cell_hex8.h to see why these are the edges
+        constexpr unsigned int edge_pairs[12][2] = { { 0, 1 }, { 0, 3 }, { 0, 4 }, { 1, 2 }, { 1, 5 }, { 2, 3 },
```
boyceg: Same story.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -278,13 +234,20 @@ getQuadratureKey(const QuadratureType quad_type,
...
     TBOX_ASSERT(elem->p_level() == 0); // higher levels are not implemented
+#endif
     if (use_adaptive_quadrature)
     {
-        const double hmax = get_elem_hmax(elem, X_node);
-        const int min_pts = elem->default_order() == FIRST ? 2 : 3;
-        const int npts = std::max(min_pts,
-                                  int(std::ceil(point_density * hmax / dx_min)));
+        const double hmax = get_max_edge_length(elem, X_node);
+        int npts = int(std::ceil(point_density * hmax / dx_min));
```
boyceg: You don't like my use of the ternary operator?

boyceg:
Looks good. I wonder if we can get the basic edge indexing info from libMesh directly?

ibtk/include/ibtk/libmesh_utilities.h
drwells: AFAICT no: this table is only available in the source file.

ibtk/src/lagrangian/FEDataManager.cpp
drwells: I don't mind :) I changed this bit to avoid one more virtual function call (to `Elem::default_order)`.

drwells:
Since we cannot get the indexing from libMesh directly (its buried in their source files) I think this is okay to merge. If there are no further comments I'll merge later today.

Inlining the indexing ourselves isn't great (but is necessary for the optimizer), but libMesh has used the same node numbering since at least 2011 so its not likely to break.

boyceg:
Unit test? :-)

drwells:
Of course! See
https://github.com/IBAMR/IBAMR/blob/083771074652a96cac29e868aba7548c26a93160/tests/IBTK/elem_hmax_01.cpp
and
https://github.com/IBAMR/IBAMR/blob/083771074652a96cac29e868aba7548c26a93160/tests/IBTK/elem_hmax_02.cpp

drwells:
@cpuelz FYI: between this patch and #507 the heart model runs about 5% faster in serial :tada:

PR #507, opened 2019-03-25
Improve performance of FEDataManager by unrolling.

drwells:
I spent some time looking at profiling data on Saturday and I noticed that setting up this particular load vector is a huge hotspot and we can benefit from unrolling loops quite a bit. This is nearly the same technique we use in deal.II for matrix-free techniques (though there we template on the number of quadrature points, which doesn't make sense for IB computations).

It looks a bit complicated, but all we do is use templates and switch statements to dispatch to a version of the loop where both the number of variables (usually NDIM, but sometimes 1) and number of basis functions are known at compile time. There is also a fallback option where we switch to run-time computation of loop bounds for the `default` case of the switch statements.

Knowing the loop bounds helps a lot for tetrahedra or other low-order elements where the number of basis functions is small: knowing the inner loop bound is always helpful.

These percentages all come from running the heart model in callgrind with one processor. This patch lowers the total number of instructions by 2.5%.  About 4.1% of total instructions were spent in interpWeighted itself (or in inlined functions): this patch lowers that to 1.6% with the new integrate_cell_rhs function taking up 0.7% of total instructions.

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -861,10 +861,12 @@ FEDataManager::spread(const int f_data_idx,
                 {
                     for (unsigned int qp = 0; qp < n_qp; ++qp)
                     {
-                        for (unsigned int i = 0; i < n_vars; ++i)
-                        {
-                            F_JxW_qp[n_vars * (qp_offset + qp) + i] += F_node[k][i] * phi_F[k][qp] * JxW_F[qp];
-                        }
+                        const int idx = n_vars * (qp_offset + qp);
+                        const double p_JxW_F = phi_F[k][qp] * JxW_F[qp];
+                        if (n_vars == NDIM)
+                            for (unsigned int i = 0; i < NDIM; ++i) F_JxW_qp[idx + i] += F_node[k][i] * p_JxW_F;
+                        else
+                            for (unsigned int i = 0; i < n_vars; ++i) F_JxW_qp[idx + i] += F_node[k][i] * p_JxW_F;
                     }
                 }
                 for (unsigned int k = 0; k < X_dof_indices[0].size(); ++k)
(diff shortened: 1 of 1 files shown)


ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1144,6 +1146,122 @@ FEDataManager::prolongData(const int f_data_idx,
...
+ * This function (the last of the three) takes the number of variables and
+ * number of basis functions as template argument which allows the compiler to
+ * unroll the two inner loops when appropriate. Setting either value to -1
+ * results in run-time selection of loop bounds (which is considerably
+ * slower).
+ */
+template <int n_vars, int n_basis>
+void
+integrate_cell_rhs_n_vars_n_basis(const unsigned int qp_offset,
+                                  const std::vector<std::vector<double> >& phi_F,
```
drwells: If we want to go even further down the rabbit hole when we rewrite FEDataInterpolation it would be great to replace `std::vector<std::vector<double>>` with a contiguous 2D array where the basis functions are indexed fastest (since we always index by quadrature points slowest).

drwells:
@boyceg This is the other 'simple performance improvement' patch I mentioned.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: Yes indeed...

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -871,10 +871,12 @@ FEDataManager::spread(const int f_data_idx,
                 {
                     for (unsigned int qp = 0; qp < n_qp; ++qp)
                     {
-                        for (unsigned int i = 0; i < n_vars; ++i)
-                        {
-                            F_JxW_qp[n_vars * (qp_offset + qp) + i] += F_node[k][i] * phi_F[k][qp] * JxW_F[qp];
-                        }
+                        const int idx = n_vars * (qp_offset + qp);
+                        const double p_JxW_F = phi_F[k][qp] * JxW_F[qp];
+                        if (n_vars == NDIM)
```
boyceg: Is this simply to make it easier for the compiler to vectorize, so that it knows the size of the loop?

Another common case would probably be `n_vars == 1`.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1494,31 +1620,25 @@ FEDataManager::interpWeighted(const int f_data_idx,
...
-                        {
-                            F_rhs_e[i](k) += F_qp[idx + i] * p_JxW_F;
-                        }
-                    }
-                }
+                integrate_cell_rhs(n_vars, n_basis, qp_offset, phi_F, JxW_F, F_qp, F_rhs_concatenated);
+
                 for (unsigned int i = 0; i < n_vars; ++i)
                 {
+                    std::copy(F_rhs_concatenated.begin() + i * n_basis,
```
boyceg: Is it worth making a helper function for this?

boyceg: (Presumably we will want to do this kind of thing in other places?)

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: Also, let's not call this a `cell`. Maybe `elem`? I would like to try to reserve `cell` for the background grid cells.

boyceg:
Looks good to me. Minor comments, taken them or leave them.

ibtk/src/lagrangian/FEDataManager.cpp
drwells: More-or-less yes. I didn't look at the assembly; I don't know if it would vectorize but it will surely unroll this loop.

drwells: Regarding `n_vars == 1`: when does that happen in this function? I thought that we always had an NDIM-dimensional force.

boyceg: You can "spread" quantities that are not forces.

boyceg: (It can make sense to spread any quantity that is a density.)

ibtk/src/lagrangian/FEDataManager.cpp
drwells: You mean with the copy?

boyceg: Yeah, going between interleaved and non-interleaved forms.

drwells: Since this is a one-liner and not applicable to spreading (and based on your comment below) I think its okay to hold off on putting this in a helper function.

ibtk/src/lagrangian/FEDataManager.cpp
drwells: I thought about this some more and figured that the best solution would be to simply add a second set of template functions for computing this. I added a test to #494 that verifies that we still calculate the same values (up to roundoff) for at least one spread kernel.

drwells:
I ended up doubling the size of this PR after @boyceg's initial review so someone should take a quick look at the force spreading implementation before we merge, just to be safe.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -867,26 +1001,9 @@ FEDataManager::spread(const int f_data_idx,
...
-                    {
-                        for (unsigned int i = 0; i < NDIM; ++i)
-                        {
-                            X_qp[NDIM * (qp_offset + qp) + i] += X_nodes[e_idx][k][i] * phi_X[k][qp];
-                        }
-                    }
-                }
+
+                sum_weighted_elem_solution(n_vars, F_dof_indices[0].size(), qp_offset, phi_F, JxW_F, F_node, F_JxW_qp);
+                sum_weighted_elem_solution<true>(NDIM, phi_X.size(), qp_offset, phi_X, {}, X_nodes[e_idx], X_qp);
```
boyceg:
```suggestion
                sum_weighted_elem_solution</*weights_are_unity*/ true>(NDIM, phi_X.size(), qp_offset, phi_X, {}, X_nodes[e_idx], X_qp);
```

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -867,26 +1001,9 @@ FEDataManager::spread(const int f_data_idx,
...
-                    for (unsigned int qp = 0; qp < n_qp; ++qp)
-                    {
-                        for (unsigned int i = 0; i < NDIM; ++i)
-                        {
-                            X_qp[NDIM * (qp_offset + qp) + i] += X_nodes[e_idx][k][i] * phi_X[k][qp];
-                        }
-                    }
-                }
+
+                sum_weighted_elem_solution(n_vars, F_dof_indices[0].size(), qp_offset, phi_F, JxW_F, F_node, F_JxW_qp);
```
boyceg: ?
```suggestion
                sum_weighted_elem_solution</*weights_are_unity*/ false>(n_vars, F_dof_indices[0].size(), qp_offset, phi_F, JxW_F, F_node, F_JxW_qp);
```

PR #508, opened 2019-03-25
Consolidate changelog entries.

drwells:
I already applied this change to the 0.4 branch.


amneetb:
No NWT in 0.4? Don't let minor design things get in the way of adding bigger things into IBAMR. We keep on doing this to ourselves every time. IBAMR should also grow in physics modules, not only just software design updates ...

drwells:
I would like to get NWT merged but this release is somewhat urgent: we really need to fix the C++14 compatibility problem with libMesh ASAP. Since that patch is still in review I think we should move on without it. We can always release again once its in.

amneetb:
Well, the patch is in review since Dec 2018. It is inefficient to keep reviewing it multiple times for small API and syntax changes. I would like to take care of the comments in a single go (which I thought I did before this latest review). Writing comments on queries sucks up a lot of time :-(

I would wait till all of the patch is reviewed before I take any further action on the branch. I don't want it to be a non-ending endeavor.

amneetb:
(This PR is also holding me off to add more features as they require stuff from this branch. It has been 2 months that I have been postponing them)

PR #511, opened 2019-03-28
Don't write some output in 3D.

drwells:
These files are not well optimized and are huge in 3D.


drwells:
Like the other 0.4 patches I will just merge this directly since I know it works (and I want to get the release out ASAP).

PR #512, opened 2019-03-28
Don't write some output in 3D.

drwells:
These files are not well optimized and are huge in 3D.

Same as #511 but for master.


boyceg:
Why not make it a boolean that is read from input but defaults to `false`? This is probably a good default for this kind of stuff for almost everyone!

drwells:
The problem is that these files are so huge in 3D that they take up all available memory. I would prefer to disable this for now in 3D unconditionally so that programs don't crash.

Issue #492, opened 2019-02-27
Explicitly partition libMesh data into interaction and non-interaction parts


drwells:
I have been doing some more profiling with callgrind (I recently learned how to use callgrind with MPI, which gives a complete picture of where the process spends it time). It looks like as we scale up to larger processor counts (e.g., 24 with the heart model) we are spending a lot of time exchanging data when interpolating the velocity. A large contributor to this imbalance is the amount of time PETSc spends sorting indices (which, IIRC, is only necessary when we add new entries into a vector that were assumed zero at construction time).

Ultimately, I believe that this is a result of the fact that we do assembly of libMesh vectors in two different ways: we do things in the 'normal' way (where we loop over the cells owned by the current processor) and we do things at interaction points (where we loop over the cells that have support on the locally owned patches). In the later case one cell may have support on multiple patches, so these are, in general, not equal (the interaction set should be a super set of the normal one). I need to do some more experiments to see what the difference is, in practice, between these, but since these sets are not the same we have to insert a lot of data into PETSc vectors corresponding to rows of the linear system that are not locally owned since these sets of dofs are not equal: hence the large amount of communication.

I think that the right way to do this is to set up our own PETSc vectors (plain `Vec`s) with the correct set of dofs for doing interaction (i.e., all dofs on cells that intersect the locally owned patches) as well as our own PETSc `VecScatter` objects which know how to accumulate the interaction vectors into the libMesh system RHS vectors. I think that if we explicitly tell PETSc what to expect we can really improve performance here.

drwells:
A few numbers, from my experiments so far:
1. with 6 processors we spent 29% of our cycles in `interpolateVelocity` and 15% of total cycles in vector assembly.
2. with 12 processors we spent 30% of our cycles in `interpolateVelocity` and 20% of total cycles in vector assembly.
3. with 24 processors we spent 24% of our cycles in `interpolateVelocity`: and 17% of total cycles in vector assembly.

That's 50%, 66%, and 71% of our time in that function being spent on communication. For comparison with a more usual workflow:
1. with 6 processors we spent 13% of our cycles in `computeLagrangianForce` and 0.2% of total cycles in vector assembly.
2. with 12 processors we spent 8% of our cycles in `computeLagrangianForce` and 0.25% of total cycles in vector assembly.
3. with 24 processors we spent 5% of our cycles in `computeLagrangianForce` and 0.21% of total cycles in vector assembly.

[...]

PR #513, opened 2019-04-04
Add a new class FEPatchData.

drwells:
This class supersedes the current `SystemDofMapCache` system: it stores all FE indices on patches, as well as localized indices for direct access to ghosted vectors in contiguous buffers (instead of `std::map<int, std::vector<std::vector<int>>>`). I use the very nice `boost::const_multi_array_ref` to return multidimensional views into aforementioned buffers that interoperate well with other functions in IBTK.

This is part of the fix to 'how do we assemble efficiently when we have a lot of nonlocal data' from #492: in particular, this implements indices for directly accessing ghosted vectors via standard pointer arithmetic.

This is definitely a work in progress: the implementation is mostly done but we don't actually use it anywhere yet. We have a bunch of stuff to do along the way:
- [ ] Consider deprecating `SystemDofMapCache` unless it is sufficiently convenient to keep around
- [ ] Add a `std::vector<FEPatchData>` member to `FEDataManager` and update it after regridding. Also add some warnings to that class that inform users that `FEDataManager` caches a lot of data and said caches should be updated if users modify the mesh in some way.
- [ ] write more tests (especially for grids with multiple element types)
- [ ] Actually use this class in `interpWeighted`
- [ ] Ensure that constraints still work when we use direct array access. libMesh sometimes modifies the array of dof indices; I need to learn how this works.
- [ ] Consider whether or not we should use a more sophisticated scattering routine instead of simply using the ghost indices directly. Would it be compatible with libMesh's constraint implementation?
- [ ] move the implementation into `FEDataManager.cpp`
- [ ] Implement a 'scratch' array function inside `SystemPatchDataView` that manages an array, indexed in the same way as the index arrays, with float data (we would use this with `get_values_for_interpolation`)
- [ ] Compute the number of dofs per cell by subtracting the index array entries instead of storing it directly

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -59,6 +59,8 @@
 #include "libmesh/enum_order.h"
 #include "libmesh/enum_quadrature_type.h"
 #include "libmesh/system.h"
+#include "libmesh/equation_systems.h"
+#include "libmesh/petsc_vector.h"
 #include "tbox/Pointer.h"
 #include "tbox/Serializable.h"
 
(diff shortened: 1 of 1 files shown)


boyceg:
One thing to be careful about is that SAMRAI has a "PatchData" concept, and so we want to make sure that it is very clear that FEPatchData is not this kind of patch data.

Also, if you are storing data in a way that is indexed by the SAMRAI cell indexing scheme ... would it make sense to go ahead and make it use the SAMRAI PatchData interface, so that SAMRAI can communicate the cached info when we reconstruct the AMR grid hierarchy? This could possibly simplify parallel redistribution, by allowing the code to simply look at the local libMesh element or node indices when setting up the libMesh partitioning.

drwells:
Yeah, the names for these classes are not great: there is really no relation to SAMRAI other than 'these elements have support on the given patch'.

drwells:
It may not be possible to store libMesh indices inside SAMRAI in an efficient way since we want to continue supporting meshes with multiple element types. We could definitely store the element ids, though.

drwells:
Lets keep the branch, but at the moment this is not a significant bottleneck so I don't think it worth worrying about in the near future.

PR #514, opened 2019-04-07
Added IBTK_MPI utility class.

abarret:
This adds the `IBTK_MPI` utility class to replace the `SAMRAI_MPI` class (which significantly changes in SAMRAI 3). Part 1 of 3 for separating out #287

### DIFF
--- a/ibtk/include/ibtk/IBTK_MPI.h
+++ b/ibtk/include/ibtk/IBTK_MPI.h
@@ -1,0 +1,331 @@
+// Filename: IBTK_MPI.h
+//
+// Copyright (c) 2002-2017, Boyce Griffith
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright notice,
+//      this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//    * Neither the name of The University of North Carolina nor the names of
+//      its contributors may be used to endorse or promote products derived from
+//      this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef included_IBTK_MPI
+#define included_IBTK_MPI
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include "SAMRAI_config.h"
+
+#include <mpi.h>
+
+#include "tbox/Utilities.h"
+
+namespace IBTK
+{
+inline MPI_Datatype
+mpi_type_id(const double)
+{
+    return MPI_DOUBLE;
+}
+
+inline MPI_Datatype
+mpi_type_id(const std::pair<double, int>)
+{
+    return MPI_DOUBLE_INT;
+}
+
+inline MPI_Datatype
+mpi_type_id(const int)
+{
+    return MPI_INT;
+}
+
+inline MPI_Datatype
+mpi_type_id(const std::pair<int, int>)
+{
+    return MPI_2INT;
+}
+
+inline MPI_Datatype
+mpi_type_id(const float)
+{
+    return MPI_FLOAT;
+}
+
+inline MPI_Datatype
+mpi_type_id(const std::pair<float, int>)
+{
+    return MPI_FLOAT_INT;
+}
+
+inline MPI_Datatype
+mpi_type_id(const char)
+{
+    return MPI_CHAR;
+}
+
+template <typename T>
+inline MPI_Datatype
+mpi_type_id(const T&)
+{
+    static_assert(!std::is_same<T, T>::value,
+                  "The given type does not have a corresponding MPI_Datatype value. At this time only char, int, "
+                  "unsigned int, float, double, std::pair<int, int>, std::pair<int, double>, and "
+                  "std::pair<int, float> are supported by IBTK_MPI.");
+    return MPI_CHAR;
+}
+
+/**
+ * @brief Provides C++ wrapper around MPI routines.
+ *
+ * The IBTK_MPI struct provides simple interfaces to common MPI routines. All function calls allow a user-provided
+ * communicator to be used.
+ *
+ * Note that this class is a utility class to group function calls in one
+ * name space (all calls are to static functions).  Thus, you should never
+ * attempt to instantiate a class of type IBTK_MPI; simply call the functions
+ * as static functions using the IBTK_MPI::function(...) syntax.
+ */
+
+struct IBTK_MPI
+{
+    /**
+     * MPI Types
+     */
+    using comm = MPI_Comm;
+    using group = MPI_Group;
+    using request = MPI_Request;
+    using status = MPI_Status;
+
+    /**
+     * Set the communicator that is used for the MPI communication routines.
+     * The default communicator is MPI_COMM_WORLD.
+     */
+    static void setCommunicator(IBTK_MPI::comm communicator);
+
+    /**
+     * Get the current MPI communicator.  The default communicator is
+     * MPI_COMM_WORLD.
+     */
+    static IBTK_MPI::comm getCommunicator();
+
+    /**
+     * Get SAMRAI World communicator.
+     */
+    static IBTK_MPI::comm getSAMRAIWorld();
+
+    /**
+     * Return the processor rank (identifier) from 0 through the number of
+     * processors minus one.
+     */
+    static int getRank(IBTK_MPI::comm communicator = getCommunicator());
+
+    /**
+     * Return the number of processors (nodes).
+     */
+    static int getNodes(IBTK_MPI::comm communicator = getCommunicator());
+
+    /**
+     * Perform a global barrier across all processors.
+     */
+    static void barrier(IBTK_MPI::comm communicator = getCommunicator());
+
+    //@{
+    /**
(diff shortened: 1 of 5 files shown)


ibtk/include/ibtk/IBTK_MPI.h
```diff
@@ -0,0 +1,344 @@
...
+
+    /**
+     * Perform a global barrier across all processors.
+     */
+    static void barrier(IBTK_MPI::comm communicator = MPI_COMM_NULL);
+
+    inline static MPI_Datatype mpi_type_id(const double)
+    {
+        return MPI_DOUBLE;
+    }
```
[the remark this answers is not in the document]
drwells: We should mark these as `inline constexpr` too since they only return a constant.

ibtk/include/ibtk/IBTK_MPI.h
```diff
@@ -0,0 +1,344 @@
...
+    /**
+     * Get the current MPI communicator.  The default communicator is
+     * MPI_COMM_WORLD.
+     */
+    static IBTK_MPI::comm getCommunicator();
+
+    /**
+     * Get SAMRAI World communicator.
+     */
+    static IBTK_MPI::comm getSAMRAIWorld();
```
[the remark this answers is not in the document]
abarret: No. In SAMRAI 3 `getSAMRAIWorld()` returns the communicator that is used upon initialization.

ibtk/src/utilities/IBTK_MPI.cpp
```diff
@@ -0,0 +1,491 @@
...
+            send[i].second = getRank(communicator);
+        }
+        MPI_Allreduce(send.data(), recv.data(), n, mpi_type_id(recv[0]), MPI_MINLOC, communicator);
+        for (int i = 0; i < n; ++i)
+        {
+            x[i] = recv[i].first;
+            rank_of_min[i] = send[i].second;
+        }
+    }
+} // minReduction
```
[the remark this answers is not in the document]
boyceg: 👍

ibtk/include/ibtk/IBTK_MPI.h
```diff
@@ -0,0 +1,344 @@
...
+     */
+    static void
+    minReduction(double* x, const int n = 1, int* rank_of_min = NULL, IBTK_MPI::comm communicator = MPI_COMM_NULL);
+    static void
+    minReduction(int* x, const int n = 1, int* rank_of_min = NULL, IBTK_MPI::comm communicator = MPI_COMM_NULL);
+    static void
+    minReduction(float* x, const int n = 1, int* rank_of_min = NULL, IBTK_MPI::comm communicator = MPI_COMM_NULL);
+    static double minReduction(double x, int* rank_of_min = NULL, IBTK_MPI::comm communicator = MPI_COMM_NULL);
+    static int minReduction(int x, int* rank_of_min = NULL, IBTK_MPI::comm communicator = MPI_COMM_NULL);
+    static float minReduction(float x, int* rank_of_min = NULL, IBTK_MPI::comm communicator = MPI_COMM_NULL);
```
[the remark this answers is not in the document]
abarret: What are we making into a template function? `minReduction` has a templated version that is `private`. These functions above just call the private templated version.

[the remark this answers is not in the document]
abarret: Also, making one of these for `std::size_t` involves doing a reduction with a `std::pair<std::size_t, int>` which I don't think has a built in `MPI_Datatype`. Is it worth creating a datatype for this?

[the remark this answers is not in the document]
drwells: You're right: there is no `MPI_UNSIGNED_LONG_INT` pair type so we have to leave this out.

abarret:
@drwells Take a look now. I think I've made all the changes that were requested.

abarret:
Not sure what this error is. Seems like it's an issue with the clang compilers and the `constexpr`. @drwells can you take a look when you get a chance?

ibtk/include/ibtk/IBTK_MPI.h
```diff
@@ -0,0 +1,332 @@
...
+
+template <typename T>
+inline constexpr static MPI_Datatype
+mpi_type_id(const T&)
+{
+    static_assert(!std::is_same<T, T>::value,
+                  "The given type does not have a corresponding MPI_Datatype value. At this time only char, int, "
+                  "unsigned int, float, double, std::pair<int, int>, std::pair<int, double>, and "
+                  "std::pair<int, float> are supported by IBTK_MPI.");
+    return -1;
```
drwells: We can just do `return MPI_CHAR;` here: that will always work.

drwells:
My suggestion doesn't work for some MPI implementations:
```

In file included from ../src/utilities/IBTK_MPI.cpp:41:
../include/ibtk/IBTK_MPI.h:96:12: error: cannot initialize return object of type 'MPI_Datatype' (aka 'ompi_datatype_t *') with an rvalue of type 'int'
    return -1;
```

We can fix the other problem by marking these functions as simply `inline` (no static or constexpr).

ibtk/src/utilities/IBTK_MPI.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+template <typename T>
+void
+IBTK_MPI::sumReduction(T* x, const int n, IBTK_MPI::comm communicator)
+{
+    communicator = communicator == MPI_COMM_NULL ? d_communicator : communicator;
+    if (n == 0 || getNodes(communicator) < 2) return;
+    std::vector<T> send(n);
+    std::copy(x, x + n, send.begin());
+    MPI_Allreduce(send.data(), x, n, mpi_type_id(send[0]), MPI_SUM, communicator);
```
drwells: We can skip the copy by setting the output buffer to `MPI_IN_PLACE`. That may not have been around when the SAMRAI guys wrote this.

ibtk/src/utilities/IBTK_MPI.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+void
+IBTK_MPI::allToOneSumReduction(int* x, const int n, const int root, IBTK_MPI::comm communicator /* = MPI_COMM_NULL */)
+{
+    communicator = communicator == MPI_COMM_NULL ? d_communicator : communicator;
+    if (getNodes(communicator) > 1)
+    {
+        std::vector<int> send(n);
+        std::copy(x, x + n, send.begin());
+        MPI_Reduce(send.data(), x, n, MPI_INT, MPI_SUM, root, communicator);
```
drwells: same

drwells:
@abarret one more thing I forgot about: please add a changelog entry.

abarret:
@drwells I've made the suggested changes and squashed the commits.

ibtk/include/ibtk/IBTK_MPI.h
```diff
@@ -0,0 +1,332 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef included_IBTK_MPI
+#define included_IBTK_MPI
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include "SAMRAI_config.h"
+
+#include "mpi.h"
```
drwells: Use `#include <mpi.h>`

ibtk/include/ibtk/IBTK_MPI.h
```diff
@@ -0,0 +1,332 @@
...
+                  "The given type does not have a corresponding MPI_Datatype value. At this time only char, int, "
+                  "unsigned int, float, double, std::pair<int, int>, std::pair<int, double>, and "
+                  "std::pair<int, float> are supported by IBTK_MPI.");
+    return MPI_CHAR;
+}
+
+/**
+ * @brief Provides C++ wrapper around MPI routines.
+ *
+ * The IBTK_MPI struct provides simple interfaces to common MPI routines. All function calls allow a communicator to be
```
drwells: should be 'user-provided communicator'

ibtk/src/utilities/IBTK_MPI.cpp
```diff
@@ -0,0 +1,326 @@
...
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include <stdlib.h>
+#include <string.h>
```
drwells: I don't think we actually need these headers anymore.

ibtk/src/utilities/IBTK_MPI.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+#include <string>
```
drwells: We should add `#include <vector>` too.

ibtk/src/utilities/IBTK_MPI.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+#include "tbox/SAMRAI_MPI.h"
+
+#include "ibtk/IBTK_MPI.h"
+#include "ibtk/app_namespaces.h"
+#include "tbox/Utilities.h"
+
+namespace IBTK
+{
+IBTK_MPI::comm IBTK_MPI::d_communicator = MPI_COMM_WORLD;
```
drwells: Since this is static IIRC we should name it `s_communicator`.

ibtk/include/ibtk/IBTK_MPI.h
```diff
@@ -0,0 +1,332 @@
...
+    /**
+     * Get SAMRAI World communicator.
+     */
+    static IBTK_MPI::comm getSAMRAIWorld();
+
+    /**
+     * Return the processor rank (identifier) from 0 through the number of
+     * processors minus one.
+     */
+    static int getRank(IBTK_MPI::comm communicator = MPI_COMM_NULL);
```
drwells: I think we can get rid of `MPI_COMM_NULL` here and elsewhere by just writing `getCommunicator()` as the default value.

ibtk/src/utilities/IBTK_MPI.cpp
```diff
@@ -0,0 +1,326 @@
...
+    return SAMRAI_MPI::commWorld;
+#else
+    return SAMRAI_MPI::getSAMRAIWorld()
+#endif
+}
+
+int
+IBTK_MPI::getNodes(IBTK_MPI::comm communicator /* = MPI_COMM_NULL */)
+{
+    communicator = communicator == MPI_COMM_NULL ? d_communicator : communicator;
```
drwells: if we change the default argument then we can get rid of all of these checks.

ibtk/src/utilities/IBTK_MPI.cpp
```diff
@@ -0,0 +1,326 @@
...
+    }
+} // allGatherSetup
+
+template <typename T>
+void
+IBTK_MPI::minMaxReduction(T* x, const int n, int* rank, MPI_Op op, IBTK_MPI::comm communicator)
+{
+    // TODO: both this approach and SAMRAI's assume that we can define
+    // byte-for-byte compatible types with, e.g.,
+    // MPI_FLOAT_INT. Check this with a unit test!
```
drwells: we can remove this comment since I opened an issue.

abarret:
Fixed

ibtk/include/ibtk/IBTK_MPI.h
```diff
@@ -0,0 +1,332 @@
...
+                  "unsigned int, float, double, std::pair<int, int>, std::pair<int, double>, and "
+                  "std::pair<int, float> are supported by IBTK_MPI.");
+    return MPI_CHAR;
+}
+
+/**
+ * @brief Provides C++ wrapper around MPI routines.
+ *
+ * The IBTK_MPI struct provides simple interfaces to common MPI routines. All function calls allow a user-provided
+ * communicator to be used. If a null communicator or none is present, the default communicator set by setCommunicator()
```
drwells: sorry, one more thing: this is out of sync with the previous changeset. I would just delete the sentence "If a null communicator..." since the default argument is already present in the declaration.

abarret:
Done. One of these days I'll get it right without all these iterations :)

PR #520, opened 2019-04-13
Conditionally disable logging of solver iterations.

drwells:
I think this is the last prerequisite patch for the test suite.

The number of solver iterations is not stable in the sense that small perturbations (the way a vector is aligned, order of operations in floating point, etc.) can change it by O(1), e.g., some solvers in the test suite take 6 iterations 90% of the time and 7 iterations 10% of the time. The best way to fix this is to simply not log the number of solver iterations.

This PR adds a new boolean that controls this. The default is to simply copy the value


amneetb:
What will happen to changes in the INSVC classes with this PR when NWT is merged?

drwells:
Nothing: the NWT branch does not change anything near where we log output.

boyceg:
I am going to go ahead and merge.

PR #523, opened 2019-04-15
Sprinkle/rigid fhd 2

drwells:
Version of #425 with autotools issues resolved.


include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+
+    /*!
+     * \brief Destructor of the class.
+     */
+    ~CIBStochasticMethod();
+
+    /*!
+     * \brief setter for KT.
+     */
+    void setkT(double kT);
```
drwells: We should read these values from the input database instead of using getters and setters.

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * \brief Class CIBFEMethod is a concrete CIBStrategy and IBMethod
+ * class which implements the motion of rigid bodies using the constraint
+ * formulation. The immersed structure is discretized using standard IB
+ * markers.
```
drwells: I think this documentation was copied form somewhere else. We should expand on what stochastic algorithms are actually implemented by this class.

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+class CIBStochasticMethod : public IBAMR::CIBMethod
+{
+public:
+    /*!
+     * \brief Constructor of the class.
+     */
+    CIBStochasticMethod(const std::string& object_name,
+                        SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db,
+                        const int no_structures = 1,
+                        bool register_for_restart = true);
```
drwells: parameters need descriptions: what are the number of structures?

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+
+    /*!
+     * \brief Destructor of the class.
+     */
+    ~CIBStochasticMethod();
```
drwells: I will go ahead and force push, in a minute, a few minor fixes. The default destructor suffices so I will go ahead and delete it.

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+     * \brief getter for Length scale.
+     */
+    double getLScale() const;
+
+    /*!
+     * \brief getter for RFD delta.
+     */
+    double getRFdelta() const;
+
+    void preprocessIntegrateData(double current_time, double new_time, int num_cycles);
```
drwells: This should be marked as `override` (I will do that)

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+
+    /*!
+     * \brief Advance the positions of the Lagrangian structures
+     */
+    void midpointStep(double current_time, double new_time);
```
drwells: also mark as `override` (I will do that)

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+
+    /*!
+     * \brief Advance the positions of the Lagrangian structure by a small increment
+     */
+    void moveLagrangianData(double delta);
+
+    /*!
+     * \brief helper function for RFD
+     */
+    void computeRFDforcesAndDisplacements();
```
drwells: I think this could be `protected`

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+
+    /*!
+     * \brief resetsets rigid body velocity of structures at the half-time positions to those at current time
+     */
+    void resetRFDVelocity();
+
+    /*!
+     * \brief configuration getter
+     */
+    void getConfiguration(const unsigned int part, Eigen::Vector3d& C, Eigen::Quaterniond& Q);
```
drwells: We should add more comments here explaining that this is the structure.

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+
+    /*!
+     * \brief initial configuration getter
+     */
+    void getInitialCOM(const unsigned int part, Eigen::Vector3d& C);
+
+    bool getReject();
```
drwells: needs documentation

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+
+    int getNumReject();
```
drwells: needs documentation

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,176 @@
...
+
+    //////////////////////////////////////////////////////////////////////////////
+
+protected:
+    //////////////////////////////////////////////////////////////////////////////
+
+private:
+    // Boltzman Constant
+    double d_kT, d_L_scale, d_rf_delta;
```
drwells: The current convention is to use C++11-style initialization in the header. I will fix this.

include/ibamr/CIBStochasticStokesSolver.h
```diff
@@ -0,0 +1,235 @@
...
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * \brief Class CIBStaggeredStokesSolver is an extension of
+ * IBAMR::StaggeredStokesSolver class that solves for the Langrange multiplier
+ * field \f$ \vec{\lambda} \f$ which imposes the rigidity constraint and solves
+ * for rigid body velocity (if it is freely-moving) along with the fluid
+ * velocity and pressure.
```
drwells: We need to update the documentation here too.

src/IB/CIBStochasticMethod.cpp
```diff
@@ -0,0 +1,463 @@
...
+                        reject++;
+                        pout << "REJECTED BY WALL\n";
+                    }
+                }
+            }
+        }
+        d_l_data_manager->getLData("X0_unshifted", ln)->restoreArrays();
+    }
+
+    reject = SAMRAI_MPI::sumReduction(reject);
```
drwells: This doesn't seem right to me. Why do we assume that we should keep track of global state, instead of state per processor?

src/IB/CIBStochasticMethod.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+            for (unsigned int d = 0; d < NDIM; ++d)
+            {
+                X_new(d) = d_center_of_mass_new[struct_handle][d] + R_dr_new(d);
+                if (!periodic_shift[d])
+                {
+                    if (X_new(d) < domain_x_lower[d] || domain_x_upper[d] < X_new(d))
+                    {
+                        reject++;
+                        pout << "REJECTED BY WALL\n";
```
drwells: We should either guard this with `d_enable_logging` (so that it can be conditionally turned off) and send it to `plog` or remove it.

src/IB/CIBStochasticMethod.cpp
```diff
@@ -0,0 +1,463 @@
...
+            if (solve_dofs[k])
+            {
+                UW[k] = a[p];
+                ++p;
+            }
+        }
+        rdvToEigen(UW, d_trans_vel_current[part], d_rot_vel_current[part]);
+        d_trans_vel_current[part] *= (d_L_scale * d_L_scale);
+        part_free_dofs_begin += num_free_dofs;
+    }
```
drwells: I don't fully understand what this does but this algorithm seems suspicious in that every single processor starts at the beginning of the global array and then does the same loop. Is there any way to avoid `U_all` here?

src/IB/CIBStochasticStokesSolver.cpp
```diff
@@ -0,0 +1,677 @@
...
+    d_sp_solver = NULL;
+    d_wide_u_var = NULL;
+    d_wide_f_var = NULL;
+    d_wide_ctx = NULL;
+    d_wide_u_idx = -1;
+    d_wide_f_idx = -1;
+    d_x_wide = NULL;
+    d_b_wide = NULL;
+    d_is_initialized = false;
+    d_reinitializing_solver = false;
```
drwells: These should also be moved to the header. I will take care of this.

src/IB/CIBStochasticStokesSolver.cpp
```diff
@@ -0,0 +1,677 @@
...
+    d_cib_strategy->getFreeRigidVelocities(&U, d_current_time);
+
+    // Create a vector that contains net external force and torque on the body.
+    // NOTE: In the initialization stage we do not need an actual force vector, a reference
+    // to U should suffice to know the required structure.
+    Vec F = U;
+
+    // Create the composite vectors.
+    Vec mv_x, mv_b;
+    std::vector<Vec> vx(3), vb(3);
```
drwells: We should just use `std::array<Vec, 3>` here.

src/IB/CIBStochasticStokesSolver.cpp
```diff
@@ -0,0 +1,677 @@
...
+} // computeRFDforcesAndDisplacements (Brennan Sprinkle)
+
+void
+CIBStochasticStokesSolver::ComputeRFD_RHS(Vec b, Vec y)
+{
+    Pointer<CIBStochasticMethod> sib_method_ops = d_cib_strategy;
+
+    // Get some constants
+    static const double gamma = d_sp_solver->getSpreadScale();
+    static const double beta = d_sp_solver->getInterpScale();
```
drwells: Why are these `static`?

src/IB/CIBStochasticMethod.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+void
+CIBStochasticMethod::moveLagrangianData(double delta)
+{
+    const int coarsest_ln = 0;
+    const int finest_ln = d_hierarchy->getFinestLevelNumber();
+    const double dt = delta;
+
+    // Fill the rotation matrix of structures with rotation angle 0.5*(W^n)*dt.
+    std::vector<Eigen::Matrix3d> rotation_mat(d_num_rigid_parts, Eigen::Matrix3d::Identity(3, 3));
```
drwells: This won't work due to the custom allocation requirements with eigen. I will fix this.

amneetb:
@brennansprinkle : Can you advice us on some of these comments on Github itself? We will take care of them in the actual code.

include/ibamr/CIBStochasticMethod.h
drwells: 'Stochastic' is pretty broad: perhaps we could pick a different name of this class based on the exact type of dynamics implemented?

src/IB/CIBStochasticMethod.cpp
amneetb: @brennansprinkle: What is the intent here? The free rigid degrees of freedom (that can actually change from time step to time step --- motion can be constrained or become free at any tilmestep) vector is a global vector with all the DoFs owned by processor zero as this is a very small vector. For a single body in the domain, the size is 3 in 2D and 6 in 3D. This is how I set it

https://github.com/IBAMR/IBAMR/blob/master/src/IB/CIBMethod.cpp#L263-L305

amneetb: @drwells: We create these MPI vectors and turn it to PETSc to do iterative solves.

brennansprinkle: Sorry, this function is entirely vestigial and should be removed as I don't believe that it is called anywhere. The original functionality (requiring a helper function that I did remove) was replaced by CIBStochasticStokesSolver::computeFUforRFD.

src/IB/CIBStochasticStokesSolver.cpp
brennansprinkle: I don't know but they shouldn't be. Will fix.

src/IB/CIBStochasticStokesSolver.cpp
brennansprinkle: agreed

include/ibamr/CIBStochasticMethod.h
brennansprinkle: no_structures is the number of rigid bodies being simulated. I made it default to 1 but this is probably bad as it may hide a good/helpful error is this value is not properly being set in main.

include/ibamr/CIBStochasticMethod.h
brennansprinkle: this function should just be removed.

src/IB/CIBStochasticMethod.cpp
amneetb: @brennansprinkle : Is there a better name than `FU`. Reminds me of Frank Underwood and the obvious.

include/ibamr/CIBStochasticMethod.h
brennansprinkle: Something like?:
Function to return the center of mass position, 'C',  and the orientation (as a quaternion), 'Q', of the rigid particle indexed by 'part'

include/ibamr/CIBStochasticMethod.h
brennansprinkle: Something like?:
Helper function used to provide a reference configuration for a rigid particle as it's initial position

include/ibamr/CIBStochasticMethod.h
brennansprinkle: Something like?:
getReject: getter for the boolean value indicating whether the current timestep has been rejected or not due to an invalid configuration (eg. a particle crossing a physical wall) 

getNumReject: getter for the total number of rejected timesteps, where a timestep is rejected due to an invalid configuration (eg. a particle crossing a physical wall)

src/IB/CIBStochasticMethod.cpp
brennansprinkle: Yeah, if I'm being honest that did cross my mind, but seemed more of a feature than a bug to me. Still, you're probably right :-) Maybe 'computeRFDForcesAndVelocities'

amneetb: What is RFD? Rigid Fluctuating Dynamics?

brennansprinkle: Random Finite Difference.
Sorry, yeah that's a bit jargony but it's what we call it in the paper so I think it should stay in there

amneetb: Agree. We should put RFD in the class description.

examples/CIB/ex5/ex_main.cpp
```diff
@@ -0,0 +1,8 @@
+#include "main_one_body.cpp"
```
boyceg: Can we put in a comment to discourage this construct (`#include`ing the `cpp` file) and a "TODO:" comment indicating that we should fix this someday?

include/ibamr/CIBStochasticMethod.h
boyceg: I think it is good to have the ability to both.

include/ibamr/CIBStochasticStokesSolver.h
```diff
@@ -0,0 +1,235 @@
+// Filename: CIBStaggeredStokesSolver.h
```
boyceg:
```suggestion
// Filename: CIBStochasticStaggeredStokesSolver.h
```
Also, do we really need a whole new solver here, or could we just massage the RHS in (a subclass of) the existing `CIBStaggeredStokesSolver`?

boyceg: If it is technically possible to do this but too much work for now, please add a TODO comment.

If there really needs to be a different solver, please address this in the class documentation.

examples/CIB/ex5/ex_main.cpp
boyceg: Oops, this is actually exactly in line with the way the other IBAMR examples are set up. :flushed:

I think we can clean this up in a later PR, after #494.

include/ibamr/CIBStochasticMethod.h
boyceg: I think it would be fine just to include a link to the relevant papers and not get into the details.

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,173 @@
...
+class CIBStandardInitializer;
+class CIBStaggeredStokesOperator;
+} // namespace IBAMR
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * \brief Class CIBFEMethod is a concrete CIBStrategy and IBMethod
```
boyceg:
```suggestion
 * \brief Class CIBStochastic is a concrete CIBStrategy and CIBMethod
```

include/ibamr/CIBStochasticMethod.h
boyceg: @brennansprinkle / @amneetb, are you able to get the structure indexing information from `LDataManager`, or is this a different notion of "structures" than what `LDataManager` provides?

include/ibamr/CIBStochasticMethod.h
boyceg: I think this is something that can be done in a separate PR. I have opened an issue (#535) to track classes that are missing support for either input file-based specification or programatic setup.

include/ibamr/CIBStochasticMethod.h
boyceg: What does "reject" mean here? I think it has something to do with crossing solid walls, but it would be helpful if the comment gave some more hints.

A separate question: Does it make sense to call `getReject()` from outside of the class? If not, perhaps this should be `private`.

boyceg: > What does "reject" mean here? I think it has something to do with crossing solid walls, but it would be helpful if the comment gave some more hints.
> 
> A separate question: Does it make sense to call `getReject()` from outside of the class? If not, perhaps this should be `private`.

Oh, I see this is addressed below.

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,173 @@
...
+     */
+    void resetRFDVelocity();
+
+    /*!
+     * \brief configuration getter
+     */
+    void getConfiguration(const unsigned int part, Eigen::Vector3d& C, Eigen::Quaterniond& Q);
+
+    /*!
+     * \brief initial configuration getter
```
boyceg:
```suggestion
     * \brief Lookup the initial center of mass C for the requested part.
```

include/ibamr/CIBStochasticMethod.h
boyceg:
```suggestion
    /**
      * @return boolean value indicating whether the current time step has been rejected or not due to an invalid configuration (eg. a particle crossing a physical wall).
      */
    bool getReject();
```

include/ibamr/CIBStochasticMethod.h
boyceg:
```suggestion
    /*!
     * @return The total number of rejected time steps encountered so far in the computation.
     */
    int getNumReject();
```

include/ibamr/CIBStochasticStokesSolver.h
```diff
@@ -0,0 +1,235 @@
...
+class INSStaggeredHierarchyIntegrator;
+class CIBStochasticMethod;
+} // namespace IBAMR
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * \brief Class CIBStaggeredStokesSolver is an extension of
```
boyceg:
```suggestion
 * \brief Class CIBStochasticStokesSolver is an extension of
```

src/IB/CIBStochasticMethod.cpp
boyceg:
```suggestion
                       if (d_enable_logging) pout << "provisional time step rejected because of particle collision with a wall\n";
```

src/IB/CIBStochasticMethod.cpp
boyceg:
```suggestion
    // Check to see if we have detected rejection on any processor.
    reject = SAMRAI_MPI::sumReduction(reject);
    if (d_enable_logging)
    {
        plog << "provisional time step rejected because of particle collision with a wall\n";
    }
```

src/IB/CIBStochasticStokesSolver.cpp
boyceg:
```suggestion
    std::array<Vec,3> vx, vb;
```

boyceg:
I just took another look. It seems like most of the changes requested are documentation. Code changes should be pretty minimal.

Two related API related things to think about before merging:

1. Is the notion of "structure" or "part" that is used in this class the same as the notion of "structure" or "part" used in `LDataManager`? If not, then we should make it clear in the class documentation. If so, then we should not require users to know how many parts are in their model in the `CIBStochasticMethod` class constructor.
2. I think that we usually put the part number last (e.g. in `IBFEMethod`), and in the interface, default the part number to `0`. This provides a (modest) simplification in the API slightly for the case in which a model only has one part. If this is expected to be a common use case for this code, I would suggest making this switch now and squashing the commit. It might be worth making the change just to keep the APIs between the different IB implementations to be similar.

amneetb:
We need this information of number of structures to set a lot of things in the constructor. Getting this information from LDataManger is not possible unless it is initialized by CIBMethod. It is chicken/egg kind of thing. Also IBFEMethod also gets number of structure information from its constructor. So it is not that different.

boyceg:
> We need this information of number of structures to set a lot of things in the constructor. Getting this information from LDataManger is not possible unless it is initialized by CIBMethod. It is chicken/egg kind of thing. Also IBFEMethod gets number of structure information from its constructor. So it is not that different.

This doesn't really answer the question, which is asking if the notion of "number of structures" is the same in the `CIBStochasticMethod` and the `LDataManager`. Are these classes using the same concept or a different concept for "number of structures"? If they are the same, I'd prefer to try to get the information out of `LDataManager`; if they are different, we should have some comments to that effect.

amneetb:
The concept of number of structure is different in the main class (CIBStochastic and CIBMethod) and the preconditioner of this class. The preconditioner can arbitrarily combine different structures from LDataManager and make it one body, while the main class treats them separately. Right now  CIBStochastic and CIBMethod treats number of structures same as like LDataManager, and this information is expected to be the same.

include/ibamr/CIBStochasticMethod.h
drwells: Yes.

boyceg:
> The concept of number of structure is different in the main class (CIBStochastic and CIBMethod) and the preconditioner of this class. The preconditioner can arbitrarily combine different structures from LDataManager and make it one body, while the main class treats them separately. Right now CIBStochastic and CIBMethod treats number of structures same as LDataManager, and this information is expected to be the same.

Since it would be a change to both `CIBMethod` and `CIBStochasticMethod`, let's just address this in another PR (c.f. #539).

include/ibamr/CIBStochasticMethod.h
boyceg: Addressing this also would affect `CIBMethod`. Let's do this in a separate PR. I've recorded the issue here: #539.

boyceg:
> 2. I think that we usually put the part number last (e.g. in `IBFEMethod`), and in the interface, default the part number to `0`. This provides a (modest) simplification in the API slightly for the case in which a model only has one part. If this is expected to be a common use case for this code, I would suggest making this switch now and squashing the commit. It might be worth making the change just to keep the APIs between the different IB implementations to be similar.

Again this seems like it also affects `CIBMethod`. Let's put this in a separate PR. I've tracked the issue here: #540.

brennansprinkle:
Sorry that took a bit longer, I wanted to wait for a permanent citation for the documentation. I've committed some updated docs.

drwells:
@brennansprinkle Thanks for following up on this. I am sorry that this has been languishing in the mean time. I will put aside some time later this week to read through the current state of the PR.

examples/CIB/Makefile.am
```diff
@@ -16,17 +16,20 @@ gtest:
 	@(cd ex2 && make gtest) || exit 1;
 	@(cd ex3 && make gtest) || exit 1;
 	@(cd ex4 && make gtest) || exit 1;
+	@(cd ex5 && make gtest) || exit 1;
```
drwells: We are at the point where removal of the gtest based test suite is imminent (we will remove it in 0.5). Lets delete this and the other test code: AFAICT `test_main.cpp` is not in this PR anyway.

examples/CIB/ex5/Makefile.am
```diff
@@ -0,0 +1,65 @@
+## Process this file with automake to produce Makefile.in
+include $(top_srcdir)/config/Make-rules
+
+## Dimension-dependent testers
+EXAMPLE_DRIVER = ex_main.cpp
+GTEST_DRIVER   = test_main.cpp
```
drwells: Same thing here: just delete all the gtest logic.

examples/CIB/ex5/Makefile.am
```diff
@@ -0,0 +1,65 @@
...
+EXTRA_DIST  = input3d.one_body boomerang_15.vertex
+#EXTRA_DIST += input3d.test
```
drwells:
```suggestion
```

examples/CIB/ex5/input3d.one_body
```diff
@@ -0,0 +1,484 @@
+// ---------------------------------------------------------------------------
```
drwells:
```suggestion
```

examples/CIB/ex5/input3d.one_body
```diff
@@ -0,0 +1,484 @@
...
+
+// random-number generator parameters
+SEED = 0                                // RNG seed (0 for clock)
+
+// ---------------------------------------------------------------------------
```
drwells:
```suggestion
```
(we do not use this kind of formatting in the input files; lets keep things uniform by not having sections delineated this way)

examples/CIB/ex5/input3d.one_body
```diff
@@ -0,0 +1,484 @@
...
+
+// physical parameters
+kT  = 0.0041419464                                // temperature
+MU  = 8.9e-04                               // viscosity (shear)
+RHO = 0.0                                // density
+L_SCALE = 2.1 				// Length scale
+
+// ---------------------------------------------------------------------------
```
drwells:
```suggestion
```

examples/CIB/ex5/input3d.one_body
drwells: Please convert all tabs to spaces in this file.

examples/CIB/ex5/main_one_body.cpp
```diff
@@ -0,0 +1,732 @@
+// Filename main.cpp
+// Created on 26 Jul 2016 by Amneet Bhalla
```
drwells: Please update authorship :)

examples/CIB/ex5/main_one_body.cpp
```diff
@@ -0,0 +1,732 @@
...
+                                          ib_method_ops,
+                                          "SP_");
+
+        /*        Pointer<CIBStaggeredStokesSolver> CIBSolver =
+                    new CIBStaggeredStokesSolver("CIBStaggeredStokesSolver",
+                                                 input_db->getDatabase("CIBStaggeredStokesSolver"),
+                                                 navier_stokes_integrator,
+                                                 ib_method_ops,
+                                                 "SP_");
+        */
```
drwells: Please remove commented-out code.

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,168 @@
...
+namespace IBAMR
+{
+/*!
+ * \brief Class CIBStochasticMethod is a concrete CIBStrategy and CIBMethod
+ * class which implements the motion of rigid bodies using the constraint
+ * formulation. The immersed structure is discretized using standard IB
+ * markers. Adds some functionality for stocastic simulations. 
+ */
+
+class CIBStochasticMethod : public IBAMR::CIBMethod
```
drwells:
```suggestion
class CIBStochasticMethod : public CIBMethod
```

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,168 @@
...
+class CIBStandardInitializer;
+class CIBStaggeredStokesOperator;
+} // namespace IBAMR
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * \brief Class CIBStochasticMethod is a concrete CIBStrategy and CIBMethod
```
drwells:
```suggestion
 * \brief CIBStochasticMethod is a concrete CIBStrategy
```
(Since `CIBMethod` inherits from `CIBStrategy` this is redundant)

include/ibamr/CIBStochasticMethod.h
```diff
@@ -0,0 +1,168 @@
...
+ * class which implements the motion of rigid bodies using the constraint
+ * formulation. The immersed structure is discretized using standard IB
+ * markers. Adds some functionality for stocastic simulations.
```
drwells: We really need more documentation in the description of this class than just `Adds some functionality for stochastic simulations`. What does it do? What models does it implement? What relevant example programs should we look at? etc.

include/ibamr/CIBStochasticMethod.h
drwells: @brennansprinkle *ping*

src/IB/CIBStochasticStokesSolver.cpp
```diff
@@ -0,0 +1,675 @@
...
+
+namespace IBAMR
+{
+namespace
+{
+// Types of refining and coarsening to perform prior to setting coarse-fine
+// boundary and physical boundary ghost cell values.
+static const std::string DATA_REFINE_TYPE = "NONE";
+static const bool USE_CF_INTERPOLATION = true;
+static const std::string DATA_COARSEN_TYPE = "CUBIC_COARSEN";
```
drwells: Since these are `const` we cannot ever change them. Could you document, in the general documentation of the class, what these are set as?

PR #524, opened 2019-04-15
Move regrid hierarchy calls

drwells:
`IBTK::HierarchyIntegrator::regridHierarchy` currently calls both `initializeCompositeHierarchyDataSpecialized` and `synchronizeHierarchyDataSpecialized` on itself and all child integrators. However, No class actually implements a nonempty `synchronizeHierarchyDataSpecialized` method aside from `IBTK::HierarchyIntegrator`, and putting everything in a routine named `initialize` is confusing. Hence I added (I am not in love with the names) `regridHierarchyBeginSpecialized` and `regridHierarchyEndSpecialized` which are called immediately before and after we do the actual regridding inside SAMRAI.

From the discussion in #343: I am now convinced that this is the correct way to handle this since a lot of classes need to do stuff immediately before or after regridding and there is no way to do it at the present aside from reimplementing `regridHierarchy` and remembering to call all setup/synch/etc functions on all child integrators.

I still need to move around some SAMRAI variables (this doesn't compile right now) so this is a work in progress.

### DIFF
--- a/ibtk/include/ibtk/HierarchyIntegrator.h
+++ b/ibtk/include/ibtk/HierarchyIntegrator.h
@@ -220,6 +220,22 @@ public:
      * GriddingAlgorithm::regidAllFinerLevels() to regrid the patch hierarchy.
      * Subclasses can control the method used to regrid the patch hierarchy by
      * overriding this public virtual member function.
+     *
+     * Before regridding, this method calls regridHierarchyBeginSpecialized()
+     * on the current integrator and all child integrators.
+     *
+     * After regridding and (optionally) checking the new grid volume, this
+     * method calls regridHierarchyEndSpecialized() on the current integrator
+     * and all child integrators. It then calls the following methods (and,
+     * therefore, the specialized methods on the current and all child
+     * integrators) in the following order:
+     *
+     * 1. initializeCompositeHierarchyData
+     * 2. synchronizeHierarchyData
+     *
+     * @warning This class assumes, but does not enforce, that this method is
+     * only called on the parent integrator. A future release of IBAMR will
+     * enforce this assumption.
      */
     virtual void regridHierarchy();
 
@@ -672,6 +688,20 @@ public:
     void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db) override;
 
 protected:
+    /*!
+     * Perform any necessary work relevant to data owned by the current
+     * integrator prior to regridding (e.g., calculating divergences). An
+     * empty default implementation is provided.
+     */
+    virtual void regridHierarchyBeginSpecialized();
+
+    /*!
+     * Perform any necessary work relevant to data owned by the current
+     * integrator after regridding (e.g., calculating divergences). An empty
+     * default implementation is provided.
+     */
+    virtual void regridHierarchyEndSpecialized();
+
     /*!
      * Virtual method to compute an implementation-specific minimum stable time
      * step size.
--- a/include/ibamr/INSCollocatedHierarchyIntegrator.h
+++ b/include/ibamr/INSCollocatedHierarchyIntegrator.h
@@ -183,12 +183,14 @@ public:
                                        bool skip_synchronize_new_state_data,
                                        int num_cycles = 1) override;
 
-    /*!
-     * Regrid the patch hierarchy.
-     */
-    void regridHierarchy() override;
-
 protected:
+    /*
+     * Since (unlike the staggered case) there is no local divergence-preserving
+     * interpolation scheme for collocated velocity fields, the velocity must
+     * always be projected onto the grid after regridding. Do that here.
+     */
+    void regridHierarchyEndSpecialized() override;
+
     /*!
      * Determine the largest stable timestep on an individual patch.
      */
--- a/include/ibamr/INSCollocatedHierarchyIntegrator.h
+++ b/include/ibamr/INSCollocatedHierarchyIntegrator.h
@@ -230,6 +230,11 @@ protected:
      */
     void setupPlotDataSpecialized() override;
 
+    /*!
+     * Project the velocity field following a regridding operation.
+     */
+    void regridProjection() override;
+
 private:
     /*!
      * \brief Default constructor.
@@ -268,11 +273,6 @@ private:
      * Reinitialize the operators and solvers used by the hierarchy integrator.
      */
     void reinitializeOperatorsAndSolvers(double current_time, double new_time);
-
-    /*!
-     * Project the velocity field following a regridding operation.
-     */
-    void regridProjection();
 
     /*!
      * Value determining the type of projection method to use.
--- a/include/ibamr/INSHierarchyIntegrator.h
+++ b/include/ibamr/INSHierarchyIntegrator.h
@@ -357,6 +357,11 @@ public:
 
 protected:
     /*!
+     * Pure virtual method to project the velocity field following a regridding operation.
+     */
+    virtual void regridProjection() = 0;
+
+    /*!
      * The constructor for class INSHierarchyIntegrator sets some default
      * values, reads in configuration information from input and restart
      * databases, and registers the integrator object with the restart manager
--- a/include/ibamr/INSStaggeredHierarchyIntegrator.h
+++ b/include/ibamr/INSStaggeredHierarchyIntegrator.h
@@ -273,6 +273,11 @@ protected:
      */
     void setupPlotDataSpecialized() override;
 
+    /*!
+     * Project the velocity field following a regridding operation.
+     */
+    void regridProjection() override;
+
 private:
     /*!
      * \brief Default constructor.
@@ -311,11 +316,6 @@ private:
      * Reinitialize the operators and solvers used by the hierarchy integrator.
      */
     void reinitializeOperatorsAndSolvers(double current_time, double new_time);
-
-    /*!
-     * Project the velocity field following a regridding operation.
-     */
-    void regridProjection();
 
     /*!
      * Determine the convective time stepping type for the current time step and
--- a/include/ibamr/INSStaggeredHierarchyIntegrator.h
+++ b/include/ibamr/INSStaggeredHierarchyIntegrator.h
@@ -201,11 +201,6 @@ public:
                                        int num_cycles = 1) override;
 
     /*!
-     * Regrid the patch hierarchy.
-     */
-    void regridHierarchy() override;
-
-    /*!
      * Setup solution and RHS vectors using state data maintained by the
      * integrator.
      */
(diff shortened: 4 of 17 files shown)


drwells:
Its not clear to me why `INSCollocatedHierarchyIntegrator` does not use the same logic for projecting the velocity: the other four NSE integrators all first check that the divergence has increased after grid refinement, but that integrator does it unconditionally. Is this an oversight or does that integrator need special considerations?

boyceg:
`INSCollocatedHierarchyIntegrator` always projects because (AFAIK) there is not a local divergence-free interpolation scheme for collocated vector fields. There are local divergence-free interpolation schemes for staggered-grid solvers, however.

drwells:
I think I fixed the remaining issues. This should work now.

I don't really have a way to test this other than the nascent test suite. @amneetb @nishantnangia did either of you have a test case for which this broke?

ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -220,9 +220,39 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
...
+     * enforce this assumption.
      */
     virtual void regridHierarchy();
 
+    /*!
+     * Perform any necessary work relevant to data owned by the current
+     * integrator prior to regridding (e.g., calculating divergences). An
+     * empty default implementation is provided.
+     */
+    virtual void regridHierarchyBeginSpecialized();
```
boyceg: I think that I usually make these kinds of methods `protected`, since they are not intended to be part of the public interface. Can we do that here?

ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -220,9 +220,39 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
...
+
+    /*!
+     * Perform any necessary work relevant to data owned by the current
+     * integrator after regridding (e.g., calculating divergences). An empty
+     * default implementation is provided.
+     */
+    virtual void regridHierarchyEndSpecialized();
```
boyceg: Same as above.

include/ibamr/INSStaggeredHierarchyIntegrator.h
```diff
@@ -201,9 +201,17 @@ class INSStaggeredHierarchyIntegrator : public INSHierarchyIntegrator
...
 
     /*!
-     * Regrid the patch hierarchy.
+     * Prepare the current hierarchy for regridding. Here we calculate the divergence.
      */
-    void regridHierarchy() override;
+    void regridHierarchyBeginSpecialized() override;
+
+    /*!
+     * Prepare the current hierarchy for regridding. Here we recalculate the
```
boyceg:
```suggestion
     * Update current hierarchy data after regridding. Here we recalculate the
```
?

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
```diff
@@ -229,9 +229,17 @@ class INSVCStaggeredHierarchyIntegrator : public INSHierarchyIntegrator
...
 
     /*!
-     * Virtual method to regrid the patch hierarchy.
+     * Prepare the current hierarchy for regridding. Here we calculate the divergence.
      */
-    virtual void regridHierarchy() override;
+    void regridHierarchyBeginSpecialized() override;
+
+    /*!
+     * Prepare the current hierarchy for regridding. Here we recalculate the
```
boyceg: Same as above.

include/ibamr/INSStaggeredHierarchyIntegrator.h
```diff
@@ -232,6 +240,21 @@ class INSStaggeredHierarchyIntegrator : public INSHierarchyIntegrator
     void removeNullSpace(const SAMRAI::tbox::Pointer<SAMRAI::solv::SAMRAIVectorReal<NDIM, double> >& sol_vec);
 
 protected:
+    /*!
+     * L1 norm of the fluid velocity before regridding.
+     */
+    double d_div_U_norm_1_pre = 0.0;
```
boyceg: Could/should these be moved into the base class?

drwells: Technically yes, but since `INSCollocatedHierarchyIntegrator` doesn't use them I thought it would be slightly better to duplicate them across two inheriting classes.

I tried to do this and to also move the 'check to see if the divergence grew' code into the base class. Unfortunately, since inheriting classes use different cell variables (e.g., `FaceVariable` instead of `CellVariable`) its not possible to make some calls to `HierarchyMathOps` without complete type information, which is why I opened #526.

drwells: I figured the duplication was slightly better than the alternatives.

drwells:
```
clang: error: clang frontend command failed due to signal (use -v to see invocation)
```
that can't be good.

amneetb:
@boyceg and @drwells : We need to make regridProjection() a public virtual function. We have made this in INSVC classes but the other classes need to to that. For INSVC one needs to have a density field to do a projection and this density field (particularly for a solid body) is sometimes not available until before the time loop. For example one calculates a level set from a FE mesh at the end of initial hierarchy setup.

boyceg:
> @boyceg and @drwells : We need to make regridProjection() a public virtual function. We have made this in INSVC classes but the other classes need to to that. For INSVC one needs to have a density field to do a projection and this density field (particularly for a solid body) is sometimes not available until before the time loop. For example one calculates a level set from a FE mesh at the end of initial hierarchy setup.

Can we safely call `regridProjection()` in user code? I am not sure. If not, can we make it `protected` instead of `public`?

amneetb:
Also IBHierarchyIntegrator always regrids before entering the time loop, whereas HierarchyIntegrator does not. INS solvers rely on HierarchyIntegrator's behavior. We should always regrid before entering the time loop.

amneetb:
regridProjection() is a safe call as projection method is an idempotent operator. It ought to be safe. If it is not something is wrong in the implementation of the function, not that it is getting called.

drwells:
@amneetb are you sure? `regridProjection` is definitely `protected` right now on master.

boyceg:
> regridProjection() is a safe call as projection method is an idempotent operator. It ought to be safe. If it is not something is wrong in the implementation of the function, not that it is getting called.

1. It is definitely not _idempotent_ for cell centered velocities, because it uses a cell-centered approximate projection, not an exact projection.
2. This is not a math question, this is a question about the implementation. This was not intended to be a public function, and I do not know if it works correctly when used in different contexts. We should be careful about moving functions from the `private` to `public` interface.

drwells:
Since the access restrictions on that function are not immediately relevant to this PR lets move that discussion to a separate issue. To emphasize: that function has never been public.

amneetb:
It is public not in master, but in NWT branch.

drwells:
@boyceg I made the suggested changes.

ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -672,6 +688,20 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
     void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db) override;
 
 protected:
+    /*!
+     * Perform any necessary work relevant to data owned by the current
+     * integrator prior to regridding (e.g., calculating divergences). An
+     * empty default implementation is provided.
+     */
+    virtual void regridHierarchyBeginSpecialized();
```
boyceg: 👍

include/ibamr/INSCollocatedHierarchyIntegrator.h
```diff
@@ -183,12 +183,14 @@ class INSCollocatedHierarchyIntegrator : public INSHierarchyIntegrator
                                        bool skip_synchronize_new_state_data,
                                        int num_cycles = 1) override;
 
-    /*!
-     * Regrid the patch hierarchy.
+protected:
+    /*
+     * Since (unlike the staggered case) there is no known divergence-free
```
boyceg: One more change to the comment here to try to clarify:
```suggestion
     * Since (unlike the staggered case) there is no local divergence-preserving
```
There are non-local interpolation schemes (e.g. involving solving Poisson problems).

include/ibamr/INSHierarchyIntegrator.h
```diff
@@ -355,6 +355,11 @@ class INSHierarchyIntegrator : public IBTK::HierarchyIntegrator
      */
     int getNumberOfCycles() const override;
 
+    /*!
+     * Pure virtual method to project the velocity field following a regridding operation.
+     */
+    virtual void regridProjection() = 0;
```
boyceg: So, uhh, should this be `public` or `protected`? :-)

drwells: :speak_no_evil:

boyceg: It seems like this is something that the solver can do as appropriate, but perhaps should not be done by external callers.

drwells: I really did mean to make this protected, as the commit states. I fixed it.

boyceg:
@drwells, are there any Navier-Stokes examples with nontrivial AMR in the test suite? `navier_stokes/ex1` comes to mind, although any of the IB examples with AMR enabled should also do the trick.

drwells:
The only NSE test for which we actually check output in the current test suite is `ex0`.

boyceg:
Sorry to ask you to do a manual check, but I would compare `navier_stokes/ex1`.

drwells:
I'll add a variant of that to #494 to check this.

drwells:
@boyceg I ran both NSE solvers for a couple thousand time steps and compared solutions via `h5dump`. At the final timestep a few dozen entries have a relative difference of 1e-5 so this patch works correctly.

amneetb:
@drwells and @boyceg : What's the bottom line on this? What should I do for INSVC in the NWT branch (and not in the master branch which is quite outdated for INSVC classes now)?

boyceg:
> @drwells and @boyceg : What's the bottom line on this? What should I do for INSVC in the NWT branch (and not in the master branch which is quite outdated for INSVC classes now)?

Like I said in #527, please make a PR with the changes that are needed (e.g. in terms of where regrid projections happen).

drwells:
The bottom line is that this implements the required functionality for NWT: it replaces the first commit with something that fixes the problem both there and also with IBFE codes.

drwells:
@boyceg Can we merge? I want to check before hitting the green button.

boyceg:
I ran some stuff last night. Let me do a quick spot check for diffs.

boyceg:
> I ran some stuff last night. Let me do a quick spot check for diffs.

Well, maybe not quick, but no diffs in output in my tests. Please merge if you are ready.

PR #525, opened 2019-04-16
Fix an uninitialized pointer warning.

drwells:
I've seen this warning one too many times and finally decided to fix it.

### DIFF
--- a/src/IB/IBFESurfaceMethod.cpp
+++ b/src/IB/IBFESurfaceMethod.cpp
@@ -890,14 +890,14 @@ IBFESurfaceMethod::computeLagrangianForce(const double data_time)
         }
         NumericVector<double>& X0_vec = X_system.get_vector("INITIAL_COORDINATES");
 
-        System* DP_system;
-        const DofMap* DP_dof_map;
-        FEDataManager::SystemDofMapCache* DP_dof_map_cache;
+        // silence some warnings by giving DP_dof_map_cache a bogus, but
+        // valid, value if we don't use jump conditions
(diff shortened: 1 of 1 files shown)


src/IB/IBFESurfaceMethod.cpp
```diff
@@ -892,12 +892,15 @@ IBFESurfaceMethod::computeLagrangianForce(const double data_time)
 
         System* DP_system;
         const DofMap* DP_dof_map;
-        FEDataManager::SystemDofMapCache* DP_dof_map_cache;
+        // silence some warnings by giving DP_dof_map_cache a bogus, but
+        // valid, value if we don't use jump conditions
+        FEDataManager::SystemDofMapCache& DP_dof_map_cache =
+            d_use_jump_conditions ? *d_fe_data_managers[part]->getDofMapCache(PRESSURE_JUMP_SYSTEM_NAME) :
+                                    F_dof_map_cache;
```
boyceg: Can we please make it `nullptr` so that if we try to use it, it crashes instead of giving bizarre results?
```suggestion
                                    nullptr;
```

drwells: I switched it to a reference instead of a pointer so I will have to go back and fix the type, but yes.

boyceg: Oh, you changed to a reference.

Can you keep it a pointer and explicitly initialize to `nullptr`?

drwells: Yup

drwells:
Fixed.

boyceg:
I will go ahead and merge.

PR #528, opened 2019-04-16
Fix libmesh 1.3.0 compat

drwells:
I am not sure how this made it through but the 0.4 release is not compatible with libMesh 1.3 (it is compatible with 1.1, 1.2, and 1.4) since I messed up the version number check here. Lets merge this and put out a 0.4.1 release quickly.


drwells:
whoops, this is onto the wrong branch. One moment.

PR #529, opened 2019-04-16
Fix compatibility with libMesh 1.3.x. (0.4.1)

drwells:
This change to the libMesh API did not occur until version 1.4.


drwells:
What's especially dumb is that I managed to fix this in #500 but not #502  :cry:

boyceg:
Shall we wait for Jenkins? (Is Jenkins processing PRs?)

boyceg:
@drwells, I am not sure why Jenkins is taking so long to complete this. Should we go ahead and merge?

drwells:
We've been busier than usual today, I suppose.

boyceg:
Looks like it was indeed just busy.

PR #531, opened 2019-04-17
Updates to a c++11 update script.

abarret:
This updates the script that replaces `UniquePtr` with `unique_ptr` to also replace `boost::array` with `std::array`. @drwells Are there any other simple changes like this that would be useful to include?

@AminKolah You can use this script to do some of the dull search and replaces on #240.


drwells:
The other onse that I can think of (that is much less common) are `boost::unordered_map`, `boost::tuple`, and `boost::make_tuple`. I am not aware of any other boost classes that we use that have been replaced by C++11 things.

PR #536, opened 2019-04-18
eliminate comparisons between signed and unsigned ints

boyceg:
This eliminates some comparisons between `int`s and `unsigned int`s in the new templated functions in `FEDataManager`.

I just changed everything to `int`, which I think should be fine in the current context of use.

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -583,35 +583,35 @@ FEDataManager::buildGhostedCoordsVector(const bool localize_data)
  */
 template <int n_vars, int n_basis, bool weights_are_unity = false>
 void
-sum_weighted_elem_solution_n_vars_n_basis(const unsigned int qp_offset,
+sum_weighted_elem_solution_n_vars_n_basis(const int qp_offset,
                                           const std::vector<std::vector<double> >& phi_F,
                                           const std::vector<double>& weights,
                                           const boost::multi_array<double, 2>& F_node,
                                           std::vector<double>& F_w_qp)
 {
-    const unsigned int n_qp = phi_F[0].size();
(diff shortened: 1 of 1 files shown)


drwells:
I'm guilty of this one. Let me take a look at what's going on here.

drwells:
Writing `int` makes sense since I compare a value against `-1`: I already assume they should be signed.

drwells:
We still need to fix the other set of template functions (`integrate_elem_rhs`). I think we can keep `qp_offset` unsigned: its only used for pointer arithmetic and is declared as unsigned in `interpWeighted` and `spread`.

boyceg:
> We still need to fix the other set of template functions (`integrate_elem_rhs`). I think we can keep `qp_offset` unsigned: its only used for pointer arithmetic and is declared as unsigned in `interpWeighted` and `spread`.

Should we prefer to avoid `unsigned int`? See discussion [here](https://www.reddit.com/r/cpp/comments/7y0o6r/is_it_a_good_practice_to_use_unsigned_values), for instance.

boyceg:
> > We still need to fix the other set of template functions (`integrate_elem_rhs`). I think we can keep `qp_offset` unsigned: its only used for pointer arithmetic and is declared as unsigned in `interpWeighted` and `spread`.
> 
> Should we prefer to avoid `unsigned int`? See discussion [here](https://www.reddit.com/r/cpp/comments/7y0o6r/is_it_a_good_practice_to_use_unsigned_values), for instance.

OK, this PR now eliminates `unsigned int` from all of these new helper functions.

drwells:
We should also change the declaration of the quadrature point offset in `spread` and `interpWeighted` to `int`.

boyceg:
> We should also change the declaration of the quadrature point offset in `spread` and `interpWeighted` to `int`.

Done.

PR #537, opened 2019-04-19
clang formatting in preparation of merging IIM implementation in IBFESurfaceMethod

boyceg:


boyceg:
Hmmm, I am not sure that this is actually going to be helpful in merging; the file in master is already basically in synch with the current results from `clang-format`.

Issue #533, opened 2019-04-18
Infrastructure for caching Eulerian data


boyceg:
We should consider making it easier to cache Eulerian data, to avoid repeatedly allocating/deallocating data on the AMR grid hierarchy, perhaps by having an `AMRDataCache`(?) object that allows callers to "check out" and "return" data with a particular layout/depth/ghost cell width.

drwells:
Should we do this before #472 ?

boyceg:
I think we need to think a little bit about whether it is a good idea to have either a global or local data management class that tracks these data. I was originally thinking that it would be a global manager class (like SAMRAI's VariableDatabase), to maximize data reuse, but perhaps it is cleaner to have a local caching class, similar to some of the FE data management classes.

boyceg:
This is addressed in #538 and #572.

boyceg:
I am going to close this because the infrastructure now exists.

PR #538, opened 2019-04-24
create class SAMRAIDataCache to cache allocated Eulerian data

boyceg:
This is a draft implementation of a SAMRAI data caching class (see #533).

### DIFF
--- a/ibtk/src/utilities/SAMRAIDataCache.cpp
+++ b/ibtk/src/utilities/SAMRAIDataCache.cpp
@@ -1,0 +1,309 @@
+// Filename: SAMRAIDataCache.cpp
+// Created on 23 Apr 2019 by Boyce Griffith
+//
+// Copyright (c) 2002-2019, Boyce Griffith
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright notice,
+//      this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//    * Neither the name of The University of North Carolina nor the names of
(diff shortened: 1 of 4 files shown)


drwells:
:+1:

I will take a look tomorrow or Friday (once I am back from VT)

ibtk/src/utilities/SAMRAIDataCache.cpp
```diff
@@ -0,0 +1,297 @@
...
+    // TODO: Make this more generic and extensible.
+    bool convertable = false;
+    if (can_convert_to<CellVariable<NDIM, T> >(var))
+    {
+        Pointer<CellDataFactory<NDIM, T> > pdat_fac = patch_descriptor->getPatchDataFactory(idx);
+        TBOX_ASSERT(pdat_fac);
+        depth = pdat_fac->getDefaultDepth();
+        ghost_width = pdat_fac->getGhostCellWidth().max();
+        TBOX_ASSERT(ghost_width == pdat_fac->getGhostCellWidth().min());
+        convertable = true;
```
drwells: I think that we could refactor this internal block (which is common to every case in the `if/else if` chain) to something like
```cpp
template <typename FactoryType>
std::pair<int, int> // depth, ghost_width
get_characteristics(const PatchDescriptor<NDIM> &patch_descriptor,
                    const int idx)
{
    auto Pointer<FactoryType> = patch_descriptor->getPatchDataFactory(idx);
    depth = pdat_fac->getDefaultDepth();
    ghost_width = pdat_fac->getGhostCellWidth().max();
    TBOX_ASSERT(ghost_width == pdat_fac->getGhostCellWidth().min());
    return std::make_pair(depth, ghost_width);
}
```

drwells:
This looks good. Could you add a simple test?

boyceg:
Yes. I have run tests with #572 and everything seems to work as expected.

@drwells, if #572 looks good to you, can you run the tests on it? (Can we set up Jenkins to do this for us automatically?)

drwells:
Is it good to go (aside from the testers) now?

PR #541, opened 2019-04-26
Mpi type wrapper test

### LINKED ISSUE #517: Add a test for our MPI type wrappers
We assume, in `IBTK_MPI`, that we can use `std::pair<float, int>` as `MPI_DATATYPE_FLOAT_INT` (and other such pairs). We should write a test that checks this.

This will be pretty easy to do with the new test infrastructure.

drwells:
Fixes #517.

### DIFF
--- a/tests/IBTK/mpi_type_wrappers.output
+++ b/tests/IBTK/mpi_type_wrappers.output
@@ -1,0 +1,25 @@
+
+size of MPI type = 8
+sizeof(double) = 8
+
+size of MPI type = 4
+sizeof(int) = 4
+
+size of MPI type = 4
+sizeof(float) = 4
+
+size of MPI type = 1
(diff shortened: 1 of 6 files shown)


tests/IBTK/mpi_type_wrappers.output
```diff
@@ -0,0 +1,25 @@
...
+sizeof(float) = 4
+
+size of MPI type = 1
+sizeof(char) = 1
+
+size of MPI type = 8
+sizeof(std::pair<float, int>) = 8
+sizeof(struct {float a; int b;}) = 8
+
+size of MPI type = 12
```
drwells: The default behavior for GCC is to keep things aligned on 8-byte boundaries, but MPI will not use any padding to decrease message sizes. Hence these are not equal.

PR #543, opened 2019-04-28
Remove some incorrect assertions.

drwells:
We sometimes call this function with edge elements: in that case these assertions don't make sense.

### DIFF
--- a/ibtk/include/ibtk/libmesh_utilities.h
+++ b/ibtk/include/ibtk/libmesh_utilities.h
@@ -986,10 +986,6 @@ template <class MultiArray>
 inline double
 get_max_edge_length(const libMesh::Elem* const elem, const MultiArray& X_node)
 {
-#ifndef NDEBUG
-    TBOX_ASSERT(elem->n_nodes() == X_node.shape()[0]);
-    TBOX_ASSERT(elem->dim() == X_node.shape()[1]);
-#endif
     const libMesh::ElemType elem_type = elem->type();
 
     double max_edge_length_2 = 0.0;


boyceg:
Why are these incorrect for `EDGE` elements?

drwells:
This assertion fails inside `IBFEMethod::spreadTransmissionForceDensity`, where we do the following:

```cpp
fe.reinit(elem);
fe.collectDataForInterpolation(elem);
const boost::multi_array<double, 2>& X_node = fe.getElemData(elem, X_sys_idx);

// Loop over the element boundaries.
for (unsigned short int side = 0; side < elem->n_sides(); ++side)
{
    // Skip non-physical boundaries.
    if (!is_physical_bdry(elem, side, boundary_info, G_dof_map)) continue;

    // Skip Dirichlet boundaries.
    if (is_dirichlet_bdry(elem, side, boundary_info, G_dof_map)) continue;

    // Construct a side element.
    std::unique_ptr<Elem> side_elem = elem->build_side_ptr(side, /*proxy*/ false);
    const bool qrule_changed = d_fe_data_managers[part]->updateSpreadQuadratureRule(
        qrule_face, d_spread_spec[part], side_elem.get(), X_node, patch_dx_min);
```
The assertion fails since it doesn't make sense to pass `X_nodes`, which stores position data for the cell, to something that estimates the length of an edge.

I thought it would be okay to remove the checks since this code 'worked' when we didn't check anything but now I am not so sure. I suspect that the length scale calculated in this function was always wrong since the indices into `X_node` will, at best, only be correct by coincidence.

boyceg:
I think the assertions are correct --- the function is expecting to get the nodal coordinates for the element that we are "measuring" --- and that it is the caller that is wrong.

boyceg:
> This assertion fails inside `IBFEMethod::spreadTransmissionForceDensity`, where we do the following:
> 
> ```c++
> fe.reinit(elem);
> fe.collectDataForInterpolation(elem);
> const boost::multi_array<double, 2>& X_node = fe.getElemData(elem, X_sys_idx);
> 
> // Loop over the element boundaries.
> for (unsigned short int side = 0; side < elem->n_sides(); ++side)
> {
>     // Skip non-physical boundaries.
>     if (!is_physical_bdry(elem, side, boundary_info, G_dof_map)) continue;
> 
>     // Skip Dirichlet boundaries.
>     if (is_dirichlet_bdry(elem, side, boundary_info, G_dof_map)) continue;
> 
>     // Construct a side element.
>     std::unique_ptr<Elem> side_elem = elem->build_side_ptr(side, /*proxy*/ false);
>     const bool qrule_changed = d_fe_data_managers[part]->updateSpreadQuadratureRule(
>         qrule_face, d_spread_spec[part], side_elem.get(), X_node, patch_dx_min);
> ```

Yes, the correct solution here is to extract the nodal data that corresponds to the side/edge element. I think that there is a way to do this using `boost::multi_array` that avoids having to copy data between representations. Let me see if I can track this down.

drwells:
I don't think it will be possible to create a view into this for most elements: for example, the nodes on a quad9 have funny numberings (0, 7, 3 and 1, 5, 2, for example) that are not at all contiguous. We should just copy the data.

boyceg:
Are you planning on changing this, or shall I do it?

drwells:
I can do it. I will get this done later today.

drwells:
This is surprisingly tricky. I am still working on a fix.

drwells:
Superceded by #552.

PR #544, opened 2019-04-28
Resize a temporary vector when needed.

drwells:
libMesh resizes this vector inside a function call.

Since the vector always had allocated enough storage this worked in release mode anyway.


ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1765,6 +1765,9 @@ FEDataManager::interpWeighted(const int f_data_idx,
                     dof_id_scratch = F_dof_indices[i];
                     F_dof_map.constrain_element_vector(F_rhs, dof_id_scratch);
                     F_vec.add_vector(F_rhs, dof_id_scratch);
+                    // libMesh sometimes resizes F_rhs inside
+                    // constrain_element_vector
+                    F_rhs.resize(F_dof_indices[0].size());
```
boyceg: Are we relying on `F_rhs` having a consistent size?

drwells: Good point: we assume, at the moment, that `F_rhs` always has the same size since all components have the same number of dofs, but there is really no reason to assume this. We might as well just resize it to the correct value at the start of each loop.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1758,6 +1757,9 @@ FEDataManager::interpWeighted(const int f_data_idx,
 
                 for (unsigned int i = 0; i < n_vars; ++i)
                 {
+                    // libMesh sometimes resizes F_rhs inside
+                    // constrain_element_vector, so ensure it has the right size:
+                    F_rhs.resize(F_dof_indices[i].size());
```
boyceg: 👍 
This seems like the right place to do this.

PR #545, opened 2019-04-28
Fix a bug in a test.

drwells:
This is what happens when I don't do any development in debug mode: this should be the last fix in the batch.

last_local_index is actually not the last local index: it is one past the last local index.


boyceg:
Seems like an odd naming choice.

Issue #546, opened 2019-04-29
Split up and merge NWT


drwells:
This is getting messy: lets collect everything in a GitHub issue so that we can easily link to PRs, branches, and commits.

For those of us joining us now: the NWT branch contains a ton of new examples (and a fair number of new features) but the history is difficult to understand because there are merge commits, application of tools like clang-format, etc. @amneetb, @boyceg, and myself are trying to split up the branch into more reasonable chunks.

Please edit this list until we have a comprehensive list of *all* the stuff in `amneetb/numerical-wave-tank` that we want to keep.
## commits with new functionality
- [ ] whatever commit added the stuff we need for 74f7494656.
- [ ] commit a11a5fa265 changes the pressure extraction algorithm for multiphase flows to second order.
- [ ] commits b142914b19, dcadb2845b, and 729f222d4c added a new Fortran routine for sign sweeping.
- [x] commit a710b9a582 added Brinkman penalization code (see #625).
- [x] commit c188e2116d enables translation (without rotation) in `ConstraintIB{Method,Kinematics}` (see #567)
- [ ] commit b300bc90ae adds some postprocessing to `ConstraintIBMethod`
- [x] commit 5c171743ef adds `FESurfaceDistanceEvaluator` (see also 16ef35b9f6) (see #580)
- [ ] commits 53cd8f1708 and a95c6b3527 fix some bugs in `IBLevelSet/ex0`
- [x] ~~commit 14addfd130 makes `regridProjection` public in some classes.~~ See the discussion in #527 for more information on why this method should remain `protected.

## commits adding new examples
- [x] commit 74f7494656 : three new multiphase flow examples (ex7, ex8, and ex9) (requires `ConstraintIBMethod::getFullLagrangeMultiplierForce` and `ConstraintIBMethod::getInterpolatedLagrangianDensity`) (we also need to update authorship here; that ended up in another commit) (see also bb6ae337367) (tracked by #547)
- [x] commit a9a3c3053a334 : fix an issue with regrid intervals in two-phase examples (this is independent of any other changes AFAICT) (tracked by #550)
- [x] commit 00bf3a616938 : four new multiphase flow examples (ex10, ex11, ex12, ex13) (requires `ConstraintIBMethod::getFullLagrangeMultiplierForce` and `ConstraintIBMethod::getInterpolatedLagrangianDensity`) (see also 5641adb47998, 25d881701ed, bb6ae337367322, and 76620bcb2050) (tracked by #548)

[...]

PR #547, opened 2019-04-29
NWT: multiphase_flow examples 7, 8, and 9

drwells:
Part of #546. As mentioned in #546 we cannot merge this until `ConstraintIBMethod::getFullLagrangeMultiplierForce` and `ConstraintIBMethod::getInterpolatedLagrangianDensity` are in the library: hence this is `do not merge` until we get that done.


drwells:
As @amneetb mentioned: removing the postprocessing calls fixes the compilation issue, so this is OK to review now.

drwells:
I force pushed a change to the commit in this PR that includes all other changes made to these examples.

boyceg:
> Looks like some of the source code files could use a run through `clang-format`. (There is some very minor formatting stuff in the comments, maybe a few other things.) Otherwise looks fine.

One other minor thing --- looks like these examples are using raw pointers in places where we probably could be using something like `std::unique_ptr` instead. There are probably existing examples that do this as well.

boyceg:
> > Looks like some of the source code files could use a run through `clang-format`. (There is some very minor formatting stuff in the comments, maybe a few other things.) Otherwise looks fine.
> 
> One other minor thing --- looks like these examples are using raw pointers in places where we probably could be using something like `std::unique_ptr` instead. There are probably existing examples that do this as well.

Finally, another minor thing --- for static functions that are not associated with a class, I think that we try to use lowercase + underscores instead of camel case, although I do not think that we are 100% consistent on this.

drwells:
I must admit I didn't read the source code. I'll take a look (and run clang-format).

boyceg:
I don't think we necessarily need to address all of this in the source before merging. We have not run modernize on the examples yet, and at some point we need to do something to cut down on the boilerplate code that shows up in all of them.

drwells:
I just rebased this on top of master to get the `ln` fix in #551.

drwells:
Per our conversations elsewhere: the examples in #418 are okay as-is. I will run clang-format and spot check this and merge it later today unless there are new objections.

drwells:
example 8 is missing `sphere3d.vertex`. @amneetb do you remember which input file we should use here? The other two copies of this file `ConstraintIB/falling_sphere/sphere3d.vertex` and `IB/explicit/ex1/sphere3d.vertex` are not equal.

Otherwise this is ready to go.

amneetb:
Looks like @nishantnangia forgot to add this here. If you type `make examples` in this directory it will generate a binary for generating sphere3d.vertex (simply executing the binary will generate the vertex file). I can do that and send it to you over the email. There are two branches multiphase-examples. Which one should I use?

drwells:
I don't see this executable. The branch is `multiphase-examples` (i.e., the first one).

amneetb:
I just emailed you the vertex file for this example. He forgot (or perhaps did not intend to) to create a separate executable for generating vertex file, which is also fine. Just doing `g++ sphereGen.cpp -o ./exe` creates the executable. No need to to have a target in the Makefile.am for this simple code.

boyceg:
Might as well have the make target, but no need to build it unless requested.

drwells:
These vertex files are huge: the 3D file is 9 MB (for comparison: `./src` is 5MB).

Does anyone object to simply generating the vertex files as part of the build process? This would make the PR a lot smaller.

boyceg:
> Does anyone object to simply generating the vertex files as part of the build process? This would make the PR a lot smaller.

This sounds like a good idea to me.

In fact, I was just thinking about whether we should aim to do this for all of the tests. It probably would not be too hard to convert the existing MATLAB scripts that build these meshes into Python scripts. It seems crazy to carry around these gigantic text files full of mesh coordinates.

drwells:
Done.

I checked that all the examples run so this is now good to go.

boyceg:
Feel free to merge once Jenkins gives the 👍.

drwells:
I also added a changelog so that @nishantnangia gets credit!

boyceg:
I created an issue (#560) on getting rid of the mesh data files.

amneetb:
> These vertex files are huge: the 3D file is 9 MB (for comparison: `./src` is 5MB).
> 
> Does anyone object to simply generating the vertex files as part of the build process? This would make the PR a lot smaller.

For this particular example, we can generate the vertex file as a part of `make examples` process, i.e, build and run the executable for generating vertex file during `make examples`.

amneetb:
> I also added a changelog so that @nishantnangia gets credit!

Thanks to both @drwells and @nishantnangia!

drwells:
of course :)

PR #550, opened 2019-04-29
NWT: Ensuring that regrid_interval is used for two-phase examples

drwells:
@amneetb This commit is independent, AFAICT, of any other changes made in NWT. It should be an easy place to start. Would you please take a look?

I checked that these examples run on this branch. We don't have any tests for this part of the library. Could you think of converting one of these into something akin to `tests/IBFE/explicit_ex4_2d`? It would be a good start to write a test that logs some useful quantities so that we know we have stable output.


amneetb:
Yes, sure. I will be happy to convert any of these examples to a test case. It probably would be best if I do this after all the reviews are done and things are merged.

drwells:
:+1: I'll hold you to it :)

boyceg:
The only changes in this PR appear to be in the setup of the input files and a little bit of code that controls cell tagging.

If these examples are independent of the new functionality, then it might be good to get some turned into tests now, so that we can verify the other changes as we finalize them.

drwells:
Hm, I think I agree with @boyceg here upon further inspection: there were no other changes on the NWT branch made to these examples. Did we do anything else to correct these examples? If not it would make more sense to add the tests now.

amneetb:
Ok, I can do this now. Assuming you are explaining it to a 2 year old, what branch should I work on, what version of libMesh and third-party library (I am assuming it is still google-test?) should I build for this purpose?

drwells:
this branch (its on `IBAMR/IBAMR`; you should be able to push commits), any version of libMesh newer than 1.1, python 3.5 to run the brand-new test suite.

boyceg:
@drwells, this could be a good time to put together some brief instructions on adding a test to the test suite.

drwells:
see `doc/testrunner.html` to get started.

drwells:
that document doesn't explicitly show how one can add a new test; that might be less obvious than I think it is. I'll write some updates but it should be enough to give @amneetb an idea about what to do.

boyceg:
Can these instructions be posted in whatever location we are primarily using for developer documentation (which I think is the GitHub project Wiki)?

drwells:
I was not aware we used the wiki. Sure!

boyceg:
We have been intending to use the GitHub Pages page to provide user-oriented documentation and the GitHub project wiki to provide developer/contributor-oriented documentation.

drwells:
Thats reasonable. I updated the relevant wiki page with a 'hello world' test example and a description of how to write new tests. @amneetb you will also need to install numdiff to run the tests (most package managers have it these days).

@amneetb if you have any questions, or things that take you more than about 15 seconds to figure out, let me know. I wrote the new test suite so everything seems very obvious to me and I may have missed something in the documentation.

amneetb:
Can you point me to the Wiki :-) ?

amneetb:
Ah, found it.

drwells:
The major goal of the test suite is that 'log files are now test outputs'; I hope the rest is just simple book-keeping.

amneetb:
One quick question, does the test suite works only with one .cpp file, i.e, the file containing `main()`? For many of the multiphase flow examples, we have multiple *.cpp and *.h files defining densities, viscosities, tagging, etc.

drwells:
Good question. The answer is 'no': one can add a new test that starts with

```cpp
#include "header.h"
```
where `header.h` is in the current directory and it works correctly. Multiple source files also work.

drwells:
Let me clarify that: to use multiple source files I think it would be best to `#include` them at the present. Writing
```make
new_test_01_SOURCES = new_test_01.cpp common.cpp
new_test_02_SOURCES = new_test_02.cpp common.cpp
```
*probably* works but I have not tried it.

amneetb:
Got it. I will give it a shot.

amneetb:
@boyceg: What should I do to update python to 3.5 and get numdiff on OSX? 

asbhalla@Amneets-MacBook-Pro:~$ python --version
Python 2.7.10

boyceg:
brew install python3

amneetb:
I guess default python is still 2.7.10, so I have to use python3 explicitly. But I didn't get numdiff. There is no brew formula for numdiff. Should I follow this instead https://pypi.org/project/Numdifftools/ ?

asbhalla@Amneets-MacBook-Pro-2:~$ python --version
Python 2.7.10
asbhalla@Amneets-MacBook-Pro-2:~$ python3 --version
Python 3.7.3
asbhalla@Amneets-MacBook-Pro-2:~$ which numdiff
asbhalla@Amneets-MacBook-Pro-2:~$ brew install numdiff
Error: No available formula with the name "numdiff" 
==> Searching for a previously deleted formula (in the last month)...
Warning: homebrew/core is shallow clone. To get complete history run:
  git -C "$(brew --repo homebrew/core)" fetch --unshallow

Error: No previously deleted formula found.
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
==> Searching taps on GitHub...
Error: No formulae found in taps.

drwells:
@amneetb numdiff is a separate executable; its not a python project. apt packages it but I guess brew does not. I think you will have to compile it yourself to use the test suite (sorry). Its an autotools project and is available at 

http://savannah.spinellicreations.com/numdiff/numdiff-5.9.0.tar.gz

amneetb:
Getting an error during compilation:
 
config.status: WARNING:  'Makefile.in' seems to ignore the --datarootdir setting
dyld: Library not loaded: /usr/local/opt/readline/lib/libreadline.7.dylib
  Referenced from: /usr/local/bin/awk
  Reason: image not found
./config.status: line 901: 30391 Done(141)               eval sed \"\$ac_sed_extra\" "$ac_file_inputs"
     30392 Abort trap: 6           | $AWK -f "$tmp/subs.awk" > $tmp/out


I don't have libreadline.7.dylib but have libreadline8.dylib. 
$ ls /usr/local/opt/readline/lib/
libhistory.8.0.dylib   libhistory.dylib       libreadline.a          
libhistory.8.dylib     libreadline.8.0.dylib  libreadline.dylib      
libhistory.a           libreadline.8.dylib    pkgconfig/     

Should I create a symbolic link libreadline.7.dylib to libreadline.8.dylib?

boyceg:
I haven’t checked, but is there a brew package for this software?

If not, @drwells, can we switch to something that is easy to get on MacOS?

amneetb:
Ah, it says `nongnu`. Doesn't that raise an eyebrow?

drwells:
I've been using numdiff for years. Let me see if there is something recommended for macs in particular.

amneetb:
Numdiff looks very interesting and useful!

https://groups.google.com/forum/#!topic/dealii/DINdUMGsxoE

Seems like deal.ii folks have installed numdiff on High-Sierra, which is what I am using (or maybe I am misreading the post).

amneetb:
There is something here for OSX (search numdiff)

https://finufft.readthedocs.io/en/latest/install.html

drwells:
I think Luca used spack to install it. I would prefer not to invoke that for a simple command-line utility.

The error message
```
dyld: Library not loaded: /usr/local/opt/readline/lib/libreadline.7.dylib
Referenced from: /usr/local/bin/awk
Reason: image not found
```
is very suspicious. This sounds like a really dumb question, but that installation of `awk` does work correctly, right?

drwells:
> Should I create a symbolic link libreadline.7.dylib to libreadline.8.dylib?
Maybe? This seems like a very odd problem to have.

boyceg:
Creating your own soft links between different library versions is not usually a good thing to do.

amneetb:
> I think Luca used spack to install it. I would prefer not to invoke that for a simple command-line utility.
> 
> The error message
> 
> ```
> dyld: Library not loaded: /usr/local/opt/readline/lib/libreadline.7.dylib
> Referenced from: /usr/local/bin/awk
> Reason: image not found
> ```
> 
> is very suspicious. This sounds like a really dumb question, but that installation of `awk` does work correctly, right?

Hmm, maybe awk is at fault here. 

asbhalla@Amneets-MacBook-Pro-2:~/Softwares/tars/numdiff-5.9.0$ awk --version
dyld: Library not loaded: /usr/local/opt/readline/lib/libreadline.7.dylib
  Referenced from: /usr/local/bin/awk
  Reason: image not found
Abort trap: 6

Boyce: Should I get awk from brew?

amneetb:
Ok, I made this work. I had to updgrade gawk through brew and had to include CFLAGS and LDFLAGS for gettext 

`./configure  CC=clang --prefix=/Users/asbhalla/Softwares/NumDiff CFLAGS=-I/usr/local/opt/gettext/include  LDFLAGS=-L/usr/local/opt/gettext/lib`

amneetb:
David can we use a separate keyword for log file that is meant for only test purposes? The plog will be dumping data to file indicated by the keyword `log_file_name`. 

`Main {
// ...
   log_file_name               = "output"
   log_all_nodes               = FALSE
// ...
}`

amneetb:
`-L` option does not work for `ln` in OSX. Suggest removing it from Makefile.am

https://github.com/deepmind/lab/issues/87

amneetb:
```
if test "../../../IBAMR" != "../.." ; then \
	  ln -L -f -s ../../../IBAMR/tests/IBTK/*input /Users/asbhalla/Softwares/IBAMR-Git/objs-dbg/tests/IBTK ; \
	  ln -L -f -s ../../../IBAMR/tests/IBTK/*output /Users/asbhalla/Softwares/IBAMR-Git/objs-dbg/tests/IBTK ; \
	fi ;
ln: illegal option -- L
usage: ln [-Ffhinsv] source_file [target_file]
       ln [-Ffhinsv] source_file ... target_dir
       link source_file target_file
ln: illegal option -- L
usage: ln [-Ffhinsv] source_file [target_file]
       ln [-Ffhinsv] source_file ... target_dir
       link source_file target_file
make[2]: *** [tests] Error 1
make[1]: *** [IBTK] Error 2
make: *** [tests] Error 2

```

drwells:
@amneetb 
> David can we use a separate keyword for log file that is meant for only test purposes?

We only have one `plog` object: I don't think there is any way to split its output. I think the best way to separate relevant test output from normal logging done by the library would be to create an output stream that writes to `output`, e.g., `std::ofstream test_output("output")` and pass that object around the test.

boyceg:
I have never tried it, but I think that it is possible to setup something that works like plog using SAMRAI::box::ParallelBuffer.

drwells:
I just rebased this on top of master to get the `ln` fix in #551.

amneetb:
Currently `attest` script is linked to the build directory only with `make tests`. This is fine. One disadvantage is that one needs to build all the tests to use this script. Users who would like to do `make tests` for one particular test won't find `attest` script in their build directory. Can we distribute this script to the build directory as a part of `make lib`? 

```
tests: lib attest.conf
        ln -L -f -s $(top_srcdir)/attest $(abs_builddir)
        @cd $@ && $(MAKE) $(AM_MAKEFLAGS) $@

```

I removed `-L` option from here in my recent commit.

amneetb:
> @amneetb
> 
> > David can we use a separate keyword for log file that is meant for only test purposes?
> 
> We only have one `plog` object: I don't think there is any way to split its output. I think the best way to separate relevant test output from normal logging done by the library would be to create an output stream that writes to `output`, e.g., `std::ofstream test_output("output")` and pass that object around the test.

Yes, I figured this trick of `std::ofstream test_output("output")` later.

drwells:
@amneetb would you mind if I rewrote this branch's history to get rid of the merge? Merge commits always make repository history much more complex and we can absolutely avoid it in this case by rebasing.

amneetb:
Not at all. The light is green from my end.

drwells:
@amneetb this crashes after 100 time steps with a segmentation fault. Could you please take a look?

amneetb:
It happens in parallel. If I increase timesteps to 200, it runs fine at timestep 100. It probably has to do with some cleanup operation when the program is trying to close. I ran with `-on_error_attach_debugger` but `lldb` does not capture the segfault on exit on my Mac. Can you run the executable on linux? I am trying some other options in the meantime

`mpiexec -np 4 ./free_falling_cyl_constraint_ib free_falling_cyl_constraint_ib.mpirun\=4.input -on_error_attach_debugger`

amneetb:
Ok, I fixed this. Give it a shot.

boyceg:
I think we should try to get the underlying bug fixed and not rely on the user using a particular regrid interval / number of time steps.

amneetb:
> I think we should try to get the underlying bug fixed and not rely on the user using a particular regrid interval / number of time steps.

I agree. 

The issue I think is that PETSc data structures related to Lagrangian data are somehow "out-of-sync" when the program closes.

drwells:
I think there may be another issue with this example: the calculated time step size goes to zero when we use `N = 64` instead of `N = 220`.

amneetb:
If by `N` you mean grid size, then yes, it is not surprising. The grid parameters are set in a specific way to satisfy CFL and other stability conditions. Why do you need to change them?

drwells:
I only did for catching the segmentation fault. Its not a big deal if we already know about this and its related to some other stability property.

amneetb:
I see. Yes, these parameters (Eulerian + Lagrangian both) are chosen to take into account CFL and ConstraintIB stability.

boyceg:
Can we push the button?

drwells:
We still need to add one change to `tests/Makefile.am`: IIRC the test is not compiled correctly at the moment.

drwells:
I must have been thinking of something else re: the makefile: this test builds correctly but fails to run (it still segmentation faults). I am investigating.

drwells:
Wait, no: my copy of the executable was left over from when I fixed the makefile and was not recompiled with the fix from #564.

boyceg:
OK --- well if this is working, let's go ahead and merge?

drwells:
It doesn't currently compile the test, so not quite. Give me a minute.

amneetb:
Let me know when this gets fixed, as I will need to follow this test example for creating the rotation test.

drwells:
That was a little trickier than I thought it would be: we have to be careful with auxiliary data files.

This should be done now.

amneetb:
@drwells: What was the issue? I could not figure it out from the forced push.

amneetb:
https://github.com/IBAMR/IBAMR/tree/master/tests/multiphase_flow

I don't see any cpp file and test input files?

boyceg:
I guess we got rid of the segfault by removing the test!

amneetb:
Why do I not think of simple solutions?

boyceg:
I had a copy of this branch that included Amneet's test. I just pushed it to `two-phase-regrid-old`. I am not sure if there are changes to the test itself that are missing; the last commit was from 4/30.

drwells:
No, there should have been a test. @boyceg I added all of the makefile code for a reason :crying_cat_face:

boyceg:
> No, there should have been a test. @boyceg I added all of the makefile code for a reason 😿

You mean it wasn't because you love Automake?

I think that the majority of the changes needed to make the test work were bug fixes in the library itself, so hopefully the old version of the branch that I found will do the trick. However, maybe this time I will defer to David to push the button.

drwells:
I simply forgot to commit the test code. I have it on another branch on my machine.

I had to move the test into a subdirectory of `tests` instead of a sub-sub directory to placate the current implementation of the 'find tests' code. I also merged all the source files while I was at it. I simply forgot to stage it.

PR #552, opened 2019-04-30
Fix side length scale

drwells:
This fixes the problem outlined in #543.


drwells:
@cpuelz this fixes the IBFE examples in debug mode.

boyceg:
I think it might be better to use the edge length instead of the overall element size, although I think that the worst case scenario here is that we may wind up using more quadrature points than we might need for "short" edges on highly distorted elements.

drwells:
For posterity's sake: @boyceg and I talked about this briefly in person and we decided that, for now, since this function is not performance critical, its okay to be a bit more conservative and use more quadrature points instead of writing something much more complex that computes a length scale for each side.

PR #553, opened 2019-04-30
fix compilation with libmesh 1.4

jabrown893:


ibtk/src/utilities/libmesh_utilities.cpp
```diff
@@ -40,6 +40,10 @@
 #include "libmesh/numeric_vector.h"
 #include "libmesh/system.h"
```
boyceg:
```suggestion
```

boyceg: Let's try to keep the `#include` blocks grouped together, so that `clang-format` will keep them in order.

jabrown893: Got it. Thanks!

PR #556, opened 2019-05-01
fix line breaks in tester message.

drwells:
I put the line breaks in weird places before.


amneetb:
Does this fix my recently added test? Probably not, right?

drwells:
Its not relevant: this just prints a better-formatted error message in case `attest` finds, e.g., an input file for a test but no matching executable.

PR #557, opened 2019-05-01
Assemble into ghost regions in interpWeighted.

drwells:
Followup to #419, though this is a lot easier now since we merged a lot of stuff in the mean time. Part of #492.

This actually passes the test suite (!) but I still need to convince myself that this is the right way to do it.

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -1541,13 +1541,29 @@ FEDataManager::interpWeighted(const int f_data_idx,
     else
     {
         // Extract local form vectors.
-        auto X_petsc_vec = static_cast<PetscVector<double>*>(&X_vec);
+        auto X_petsc_vec = dynamic_cast<PetscVector<double>*>(&X_vec);
+        TBOX_ASSERT(X_petsc_vec != nullptr);
         const double* const X_local_soln = X_petsc_vec->get_array_read();
+        // Since we do a lot of assembly in this routine into off-processor
+        // entries we will directly insert into the ghost values (and then
+        // scatter in the calling function with the usual batch function).
+        F_vec.zero();
+        auto F_petsc_vec = dynamic_cast<PetscVector<double>*>(&F_vec);
+        Vec F_local_form = nullptr;
+        double *F_local_soln = nullptr;
+        const bool is_ghosted = F_vec.type() == GHOSTED;
+        if (is_ghosted)
+        {
+            TBOX_ASSERT(F_petsc_vec != nullptr);
+            int ierr = VecGhostGetLocalForm(F_petsc_vec->vec(), &F_local_form);
+            IBTK_CHKERRQ(ierr);
+            ierr = VecGetArray(F_local_form, &F_local_soln);
+            IBTK_CHKERRQ(ierr);
+        }
 
         // Loop over the patches to interpolate values to the element quadrature
         // points from the grid, then use these values to compute the projection
         // of the interpolated velocity field onto the FE basis functions.
-        F_vec.zero();
         DenseVector<double> F_rhs;
         // Assemble F_rhs_e's vectors in an interleaved format (see the implementation):
         std::vector<double> F_rhs_concatenated;
@@ -1719,7 +1735,18 @@ FEDataManager::interpWeighted(const int f_data_idx,
 
                     dof_id_scratch = F_dof_indices[i];
                     F_dof_map.constrain_element_vector(F_rhs, dof_id_scratch);
-                    F_vec.add_vector(F_rhs, dof_id_scratch);
+                    if (is_ghosted)
+                    {
+                        for (unsigned int i = 0; i < dof_id_scratch.size(); ++i)
+                        {
+                            F_local_soln[F_petsc_vec->map_global_to_local_index(dof_id_scratch[i])]
+                                += F_rhs(i);
+                        }
+                    }
+                    else
+                    {
+                        F_vec.add_vector(F_rhs, dof_id_scratch);
+                    }
                 }
                 qp_offset += n_qp;
             }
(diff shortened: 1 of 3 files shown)


ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1541,13 +1541,29 @@ FEDataManager::interpWeighted(const int f_data_idx,
     else
     {
         // Extract local form vectors.
-        auto X_petsc_vec = static_cast<PetscVector<double>*>(&X_vec);
+        auto X_petsc_vec = dynamic_cast<PetscVector<double>*>(&X_vec);
```
boyceg: We should probably rewrite all of this to just take in `PetscVector`s instead of `NumericVector`s.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1719,14 +1735,40 @@ FEDataManager::interpWeighted(const int f_data_idx,
 
                     dof_id_scratch = F_dof_indices[i];
                     F_dof_map.constrain_element_vector(F_rhs, dof_id_scratch);
-                    F_vec.add_vector(F_rhs, dof_id_scratch);
+                    if (is_ghosted)
+                    {
+                        for (unsigned int i = 0; i < dof_id_scratch.size(); ++i)
+                        {
+                            F_local_soln[F_petsc_vec->map_global_to_local_index(dof_id_scratch[i])]
```
boyceg: Can we avoid this lookup for DOF ids that are owned by the processor?

boyceg: (We would still need to do the lookup for DOF ids owned by other processors, to find where those entries are stored in the ghost region.)

src/IB/IBFEMethod.cpp
```diff
@@ -988,7 +988,8 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
                                                  /*close_F*/ false,
                                                  /*close_X*/ false);
     }
-    batch_vec_assembly(d_U_rhs_vecs);
+    batch_vec_ghost_update(d_U_rhs_vecs, ADD_VALUES, SCATTER_REVERSE);
```
boyceg: Do we need to have a switch depending on whether the vector is `GHOSTED` or not?

ibtk/src/lagrangian/FEDataManager.cpp
drwells: We might as well since we already `static_cast` to that derived class: no one's code is going to work if they supply any other vector here.

ibtk/src/lagrangian/FEDataManager.cpp
drwells: Yes and no. If you look at the code for `map_global_to_local_index` it starts with something like

```cpp
if (proc_dof_start <= dof && dof < proc_dof_end)
    return dof - proc_dof_start;
```
so the translation is cheap for on-processor dofs: libMesh stores a `map` for dofs in the ghost region which is more expensive to use.

We could avoid it by inlining that check ourselves (which I think the compiler may actually do) but I would prefer not to do it since we are going to cache the indexing anyway in #513.

src/IB/IBFEMethod.cpp
drwells: Yup. I will fix this.

I also need to add some more tests that verify things work with both ghosted and unghosted vectors.

boyceg: > I also need to add some more tests that verify things work with both ghosted and unghosted vectors.

Does it make sense for there to be a flag in `IBFEMethod` that determines the vector type?

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: Oh yeah, now I remember seeing this when I was looking at the libMesh code. I agree there is no need for us also to have that check here.

src/IB/IBFEMethod.cpp
drwells: Hm. IBFEMethod already passes ghosted vectors to this function, but CIBFEMethod does not. I would prefer to use ghosted vectors internally for this type of assembly.

What would this flag do? I am not sure that I understand your proposal.

boyceg: Sure: The flag would determine if we use ghosted or non-ghosted vectors in this interpolation routine. Right now we always use ghosted vectors in `IBFEMethod` for this stuff, but it should be easy enough to switch to non-ghosted. This would make it easy to measure the effect of switching from using `VecSetValues` and the internal stash, to using fully local assembly followed by a scatter.

(Note that the coordinate vectors will always have to be ghosted so that we know where the interaction points are.)

drwells: :+1:

boyceg: To try to avoid confusion: My proposal is that in `IBFEMethod`, we would dynamically determine the type of `d_U_IB_rhs_vecs` when we set those vectors up.

This could also help to test/verify the implementation is doing the right thing.

drwells:
The flag was added in #574.

src/IB/IBFEMethod.cpp
boyceg: I think that we still need to have some logic here to decide how to assemble, and also possibly in `IBFEMethod` and other places where we batch assemble following calls to `interpWeighted()`.

drwells: Absolutely. In fact, the new test that does not use ghosted vectors for assembly now fails on this branch.

drwells:
I am running the heart model with 4 processors right now to benchmark this. It looks like this is perhaps 1% slower than the current approach. I will have to do some more extensive profiling to see exactly what is slowing it down.

drwells:
d'oh, I used second order elements in one place and first order in the other. Let me try that again.

boyceg:
What's up with the timings?

drwells:
I think I need to do some more comprehensive timings before I can accurately assess the changes. I am going to run the heart model with valgrind; I also should finally learn how to use open speed shop to get 'real' profiling results on a cluster. I will handle at least the first one tomorrow.

drwells:
Some tentative results, via callgrind and four processes (but *very* coarse grid resolution, N = 64 on finest level)
1. doing global-to-local index translation via libMesh's petsc vector is pretty cheap: we now spend about 0.3% of our total time doing it.
2. We spend 4% of our time doing vector assembly on `master` vs. 1.2% with this patch.

I expect that this will perform better with more processes. I finally need to learn how to use a 'real' MPI profiler; I am working on that now.

boyceg:
@drwells, is there anything else we want to do in this branch before merging?

drwells:
Nope. I am looking at the callgrind output now (it ran overnight): I'll merge now but post a summary of the results once I finish reading it.

drwells:
It looks like this, with callgrind and four processors, lowers the total number of instructions executed by about 3%. Since this is a bottleneck I suspect that we should perform at least this much better when we run on a 'real' MPI network, and the percentage should increase a little as we add more MPI processes since we are decreasing the amount of traffic.

PR #558, opened 2019-05-02
Routines for interpolating from face/side data to node data

boyceg:


drwells:
@abarret would you be willing to write some tests for this?

boyceg:
In fact I think there is a bug in the error checking. I am planning on writing some tests but may not get to it this week. If anyone else wants to do it in the meantime, let me know.

drwells:
@boyceg what is the plan for this PR?

boyceg:
@abarret, @xxjaehoxx, we might want to revisit this along with #1408. I think that it may be easier to deal with the mean flows for _analysis_ purposes if they are interpolated onto nodes.

drwells:
This is mostly done - I'll find some time to wrap it up ASAP.

drwells:
Quick poll - how about we change the default viz. format for staggered grids to nodal data? It makes more sense than cell-centered. We can make cell-centered an input database option to restore the previous behavior.

boyceg:
Fine with me! Also for vorticity and force!

Pressure is actually cell centered so seems to make sense to leave it.

abarret:
The plots in visit look so much nicer with nodal values.

drwells:
This is done - there were some minor indexing bugs (see 7344b55b7491a679ad74dcf3fea992ebc606159d) that needed to be fixed but we can now do second-order accurate interpolation from face or side data to nodal. I'll redo the viz output in the next set of patches.

boyceg:
@abarret can you review / approve? I cannot do it because I started the PR.

drwells:
I already 'approved' it so I can just merge my own changes 👿

drwells:
The test suite passes and this only adds new features so I'll just merge it as-is.

ibtk/src/math/HierarchyMathOps.cpp
```diff
@@ -1930,6 +1930,102 @@ HierarchyMathOps::interp(const int dst_idx,
...
 } // interp
 
+void
+HierarchyMathOps::interp(const int dst_idx,
+                         const Pointer<NodeVariable<NDIM, double> > /*dst_var*/,
+                         const int src_idx,
+                         const Pointer<FaceVariable<NDIM, double> > /*src_var*/,
+                         const Pointer<HierarchyGhostCellInterpolation> src_ghost_fill,
+                         const double src_ghost_fill_time,
+                         const bool src_cf_bdry_synch)
```
boyceg, line 1940: I should have noticed this earlier, but for this `interp` function, it probably is necessary also to have an option for `dst_cf_bdry_synch`. Currently I don't think that the nodal representation will generally have consistent values at coarse-fine interfaces.

PR #561, opened 2019-05-03
update Sublime project template

boyceg:
These settings tell `cppcheck` to assume that header (`.h`) files are in C++ and not C.


boyceg:
I am going to go ahead and merge because this only affects a template project file.

PR #564, opened 2019-05-06
Do not remove an unused patch data index.

drwells:
Fix for the issue in #550: we were removing patch data that was not allocated which causes SAMRAI to segfault, even in debug mode.

I am writing up some notes on how to debug things with valgrind (and other tools) with MPI so that everyone can learn the tricks that I just learned.

### DIFF
--- a/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
+++ b/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
@@ -186,7 +186,10 @@ IBHydrodynamicSurfaceForceEvaluator::~IBHydrodynamicSurfaceForceEvaluator()
     var_db->removePatchDataIndex(d_ls_solid_idx);
     var_db->removePatchDataIndex(d_u_idx);
     var_db->removePatchDataIndex(d_p_idx);
-    var_db->removePatchDataIndex(d_mu_idx);
+    if (!d_mu_is_const)
+    {
+        var_db->removePatchDataIndex(d_mu_idx);
+    }
 
     return;
 } // ~IBHydrodynamicSurfaceForceEvaluator


amneetb:
Thanks, David for doing the heavy lifting. Much appreciated.

drwells:
See also https://github.com/IBAMR/IBAMR/wiki/Debugging-Tips

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -186,7 +186,10 @@ IBHydrodynamicSurfaceForceEvaluator::~IBHydrodynamicSurfaceForceEvaluator()
     var_db->removePatchDataIndex(d_ls_solid_idx);
     var_db->removePatchDataIndex(d_u_idx);
     var_db->removePatchDataIndex(d_p_idx);
-    var_db->removePatchDataIndex(d_mu_idx);
+    if (!d_mu_is_const)
```
boyceg: Should we default the values of these cached patch data indices to `IBTK::invalid_index` and check for that instead?

OTOH, it seems unlikely that this logic will ever change --- if mu is non-constant, then there needs to be a patch data index to store the result.

drwells: The indices already have the correct default value. I think this is a little clearer than checking the index since it mirrors the condition above where we decide whether or not to allocate, but I could go either way on this one. Do you prefer one or the other?

boyceg: I do not have a strong preference for this particular implementation. In general, I guess that I prefer a construct like:
```suggestion
    if (d_mu_idx != IBTK::invalid_index)
```
so that the implementation can change the logic that determines whether the index is set up without having to also change the logic to check that the index should be removed.

Again, in this case, I don't really care that much.

Random thought: Should we have a first-class data type for patch data indices that generates an error or warning if you try to convert an invalid index to an integer?

drwells:
I am lazy so lets keep it as I originally wrote it :)

PR #562, opened 2019-05-03
routines for computing mean flow quantities and turbulent kinetic energy

boyceg:

PR #565, opened 2019-05-06
fix some warning messages

boyceg:


drwells:
This looks good ATM: adding `const` should always be okay (as long as things compile) and we do not use `d_ec_idx` anywhere.

boyceg:
I get one more unused variable warning, but it is fixed in #562.

PR #566, opened 2019-05-06
Remove some unnecessary pointers.

drwells:
The type of these objects is not dynamic, so we might as well just use values and delete the destructor.

### DIFF
--- a/include/ibamr/StaggeredStokesSolver.h
+++ b/include/ibamr/StaggeredStokesSolver.h
@@ -37,6 +37,7 @@
 
 #include <vector>
 
+#include "LocationIndexRobinBcCoefs.h"
 #include "PoissonSpecifications.h"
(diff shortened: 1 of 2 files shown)


include/ibamr/StaggeredStokesSolver.h
```diff
@@ -67,11 +68,6 @@ class StaggeredStokesSolver : public virtual IBTK::GeneralSolver
      */
     StaggeredStokesSolver();
 
-    /*!
-     * \brief Destructor.
-     */
-    ~StaggeredStokesSolver();
```
abarret: I think we've adopted the ` = default` syntax instead of removing the definition completely.

drwells:
That makes sense, but since its an implementation detail I would rather remove the allocation.

PR #548, opened 2019-04-29
NWT: multiphase_flow examples 10, 11, 12, and 13

drwells:
Part of #546. As mentioned in #546 we cannot merge this until `ConstraintIBMethod::getFullLagrangeMultiplierForce` and `ConstraintIBMethod::getInterpolatedLagrangianDensity` are in the library: hence this is do not merge until we get that done.

PR #567, opened 2019-05-06
Allowing for rotation even when translation locked

drwells:
This change is required for #548.

I am not familiar enough with this part of the code to evaluate this change. @amneetb would it be possible to write a simple test that verifies we can rotate a stationary object correctly?


amneetb:
For just a rigid body the code is very simple and I trust it. The more complex situation is a rotation in the presence of fluid, and ex11 tests that (free decay test of a rectangular barge). We compared it with experimental results. I can rerun the simulation, compare it with experiments again, and record the rotational angle for few 100 time steps. Would that be useful?  

(ConstraintIB does not work as a standalone class. It needs to do some FSI problem)

drwells:
> I can rerun the simulation, compare it with experiments again, and record the rotational angle for few 100 time steps. Would that be useful?

Sounds good to me.

drwells:
> For just a rigid body the code is very simple and I trust it. 

I read this comment again and I think I need to clarify part of the reason we want to write tests. I don't doubt that all of this code already works. The main point of adding tests at this stage is that we can guarantee it will *continue* to work even after the SAMRAI 3 transition, optimizing other algorithms, new MPI versions, etc. Even very simple tests are valuable for this purpose.

amneetb:
@drwells : I have time today to work on the rotation test. What branch should I checkout? Also, the said branch has test/multiphase included in it or not? I would strongly prefer to have that.

drwells:
@amneetb The changes are on this branch. It does not include the examples; those are in #548.

amneetb:
I am not sure how I will add the test then. Should I add a new test directory named rotation test?

drwells:
Just write a test that exposes the new rotation functionality: it can be super simple and would be better if it were not a converted example (there is not that much new code on this branch).

Since the new features are in `src/IB` I suggest adding the test in `tests/IB`.

amneetb:
@drwells : The rotational code in ConstrainIBMethod relies on angular momentum conservation principle. To test this code you need to solve an actual FSI problem e.g., a free falling plate in air etc. To do this you need a different mass of the body than air, which ties in multiphase code. You are probably imagining "rotating a line in a plane" kind of a simple example. This class can't do this. Rotational tests for FSI problems are rare in the literature :-/

drwells:
Would it be enough to use manufactured solutions, i.e., prescribed density and velocity fields?

amneetb:
ConstraintIBMethod does only FSI, prescribed kinematics or free moving bodies in liquids. Prescribed kinematics is a no brainer. Also, the new feature added is for freely rotating bodies. The only reason this code was added so late was that we could not do different mass densities of bodies and fluids to test free rotation FSI before. Only recently we are able to do these kinds of cases, thanks to multiphase flow infrastructure.

amneetb:
If you want to test the new code in this PR, it has to be a freely rotating body. i.e., no prescribed velocities. Setting up prescribed densities means you are pulling in all of the multiphase flow infrastructure, so might as well do the rotating barge case?

boyceg:
Yes — rotating barge.

drwells:
It sounds like there is no way to test this in isolation, so the barge example would be reasonable if that is the best way to make sure all the new code gets run.

amneetb:
I am happy to write a test for rotating barge. I think we should replay this PR on top of master then? 
I am still not sure about the test directories name. I would need to create a new directory named tests/rotating_barge?

drwells:
We don't need separate test directories for separate tests; e.g., all of the IBTK tests are in `tests/IBTK`. I would prefer not to proliferate test directories (we do not want hundreds of test directories).

I will rebase in a moment.

drwells:
rebased.

amneetb:
Done.

drwells:
I'll take a look :)

drwells:
Do we have code to generate the barge vertex file? It would be pretty straightforward to run that (so that we do not need to include a 5000 line input file) at the start of the test.

amneetb:
https://github.com/IBAMR/IBAMR/blob/amneetb/numerical_wave_tank/examples/multiphase_flow/ex11/Barge2d.m

boyceg:
Random idea --- while we are in the process of updating these scripts to use either Python or C++, could we detect whether MATLAB is installed and run the script from the command line, and tag things that "should" be updated?

drwells:
Most matlab scripts aren't really well suited for batch processing; we would have to modify them to get that to work. I would prefer to just keep converting them as they pop up.

drwells:
@amneetb The test doesn't pass (it fails to run): would you please take a look?

amneetb:
It was running before my addition of bargeGen stuff. Since `attest` leaves no evidence of crime (not even the output file or log file or any temporary directory) I don't know what I need to do to fix it.

If I run makes tests, I see the vertex file. The executable run, and if I `numdiff` on output file and `rotating_barge_cib.mpirun=4.output`, it says the files are equivalent. AFAIK, I did everything correctly.

drwells:
Try running with `attest --keep-work-directories` to preserve the temporary directories it keeps. Making `attest` less laconic is on the todo list.

drwells:
The temp directories are usually in `/tmp/`.

amneetb:
I still don't see any directory. Probably it will take you 2 seconds to figure this out?

drwells:
I'll take a look.

tests/multiphase_flow/Makefile.am
```diff
@@ -1,13 +1,24 @@
...
+rotating_barge_cib_CXXFLAGS = $(AM_CXXFLAGS) -DNDIM=2 -DSOURCE_DIR=\"$(abs_srcdir)\"
+rotating_barge_cib_LDADD = $(IBAMR_LDFLAGS) $(IBAMR2d_LIBS) $(IBAMR_LIBS)
+rotating_barge_cib_SOURCES = rotating_barge_cib.cpp
+
+bargeGen_CXXFLAGS = $(AM_CXXFLAGS)
+bargeGen_LDADD =
+bargeGen_SOURCES = bargeGen.cpp
+
+barge2d.vertex: bargeGen
+	./bargeGen
```
drwells: This isn't going to work for the test suite since tests are always executed in a temporary directory. We will need to generate this file for the test.

tests/multiphase_flow/rotating_barge_cib.cpp
```diff
@@ -94,15 +94,6 @@ void output_data(Pointer<PatchHierarchy<NDIM> > patch_hierarchy,
 int
 main(int argc, char* argv[])
 {
-    // Several parts of the code (such as LDataManager) expect mesh files,
-    // specified in the input file, to exist in the current working
-    // directory. Since tests are run in temporary directories we need to regenerate these input
-    // to work.
```
drwells: I left this comment in the test for a reason :)

drwells:
The test looks good and passes for @amneetb and myself so lets merge once the CI build finishes.

PR #570, opened 2019-05-07
avoid using raw pointers in IBTK

boyceg:


drwells:
I was going to tackle these slowly, but here they all are!

boyceg:
This is what I do when I am home with a sick kid.

boyceg:
FWIW, I am planning to do the same thing for the IBAMR classes in a separate PR.

abarret:
My only concern with these changes is that it's not immediately clear to me which class owns the underlying data without looking at the implementation. Should we update function signatures with a `unique_ptr` plus `std::move` syntax instead of passing in raw pointers? As a specific example, you can look at `HierarchyIntegrator::registerGhostfillRefineAlgorithm`.

boyceg:
> My only concern with these changes is that it's not immediately clear to me which class owns the underlying data without looking at the implementation. Should we update function signatures with a `unique_ptr` plus `std::move` syntax instead of passing in raw pointers? As a specific example, you can look at `HierarchyIntegrator::registerGhostfillRefineAlgorithm`.

I am pretty sure that we are passing ownership in the vast majority of cases, and so `std::unique_ptr` would be appropriate. I was thinking about making that change at the same time. Let me take another look now.

abarret:
> I am pretty sure that we are passing ownership in the vast majority of cases, and so `std::unique_ptr` would be appropriate.

Yes. I agree; these classes assume ownership. But is it clear from a user's perspective that these classes assume ownership, by e.g. invalidation pointers?

boyceg:
> > I am pretty sure that we are passing ownership in the vast majority of cases, and so `std::unique_ptr` would be appropriate.
> 
> Yes. I agree; these classes assume ownership. But is it clear from a user's perspective that these classes assume ownership, by e.g. invalidation pointers?

Not currently.

It looks like this is only relevant for `HierarchyIntegrator`. Perhaps we could deprecate the raw pointer-based interface and add a preferred interface that uses `std::unique_ptr`.

The problem with changing to `std::unique_ptr` is that it changes the API, which breaks, for instance, `IBHierarchyIntegrator`.

boyceg:
OK, looking at this some more; really this is shared, but with `HierarchyIntegrator` taking ownership for ultimately cleaning up the allocated memory. In particular, we cannot blindly convert to `std::unique_ptr` in the API and have everything just work, because in some places we keep a hold of the pointer and do things with it.

It would be OK to convert to a `std::shared_ptr`, but that requires a lot of changes in the library, and could require changes in user code.

I think if we make those changes, they should go in a different PR.

Thoughts?

abarret:
There might be a similar issue with `CoarsenPatchStrategySet` and `RefinePatchStrategySet`.

In `HierarchyIntegrator` it's already not clear who has ownership of the data. Maybe we should make a note of it, and leave the `std::shared_ptr` changes until after the SAMRAI3 changes, which will break every API anyway?

I think it would be nice to get @drwells opinion on this.

boyceg:
> There might be a similar issue with `CoarsenPatchStrategySet` and `RefinePatchStrategySet`.

Those are slightly better because there is a flag in the constructor that says whether you want to assign ownership or not.

At some point, `CoarsenPatchStrategy` and `RefinePatchStrategy` could not be used with `tbox::Pointer`, which is why these functions all use raw pointers. I am not sure, but I think that the similar APIs in SAMRAI 3 now use `std::shared_ptr` with these classes.

ibtk/src/utilities/StandardTagAndInitStrategySet.cpp
```diff
@@ -56,13 +56,8 @@ namespace IBTK
...
-    {
-        for (const auto& strategy : d_strategy_set)
-        {
-            delete strategy;
-        }
-    }
+    if (!d_managed)
+        for (auto& strategy : d_strategy_set) strategy.release();
     return;
 } // ~StandardTagAndInitStrategySet
```
drwells: What is going on here? Won't the calls to `release` create memory leaks? I do not understand why we need this destructor.

ibtk/include/ibtk/StandardTagAndInitStrategySet.h
```diff
@@ -340,7 +340,7 @@ class StandardTagAndInitStrategySet : public SAMRAI::mesh::StandardTagAndInitStr
     /*!
      * \brief The set of SAMRAI::xfer:StandardTagAndInitStrategy objects.
      */
-    std::vector<SAMRAI::mesh::StandardTagAndInitStrategy<NDIM>*> d_strategy_set;
+    std::vector<std::unique_ptr<SAMRAI::mesh::StandardTagAndInitStrategy<NDIM> > > d_strategy_set;
```
drwells: We should now be able to get rid of `d_managed` since we are managing memory now.

ibtk/include/ibtk/CoarsenPatchStrategySet.h
```diff
@@ -160,7 +160,7 @@ class CoarsenPatchStrategySet : public SAMRAI::xfer::CoarsenPatchStrategy<NDIM>
     /*!
      * \brief The set of SAMRAI::xfer:CoarsenPatchStrategy objects.
      */
-    std::vector<SAMRAI::xfer::CoarsenPatchStrategy<NDIM>*> d_strategy_set;
+    std::vector<std::unique_ptr<SAMRAI::xfer::CoarsenPatchStrategy<NDIM> > > d_strategy_set;
```
drwells: We should also get rid of `d_managed` here.

ibtk/include/ibtk/LaplaceOperator.h
```diff
@@ -67,9 +67,9 @@ class LaplaceOperator : public LinearOperator
     LaplaceOperator(std::string object_name, bool homogeneous_bc = false);
 
     /*!
-     * \brief Destructor.
+     * \brief Default destructor.
```
drwells:
```suggestion
     * \brief destructor.
```
(or `defaulted destructor`)

ibtk/include/ibtk/PoissonSolver.h
```diff
@@ -65,9 +65,9 @@ class PoissonSolver : public virtual GeneralSolver
     PoissonSolver() = default;
 
     /*!
-     * \brief Destructor.
+     * \brief Default destructor.
```
drwells:
```suggestion
     * \brief destructor.
```

ibtk/src/lagrangian/LSiloDataWriter.cpp
```diff
@@ -1575,11 +1575,8 @@ LSiloDataWriter::writePlotData(const int time_step_number, const double simulati
             block_names_per_proc[ln].resize(mpi_nodes);
             mb_names_per_proc[ln].resize(mpi_nodes);
             ucd_mesh_names_per_proc[ln].resize(mpi_nodes);
-        }
 
-        // Set the values for the root process.
-        if (mpi_rank == SILO_MPI_ROOT)
-        {
+            // Set the values for the root process.
```
drwells: Why was this `if` statement deleted?

ibtk/src/utilities/CoarsenPatchStrategySet.cpp
```diff
@@ -56,13 +56,8 @@ namespace IBTK
...
 {
-    if (d_managed)
-    {
-        for (const auto& strategy : d_strategy_set)
-        {
-            delete strategy;
-        }
-    }
+    if (!d_managed)
+        for (auto& strategy : d_strategy_set) strategy.release();
```
drwells: Same issue with `release`.

ibtk/src/utilities/RefinePatchStrategySet.cpp
```diff
@@ -54,13 +54,8 @@ namespace IBTK
...
 {
-    if (d_managed)
-    {
-        for (const auto& strategy : d_strategy_set)
-        {
-            delete strategy;
-        }
-    }
+    if (!d_managed)
+        for (auto& strategy : d_strategy_set) strategy.release();
```
drwells: same issue with `release`.

ibtk/src/utilities/StandardTagAndInitStrategySet.cpp
boyceg: In the current interface, the caller gets to decide if they hand over management of the provided objects.

ibtk/include/ibtk/StandardTagAndInitStrategySet.h
boyceg: This will require changing the interface for the constructor -- currently callers get to decide if they want to hand over management.

ibtk/include/ibtk/CoarsenPatchStrategySet.h
boyceg: Same as above.

ibtk/include/ibtk/LaplaceOperator.h
boyceg: Why not "default"?

ibtk/include/ibtk/PoissonSolver.h
boyceg: Why not "default"?

ibtk/src/lagrangian/LSiloDataWriter.cpp
boyceg: Because it is redundant.

drwells:
okay, I understand the `release` code now. I think it would be better to keep the old version with `delete` in those cases.

boyceg:
> okay, I understand the `release` code now. I think it would be better to keep the old version with `delete` in those cases.

Yeah, I guess that is safer if we ever change the interface to use `std::unique_ptr` explicitly.

boyceg:
> > okay, I understand the `release` code now. I think it would be better to keep the old version with `delete` in those cases.
> 
> Yeah, I guess that is safer if we ever change the interface to use `std::unique_ptr` explicitly.

OK, I reverted those changes.

ibtk/include/ibtk/PoissonSolver.h
drwells: Writing 'default destructor.' like this makes it sound like, to me, there might be more than one destructor (just like how we label the default constructor). It seems odd but its not wrong.

boyceg: But it is literally the default destructor!

drwells: You're right. I think this is just a personal problem I have...

boyceg: [cppreference](https://en.cppreference.com/w/cpp/language/destructor) also calls it the defaulted destructor, which I think is perverse.

drwells: Thats the term I prefer :smile_cat:

ibtk/src/solvers/impls/CCPoissonHypreLevelSolver.cpp
```diff
@@ -1199,23 +1199,17 @@ CCPoissonHypreLevelSolver::copyToHypre(const std::vector<HYPRE_StructVector>& ve
                                        const CellData<NDIM, double>& src_data,
                                        const Box<NDIM>& box)
 {
+    const bool copy_data = src_data.getGhostBox() != box;
+    std::unique_ptr<CellData<NDIM, double> > src_data_box(
+        copy_data ? new CellData<NDIM, double>(box, src_data.getDepth(), 0) : nullptr);
+    CellData<NDIM, double>* hypre_data =
+        copy_data ? src_data_box.get() : const_cast<CellData<NDIM, double>*>(&src_data);
```
drwells: We could make this a little clearer by using a reference instead of a pointer.

boyceg:
> Would you be willing to go back and de-pointerize a few things? We do not technically need to use pointers for the members in
> 
> * `CartCelDoubleQuadraticCFInterpolation`
> * `CartSideDoubleQuadraticCFInterpolation`

I will change these.

> * `LaplaceOperator`
> * `PoissonFACPreconditionerStrategy`

I will not change these, since we do not know in the parent class the best choice for subclasses.
 
> I agree with @abarret in that it might be worthwhile to break some interfaces here for the sake of progress. If we assume ownership of a pointer (e.g., in `HierarchyIntegrator::registerProlongRefineAlgorithm`) then we should update the interface to take a `unique_ptr` as an argument.

Let's make an issue for this and do it in a different PR (and probably not by me!).

drwells:
I am running this through the test suite now.

drwells:
The test suite is OK. @abarret are you happy with the final change set?

PR #571, opened 2019-05-08
Actually add the two phase regrid test.

drwells:
This should have been in #550, but I forgot to actually add the test.


drwells:
attest is currently limited to finding tests in `tests/` and `tests/*/`, so I had to do the following:
1. I moved the test up a subdirectory.
2. To keep things tidy I merged all of the source and header files into one giant 2000 line test.
3. Since tests don't run in the source directory (we always create a temporary directory) I had to "hard code" the absolute path to the source directory to get the other input data file.

tests/multiphase_flow/free_falling_cyl_cib.cpp
```diff
@@ -0,0 +1,2202 @@
...
+     */
+    RigidBodyKinematics& operator=(const RigidBodyKinematics& that);
+
+    // No need for restart code in this example
+
+    /*!
+     * \brief Set rigid body velocity.
+     */
+    void setRigidBodySpecificVelocity(const double time,
+
```
drwells: Here is the second half of the trick to get the absolute path to an input file.

drwells:
The first half of the trick is in https://github.com/IBAMR/IBAMR/blob/master/tests/multiphase_flow/Makefile.am#L6, where we pass the absolute source directory as a macro definition. I wish there was a way to automatically set this up for all automake executables but I don't see how to do it.

boyceg:
> attest is currently limited to finding tests in `tests/` and `tests/*/`, so I had to do the following:
> 
> 1. I moved the test up a subdirectory.
> 2. To keep things tidy I merged all of the source and header files into one giant 2000 line test.

This seems like it would make it harder to maintain the test source code.

What about naming the subdirectories:
```
tests/IBFE_test0
tests/IBFE_test1
...
tests/multiphase_test0
tests/multiphase_test1
tests/multiphase_test2
...
```
Would this work with attest?

> 3. Since tests don't run in the source directory (we always create a temporary directory) I had to "hard code" the absolute path to the source directory to get the other input data file.

Can you get Automake to deposit all of the needed files into the build directory?

boyceg:
> The first half of the trick is in https://github.com/IBAMR/IBAMR/blob/master/tests/multiphase_flow/Makefile.am#L6, where we pass the absolute source directory as a macro definition. I wish there was a way to automatically set this up for all automake executables but I don't see how to do it.

What do you use the source directory name for?

amneetb:
> > 1. I moved the test up a subdirectory.
> > 2. To keep things tidy I merged all of the source and header files into one giant 2000 line test.
> 
> This seems like it would make it harder to maintain the test source code.

Yeah, I would strongly prefer to separate the helper code from main.cpp. Right now main.cpp is a giant code file!

drwells:
Concatenating everything into one file isn't pretty but we are not supposed to need to update these tests (except for code incompatibilities): I can split things up again if we think thats necessary.

I would prefer to keep all of the source code in one test (or at least one ultimate object file instead of several) so that its clear what the dependencies between files and tests are. This will make fixing failing tests easier in the future. If we have common code we can just put it in a shared header and `#include` it.

Its ultimately not significant whether we do
```
tests/multiphase_test_0/test.cpp
```
or
```
tests/multiphase/test_0.cpp
```
but the later seems more logical to me and also avoids 1000s of lines of autotools churn since we don't have to keep setting up new directories (this can actually make things slower when we get a ton of targets).

boyceg:
> Concatenating everything into one file isn't pretty but we are not supposed to need to update these tests (except for code incompatibilities): I can split things up again if we think thats necessary.
> 
> I would prefer to keep all of the source code in one test (or at least one ultimate object file instead of several) so that its clear what the dependencies between files and tests are. This will make fixing failing tests easier in the future. If we have common code we can just put it in a shared header and `#include` it.
> 
> Its ultimately not significant whether we do
> 
> ```
> tests/multiphase_test_0/test.cpp
> ```
> 
> or
> 
> ```
> tests/multiphase/test_0.cpp
> ```
> 
> but the later seems more logical to me and also avoids 1000s of lines of autotools churn since we don't have to keep setting up new directories (this can actually make things slower when we get a ton of targets).

Does it make sense to split the difference by having a basic file (`test_0.cpp`) that `#include`s separate files from a subdirectory (maybe `tests/multiphase/test_0_src`)?

drwells:
All right: that adds in all the test code as headers.

drwells:
My main goal is to avoid unnecessary subdirectories and extra object files. I think this is a reasonable solution.

tests/multiphase_flow/free_falling_cyl_cib.cpp
boyceg: I don't follow why this trick is needed. Why can't we dump the input files in the right place automatically?

boyceg:
> All right: that adds in all the test code as headers.

I am not sure I understand why having multiple object files per test is something to avoid, but putting the relevant code in headers seems like a reasonable way to break things up.

drwells:
Come to think of it: I do have a good reason to avoid subsubdirectories but I cannot think of any reason why we cannot handle multiple object files in a single test. `automake` can surely handle the case where two executables depend on the same object file: it simply won't compile it again (I have not tested this but I assume it is true).

My best guess is I was thinking of how we do things in deal.II, where we pick up tests by searching for files (so every source file corresponds to exactly one test) automatically, but since that won't work with automake we have to list them explicitly. While we don't need this now, I still think it would be good to try and keep this invariant if possible in case we get a chance to clean this up dramatically with BuildSystem.

boyceg:
> Come to think of it: I do have a good reason to avoid subsubdirectories but I cannot think of any reason why we cannot handle multiple object files in a single test. `automake` can surely handle the case where two executables depend on the same object file: it simply won't compile it again (I have not tested this but I assume it is true).
> 
> My best guess is I was thinking of how we do things in deal.II, where we pick up tests by searching for files (so every source file corresponds to exactly one test) automatically, but since that won't work with automake we have to list them explicitly. While we don't need this now, I still think it would be good to try and keep this invariant if possible in case we get a chance to clean this up dramatically with BuildSystem.

Setting things up to automatically build this stuff is a good reason in my book.

I am still confused about the `SOURCE_DIR` stuff.

drwells:
My use of string-concatenation with the C preprocessor is not great.

The problem is that some objects expect, as part of their input arguments, names of input files (not streams). These input files are not generally available in the build directory. Additionally, since `attest` runs tests in a temporary directory (and only copies the executable and the input file) they are not available in the `CWD` of the actual test process.

To get around this we simply modify the executable to copy the input file from the source directory into the actual run directory. Passing
```make
-DSOURCE_DIR=\"$(abs_srcdir)\"
```
makes the absolute source directory available to the preprocessor as a string constant. Hence
```c
SOURCE_DIR "/file.vertex"
```
correctly expands to the absolute path to that file after the preprocessor applies string literal concatenation.

Like the single source file rule, this could be avoided via more make rules, but this choice moves handling of extra input files into the executable and does not require complicating `attest` and the build system in an ad-hoc way for each test.

boyceg:
Is there a reason that you don't want to copy or link the files directly, as done in the examples?

(My preference is to copy, so that edits that happen in the build directory do not get propagated back into the source directory.)

drwells:
I did that because I prefer linking (i.e., its a matter of taste).

I frequently edit the input or output files directly from the build directory while working on a test and, as a result, if we copied (instead of linked) them my changes are clobbered when I recompile.

There is a small speed bonus to linking instead of copying but its not relevant when we have only a dozen tests or so.

boyceg:
> I did that because I prefer linking (i.e., its a matter of taste).
> 
> I frequently edit the input or output files directly from the build directory while working on a test and, as a result, if we copied (instead of linked) them my changes are clobbered when I recompile.
> 
> There is a small speed bonus to linking instead of copying but its not relevant when we have only a dozen tests or so.

So I take the opposite view that linking makes it more likely that spurious changes get committed to the input files in the repo, but I certainly agree that it is more convenient to tinker with these files when they are linked.

That said ... I still don't understand why you want to pass in the source directory to the executable instead of setting up a link to the data files in the makefile ...

drwells:
> So I take the opposite view that linking makes it more likely that spurious changes get committed to the input files in the repo, but I certainly agree that it is more convenient to tinker with these files when they are linked.

Ultimately both work. Would you prefer that I change it? That would make the semantics the same as the examples. Linking a thousand tests would be cheaper than copying a thousand files but we are nowhere near that yet.

> That said ... I still don't understand why you want to pass in the source directory to the executable instead of setting up a link to the data files in the makefile ...

We would have to add both custom make logic and custom `attest` logic for this to work (since `attest` sets up its own work directories): by doing the copy in the executable we fix things in one place instead of two.

boyceg:
> > So I take the opposite view that linking makes it more likely that spurious changes get committed to the input files in the repo, but I certainly agree that it is more convenient to tinker with these files when they are linked.
> 
> Ultimately both work. Would you prefer that I change it? That would make the semantics the same as the examples. Linking a thousand tests would be cheaper than copying a thousand files but we are nowhere near that yet.

It is fine with me to link in the tests but copy in the examples. Typical users will not be modifying tests.

We also can revisit the question of copying vs. linking for the examples.

> > That said ... I still don't understand why you want to pass in the source directory to the executable instead of setting up a link to the data files in the makefile ...
> 
> We would have to add both custom make logic and custom `attest` logic for this to work (since `attest` sets up its own work directories): by doing the copy in the executable we fix things in one place instead of two.

I guess it isn't a big deal as long as we know how to do it portably.

boyceg:
@drwells, go ahead and merge whenever you are ready.

PR #574, opened 2019-05-08
Add a switch for the vector type for U's RHS.

drwells:
This allows us to switch back to the old VecCache based implementation.

I added two tests where we explicitly use one option or the other when setting up IBFEMethod. They both work.

Part of #492.

### DIFF
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -106,6 +106,19 @@ namespace IBAMR
  * By default, the libMesh data is partitioned once at the beginning of the
  * computation by libMesh's default partitioner.
  *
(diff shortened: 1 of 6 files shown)


boyceg:
Is this missing the source for the test?

drwells:
No :) We can add a new test simply by adding a new pair of input and output files whose base name corresponds to a test that already exists. We are now up to 10 different interpolation tests that all use the same executable.

Its good that you checked, though, given the disaster that occurred the last time I tried to add a test.

drwells:
i.e., since the input file begins with `interpolate_velocity_01` it will be paired with the test executable of that name.

Issue #569, opened 2019-05-06
Convert all gtest tests into new-style tests


drwells:
Now that the new test suite is off the ground its time to start working on this: we should convert all tests that run in a minute or less to new-style tests. We can reproduce the assertions by writing the values to `plog`.

A lot of gtest tests simply check that the program completes successfully. We should check that the output for those tests hasn't changed recently and then convert them into new-style tests by, e.g., changing calls to `pout` to calls to `plog`.

## IBTK tests
- [x] CCLaplace (see #578)
- [x] CCPoisson (see #568)
- [x] PETScOps (see #635)
- [x] PhysBdryOps (see #635)
- [x] SCLaplace (see #635)
- [x] VCLaplace (see #635)
- [x] VCViscousSolver (see #635)

## advection-diffusion tests
- [x] `adv_diff/ex0` (#632)
- [x] `adv_diff/ex1` (#632)
- [x] `adv_diff/ex2` (#632)

## advection tests
- [x] `advect/test_main.cpp` (#577)

## CIBFE tests
- [x] `CIBFE/ex0`
- [x] `CIBFE/ex1`

CIBFEMethod is not implemented yet: see #721.

## ConstraintIB tests
As many of the examples as can be reasonably converted, though none of these presently record output.

## IB tests
- [x] `IB/explicit/ex0` (#595)
Any other tests that run quickly.

## IBFE tests
- [x] `IBFE/explicit/ex0`
Any other tests that run quickly. See #283, #284.

## level_set tests
Any tests that run quickly.

## multiphase_flow tests
Any tests that run quickly; see #288.

## navier-stokes tests
- [x] `navier_stokes/ex0` (#767)
Any tests that run quickly.

Finally, after all this is done:
- [x] delete all gtest-based code and restore the old way of doing examples. (#772)

drwells:
Once this is done we should go back and merge `ex_main.cpp` and `example.cpp` back together.

drwells:
We got close to finishing this, but its ultimately not a high priority item, so this got moved to 0.6.

PR #577, opened 2019-05-09
Convert the advection gtest to a new style test.

drwells:
More progress towards #569.


drwells:
Since this only adds a new test (and changes nothing in the library) and the other principal developers are otherwise occupied, I will merge this so that we can get on with modernizing the test suite.

PR #580, opened 2019-05-10
Fe surface distance evaluator

drwells:
Another spinoff from #418. This class is needed in a few different places.

I think this is the last (or nearly last) patch we can pull from the wave tank branch easily.

### DIFF
--- a/include/ibamr/FESurfaceDistanceEvaluator.h
+++ b/include/ibamr/FESurfaceDistanceEvaluator.h
@@ -1,0 +1,225 @@
+// Filename FESurfaceDistanceEvaluator.h
+// Created on Sep 5, 2018 by Nishant Nangia and Amneet Bhalla
+//
+// Copyright (c) 2002-2018, Nishant Nangia and Amneet Bhalla
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright notice,
+//      this list of conditions and the following disclaimer.
+//
+//    * Redistributions in binary form must reproduce the above copyright
+//      notice, this list of conditions and the following disclaimer in the
+//      documentation and/or other materials provided with the distribution.
+//
+//    * Neither the name of The University of North Carolina nor the names of
+//      its contributors may be used to endorse or promote products derived from
+//      this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+
+/////////////////////// INCLUDE GUARD ////////////////////////////////////
+
+#ifndef included_FESurfaceDistanceEvaluator
+#define included_FESurfaceDistanceEvaluator
+
+///////////////////////////// INCLUDES ///////////////////////////////////
+#include <map>
+#include <vector>
+
+#include "PatchHierarchy.h"
+#include "ibamr/IBFEMethod.h"
+#include "ibtk/IndexUtilities.h"
+#include "ibtk/ibtk_utilities.h"
+#include "ibtk/libmesh_utilities.h"
+#include "libmesh/boundary_mesh.h"
+#include "tbox/Pointer.h"
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * \brief class FESurfaceDistanceEvaluator is a utility class which is used to identify
+ * which line (triangle) in 2D (3D) elements are intersecting which grid cells.
+ */
+class FESurfaceDistanceEvaluator
+{
+public:
+    static const double s_large_distance;
+
+    /*!
+     * The only constructor of this class.
+     */
+    FESurfaceDistanceEvaluator(const std::string& object_name,
+                               SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > patch_hierarchy,
+                               SAMRAI::tbox::Pointer<IBAMR::IBFEMethod> ibfe_method,
+                               const libMesh::BoundaryMesh& bdry_mesh,
+                               const int part,
+                               const int gcw = 1);
+
+    /*!
+     * Destructor for this class.
+     */
+    ~FESurfaceDistanceEvaluator();
+
+    /*!
+     * Map the triangles intersecting a particular grid cell.
+     */
+    void mapIntersections();
+
+    /*!
+     * Get the map maintaining triangle-cell intersection and neighbors.
+     */
+    const std::map<SAMRAI::pdat::CellIndex<NDIM>, std::set<libMesh::Elem*>, IBTK::CellIndexFortranOrder>&
+    getNeighborIntersectionsMap();
+
+    /*!
+     * Compute the signed distance in the viscinity of the finite element mesh.
+     */
+    void computeSignedDistance(int n_idx, int d_idx);
+
+    /*!
+     * Update the sign of the distance function away from the finite element mesh.
+     */
+    void updateSignAwayFromInterface(int d_idx);
+
+    /*!
+     * Check whether the grown box and line element intersect in 2D.
+     *
+     *     (box_tl)   *--------------*  (box_tr)
+     *                |              |
+     *                |              |
+     *                |              |
+     *                |              |
+     *                |              |
+     *     (box_bl)   *--------------*  (box_br)
+     */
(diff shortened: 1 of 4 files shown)


include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+     *                  |  |              |                   / \
+     *                  ^  |      *       |                  /   \
+     *                     |   box_center |                 /     \
+     *                     |              |                /       \
+     *                     |              |               /         \
+     *                      --------------               *-----------*
+     *                           <------->
+     *                            box_half_dx[0]      vert1         vert2
+     */
+    static bool checkIntersection3D(const IBTK::Vector3d& box_center,
```
amneetb: @drwells : I hope you liked my ASCII art.

drwells: It makes the point clearly, so yes :)

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+namespace IBAMR
+{
+/*!
+ * \brief class FESurfaceDistanceEvaluator is a utility class which is used to identify
+ * which line (triangle) in 2D (3D) elements are intersecting which grid cells.
+ */
+class FESurfaceDistanceEvaluator
+{
+public:
+    static const double s_large_distance;
```
boyceg: If this is meant to be used outside of the class, let's stick with the standard convention:
```suggestion
    static const double LARGE_DISTANCE;
```

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+
+    /*!
+     * The only constructor of this class.
+     */
+    FESurfaceDistanceEvaluator(const std::string& object_name,
+                               SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > patch_hierarchy,
+                               SAMRAI::tbox::Pointer<IBAMR::IBFEMethod> ibfe_method,
```
boyceg: Does this need the full `IBFEMethod` implementation, or is it good enough to have the `FEDataManager` or even just the `Mesh`?

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+                               const libMesh::Mesh& mesh,
+                               const libMesh::BoundaryMesh& bdry_mesh,
```
boyceg: Do we need to have `BoundaryMesh` in the interface here --- can we just get the boundary mesh internally, or does the `BoundaryMesh` need to be properly set up for this to work?

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+                               const int part,
```
boyceg: If you have the part number and you have to pass in the `IBFEMethod` implementation, can we extract the `Mesh` internally?

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+                               const int gcw = 1,
+                               bool use_extracted_bdry_mesh = true);
+
+    /*!
+     * Destructor for this class.
+     */
+    ~FESurfaceDistanceEvaluator();
+
+    /*!
+     * Map the triangles intersecting a particular grid cell.
```
boyceg:
```suggestion
     * Map the triangles intersecting a particular Cartesian grid cell.
```
Also, is this specialized to triangular meshes? If not, suggest:
```suggestion
     * Map the elements from the finite element mesh intersecting a particular Cartesian grid cell.
```

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+                               const libMesh::Mesh& mesh,
+                               const libMesh::BoundaryMesh& bdry_mesh,
+                               const int part,
+                               const int gcw = 1,
+                               bool use_extracted_bdry_mesh = true);
+
+    /*!
+     * Destructor for this class.
+     */
+    ~FESurfaceDistanceEvaluator();
```
boyceg: Probably can do:
```suggestion
    ~FESurfaceDistanceEvaluator() = default;
```

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+
+    /*!
+     * Map the triangles intersecting a particular grid cell.
+     */
+    void mapIntersections();
+
+    /*!
+     * Get the map maintaining triangle-cell intersection and neighbors.
+     */
+    const std::map<SAMRAI::pdat::CellIndex<NDIM>, std::set<libMesh::Elem*>, IBTK::CellIndexFortranOrder>&
```
boyceg:
```suggestion
    using intersection_map = std::map<SAMRAI::pdat::CellIndex<NDIM>, std::set<libMesh::Elem*>, IBTK::CellIndexFortranOrder>;
    const intersection_map&
```

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+    getNeighborIntersectionsMap();
+
+    /*!
+     * Compute the signed distance in the viscinity of the finite element mesh.
+     */
+    void computeSignedDistance(int n_idx, int d_idx);
```
boyceg: What are `n_idx` and `d_idx`?

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+
+    /*!
+     * Update the sign of the distance function away from the finite element mesh.
```
boyceg: It might be helpful to briefly describe how this is different from `computeSignedDistance()`, and how they are meant to be used.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -0,0 +1,1310 @@
...
+                        {
+                            fe_bdry->reinit(elem, s);
+                            const unsigned int n_qp = qrule_bdry->n_points();
+                            const std::vector<libMesh::Point>& bdry_normals = fe_bdry->get_normals();
+                            TBOX_ASSERT(n_qp == 1); // for now
+                            unsigned int qp = 0;
+                            for (int d = 0; d < NDIM; ++d)
+                            {
+                                avg_unit_normal(d) += bdry_normals[qp](d);
+                            }
```
amneetb: Since this class is of interest, let's use the normal in the right way. We should follow the angle weighted normal as explained here:
https://ieeexplore.ieee.org/document/1407857
https://www.sciencedirect.com/science/article/pii/S0021999114004495

@drwells : Would like to implement this?

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+     *
+     *     (box_tl)   *--------------*  (box_tr)
+     *                |              |
+     *                |              |
+     *                |              |
+     *                |              |
+     *                |              |
+     *     (box_bl)   *--------------*  (box_br)
+     */
+    static bool checkIntersection2D(const IBTK::Vector3d& box_bl,
```
boyceg: It might make sense to put these in `libmesh_utilities.h`?

Also, @drwells, could we use the box class here that you are using for the new partitioner?

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+private:
+    /*!
+     * Default constructor is not implemented and should not be used.
+     */
+    FESurfaceDistanceEvaluator();
+
+    /*!
+     * Default assignment operator is not implemented and should not be used.
+     */
+    FESurfaceDistanceEvaluator& operator=(const FESurfaceDistanceEvaluator& that);
```
boyceg:
```suggestion
    FESurfaceDistanceEvaluator& operator=(const FESurfaceDistanceEvaluator&) = delete;
```

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -0,0 +1,249 @@
...
+
+    /*!
+     * Default copy constructor is not implemented and should not be used.
+     */
+    FESurfaceDistanceEvaluator(const FESurfaceDistanceEvaluator& from);
```
boyceg:
```suggestion
    FESurfaceDistanceEvaluator(const FESurfaceDistanceEvaluator&) = delete;
```

include/ibamr/FESurfaceDistanceEvaluator.h
boyceg: Looking at the implementation, it appears that you could _possibly_ just pass in the `EquationSystems` object and omit `ibfe_method`, `mesh`, and `bdry_mesh`.

amneetb:
It will be efficient (to avoid back-and-forth) if you guys can make these changes according to the design you have in mind. I approve all of the above changes. Some clarifications are below:

`n_idx` is the patch data index to plot the number of elements per Cartesian Cell in VisIt and 'd_idx` is the patch data index to hold the actual distance function.  

LARGE_DISTANCE should be set by the user according to their application. For example if distance is calculated for only 4 cells near the interface, then I generally set LARGE DISTANCE to 5*h. But for some applications, LARGE DISTANCE can actually be a very large value.

amneetb:
What I had in my mind with all the design decisions I had chosen are already there. I can't think of anything better, but I am open to and welcome new changes.

boyceg:
> What I had in my mind with all the design decisions I had chosen are already there. I can't think of anything better, but I am open to and welcome new changes.

The design is fine.

My questions were trying to get at whether this thing really needed any state or not --- like could it just be a collection of static functions? --- and if it needed state, how much needed to be provided.

If you can update the comments to address the questions about uses/limitations of the class, I will try to strip out unneeded stuff in the interface, and then we can get this merged.

amneetb:
Done.

In case you want to see the current usage of the class:
https://github.com/IBAMR/IBAMR/blob/amneetb/numerical_wave_tank/examples/IBLevelSet/ex0/example.cpp#L829-L868

drwells:
We discussed this by email but I don't think we reached a conclusion regarding who is going to finish addressing all of the review comments, add tests, etc. Should we explicitly assign this to a student or to someone else?

drwells:
One thing that we should add is (from a conversation with @amneetb) the use of angle-weighted normals in the case where multiple faces/edges are at an equal distance from the point: see

"An efficient immersed boundary treatment for complex moving object" by Chen Liu and Changhong Hu (JCP, 2014)

drwells:
We should try to make sure that this is still second order accurate even if we have higher degree finite elements and isoparametric elements.

drwells:
We should add a reference to the relevant paper that describes the algorithm used here.

drwells:
Superceded by #744.

PR #582, opened 2019-05-14
Fix a memory leak in LData.

drwells:
This was caught by GCC's leak sanitizer. Also add a test to make sure that the problem stays fixed.


ibtk/src/lagrangian/LData.cpp
```diff
@@ -139,36 +119,14 @@ LData::LData(Pointer<Database> db) : d_name(db->getString("d_name")), d_depth(db
     if (num_ghost_nodes > 0)
     {
         db->getIntegerArray("d_nonlocal_petsc_indices",
-                            d_nonlocal_petsc_indices.empty() ? nullptr : &d_nonlocal_petsc_indices[0],
+                            d_nonlocal_petsc_indices.empty() ? nullptr : d_nonlocal_petsc_indices.data(),
```
boyceg: Do we still need this check with `data()`? I think we can just do:
```suggestion
                            d_nonlocal_petsc_indices.data(),
```

drwells:
boost + std::array + glibcxx debugging don't get along. Let me see if the CI server likes this new version more.

drwells:
Switching to `std::vector` made GCC happy and the tests pass so I will merge.

boyceg:
Now getting some lovely warnings:
```
/Users/boyceg/code/IBAMR/ibtk/include/ibtk/LData.h:366:80: warning: braces around scalar initializer [-Wbraced-scalar-init]
    boost::multi_array_ref<double, 1> d_boost_array {nullptr, std::vector<int>{{0}}};
```

drwells:
I forgot that clang does that. We should enable ‘-Werror’ on the testers to catch these things.

boyceg:
> I forgot that clang does that. We should enable ‘-Werror’ on the testers to catch these things.

Maybe, maybe not. Does Boost build without warnings on the system running Jenkins? Sometimes it can be a pretty bad offender.

drwells:
You're right (so is the old copy of eigen).

PR #584, opened 2019-05-15
Add a test for RT0 coarsening.

drwells:
@boyceg This is most of what we need to test the RT0 coarsening code.

The test doesn't do anything but write some plotting output at the moment, but I am *pretty sure* that I am calling the coarsening operator correctly.

For reasons I don't fully understand the 'inverse ghost zone' option in visit disables plotting the finest grid level, which lets us visualize the coarser level (which is relevant). 

Here is a plot of the velocity field magnitude on the coarse level:
visit0000

Here is what we get from the coarsening operator:
visit0001

I don't understand why this coarsening operator needs such a large ghost region. It also appears, if I have done everything correctly, that we are not correctly filling the bottom part of the patch.


tests/IBTK/rt0_coarsen_01.cpp
```diff
@@ -0,0 +1,194 @@
...
+            const Box<NDIM> patch_box_0 = level_0->getPatch(0)->getBox();
+            const Box<NDIM> patch_box_1 = level_1->getPatch(0)->getBox();
+
+            // TODO we should probably just read this from the input file.
+            IntVector<NDIM> ratio;
+            for (int d = 0; d < NDIM; ++d)
+            {
+                ratio(d) = 4;
+            }
+            IBTK::CartSideDoubleRT0Coarsen coarsen_op(ratio);
```
boyceg: The argument to the constructor for `CartSideDoubleRT0Coarsen` is the ghost cell width. I think that you should be able to use the default value of 0.

tests/IBTK/rt0_coarsen_01.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+        // The rest is just book-keeping, this is the actual test:
+        {
+            Pointer<PatchLevel<NDIM> > level_0 = patch_hierarchy->getPatchLevel(0);
+            Pointer<PatchLevel<NDIM> > level_1 = patch_hierarchy->getPatchLevel(1);
+
+            // there should only be one patch on each patch level
+            Pointer<SideData<NDIM, double> > u_fc_0_data = level_0->getPatch(0)->getPatchData(u_sc_idx);
+            Pointer<SideData<NDIM, double> > u_fc_1_data = level_1->getPatch(0)->getPatchData(u_sc_idx);
+            const Box<NDIM> patch_box_0 = level_0->getPatch(0)->getBox();
```
boyceg: This will break in parallel (I think). Should we assert that it is running only on one processor?

tests/IBTK/rt0_coarsen_01.cpp
```diff
@@ -0,0 +1,194 @@
...
+            }
+            IBTK::CartSideDoubleRT0Coarsen coarsen_op(ratio);
+            coarsen_op.coarsen(*level_0->getPatch(0), *level_1->getPatch(0),
+                               u_sc_idx, u_sc_idx, patch_box_0, ratio);
+            u_fc_1_data->fillAll(0.0);
+        }
+
+#ifdef DO_PLOT
+        visit_writer->registerPlotQuantity(u_cc_var->getName(), "VECTOR", u_cc_idx);
+        hier_math_ops.interp(u_cc_idx, u_cc_var, u_sc_idx, u_sc_var, nullptr, 0.0, false);
```
boyceg: Since everything is happening only on a single patch, you could alternatively use `PatchMathOps` here.

boyceg:
> I don't understand why this coarsening operator needs such a large ghost region. It also appears, if I have done everything correctly, that we are not correctly filling the bottom part of the patch.

Looks like it. I will see if I can spot anything in the implementation of the coarsening operator while I wait for my delayed flight.

Can you update the test to use local refinement?

tests/IBTK/rt0_coarsen_01.cpp
```diff
@@ -0,0 +1,194 @@
...
+#define DO_PLOT
+#ifdef DO_PLOT
+        Pointer<CellVariable<NDIM, double> > u_cc_var = new CellVariable<NDIM, double>("u_cc", NDIM);
+        const int u_cc_idx = var_db->registerVariableAndContext(u_cc_var, ctx);
+        Pointer<CellVariable<NDIM, double> > exact_cc_var = new CellVariable<NDIM, double>("exact_cc", NDIM);
+        const int exact_cc_idx = var_db->registerVariableAndContext(exact_cc_var, ctx);
+#endif
+
+        gridding_algorithm->makeCoarsestLevel(patch_hierarchy, 0.0);
+        const int tag_buffer = std::numeric_limits<int>::max();
```
boyceg: I think that setting this `numeric_limits<int>::max()` will effectively refine the entire domain --- is that the goal?

You also can specify the region to be refined in the input file using `REFINED_BOXES`. There are some examples of this in the repo.

boyceg:
> > I don't understand why this coarsening operator needs such a large ghost region. It also appears, if I have done everything correctly, that we are not correctly filling the bottom part of the patch.
> 
> Looks like it. I will see if I can spot anything in the implementation of the coarsening operator while I wait for my delayed flight.
> 
> Can you update the test to use local refinement?

OK, I am 99% sure that the red zone is from the inverse ghost zone operator. If I go to "Controls", then "Subset ..." (Super-U on Mac), then unselect level 1, and I get a nice clean view of the solution magnitude on level 0. I am not sure that I believe it, but I see zero difference between the exact and computed solution. (Is this a field that we should reproduce exactly?)

drwells:
We should reproduce this field exactly: its
```
u {
   function_0 = "2*X_0 + 1"
   function_1 = "3 - 3*X_1"
}
```
which can be interpolated exactly by RT0.

I still see the red lines even when I remove the first layer. I'll address the other comments tomorrow. In particular: we might want to write a second test that uses a less crude grid that doesn't assume we have exactly two patches.

boyceg:
> I still see the red lines even when I remove the first layer. I'll address the other comments tomorrow. In particular: we might want to write a second test that uses a less crude grid that doesn't assume we have exactly two patches.

It seems odd that it works for me but not for you. I would suggest adding in an error variable and actually evaluating its norm in the code. This could just be a weird VisIt graphing thing. You can also compute the error directly in VisIt. When I do that, I get red bars in the same places --- however, VisIt reports the value of the red is 0. :-)

tests/IBTK/rt0_coarsen_01.cpp
drwells: The goal is to refine the entire domain (I want a domain with two patches, one on each level). I will look for refined boxes.

tests/IBTK/rt0_coarsen_01.cpp
drwells: I needed this to placate a check inside `CartSideDoubleRT0Coarsen.cpp`:
```cpp
    {
        if (d_gcw(d) + 1 < ratio(d))
        {
            TBOX_ERROR(
                "CartSideDoubleRT0Coarsen::coarsen():\n"
                << "   invalid refinement ratio between coarse and fine index spaces for specified ghost cell width"
                << std::endl);
        }
    }
```
since ratio is `{4, 4}` here.

drwells:
> Can you update the test to use local refinement?

I would prefer to do this in a second test once I am convinced that we can do this with uniform refinement (i.e., I will write `rt0_coarsen_02` once everything here works).

> It seems odd that it works for me but not for you. I would suggest adding in an error variable and actually evaluating its norm in the code.

I still see the red lines when I use visit's subset function to hide certain patches. I'll compute the error.

tests/IBTK/rt0_coarsen_01.cpp
boyceg: > I needed this to placate a check inside `CartSideDoubleRT0Coarsen.cpp`:
> 
> ```c++
>     for (unsigned int d = 0; d < NDIM; ++d)
>     {
>         if (d_gcw(d) + 1 < ratio(d))
>         {
>             TBOX_ERROR(
>                 "CartSideDoubleRT0Coarsen::coarsen():\n"
>                 << "   invalid refinement ratio between coarse and fine index spaces for specified ghost cell width"
>                 << std::endl);
>         }
>     }
> ```
> 
> since ratio is `{4, 4}` here.

Sorry, I should have looped back. A given coarse DOF on face `d` talks to `ratio(d)-1` DOFs to the "left" and `ratio(d)-1` DOFs to the "right".

boyceg:
> > Can you update the test to use local refinement?
> 
> I would prefer to do this in a second test once I am convinced that we can do this with uniform refinement (i.e., I will write `rt0_coarsen_02` once everything here works).

Absolutely. We should have both! :-)

drwells:
@boyceg I still see the red bars, but I didn't understand what visit was doing before: these values correspond to the largest floating point number (i.e., SAMRAI's default fill value).

Did you change the ghost cell width vector when you ran it?

boyceg:
Nope. Just exactly what is in your branch.

drwells:
Weird. Can you still see the values if you use `h5dump`?

drwells:
a-ha, I figured it out: you ran it in release mode. This problem only shows up in debug mode.

boyceg:
Nope I am in debug mode...

drwells:
Could you try using `h5dump` to see if it is actually a visit problem? I can definitely see wrong values when I just check the output that way.

boyceg:
I am looking at the FORTRAN kernel and am wondering if this needs to be completely reworked. I think it is grabbing values from ghost cells at the physical boundary that are not guaranteed to be setup correctly. It also probably will not work with local refinement.

For now: let's test out `CONSERVATIVE_COARSEN` along with the existing RT0 refinement operators, and revisit RT0 coarsening once I have some time to finish it up.

boyceg:
There are various refinement and coarsening operators available for the different data types.

Shall we have separate tests for each operator for each data centering, or should there be one test that loops over everything?

Similarly, should we have one test that does both local and uniform refinement, or split it up?

This actually a key thing for SAMRAI 3 migration, because there are some nontrivial API changes related to this stuff.

boyceg:
Ultimately, I think that we will want to have the following cases:

1. boundary conditions: physical vs. periodic
2. uniform refinement vs. local refinement

With local refinement, we might want to have tests that refine both at the boundary and in the interior. If we want to be exhaustive, we could have tests that set up local refinement only in the middle of each physical boundary, and then at each edge, and then at each corner. This kind of tests we would want to have some loops rather than having separate tests.

drwells:
> For now: let's test out CONSERVATIVE_COARSEN along with the existing RT0 refinement operators, and revisit RT0 coarsening once I have some time to finish it up.

Do we actually have RT0 refinement operators? I only see the coarsening/prolongation code in IBTK.

> Shall we have separate tests for each operator for each data centering, or should there be one test that loops over everything?

If we can, with reasonable effort, put everything in one executable then that is preferable to me. I think it would be easiest to have one test with uniform refinement and one test with adaptive refinement so that we only need to set up the SAMRAI stuff once. We should also have variants for periodic and dirichlet (so 4 tests determined by grid layout). The choice isn't too important: we should just do whatever lets us write relevant tests easily.

boyceg:
RT0-style refinement is provided by `SPECIALIZED_LINEAR_REFINE`. (This is a terrible name --- I apologize.)

drwells:
Ah. This is a good chance to rename it.

boyceg:
Actually, it is even worse that that --- it is `SPECIALIZED_CONSTANT_REFINE`. `SPECIALIZED_LINEAR_REFINE` is a very odd thing that does linear interpolation in the normal direction (like RT0) but slope-limited linear interpolation in the tangential direction (unlike RT0).

drwells:
I was about to ask about the description of the class that mentioned exactly that. Should we go with this or rewrite it as 'true' RT0?

boyceg:
> I was about to ask about the description of the class that mentioned exactly that. Should we go with this or rewrite it as 'true' RT0?

I am pretty sure that it is "true" RT0. I am about to push a new branch that just renames this class and associated FORTRAN functions.

drwells:
Wouldn't RT0 be piecewise constants in the tangential direction?

boyceg:
> Wouldn't RT0 be piecewise constants in the tangential direction?

Yes. That is what `SPECIALIZED_CONSTANT_REFINE` does. (That is why CONSTANT is in the name.)

drwells:
oh, duh. My eyes jumped too fast to the description of `SPECIALIZED_LINEAR_REFINE`.

drwells:
@boyceg one more thought on the discussion about tests from above: it should be possible to write a single test executable that does this for all four cases since we can read in topological/refinement information purely from input files.

drwells:
I am going to try and wrap this up tomorrow: there is not a lot left to do since we can copy and paste the refinement test to get most of what we need.

We should add a sanity check on ghost widths/box sizes inside `coarsen` in debug mode: I got a segmentation fault inside some fortran code when I tried to run this in debug mode with a ghost width of zero (the last line of computation in `subroutine scrt0coarsen2d`).

drwells:
@boyceg is there a straightforward way to populate ghost data with exact values, e.g., like what we do with CartGridFunction? Its not clear to me how we should set solution values at ghost points (and `CartGridFunction` doesn't appear to do this).

boyceg:
> @boyceg is there a straightforward way to populate ghost data with exact values, e.g., like what we do with CartGridFunction? Its not clear to me how we should set solution values at ghost points (and `CartGridFunction` doesn't appear to do this).

Not currently.

From the linear algebra standpoint, you can view these operators as implicitly defining some matrix `A` that goes from degrees of freedom on the fine level to degrees of freedom on the coarse level. Physical boundary ghost cell values are not "degrees of freedom".

To simplify the implementation, however, I _usually_ write code that is oblivious to the presence of physical boundaries (or coarse-fine boundaries in the case of local mesh refinement). This approach gives an operator `A0` that operates on the "extended" domain that includes both "degrees of freedom" and ghost values.

To form the action of `A`, we need an "extension operator" `E` that will fill in the ghost values. We then evaluate `A` as `A0 * E`. There are various implementations of these extension operators in the library.

In this particular case, the intention is for the coarsening operator to be discretely adjoint with the corresponding RT0 refine operator. It is a little linear algebra exercise to figure out what the appropriate definition is for `E`.

The alternative is to make the coarsening operator aware of where physical boundaries are located. This actually seems easier in this case, because there also are modifications to the operator that are probably needed at the coarse-fine interface that are not currently implemented.

At any rate, I would maybe be inclined to close this PR until someone has the time to do the algebra exercise and, if necessary, change the way that physical boundaries and coarse-fine interfaces are handled with this. In the meantime, we can continue using `CONSERVATIVE_COARSEN` for side-centered data.

drwells:
The problems with this test are fixed in #1122. I am migrating the test written here to that PR.

Issue #554, opened 2019-04-30
Remove deprecated libMesh calls in examples


drwells:
I forgot to look for deprecated libMesh functions in the examples in #500. We should go back and get rid of, e.g., calls to `get_node`.

boyceg:
In fact, I think that the examples do not currently compile with libMesh master unless libMesh is compiled with deprecated functions enabled.

drwells:
This is pretty easy to do so I will label it as `starter project`.

The easiest way to do this is to compile libMesh without support for deprecated functions. That will turn all of our usages into compile errors. The substitutions are all the same as those in 589f73621caf4fd358906454bff6c165eb74246c.

PR #587, opened 2019-05-17
remove some deprecated libMesh functions

boyceg:
This may break compatibility with some older versions of libMesh, but seems to be needed to avoid configuring libMesh `master` with `--enable-deprecated`.


drwells:
Fixes #554.

`neighbor_ptr` and `node_ref` are both in 1.1 so we should be OK.

PR #590, opened 2019-05-18
fix RT0 refine operator

boyceg:


boyceg:
@drwells, can you see if this makes any difference in your RT0 test? I am setting up the refine operator to simply copy data where possible (i.e. at fine DOFs that overlap a coarse grid DOF), and only to interpolate where needed (i.e. at fine DOFs that are between two coarse grid DOFs).

drwells:
> I am setting up the refine operator to simply copy data where possible (i.e. at fine DOFs that overlap a coarse grid DOF), and only to interpolate where needed (i.e. at fine DOFs that are between two coarse grid DOFs).

This makes sense. Do you mind if I simply add the test to this branch?

boyceg:
Please do; then I can test it and make changes to the implementation if needed.

drwells:
Aren't unit tests nice? :smile_cat:

drwells:
There is something wrong with the indexing. I can take a look if you need me to.

Here is what happens when we try to refine `u = 42`:
visit0000

we get the same picture when the velocity field is `(0, 42)` (but rotated by 90 degrees).

boyceg:
I think that this is a bug in the patch hierarchy setup in the tester, not in the refine operator per se.

The coarse patch is 4-by-4 with (0,0) in the bottom corner, it spans the cell indices [(0,0),(3,3)].

It looks like the fine patch is being set to be [(0,0),(19,19)], which corresponds to refining the coarse box [(0,0),(4,4)] instead of the coarse box [(0,0),(3,3)]. The fine box should be [(0,0),(15,15)].

You can see in the VisIt file that something is wrong, because the domain should span [0,1]^2, but it actually is spanning [0,1.25]^2.

boyceg:
If I add:
```
            // check that the fine patch and the coarse patch cover the same index space
            TBOX_ASSERT(Box<NDIM>::refine(patch_box_0, ratio) == patch_box_1);
```
I find that the coarse and fine patches do not cover the same index space.

If that is fixed, I think that the code will work.

boyceg:
Aha, it was because `REFINE_BOXES` was not setup correctly.

For some reason, SAMRAI allows you to set up `REFINE_BOXES` to refine coarse locations that are "outside" of the actual coarse boxes, which will generate fine boxes that are not contained by the coarse boxes.

boyceg:
It appears that this is the only test input file that suffered from that particular issue.

Do you want to clean up the history and then merge?

drwells:
Oh. Wild!

If the graphical output is correct then we should move on to calculating pointwise errors. They should be zero if we have a solution in the RT0 space. I'll fix up the test to do that.

tests/IBTK/rt0_refine_01_2d.4.output
```diff
@@ -0,0 +1,24 @@
...
+max norm of exact - refined: 0.272461
+test results for non_rt
+max norm of u_sc: 1.93848
+max norm of exact - refined: 0.0697632
+test results for non_rt
+max norm of u_sc: 1.93848
+max norm of exact - refined: 0.0175438
+test results for non_rt
+max norm of u_sc: 1.93848
+max norm of exact - refined: 0.00439239
```
drwells: I am surprised that this is second order, while `rt0_refine_01_2d.2.output` is first order. @boyceg did we expect this?

drwells:
I fixed the test and expanded it to work in 2D and 3D with several refinement ratios (2,2, 2,4, and 4,4 in 2D, similar in 3D).

tests/IBTK/rt0_refine_01_2d.4.output
drwells: If the answer is not obvious let me know and I will investigate.

boyceg: I think that this is first order. The refinement ratio is 4, and you see the error go down by a factor of 4 when you refine by a factor of 4.

drwells: oh, duh. This makes sense.

drwells:
Since we get no error in the RT0 space and first order convergence outside of it I think this is ready.

boyceg:
Is there a test with local refinement?

drwells:
Not yet. I'll go write one.

drwells:
We now have ten tests covering refinement across the entire domain or refinement in just the corner.

drwells:
Merging the other PR created a merge conflict. I'll fix; give me a minute.

drwells:
Fixed.

boyceg:
I just updated the description of a565eb2 to remove WIP. Let's merge once Jenkins finishes.

drwells:
Jenkins is happy so lets merge.

This PR makes the test suite 35% bigger!

PR #592, opened 2019-05-19
flexible interlevel transfer operator specification for incompressible Navier-Stokes solvers

boyceg:


abarret:
This looks good. It's conceivable that this could cause user code to break if they have a custom implementation of `INSHierarchyIntegrator`. Can you add a changelog entry?

boyceg:
> This looks good. It's conceivable that this could cause user code to break if they have a custom implementation of `INSHierarchyIntegrator`. Can you add a changelog entry?

Let me make a delegating constructor.

boyceg:
How about this? This will assign a default value for code using the old constructor, but that code did not assume that those default values were available anyway, and so I think that it should be backward compatible.

Issue #588, opened 2019-05-17
Make attest less laconic


drwells:
We should add an equivalent of `ctest`'s `V` or `VV` flags to `attest` so that, when we run in verbose mode and a test fails, we can print the error message to the screen. This would be more convenient then digging through the temporary directory `attest` creates.

amneetb:
https://github.com/IBAMR/IBAMR/blob/master/tests/multiphase_flow/rotating_barge_cib.cpp#L130

Dumb question: Anything before this line is executed just by processor 0? When does the SIMD part start? With `mpirun -np 4 rotating_barge_cib free_falling_cyl_cib.mpirun=4.input` won't all the 4 processors write out this file?🤔

drwells:
All four processors attempt to write, but only one gets the file handle and can actually produce output (the other three wait for the file handle from the OS). I think. I'll go back and add barriers so that this is clearer.

amneetb:
Why barrier? Why not just have processor 0 write this out? Once the file is written, no need to write this again from other processors.

drwells:
You're right, we don't need barriers. Either way: its now fixed.

drwells:
#1289 implemented verbose mode so this is now fixed.

PR #596, opened 2019-05-22
Guard writing to input files with a rank check.

drwells:
From a comment in #588: the current version works (perhaps by coincidence) but we should make it explicit.

### DIFF
--- a/tests/multiphase_flow/rotating_barge_cib.cpp
+++ b/tests/multiphase_flow/rotating_barge_cib.cpp
@@ -86,12 +86,19 @@
 int
 main(int argc, char* argv[])
 {
+    // Initialize PETSc, MPI, and SAMRAI.
+    PetscInitialize(&argc, &argv, NULL, NULL);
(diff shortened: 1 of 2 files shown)


tests/multiphase_flow/rotating_barge_cib.cpp
```diff
@@ -127,12 +134,7 @@ main(int argc, char* argv[])
             }
         }
     }
-
-    // Initialize PETSc, MPI, and SAMRAI.
-    PetscInitialize(&argc, &argv, NULL, NULL);
-    SAMRAI_MPI::setCommunicator(PETSC_COMM_WORLD);
-    SAMRAI_MPI::setCallAbortInSerialInsteadOfExit();
-    SAMRAIManager::startup();
+    SAMRAI_MPI::barrier();
```
boyceg: I don't think it is doing any harm, but I also don't think you need that call to `SAMRAI::barrier()`.
```suggestion
```

PR #598, opened 2019-05-24
Optimize bspline 3 spread

drwells:
I noticed this when reading some callgrind output: these loop bounds can be fixed by moving around some index computations. callgrind reports that we spend about 9% of our instructions in this function. These changes make it cost about 25% less: a very easy to implement performance improvement.

This change was easy to do since we have tests for the spread kernels. We can do the same thing for interpolation but we will have to come up with better interpolation kernel tests first.

### DIFF
--- a/ibtk/src/lagrangian/LEInteractor.cpp
+++ b/ibtk/src/lagrangian/LEInteractor.cpp
@@ -4264,6 +4264,7 @@ LEInteractor::spread(double* const q_data,
     const int stencil_size = getStencilSize(spread_fcn);
     const int min_ghosts = getMinimumGhostWidth(spread_fcn);
     const int q_gcw_min = q_gcw.min();
+    TBOX_ASSERT(0 <= q_gcw_min);
     bool patch_touches_physical_bdry = false;
     for (unsigned int d = 0; d < NDIM; ++d)
     {
--- a/ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
+++ b/ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
@@ -2450,6 +2450,7 @@ c
       INTEGER depth
       INTEGER nindices
       INTEGER ilower0,iupper0,ilower1,iupper1
+c     This assumes nugc0, nugc1, nugc2 are all positive, which *should* always be true.
       INTEGER nugc0,nugc1
 
       INTEGER indices(0:nindices-1)
@@ -2501,11 +2502,6 @@ c
             ic_upper(d) = ic_center(d)+1
          enddo
 
-         ic_lower(0) = max(ic_lower(0),ilower0-nugc0)
-         ic_upper(0) = min(ic_upper(0),iupper0+nugc0)
-
-         ic_lower(1) = max(ic_lower(1),ilower1-nugc1)
-         ic_upper(1) = min(ic_upper(1),iupper1+nugc1)
 c
 c     Compute the spreading weights.
 c
--- a/ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
+++ b/ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
@@ -2505,16 +2505,18 @@ c
 c
 c     Compute the spreading weights.
 c
-         do ic0 = ic_lower(0),ic_upper(0)
-            X_cell(0) = x_lower(0)+(dble(ic0-ilower0)+0.5d0)*dx(0)
-            w0(ic0-ic_lower(0)) =
+         do ic0 = 0,2
+            X_cell(0) = x_lower(0)+(dble(ic0 + ic_lower(0) - ilower0)
+     &           + 0.5d0)*dx(0)
+            w0(ic0) =
      &           lagrangian_bspline_3_delta(
      &           (X(0,s)+Xshift(0,l)-X_cell(0))/dx(0))
          enddo
 
-         do ic1 = ic_lower(1),ic_upper(1)
-            X_cell(1) = x_lower(1)+(dble(ic1-ilower1)+0.5d0)*dx(1)
-            w1(ic1-ic_lower(1)) =
+         do ic1 = 0,2
+            X_cell(1) = x_lower(1)+(dble(ic1 + ic_lower(1) - ilower1)
+     &      + 0.5d0)*dx(1)
+            w1(ic1) =
      &           lagrangian_bspline_3_delta(
      &           (X(1,s)+Xshift(1,l)-X_cell(1))/dx(1))
          enddo
@@ -2522,11 +2524,12 @@ c
 c     Spread V onto u.
 c
          do d = 0,depth-1
-            do ic1 = ic_lower(1),ic_upper(1)
-               do ic0 = ic_lower(0),ic_upper(0)
-                  u(ic0,ic1,d) = u(ic0,ic1,d)+(
-     &                 w0(ic0-ic_lower(0))*
-     &                 w1(ic1-ic_lower(1))*
+            do ic1 = 0,2
+               do ic0 = 0,2
+                  u(ic0 + ic_lower(0), ic1 + ic_lower(1), d) =
+     &                 u(ic0 + ic_lower(0), ic1 + ic_lower(1), d) + (
+     &                 w0(ic0)*
+     &                 w1(ic1)*
      &                 V(d,s)/(dx(0)*dx(1)))
                enddo
             enddo
--- a/ibtk/src/lagrangian/fortran/lagrangian_interaction3d.f.m4
+++ b/ibtk/src/lagrangian/fortran/lagrangian_interaction3d.f.m4
@@ -2798,6 +2798,7 @@ c
       INTEGER depth
       INTEGER nindices
       INTEGER ilower0,iupper0,ilower1,iupper1,ilower2,iupper2
+c     This assumes nugc0, nugc1, nugc2 are all positive, which *should* always be true.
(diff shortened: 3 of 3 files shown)


ibtk/src/lagrangian/LEInteractor.cpp
```diff
@@ -4264,6 +4264,7 @@ LEInteractor::spread(double* const q_data,
     const int stencil_size = getStencilSize(spread_fcn);
     const int min_ghosts = getMinimumGhostWidth(spread_fcn);
     const int q_gcw_min = q_gcw.min();
+    TBOX_ASSERT(0 <= q_gcw_min);
```
drwells: Since we error out if `q_gcw_min < min_ghosts` below this is redundant, but I added it anyway since it makes it clearer that this value has to be nonnegative.

ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
```diff
@@ -2501,36 +2502,34 @@ c
             ic_upper(d) = ic_center(d)+1
          enddo
 
-         ic_lower(0) = max(ic_lower(0),ilower0-nugc0)
-         ic_upper(0) = min(ic_upper(0),iupper0+nugc0)
-
-         ic_lower(1) = max(ic_lower(1),ilower1-nugc1)
-         ic_upper(1) = min(ic_upper(1),iupper1+nugc1)
```
drwells: From above: since `nugcX` must be nonnegative we can skip all of this and make the loop bounds fixed instead, since `ic_upper(N) - ic_lower(N) == 2`.

boyceg: Notice that `ic_lower` and `ic_upper` give the cell indices of the box that corresponds to the stencil of the regularized kernel function. That box need not be contained in the ghost box (which is determined by `ilowerX-nugcX` and `iupperX+nugcX`). These calls to `min` and `max` are "trimming" the stencil box, to ensure that it is contained in the ghost box.

Two alternatives are:
1. Make sure that the ghost width is large enough that we can ensure that we don't have to do this check. (This may require excluding some IB interaction points that are "far" from the patch interior, and hence cannot contribute to forces in the patch interior.)
2. Move this logic down into the loop where values are actually "spread" --- although putting if statements in the inner loop will probably eliminate the ability of the compiler to generate vectorized code.

drwells: You're right. I think I know how to fix this. I am surprised that the spread test didn't fail.

drwells:
I have a question about the first alternative. We currently have the following check in `LEInteractor.cpp`:
```cpp
    const int stencil_size = getStencilSize(spread_fcn);
    const int min_ghosts = getMinimumGhostWidth(spread_fcn);
    const int q_gcw_min = q_gcw.min();
    bool patch_touches_physical_bdry = false;
    {
        patch_touches_physical_bdry = patch_touches_physical_bdry || patch_touches_lower_physical_bdry[d];
        patch_touches_physical_bdry = patch_touches_physical_bdry || patch_touches_upper_physical_bdry[d];
    }
    if (patch_touches_physical_bdry && q_gcw_min < min_ghosts)
    {
        TBOX_ERROR("LEInteractor::spread(): insufficient ghost cells at physical boundary:"
                   << "  kernel function          = "
                   << spread_fcn
                   << "\n"
                   << "  kernel stencil size      = "
                   << stencil_size
                   << "\n"
                   << "  minimum ghost cell width = "
                   << min_ghosts
                   << "\n"
                   << "  ghost cell width         = "
                   << q_gcw_min
                   << "\n");
    }
```
This guarantees that we can reasonably evaluate these kernels along the physical boundary. Suppose we want to evaluate at a point that is on an interior boundary between two patches: won't we still need the same ghost depth to calculate this correctly? i.e., shouldn't we always have this assertion (regardless of whether or not we are at a physical boundary) on the number of ghost points?

boyceg:
> I have a question about the first alternative. We currently have the following check in `LEInteractor.cpp`:
> 
> ```c++
>     const int stencil_size = getStencilSize(spread_fcn);
>     const int min_ghosts = getMinimumGhostWidth(spread_fcn);
>     const int q_gcw_min = q_gcw.min();
>     bool patch_touches_physical_bdry = false;
>     for (unsigned int d = 0; d < NDIM; ++d)
>     {
>         patch_touches_physical_bdry = patch_touches_physical_bdry || patch_touches_lower_physical_bdry[d];
>         patch_touches_physical_bdry = patch_touches_physical_bdry || patch_touches_upper_physical_bdry[d];
>     }
>     if (patch_touches_physical_bdry && q_gcw_min < min_ghosts)
>     {
>         TBOX_ERROR("LEInteractor::spread(): insufficient ghost cells at physical boundary:"
>                    << "  kernel function          = "
>                    << spread_fcn
>                    << "\n"
>                    << "  kernel stencil size      = "
>                    << stencil_size
>                    << "\n"
>                    << "  minimum ghost cell width = "
>                    << min_ghosts
>                    << "\n"
>                    << "  ghost cell width         = "
>                    << q_gcw_min
>                    << "\n");
>     }
> ```
> 
> This guarantees that we can reasonably evaluate these kernels along the physical boundary.

The purpose of this check is related to the question about ghost cell values at physical boundaries.

The basic idea for the implementation of spreading is that we are only trying to get the right values in the interior of each Cartesian grid patch (including degrees of freedom that may be located on the patch boundary). We generally do not have enough Lagrangian ghost data to fill in the values in the ghost cell regions of the Cartesian grid patches, _except_ at physical boundaries.

We want to avoid having to write special-purpose code to handle spreading at physical boundaries. To do this, we implement the overall spreading operator `S` as `S = E^T * S0`, where `S0` is a spreading operator that is oblivious to the presence of physical boundaries. In particular, it will spread forces into the ghost region, including at physical boundaries. To avoid "losing" those force contributions, we then apply the transpose of an extension operator `E` that is used to extrapolate the solution from the interior of the domain to the interior + physical boundary ghost region.

It may be easier to see the motivation for this by considering interpolation, where we are faced with essentially the same issue. We want an operator `J` that will interpolate data from the interiors of the patches to the Lagrangian mesh, but it is easier to write code for an operator `J0` that interpolates from the extended domain. The extension operator `E` defines the ghost values from the interior values, and then `J = J0 * E`.

One reason to use `E^T` in spreading is to maintain a discrete adjoint property for `S` and `J`.

In order to capture all of the forces at the physical boundary, we need for the ghost region to include the full support of the kernel function.

> Suppose we want to evaluate at a point that is on an interior boundary between two patches: won't we still need the same ghost depth to calculate this correctly? i.e., shouldn't we always have this assertion (regardless of whether or not we are at a physical boundary) on the number of ghost points?

No, we don't. In fact, in the current implementation, the Cartesian grid data can have no ghost cells and everything is fine except at the physical boundary. However, we do need data for every Lagrangian interaction point that could possibly interact with the given grid patch --- and that will generally include interaction points that are in the ghost region of the Cartesian grid patch.

Does this make sense?

boyceg:
To simplify these implementation (and hopefully obtain better vectorization), we could do the following:
1. Filter out any Lagrangian interaction points that cannot interact with the interior of the patch data. (I think that there is code in `LEInteractor` that does something like this already.)
2. Allocate a temporary buffer (if needed) that includes a ghost region that is large enough to avoid truncating the kernel stencil for any interaction point that is being processed.

I think that it would suffice to exclude Lagrangian points that are more than `kernel_size / 2` away from a patch boundary, and to extend the ghost region to `kernel_size`. This would greatly simplify the implementation of the actual numerical routines, and hopefully allow for better vectorization.

We could probably do the same thing in interpolation --- filter out Lagrangian interaction points that are outside of the patch, and require that the ghost region be large enough to avoid worrying about the support of the kernel.

drwells:
This helps a lot. Let me make sure I understand the non-ghosted case.

If we have two patches `A` and `B` and an interaction point `X` spread onto points `o`
```
+--------------------+--------------------+
|                    |                    |
|                   o|                    |
|                    |                    |
|         A      o  X| o        B         |
|                    |                    |
|                   o|                    |
|                    |                    |
|                    |                    |
+--------------------+--------------------+
```
then, on patch `A`, we treat the rightmost `o` as having weight zero (and vice versa on patch `B`) and do not consider any ghost information at any point. Since spreading is linear the sum is correct (the spread of the sum is the sum of the spreads). Is this right?

(I wrote this before you posted the proposal above.)

boyceg:
> This helps a lot. Let me make sure I understand the non-ghosted case.
> 
> If we have two patches `A` and `B` and an interaction point `X` spread onto points `o`
> 
> ```
> +--------------------+--------------------+
> |                    |                    |
> |                   o|                    |
> |                    |                    |
> |         A      o  X| o        B         |
> |                    |                    |
> |                   o|                    |
> |                    |                    |
> |                    |                    |
> +--------------------+--------------------+
> ```
> 
> then, on patch `A`, we treat the rightmost `o` as having weight zero (and vice versa on patch `B`) and do not consider any ghost information at any point. Since spreading is linear the sum is correct (the spread of the sum is the sum of the spreads). Is this right?

I think that right now, stuff will be spread to whatever locations are present in the ghost region --- so it will depend on the size of the ghost buffer. But we do not guarantee that the ghost values are set _correctly_ (because we cannot without getting additional Lagrangian ghost data).

drwells:
I have a followup patch to this that implements the same idea, but for interpolation (and it does it correctly). Its not as good as vectorization, but its an improvement that doesn't require sorting out the ghost data situation at the moment, so I think its worth doing now.

Let me close this in favor of the discussion in #615 since this PR will be replaced by something better.

PR #599, opened 2019-05-24
run clang-format on current master

boyceg:
In working on merging NWT, I am running into some diffs that look like they are just related to clang-formatting.

This PR runs clang-format on everything in the current IBAMR master. There are a lot of changes. It is OK with me if we do not want to do this, but I am not sure how best to handle the formatting-related diffs that remain between current master and the NWT branch.


drwells:
We should definitely do this!

However: we will need to add another travis instance that runs the blessed version of clang-format on all PRs and complains if PRs don't format their code correctly.

boyceg:
Do you want to add the travis job before or after running clang-format? :-D

drwells:
We can add it after this PR. It won't be the end of the world if we commit a few changes that are not clang-formatted (but we should try really hard to remember until we get the CI bot going)

abarret:
There is a Jenkins job that hasn't been enabled to run clang-format on files and email a patch file to users. It's ready to go, just needs to be turned on.

drwells:
@abarret does it fail when the indentation is off?

abarret:
It should fail whenever clang-format reports a diff. I haven't tested it because I don't know how to test Jenkins jobs...

drwells:
@abarret We should just test it manually with a poorly-indented PR: I will be satisfied as long as it fails when someone leaves some trailing spaces.

boyceg:
OK, I will merge once Jenkins finishes.

boyceg:
If one of you wants to merge once Jenkins is done, feel free. Otherwise I will do this later tonight.

drwells:
I am still bisecting the restart issue. I'll take a look at this while I do.

drwells:
that was quick to finish.

PR #601, opened 2019-05-28
Underlinkage in the build system

drwells:
@abarret brought this to my attention: the test suite fails to compile with some configurations. I can get things to consistently work with one set of libraries on my workstation and consistently fail with another set.

I think we can fix this by compiling the tests in a way that is identical to how we compile the examples, but things don't work yet.


drwells:
My best guess is that deleting the output generation code in some examples causes linkage problems where we get static SAMRAI objects linked into the executable but we do not get SAMRAI's dependencies (HDF5 and silo). This is only a problem since SAMRAI uses static linking (i.e., we must resolve its dependencies at link time).

drwells:
Yup: if we copy and paste, for example, `tests/IB/explicit_ex0.cpp` into `examples/IB/explicit/ex0/ex_main.cpp` then compilation of that example fails. The problem is underresolved linking, which is completely independent of the test suite (its a bug in the build system).

@abarret would you please try applying the following diff
```
diff --git a/examples/IB/explicit/ex0/example.cpp b/examples/IB/explicit/ex0/example.cpp
index ff87671..0b0e0f3 100644
--- a/examples/IB/explicit/ex0/example.cpp
+++ b/examples/IB/explicit/ex0/example.cpp
@@ -60,6 +60,8 @@
 
 #include <array>
 
+#define ENABLE_OUTPUT 0
+
 // Function prototypes
 void output_data(Pointer<PatchHierarchy<NDIM> > patch_hierarchy,
                  Pointer<INSHierarchyIntegrator> navier_stokes_integrator,
@@ -272,6 +274,7 @@ run_example(int argc, char* argv[], std::vector<double>& u_err, std::vector<doub
         Pointer<Database> input_db = app_initializer->getInputDatabase();
 
         // Get various standard options set in the input file.
+#if ENABLE_OUTPUT
         const bool dump_viz_data = app_initializer->dumpVizData();
         const int viz_dump_interval = app_initializer->getVizDumpInterval();
         const bool uses_visit = dump_viz_data && app_initializer->getVisItDataWriter();
@@ -290,6 +293,7 @@ run_example(int argc, char* argv[], std::vector<double>& u_err, std::vector<doub
 
         const bool dump_timer_data = app_initializer->dumpTimerData();
         const int timer_dump_interval = app_initializer->getTimerDumpInterval();
+#endif
 
         // Create major algorithm and data objects that comprise the
         // application.  These objects are configured from the input database
@@ -399,6 +403,7 @@ run_example(int argc, char* argv[], std::vector<double>& u_err, std::vector<doub
             time_integrator->registerBodyForceFunction(f_fcn);
         }
 
+#if ENABLE_OUTPUT
         // Set up visualization plot file writers.
         Pointer<VisItDataWriter<NDIM> > visit_data_writer = app_initializer->getVisItDataWriter();
         Pointer<LSiloDataWriter> silo_data_writer = app_initializer->getLSiloDataWriter();
@@ -408,6 +413,7 @@ run_example(int argc, char* argv[], std::vector<double>& u_err, std::vector<doub
             ib_method_ops->registerLSiloDataWriter(silo_data_writer);
         }
+#endif
 
         // Initialize hierarchy configuration and data on all patches.
         time_integrator->initializePatchHierarchy(patch_hierarchy, gridding_algorithm);
@@ -435,7 +441,9 @@ run_example(int argc, char* argv[], std::vector<double>& u_err, std::vector<doub
 
         const int p_idx = var_db->mapVariableAndContextToIndex(p_var, p_ctx);
         const int p_cloned_idx = var_db->registerClonedPatchDataIndex(p_var, p_idx);
+#if ENABLE_OUTPUT
         visit_data_writer->registerPlotQuantity("P error", "SCALAR", p_cloned_idx);
+#endif
 
         const int coarsest_ln = 0;
         const int finest_ln = patch_hierarchy->getFinestLevelNumber();
@@ -448,6 +456,7 @@ run_example(int argc, char* argv[], std::vector<double>& u_err, std::vector<doub
         // Write out initial visualization data.
         int iteration_num = time_integrator->getIntegratorStep();
         double loop_time = time_integrator->getIntegratorTime();
+#if ENABLE_OUTPUT
         if (dump_viz_data && uses_visit)
         {
             pout << "\n\nWriting visualization files...\n\n";
@@ -455,6 +464,7 @@ run_example(int argc, char* argv[], std::vector<double>& u_err, std::vector<doub
             visit_data_writer->writePlotData(patch_hierarchy, iteration_num, loop_time);
             silo_data_writer->writePlotData(iteration_num, loop_time);
         }
+#endif
 
         // Main time step loop.
         double loop_time_end = time_integrator->getEndTime();
@@ -484,6 +494,7 @@ run_example(int argc, char* argv[], std::vector<double>& u_err, std::vector<doub
             // processing.
             iteration_num += 1;
             const bool last_step = !time_integrator->stepsRemaining();
+#if ENABLE_OUTPUT
             if (dump_viz_data && uses_visit && (iteration_num % viz_dump_interval == 0 || last_step))
             {
                 pout << "\nWriting visualization files...\n\n";
@@ -510,6 +521,7 @@ run_example(int argc, char* argv[], std::vector<double>& u_err, std::vector<doub
                             loop_time,
                             postproc_data_dump_dirname);
             }
+#endif
 
             // Compute velocity and pressure error norms.
             const int coarsest_ln = 0;
@@ -591,6 +603,7 @@ run_example(int argc, char* argv[], std::vector<double>& u_err, std::vector<doub
     return true;
 } // main
 
+#if ENABLE_OUTPUT
 void
 output_data(Pointer<PatchHierarchy<NDIM> > patch_hierarchy,
             Pointer<INSHierarchyIntegrator> navier_stokes_integrator,
@@ -637,3 +650,4 @@ output_data(Pointer<PatchHierarchy<NDIM> > patch_hierarchy,
     VecDestroy(&X_lag_vec);
     return;
 } // output_data
+#endif
```
and see if that causes compilation of that particular example program to fail for you to double-check?

drwells:
@boyceg I think I know how to fix this but it will probably require breaking functionality in the `PACKAGE_` macros. We don't use these macros consistently anyway so I think its okay to ignore this system and just do what autotools wants to do. Is this okay?

boyceg:
Sure.

abarret:
> @abarret would you please try applying the following diff

The example fails for me in this case. Seems like you found it.

drwells:
Thats good.

A quick fix to the problem induced with my `.diff` file is to simply add `-lz -lhdf5 -lhdf5_hl` to the end of `LIBS` in the generated Makefile. I thought it would work, but providing the full path to a shared object file doesn't seem to work.

Its a bit of a mystery to me how things work for the examples now. I don't see how, but something is causing us to get very lucky!

drwells:
@abarret I uploaded my fix. Please try it out.

This works for me with my 'native toolchain build'. I will try with my spack-based toolchain too and post again if it does not work.

drwells:
@abarret and I worked on this this afternoon. I think I have now figured out the last few things to fix (but I will open a followup issue).

The way we handle libMesh is different from other interdependent libraries; I added a comment explaining that. If we do not configure libMesh near the beginning of the dependency list then weird things happen.

drwells:
Fun fact: `Silo` states that `libsz` is an optional dependency, and will happily try to link against it, but it doesn't use it anywhere AFAICT.

boyceg:
I think it depends on how Silo is configured.

drwells:
I searched silo for any functions beginning with `SZ_`, which are the only defined symbols in `libsz.so` and didn't find any. If it does support it, its not in 4.10.2 unless I am missing something silly.

drwells:
@abarret would you please try this branch? I fixed the libMesh issue we encountered this afternoon.

abarret:
This works for me.

drwells:
This doesn't work on my laptop since silo doesn't see HDF5 flags. I think I know how to fix it.

drwells:
I think the first failure here is spurious:
```
==================================
Configuring required package PETSc
==================================
configure: using PETSC_DIR: /srv/clang-sfw/c++11_libs/petsc/3.7.2
configure: using PETSC_ARCH: linux-opt
grep: /srv/clang-sfw/c++11_libs/petsc/3.7.2/linux-opt/lib/petsc/conf/petscvariables: No such file or directory
grep: /srv/clang-sfw/c++11_libs/petsc/3.7.2/linux-opt/lib/petsc/conf/petscvariables: No such file or directory
grep: /srv/clang-sfw/c++11_libs/petsc/3.7.2/linux-opt/lib/petsc/conf/petscvariables: No such file or directory
```
Those checks should work.

drwells:
@abarret if you have time could you try this again? I think that I have resolved all issues with silo linkage now.

drwells:
Jenkins retest this please

drwells:
I fixed another error. My workstation and laptop are now happy, as is the CI tester. If this still fixes @abarret's issue then I suggest we merge.

drwells:
I got the error with a nonexistent PETSc directory because that directory does not actually exist; we set a 'wrong' value for PETSc's directory and then override it when configuring libMesh to get a working configuration. The last force push fixes this.

drwells:
@abarret does this work for you? If so I will merge.

abarret:
This works for me.

PR #605, opened 2019-05-31
Add a missing call to PetscFinalize().

drwells:
For some reason mpich was okay with omitting this so I did not catch it before.


drwells:
Since this is a trivial fix (and its only in the test suite) I will merge myself.

PR #607, opened 2019-06-04
NWT: necessary changes for Brinkman code.

drwells:
These changes are all necessary for the Brinkman code in #418. Part of #546.

### DIFF
--- a/ibtk/include/ibtk/IndexUtilities.h
+++ b/ibtk/include/ibtk/IndexUtilities.h
@@ -35,6 +35,8 @@
 
 /////////////////////////////// INCLUDES /////////////////////////////////////
 
+#include "ibtk/ibtk_utilities.h"
+
 #include "Box.h"
 #include "CartesianGridGeometry.h"
 #include "CartesianPatchGeometry.h"
@@ -139,6 +141,17 @@ public:
     getCellIndex(const DoubleArray& X,
                  const SAMRAI::tbox::Pointer<SAMRAI::geom::CartesianGridGeometry<NDIM> >& grid_geom,
                  const SAMRAI::hier::IntVector<NDIM>& ratio);
+
+    /*!
+     * \return The spatial coordinate of the given side center.
+     *
+     * @param patch The patch on which the cell lives.
+     *
+     * @param side_idx The SideIndex describing the current side.
+     */
+    static IBTK::VectorNd
+    getSideCenter(const SAMRAI::hier::Patch<NDIM> &patch,
+                  const SAMRAI::pdat::SideIndex<NDIM> &side_idx);
 
     /*!
      * \brief Map (i,j,k,d) index for a DOF defined for a SAMRAI variable
--- a/include/ibamr/IBHydrodynamicSurfaceForceEvaluator.h
+++ b/include/ibamr/IBHydrodynamicSurfaceForceEvaluator.h
@@ -33,16 +33,13 @@
 #ifndef included_IBHydrodynamicSurfaceForceEvaluator
 #define included_IBHydrodynamicSurfaceForceEvaluator
 
-#include <ibamr/AdvDiffHierarchyIntegrator.h>
-#include <ibamr/INSHierarchyIntegrator.h>
-
 #include "ibtk/ibtk_macros.h"
-#include <ibtk/LData.h>
-#include <ibtk/LDataManager.h>
-#include <ibtk/ibtk_utilities.h>
+#include "ibtk/ibtk_utilities.h"
 
 #include "Box.h"
+#include "CellVariable.h"
 #include "RobinBcCoefStrategy.h"
+#include "tbox/DescribedClass.h"
 
 IBTK_DISABLE_EXTRA_WARNINGS
 #include "Eigen/Core"
@@ -52,10 +49,12 @@ IBTK_ENABLE_EXTRA_WARNINGS
 #include <map>
 #include <vector>
 
-namespace IBTK
+/////////////////////////////// NAMESPACE ////////////////////////////////////
+namespace IBAMR
 {
-class LData;
-} // namespace IBTK
+class AdvDiffHierarchyIntegrator;
+class INSHierarchyIntegrator;
+} // namespace IBAMR
 namespace SAMRAI
 {
 namespace hier
@@ -91,7 +90,7 @@ namespace IBAMR
  *
  * \note  This class will work with both the constant and variable viscosity
  */
-class IBHydrodynamicSurfaceForceEvaluator
+class IBHydrodynamicSurfaceForceEvaluator : public virtual SAMRAI::tbox::DescribedClass
 {
 public:
     /*!
@@ -109,9 +108,44 @@ public:
     virtual ~IBHydrodynamicSurfaceForceEvaluator();
 
     /*!
-     * \brief Compute the hydrodynamic force via surface integration.
+     * \brief Compute the hydrodynamic force via surface integration after all the hierarchy
+     * variables have been integrated.
+     *
+     * \note The users should call this routine if the hydrodynamic forces need to be computed just
+     * as a postprocessing step.
+     *
+     * @note This function uses 3D vectors even in 2D so that the torque makes sense.
      */
-    virtual void computeHydrodynamicForce();
+    virtual void computeHydrodynamicForceTorque(IBTK::Vector3d& pressure_force,
+                                                IBTK::Vector3d& viscous_force,
+                                                IBTK::Vector3d& pressure_torque,
+                                                IBTK::Vector3d& viscous_torque,
+                                                const IBTK::Vector3d& X0);
+
+    /*!
+     * \brief Compute the hydrodynamic force via surface integration while the integration of
+     * variables is happening during a given timestep.
+     *
+     * @note This function uses 3D vectors even in 2D so that the torque makes sense.
+     */
+    virtual void computeHydrodynamicForceTorque(IBTK::Vector3d& pressure_force,
+                                                IBTK::Vector3d& viscous_force,
+                                                IBTK::Vector3d& pressure_torque,
+                                                IBTK::Vector3d& viscous_torque,
+                                                const IBTK::Vector3d& X0,
+                                                double time,
+                                                double current_time,
+                                                double new_time);
+    /*!
+     * \brief Set the surface contour value <em> s <\em> of the level set field, such that \f$ \phi(s) = \partial \Omega
+     * \f$.
+     */
+    virtual void setSurfaceContourLevel(double s = 0);
+
+    /*!
+     * \brief Indicate if the force and torque results need to be written on a file.
+     */
+    void writeToFile(bool write_to_file = true);
 
 private:
     /*!
@@ -143,7 +177,9 @@ private:
      * Fill required patch data and ghost cells.
      */
     void fillPatchData(SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > patch_hierarchy,
-                       const double fill_time);
+                       double fill_time,
+                       bool use_current_ctx,
+                       bool use_new_ctx);
 
     /*!
      * \brief Object name.
@@ -196,12 +232,25 @@ private:
     double d_surface_contour_value = 0.0;
 
     /*!
-     * \brief File streams associated for the output.
+     * \brief Whether to write results on a text file.
+     */
+    bool d_write_to_file = false;
+
+    /*!
+     * \brief File streams associated for the output of hydrodynamic force.
      *
      * \note Columns 1-3 represent sum of -p.n dA. Columns 4-6 represent sum of n.(grad U + grad U^T) dA.
      *
      */
     std::unique_ptr<std::ofstream> d_hydro_force_stream;
+
+    /*!
+     * \brief File streams associated for the output of hydrodynamic torque.
+     *
+     * \note Columns 1-3 represent sum of r X -p.n dA. Columns 4-6 represent sum of r x n.(grad U + grad U^T) dA.
+     *
+     */
+    std::unique_ptr<std::ofstream> d_hydro_torque_stream;
 };
 } // namespace IBAMR
 
--- a/include/ibamr/IBHydrodynamicSurfaceForceEvaluator.h
+++ b/include/ibamr/IBHydrodynamicSurfaceForceEvaluator.h
@@ -142,11 +142,6 @@ public:
      */
     virtual void setSurfaceContourLevel(double s = 0);
 
-    /*!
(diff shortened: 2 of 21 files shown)


drwells:
Lets move the output stream stuff into a separate function (so that this class is not responsible for writing output files), i.e., `log_torque(Vector3d &foo, Vector3d &barr, std::ostream &out)`.

include/ibamr/IBHydrodynamicSurfaceForceEvaluator.h
```diff
@@ -103,9 +102,40 @@ class IBHydrodynamicSurfaceForceEvaluator
...
+                                                IBTK::Vector3d& viscous_torque,
+                                                const IBTK::Vector3d& X0,
+                                                double time,
+                                                double current_time,
+                                                double new_time);
+    /*!
+     * \brief Set the surface contour value <em> s <\em> of the level set field, such that \f$ \phi(s) = \partial \Omega
+     * \f$.
      */
-    virtual void computeHydrodynamicForce();
```
drwells: We should at least log that we broke this interface under 'incompatible changes' and fix any relevant examples to not use it.

include/ibamr/IBHydrodynamicSurfaceForceEvaluator.h
```diff
@@ -137,7 +167,9 @@ class IBHydrodynamicSurfaceForceEvaluator
      * Fill required patch data and ghost cells.
      */
     void fillPatchData(SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > patch_hierarchy,
-                       const double fill_time);
+                       double fill_time,
+                       bool use_cuurent_ctx,
```
drwells:
```suggestion
                       bool use_current_ctx,
```

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -70,6 +68,31 @@ sign(const double X)
...
+            side_coord[d] = patch_X_lower[d] + patch_dx[d] * (static_cast<double>(side_idx(d) - patch_lower_idx(d)));
+        }
+        else
+        {
+            side_coord[d] =
+                patch_X_lower[d] + patch_dx[d] * (static_cast<double>(side_idx(d) - patch_lower_idx(d)) + 0.5);
+        }
+    }
+    return;
+} // get_physical_coordinate
```
drwells: Do we already have a function that does this somewhere, like in `IndexUtilities`?

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -266,6 +336,10 @@ IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForce()
                     IBTK::Vector3d n = IBTK::Vector3d::Zero();
                     n(axis) = sign(phi_upper - phi_lower);
```
drwells: I think we can use `std::signbit` here.

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -336,25 +409,73 @@ IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForce()
...
 /////////////////////////////// PROTECTED ////////////////////////////////////
 
 /////////////////////////////// PRIVATE //////////////////////////////////////
 
 void
 IBHydrodynamicSurfaceForceEvaluator::fillPatchData(Pointer<PatchHierarchy<NDIM> > patch_hierarchy,
-                                                   const double fill_time)
+                                                   const double fill_time,
+                                                   bool use_current_ctx,
+                                                   bool use_new_ctx)
```
drwells: We should make the arguments to this the same as `computeHydrodynamicForceTorque`: it has the same logic for determining which context to use.

We should also log that this improvement (i.e., we no longer assume that we are at the current time) is incompatible.

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -196,29 +219,76 @@ IBHydrodynamicSurfaceForceEvaluator::~IBHydrodynamicSurfaceForceEvaluator()
...
     for (int ln = coarsest_ln; ln <= finest_ln; ++ln)
     {
-        patch_hierarchy->getPatchLevel(ln)->allocatePatchData(d_ls_solid_idx, integrator_time);
-        patch_hierarchy->getPatchLevel(ln)->allocatePatchData(d_u_idx, integrator_time);
-        patch_hierarchy->getPatchLevel(ln)->allocatePatchData(d_p_idx, integrator_time);
-        patch_hierarchy->getPatchLevel(ln)->allocatePatchData(d_mu_idx, integrator_time);
+        patch_hierarchy->getPatchLevel(ln)->allocatePatchData(d_ls_solid_idx, time);
+        patch_hierarchy->getPatchLevel(ln)->allocatePatchData(d_u_idx, time);
+        patch_hierarchy->getPatchLevel(ln)->allocatePatchData(d_p_idx, time);
+        patch_hierarchy->getPatchLevel(ln)->allocatePatchData(d_mu_idx, time);
```
drwells: We don't need to allocate here if the density is constant.

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -336,25 +409,73 @@ IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForce()
     }
```
drwells: Line 408: we don't need to deallocate if the density is constant.

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -396,16 +520,23 @@ IBHydrodynamicSurfaceForceEvaluator::fillPatchData(Pointer<PatchHierarchy<NDIM>
             p_vc_ins_hier_integrator->getTransportedViscosityVariable();
         Pointer<CellVariable<NDIM, double> > mu_ins_var = p_vc_ins_hier_integrator->getViscosityVariable();
         RobinBcCoefStrategy<NDIM>* mu_bc_coef = nullptr;
-        int mu_current_idx = -1;
+        int mu_idx = -1;
```
drwells:
```suggestion
        int mu_idx = IBTK::invalid_index;
```

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -428,55 +559,58 @@ IBHydrodynamicSurfaceForceEvaluator::fillPatchData(Pointer<PatchHierarchy<NDIM>
...
-            hier_p_bdry_fill->setHomogeneousBc(false);
-            hier_p_bdry_fill->fillData(fill_time);
-        }
-        else
-        {
-            TBOX_ERROR(d_object_name << "::IBHydrodynamicSurfaceForceEvaluator():\n"
-                                     << " no valid pressure boundary condition object registered with INS integrator.\n"
-                                     << " This statement should not have been reached");
-        }
+    // Fill ghost cells for pressure
```
drwells: Note that we need to do this regardless of whether nor not the density is constant, so this fixes a bug.

drwells:
@amneetb and I just read through this together and we are both happy.

@amneetb has test code for a hydrostatic suspended cylinder with zero torque. Otherwise I have only minor comments for things to change. I can handle these little changes.

drwells:
@amneetb we have five examples that still use `computeHydrodynamicForce`:
```
[drwells@learch examples]$ grep -Ri computeHydrodynamicForce
CIB/ex1/example.cpp:            hydro_force->computeHydrodynamicForce(u_idx,
ConstraintIB/eel2d/example.cpp:            hydro_force->computeHydrodynamicForce(u_idx,
ConstraintIB/flow_past_cylinder/example.cpp:            hydro_force->computeHydrodynamicForce(u_idx,
ConstraintIB/moving_plate/example.cpp:            hydro_force->computeHydrodynamicForce(u_idx,
multiphase_flow/ex1/example.cpp:            hydro_force_evaluator->computeHydrodynamicForce();
multiphase_flow/ex2/example.cpp:            hydro_force_evaluator->computeHydrodynamicForce();
```
Do you have a recommended change that preserves the behavior of these examples?

drwells:
(compilation fails because we cannot compile these examples any longer)

drwells:
Clarification: most of these examples don't use this class (I only searched for the function, not class, name). The only examples we need to update are the two multiphase flow examples.

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
drwells: (I added this to IndexUtilities)

include/ibamr/IBHydrodynamicSurfaceForceEvaluator.h
```diff
@@ -190,12 +222,25 @@ class IBHydrodynamicSurfaceForceEvaluator
     double d_surface_contour_value = 0.0;
 
     /*!
-     * \brief File streams associated for the output.
+     * \brief Whether to write results on a text file.
+     */
+    bool d_write_to_file = false;
+
+    /*!
+     * \brief File streams associated for the output of hydrodynamic force.
```
drwells: As noted above, lets remove this and the other logging in favor of a method taking a `std::ostream &` argument.

ibtk/include/ibtk/IndexUtilities.h
```diff
@@ -140,6 +140,18 @@ class IndexUtilities
...
+    /*!
+     * \return The spatial coordinate of the given side center.
+     *
+     * @param patch The patch on which the cell lives.
+     *
+     * @param side_idx The SideIndex describing the current side.
+     */
+    IBTK::VectorNd
+    getSideCenter(const SAMRAI::hier::Patch<NDIM> &patch,
+                  const SAMRAI::pdat::SideIndex<NDIM> &side_idx);
```
drwells: We should add a short test for this.

drwells:
@amneetb Could you please send me the test code you mentioned you had for the non-rotating cylinder?

amneetb:
> @amneetb Could you please send me the test code you mentioned you had for the non-rotating cylinder?

I need to rewrite it. I was hacking the NWT branch to do a quick-and-dirty check. If you can commit these changes to some branch, I can add the test.

drwells:
Aside from the trivial merge conflict this is ready to go once we fix the examples: lets merge this once the testers are happy. @amneetb how should we fix the first two multiphase flow examples?

amneetb:
> @amneetb how should we fix the first two multiphase flow examples?

I think we decided that we can pass an `std::ostream` argument to this function and that `main.cpp` should manage this stream object.

drwells:
@amneetb Agreed; however, the issue with the two examples is that they use `IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForce` which we removed.

amneetb:
> `IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForce` which we removed.

Oh this is easy. We just use `IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForceTorque` in its place which is what I switched to in the NWT branch. Notice that we are already passing force and torque vectors in this function. We can just use this to populate an ostream file created in main.cpp. 

Something like 

`force_stream << time << pressure_force[0] << "\t" << pressure_force[1] << "\t"  << viscous_force[0] << "\t" << viscous_force[1] << std::endl; `

and 

`torque_stream << time << pressure_torque[2] << "\t"  << viscous_torque[2] << std::endl;`

This should be enough for these two-dimensional examples.

amneetb:
> Something like
> 
> `force_stream << time << pressure_force[0] << "\t" << pressure_force[1] << "\t" << viscous_force[0] << "\t" << viscous_force[1] << std::endl; `
> 
> and
> 
> `torque_stream << time << pressure_torque[2] << "\t" << viscous_torque[2] << std::endl;`

We can also write just one file:

`force_torque_stream << time << pressure_force[0] << "\t" << pressure_force[1] << "\t" << viscous_force[0] << "\t" << viscous_force[1] << "\t" << pressure_torque[2] << "\t" << viscous_torque[2] << std::endl; `

amneetb:
https://github.com/drwells/IBAMR/blob/brinkman-dependencies/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp#L345

Note that we need to make sure cross product works in 2d also and does not error out. Torque is 3D even for 2D simulations (out of plane direction). I had IBTK::Vector3d instead of Vector2d for this reason in this routine.

boyceg:
I was going to ask about that cross product routine in the implementation, which seems a bit odd. Could we just use 3D vectors in all cases, with the native cross product?

amneetb:
Yes, please. I always stick to Vector3d for all of the rigid body dynamics code. It makes even physical sense because of torque.

drwells:
I'll change that back. Shouldn't `X0` still be an N-dimensional vector, though?

amneetb:
> I'll change that back. Shouldn't `X0` still be an N-dimensional vector, though?

Yes, but it not worth saving the memory of a double. You need `r` from `X0` which needs to be 2d vector for cross product. I would suggest sticking to vector3d for everything.

drwells:
Easy enough.

amneetb:
> @amneetb Could you please send me the test code you mentioned you had for the non-rotating cylinder?

Here you go:
[check_hydro_force.zip](https://github.com/IBAMR/IBAMR/files/3312427/check_hydro_force.zip)

It sets a linear pressure field with slope -1, i.e, increases downwards. Net pressure force on the body (cylinder in this case) i

`{Px, Py} = integral  [-grad(p) dv] = (0, volume of body)`

drwells:
Ok, that should fix:
1. all 2D vs 3D issues
2. merge conflicts
3. test addition

So this is ready!

amneetb:
Thanks!

PS: This block could be avoided 

https://github.com/drwells/IBAMR/blob/brinkman-dependencies/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp#L289-L292 

if we initialize `IBTK::Vector3d r_vec = IBTK::Vector3d::Zero()`

PPS: We can also add another line of check in the code for pressure torque. It should be zero analytically, as pressure forces pass through the COM of the cylinder.

drwells:
Good call. I did both.

amneetb:
Thanks! Next stop: `BrinkmanPenalizationRigidBodyDynamics.{h,cpp}`

drwells:
Since @boyceg approved and @amneetb is happy lets merge.

PR #611, opened 2019-06-13
Clean up macro definitions

drwells:
Fixes #610: we no longer print thousands of totally irrelevant Eigen warnings with newer GCC compilers.

While `_Pragma` is a keyword in C++11 (so this should always work), I added a test for it anyway to make sure all our compilers can, in fact, successfully compile `_Pragma("GCC diagnostic pop")` et al to be sure.

PR #613, opened 2019-06-14
Sort includes

drwells:
Related to #611: this PR will make that PR a bit simpler since it sorts all headers (i.e., third-party headers are no longer interleaved with IBAMR headers).

I used the standard order, which is most specific (IBAMR, IBTK, then SAMRAI) to most general (boost and system headers). I had to add a small number of `#include`s to fix compilation.


drwells:
@boyceg I would like to get this merged ASAP since we will need to update other branches accordingly. Does it look okay?

drwells:
@marshallrdavey is trying to debug something and the insane eigen output is getting in the way. Since this doesn't change anything substantial (just sorting headers) I will merge as-is now so that we can fix his issue.

boyceg:
Looks good to me. It will be great not to have all of those warnings!

PR #616, opened 2019-06-18
Avoid using Triangle in the interpolation tests.

drwells:
@abarret pointed this out to me a few weeks ago.

On some machines we end up with more elements than others when we generate meshes with Triangle. Avoid this by just using a built-in libMesh mesh generator.


drwells:
Since this only pertains to the test suite I am going to go ahead and merge (I have two followup PRs that need this)

PR #617, opened 2019-06-18
Add 3d interpolation tests

drwells:
Followup to #616: this adds fourteen 3D interpolation tests.


drwells:
Again, like #616, since this only pertains to the test suite, I will merge as-is.

PR #618, opened 2019-06-20
Use F95's FLOOR instead of lagrangian_floor.

drwells:
This can be better inlined by the compiler: callgrind reports that we use about 10% less instructions in the IB kernels with this switch. More precisely: since `lagrangian_floor` is `EXTERNAL` *and* in another translation unit the compiler cannot optimize it without something very fancy like `-fwhole-program` or link-time optimization.

One minor change in this patch is that `lagrangian_floor(-42)` is `-43` while `floor(-42)` is `-42`. I don't think this is a big deal since we are `floor`ing structure coordinates which are not going to end up at integer multiples of `dx` anyway.

### DIFF
--- a/ibtk/src/lagrangian/fortran/lagrangian_delta.f.m4
+++ b/ibtk/src/lagrangian/fortran/lagrangian_delta.f.m4
@@ -36,23 +36,6 @@ c     SUCH DAMAGE.
 c
 define(REAL,`double precision')dnl
 define(INTEGER,`integer')dnl
-ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
-c
-c     Returns the greatest integer less than or equal to x.
-c
-ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
-c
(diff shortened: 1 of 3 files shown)


boyceg:
I can’t remember why I introduced this function. It is fine with me to remove it if it does not break anything.

drwells:
Hm. It won't be hard to make sure that we have at least one test per kernel (particularly with #619). I will check that we have a test for everything we change here (and that they still pass) before merging this.

boyceg:
We possibly could make some M4 macros for all of the nonstandard scalar functions to get better performance here as well.

Also, I am not opposed to making these kernels C or C++, although I suspect that we can make them faster by sticking with FORTRAN.

drwells:
I think Fortran is the right choice since we use multidimensional arrays and we do not want aliasing. It would be a lot of work to get either of those in C++.

drwells:
> We possibly could make some M4 macros for all of the nonstandard scalar functions to get better performance here as well.

I think that the compiler will inline things appropriately as long as they are in the same translation unit, so we don't necessarily need macros. What other functions did you have in mind?

PR #619, opened 2019-06-20
Add more spread and interpolation tests.

drwells:
This commit adds tests for some of the kernels which are getting upgraded.


drwells:
We now have a basic test for each kernel \o/ :tada:

drwells:
Since this change is only in the test suite I will merge as-is.

PR #625, opened 2019-06-24
NWT: Brinkman

drwells:
Next part of #546.

I think the only thing missing is a few more sentences on what this class actually does. @amneetb could you write a paragraph or two and give a link to a citation, please?

### DIFF
--- a/include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
+++ b/include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
@@ -69,8 +69,8 @@ namespace IBAMR
  * \brief BrinkmanPenalizationRigidBodyDynamics provides an implementation of Brinkman penalization
  * body force for a rigid body motion in the momentum equation.
  *
- * The penalization force is taken to be \f$ \frac{\chi}{\kappa}(\bm{u}^{n+1}
- * - \bm{u}_b) \f$. The class computes the coefficient \f$
+ * The penalization force is taken to be \f$ \frac{\chi}{\kappa}(\bm{u}_b
+ * - \bm{u}^{n+1}) \f$. The class computes the coefficient \f$
  * \frac{\chi}{\kappa}$ of the fluid velocity \f$ \bm{u}^{n+1} \f$ for the
  * variable-coefficient INS solvers of INSVCStaggeredHierarchyIntegrator. This
  * is done in the BrinkmanPenalizationRigidBodyDynamics::demarcateBrinkmanZone
@@ -79,7 +79,7 @@ namespace IBAMR
  * rigid body velocity \f$\bm{u}_b\f$ is computed through Newton's law of
  * motion by netting the hydrodynamic and external forces and torques on the
  * body in the BrinkmanPenalizationRigidBodyDynamics::computeBrinkmanVelocity
- * method. A simple forward-Euler scheme is imployed for the second-law of
+ * method. A simple forward-Euler scheme is employed for the second-law of
  * motion.
  *
  * For further information on applications of this class see
@@ -87,8 +87,6 @@ namespace IBAMR
  */
 class BrinkmanPenalizationRigidBodyDynamics : public BrinkmanPenalizationStrategy
 {
-    ////////////////////////////// PUBLIC ////////////////////////////////////////
-
 public:
     /*
      * \brief Constructor of the class.
@@ -103,7 +101,7 @@ public:
     /*
      * \brief Destructor of the class.
      */
-    ~BrinkmanPenalizationRigidBodyDynamics();
+    ~BrinkmanPenalizationRigidBodyDynamics() = default;
 
     /*!
      * \brief Set initial conditions for the rigid body.
@@ -284,8 +282,6 @@ public:
         return;
     } // getHydrodynamicForceTorque
 
-    /////////////////////////////// PROTECTED ////////////////////////////////////
-
 protected:
     // Center of mass.
     Eigen::Vector3d d_center_of_mass_initial = Eigen::Vector3d::Zero(),
@@ -326,8 +322,6 @@ protected:
     // Routines to get prescribed kinematics and additional external forces and torques.
     KinematicsFcnData d_kinematics_fcn_data;
     ExternalForceTorqueFcnData d_ext_force_torque_fcn_data;
-
-    /////////////////////////////// PROTECTED ////////////////////////////////////
 
 private:
     /*!
--- a/include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
+++ b/include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
@@ -68,6 +68,22 @@ namespace IBAMR
 /*!
  * \brief BrinkmanPenalizationRigidBodyDynamics provides an implementation of Brinkman penalization
  * body force for a rigid body motion in the momentum equation.
+ *
+ * The penalization force is taken to be \f$ \frac{\chi}{\kappa}(\bm{u}^{n+1}
+ * - \bm{u}_b) \f$. The class computes the coefficient \f$
+ * \frac{\chi}{\kappa}$ of the fluid velocity \f$ \bm{u}^{n+1} \f$ for the
+ * variable-coefficient INS solvers of INSVCStaggeredHierarchyIntegrator. This
+ * is done in the BrinkmanPenalizationRigidBodyDynamics::demarcateBrinkmanZone
+ * method. Here \f$ \chi \f$ is the body indicator function and \f$\kappa \sim
+ * \Delta t/ \rho \ll 1 \f$ is the vanishing permeability of the body. The
+ * rigid body velocity \f$\bm{u}_b\f$ is computed through Newton's law of
+ * motion by netting the hydrodynamic and external forces and torques on the
+ * body in the BrinkmanPenalizationRigidBodyDynamics::computeBrinkmanVelocity
+ * method. A simple forward-Euler scheme is imployed for the second-law of
+ * motion.
+ *
+ * For further information on applications of this class see
+ * https://arxiv.org/abs/1904.04078.
  */
 class BrinkmanPenalizationRigidBodyDynamics : public BrinkmanPenalizationStrategy
 {
--- a/include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
+++ b/include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
@@ -1,0 +1,356 @@
+// Filename: BrinkmanPenalizationRigidBodyDynamics.h
+// Created on 05 Dec 2018 by Amneet Bhalla
+//
+// Copyright (c) 2002-2018, Amneet Bhalla
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+//    * Redistributions of source code must retain the above copyright notice,
+//      this list of conditions and the following disclaimer.
+//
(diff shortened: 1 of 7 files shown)


drwells:
Another question for @amneetb: on the NWT branch we only have one example that directly uses this class, and that example is pretty complicated. Do we have any simpler test code for this?

src/IB/BrinkmanPenalizationRigidBodyDynamics.cpp
```diff
@@ -74,56 +76,9 @@ set_rotation_matrix(const Eigen::Vector3d& rot_vel,
...
-
-    Eigen::Vector3d W;
-
-    W[0] = (b3 * c2 * d1 - b2 * c3 * d1 - a3 * c2 * d2 + a2 * c3 * d2 + a3 * b2 * d3 - a2 * b3 * d3) / dnr;
-    W[1] = -(b3 * c1 * d1 - b1 * c3 * d1 - a3 * c1 * d2 + a1 * c3 * d2 + a3 * b1 * d3 - a1 * b3 * d3) / dnr;
-    W[2] = (b2 * c1 * d1 - b1 * c2 * d1 - a2 * c1 * d2 + a1 * c2 * d2 + a2 * b1 * d3 - a1 * b2 * d3) / dnr;
-
-    return W;
+    Eigen::FullPivLU<Eigen::Matrix3d> lu(T);
+    return lu.solve(L);
```
amneetb: Don't you like my pen-and-paper solution? ;-)

drwells: While your rendition of Cramer's rule was beautiful, I really would prefer to use a more stable algorithm.

amneetb:
Here you go:

`The penalization force is taken to be \f$ \frac{\chi}{\kappa}(\bm{u}^{n+1} - \bm{u}_b) \f$. The class computes the coefficient \f$ \frac{\chi}{\kappa}$ of the fluid velocity \f$ \bm{u}^{n+1} \f$ for INSVCStaggeredHierarchyIntegrator in the demarcateBrinkmanZone method. Here \f$ \chi \f$ is the body indicator function and \f$\kappa \sim \Delta t/ \rho\f \ll 1 $ is the vanishing permeability of the body. The rigid body velocity \f$\bm{u}_b\f$ is computed  through Newton's law of motion by netting the hydrodynamic and external forces and torques in the computeBrinkmanVelocity method. A simple forward-Euler scheme is imployed for the second-law of motion.`

Reference:
https://arxiv.org/abs/1904.04078

amneetb:
> Another question for @amneetb: on the NWT branch we only have one example that directly uses this class, and that example is pretty complicated. Do we have any simpler test code for this?

Yeah, I don't have a simple example yet. But I plan to add something simple in the future.

drwells:
Lets go ahead with this for now so that we can keep making progress on the NWT branch, but I will hold you to writing a simple test code in the future :)

include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
```diff
@@ -0,0 +1,372 @@
...
+{
+    ////////////////////////////// PUBLIC ////////////////////////////////////////
+
+public:
+    /*
+     * \brief Constructor of the class.
+     */
+    BrinkmanPenalizationRigidBodyDynamics(std::string object_name,
+                                          SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > ls_solid_var,
+                                          SAMRAI::tbox::Pointer<IBAMR::AdvDiffHierarchyIntegrator> adv_diff_solver,
```
boyceg: This is the *level set* solver, right? Should we name it as such?

include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
```diff
@@ -0,0 +1,372 @@
...
+    /*!
+     * \brief Register user defined velocity function data.
+     */
+    void registerKinematicsFunction(const KinematicsFcnData& data);
+
+    /*!
+     * \brief Set what rigid DOFs need to be solved for this
+     * particular structure.
+     */
+    virtual void setSolveRigidBodyVelocity(const IBTK::FreeRigidDOFVector& solve_rigid_dofs);
```
boyceg: At some point, should we make a common RBD base class to avoid reimplementing this?

include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
```diff
@@ -0,0 +1,372 @@
...
+
+    /*!
+     * \brief Write out object state to the given database.
+     */
+    void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db) override;
+
+    /*!
+     * \brief Get initial center of mass of the body.
+     */
+    const Eigen::Vector3d& getInitialCOMPosn() const
```
boyceg: Should these be marked as `inline`? (And possibly moved to a `private/BrinkmanPenalizationRigidBodyDynamics-inl.h` file? Although I do not care about this as much as I use to.)

include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
amneetb: This is used as part of IBLevelSetHI, but I think it is more appropriate to name it by the numerical technique.

include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
```diff
@@ -0,0 +1,372 @@
...
+                                    Eigen::Vector3d& hydro_torque_viscous) const
+    {
+        hydro_force_pressure = d_hydro_force_pressure;
+        hydro_force_viscous = d_hydro_force_viscous;
+        hydro_torque_pressure = d_hydro_torque_pressure;
+        hydro_torque_viscous = d_hydro_torque_viscous;
+        return;
+    } // getHydrodynamicForceTorque
+
+    /////////////////////////////// PROTECTED ////////////////////////////////////
```
boyceg:
```suggestion
```

include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
```diff
@@ -0,0 +1,372 @@
...
+ * body in the BrinkmanPenalizationRigidBodyDynamics::computeBrinkmanVelocity
+ * method. A simple forward-Euler scheme is imployed for the second-law of
+ * motion.
+ *
+ * For further information on applications of this class see
+ * https://arxiv.org/abs/1904.04078.
+ */
+class BrinkmanPenalizationRigidBodyDynamics : public BrinkmanPenalizationStrategy
+{
+    ////////////////////////////// PUBLIC ////////////////////////////////////////
```
boyceg:
```suggestion
```

include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
```diff
@@ -0,0 +1,372 @@
...
+
+    // Forces and torques on the body.
+    Eigen::Vector3d d_hydro_force_pressure, d_hydro_force_viscous, d_hydro_torque_pressure, d_hydro_torque_viscous,
+        d_ext_force, d_ext_torque;
+
+    // Routines to get prescribed kinematics and additional external forces and torques.
+    KinematicsFcnData d_kinematics_fcn_data;
+    ExternalForceTorqueFcnData d_ext_force_torque_fcn_data;
+
+    /////////////////////////////// PROTECTED ////////////////////////////////////
```
boyceg:
```suggestion
```

include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
amneetb: Actually it is used by INSVCHI and not by BLevelSetHI directly.

include/ibamr/BrinkmanPenalizationStrategy.h
```diff
@@ -0,0 +1,175 @@
...
+
+namespace IBAMR
+{
+/*!
+ * \brief BrinkmanPenalizationStrategy is an abstract class that provides an interface
+ * to implement Brinkman penalization body force in the momentum equation.
+ */
+class BrinkmanPenalizationStrategy : public SAMRAI::tbox::Serializable
+{
+    ////////////////////////////// PUBLIC ////////////////////////////////////////
```
boyceg:
```suggestion
```

include/ibamr/BrinkmanPenalizationStrategy.h
```diff
@@ -0,0 +1,175 @@
...
+
+    /*
+     * \brief Get the Brinkman coefficient.
+     */
+    double getBrinkmanCoefficient() const
+    {
+        return d_chi;
+    } // getBrinkmanPenaltyFactor
+
+    /////////////////////////////// PROTECTED ////////////////////////////////////
```
boyceg:
```suggestion
```

include/ibamr/BrinkmanPenalizationStrategy.h
```diff
@@ -0,0 +1,175 @@
...
+     * \brief Write out object state to the given database.
+     *
+     * \note An empty default implementation is provided.
+     */
+    void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db) override;
+
+    /*
+     * \brief Get the name of the object.
+     */
+    const std::string& getName() const
```
boyceg: Save comments as above.

include/ibamr/BrinkmanPenalizationStrategy.h
```diff
@@ -0,0 +1,175 @@
...
+     */
+    double d_current_time = std::numeric_limits<double>::quiet_NaN(),
+           d_new_time = std::numeric_limits<double>::quiet_NaN();
+
+    /*
+     * Brinkman coefficient.
+     */
+    double d_chi = 1e8;
+
+    /////////////////////////////// PROTECTED ////////////////////////////////////
```
boyceg:
```suggestion
```

src/IB/BrinkmanPenalizationRigidBodyDynamics.cpp
```diff
@@ -0,0 +1,516 @@
...
+    bool from_restart = RestartManager::getManager()->isFromRestart();
+    if (from_restart) getFromRestart();
+    if (input_db) getFromInput(input_db, from_restart);
+
+    d_hydro_force_eval->setSurfaceContourLevel(d_contour_level);
+
+    return;
+} // BrinkmanPenalizationRigidBodyDynamics
+
+BrinkmanPenalizationRigidBodyDynamics::~BrinkmanPenalizationRigidBodyDynamics()
```
boyceg: Should we default this?

include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
drwells: Functions defined in the class declaration are implicitly inline so its not needed.

src/IB/BrinkmanPenalizationRigidBodyDynamics.cpp
drwells: Yup.

include/ibamr/BrinkmanPenalizationRigidBodyDynamics.h
drwells: I am ambivalent about the name so I kept it as-is since it works with rigid bodies.

drwells:
I am pretty ambivalent about putting inline functions in the class declaration: its purely cosmetic and doesn't really impact readability when there is only one line of code. I made all other suggested changes aside from the rename.

PR #626, opened 2019-06-24
NWT: correctly normalize quaternions.

drwells:
This was done on the NWT branch, though its a bug in IBAMR. In Eigen, `Quaterniond::normalize()` normalizes the quaternion, while `Quaterniond::normalized()` copies the current object and normalizes the copy.

I searched through the rest of the library to make sure that we always do this correctly.


amneetb:
Looks good. These are the same changes I made in NWT, right? If so, go ahead and merge.

drwells:
@amneetb I made the same fix in `CIBFEMethod`; the rest of the changes are in NWT.

drwells:
Eigen's documentation states that 

```
    Operations interpreting the quaternion as rotation have undefined behavior if the quaternion is not normalized.
```
so normalizing can never hurt and in this case is required. Lets merge.

PR #633, opened 2019-06-25
run clang-format

boyceg:
Minor changes in the library along with updated header organization for all examples and tests.


drwells:
Most of these are header order changes. Lets merge once the testers are done.

PR #634, opened 2019-06-25
fix contrib dir path in ibtk

boyceg:


boyceg:
I am surprised that "bundled" versions of Boost, Eigen, and muParser worked when compiling the IBTK library.

drwells:
I have definitely used these. How bizarre!

drwells:
There is an easy explanation for boost and eigen: the second argument refers to the build directory. Since there is nothing to build, the second argument is never used.

For muParser: this is a link to the directory in which we build it: in principle we could set this value to something completely arbitrary since `configure` will correctly set up the path to the library based on the supplied value.

boyceg:
Without this change, IBTK builds fine, but IBTK examples do not.

(Because of reasons, running `make examples` in the IBAMR build directory only compiles the IBAMR examples and not the IBTK examples.)

drwells:
Hm, I must have misunderstood the IBTK configuration script.

drwells:
I don't fully understand how the paths work but I can verify that the problem exists on master and this branch fixes it.

boyceg:
> I don't fully understand how the paths work but I can verify that the problem exists on master and this branch fixes it.

Because IBTK is a full-fledged Automake package, its top-level build directory is `/path/to/ibamr-build-dir/ibtk`.

PR #635, opened 2019-06-25
Convert the rest of the IBTK tests.

drwells:
Part of #569. I am marking this as WIP since there are two tests that I can't figure out:
1. laplace_02_3d (the 3D side-centered laplace solver) doesn't converge in the infinity norm.
2. ~vc_viscous_solver_3d is only first order in the infinity norm but we have uniform grid refinement.~

These seem wrong, particularly since things work in 2D.


amneetb:
> 2\. vc_viscous_solver_3d is only first order in the infinity norm but we have uniform grid refinement.

Can you please check grid convergence with higher base resolution. With low viscosity, things are not second-order accurate on lower grid resolutions, but reaches 2nd order accuracy on finer grids. We have mentioned this in our paper.

drwells:
I tried this up to N = 256 and only got first order. Let me double-check.

amneetb:
> I tried this up to N = 256 and only got first order. Let me double-check.

See Fig.6 here https://www.sciencedirect.com/science/article/pii/S0021999119302256 and the related discussion for lower grid resolution.

drwells:
It takes this example a few more steps to get to second order:
```
[drwells@auricle IBTK]$ mpirun -np 1 ./vc_viscous_solver_3d ./vc_viscous_solver_3d.input 1>/dev/null; cat output
|e|_oo = 0.00983525
|e|_2  = 0.00290116
|e|_1  = 0.00368274
|r|_oo = 1.20508e-11
|r|_2  = 3.00537e-12
|r|_1  = 3.85195e-12
[drwells@auricle IBTK]$ mpirun -np 1 ./vc_viscous_solver_3d ./vc_viscous_solver_3d.input 1>/dev/null; cat output
|e|_oo = 0.00351775
|e|_2  = 0.000722825
|e|_1  = 0.000920591
|r|_oo = 2.42721e-11
|r|_2  = 3.22049e-12
|r|_1  = 3.78807e-12
[drwells@auricle IBTK]$ mpirun -np 1 ./vc_viscous_solver_3d ./vc_viscous_solver_3d.input 1>/dev/null; cat output
|e|_oo = 0.00114713
|e|_2  = 0.000180554
|e|_1  = 0.000230043
|r|_oo = 1.38101e-10
|r|_2  = 1.16218e-11
|r|_1  = 1.25067e-11
[drwells@auricle IBTK]$ mpirun -np 1 ./vc_viscous_solver_3d ./vc_viscous_solver_3d.input 1>/dev/null; cat output
|e|_oo = 0.000354218
|e|_2  = 4.51295e-05
|e|_1  = 5.7506e-05
|r|_oo = 4.64354e-10
|r|_2  = 4.78152e-11
|r|_1  = 5.22006e-11
```
These were with 32^3, 64^3, 128^3, and 256^3 (the last one takes a very long time on my workstation).

That leaves the other laplace solver which doesn't converge at all.

amneetb:
Great! It does what we say in the paper😁

Is the other test also committed by me? 🤔

amneetb:
One way to accelerate the test is to use higher base viscosity. I think it uses mu = 1 + sinusoidal. If we set mu = 10+ sinusoidal, we don’t need to go to higher grids. However, I do not Mathematica scripts to generate the RHS we stems from 

div . mu . (grad u + grad u^T) = RHS

nishantnangia:
I actually have the Mathematica script ;-) I'll send it to your email, Amneet.

amneetb:
Thanks! Can you also email me Mathematica which I also don't have ☹️

nishantnangia:
I only have Mathematica through a Northwestern license, so don't tell anyone 🤐

drwells:
@amneetb no, this example (`laplace_02_3d`) has been around for a long time. Maybe there is something wrong in the input file.

drwells:
Plotting a contour plot of the error for the failing test is quite illustrative:
visit0000

it looks like this test doesn't handle periodic boundary conditions correctly.

boyceg:
What is the original example?

drwells:
Its the side-centered Laplace solver (`ibtk/examples/SCLaplace/example.cpp`). I have to admit that I haven't looked at this (past drawing that picture) much and its totally possible that there is something silly going on.

drwells:
Yup: the input function is not periodic, so we run into problems at boundaries. I am not sure why its so much more pronounced at the coarse-fine interface, but here we are.

drwells:
Since this just updates tests I will merge as-is.

PR #637, opened 2019-06-26
add support for velocity jumps in IBFESurfaceMethod

boyceg:
This is the successor to #240.


src/IB/IBFESurfaceMethod.cpp
```diff
@@ -164,10 +166,22 @@ libmesh_restart_file_name(const std::string& restart_dump_dirname,
...
+const std::string IBFESurfaceMethod::VELOCITY_SYSTEM_NAME = "IB velocity system";
 const std::string IBFESurfaceMethod::NORMAL_VELOCITY_SYSTEM_NAME = "IB normal velocity system";
-const std::string IBFESurfaceMethod::PRESSURE_JUMP_SYSTEM_NAME = "IB [[p]] system";
 const std::string IBFESurfaceMethod::TANGENTIAL_VELOCITY_SYSTEM_NAME = "IB tangential velocity system";
-const std::string IBFESurfaceMethod::VELOCITY_SYSTEM_NAME = "IB velocity system";
+const std::string IBFESurfaceMethod::PRESSURE_JUMP_SYSTEM_NAME = "IB [[p]] system";
+const std::string IBFESurfaceMethod::WSS_SYSTEM_NAME = "One sided interior/exterior shear stress system";
+const std::string IBFESurfaceMethod::P_SYSTEM_NAME = "One sided interior/exterior pressure system";
+const std::string IBFESurfaceMethod::TAU_SYSTEM_NAME = "traction system";
+const boost::array<std::string, NDIM> IBFESurfaceMethod::VELOCITY_JUMP_SYSTEM_NAME = {
```
drwells:
```suggestion
const std::array<std::string, NDIM> IBFESurfaceMethod::VELOCITY_JUMP_SYSTEM_NAME = {
```

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -111,7 +120,14 @@ class IBFESurfaceMethod : public IBStrategy
     static const std::string NORMAL_VELOCITY_SYSTEM_NAME;
     static const std::string PRESSURE_JUMP_SYSTEM_NAME;
     static const std::string TANGENTIAL_VELOCITY_SYSTEM_NAME;
+    static const boost::array<std::string, NDIM> VELOCITY_JUMP_SYSTEM_NAME;
```
drwells:
```suggestion
    static const std::array<std::string, NDIM> VELOCITY_JUMP_SYSTEM_NAME;
```

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -547,15 +615,19 @@ class IBFESurfaceMethod : public IBStrategy
     IBTK::FEDataManager::SpreadSpec d_default_spread_spec;
     std::vector<IBTK::FEDataManager::InterpSpec> d_interp_spec;
     std::vector<IBTK::FEDataManager::SpreadSpec> d_spread_spec;
-    bool d_use_jump_conditions = false;
-    bool d_perturb_fe_mesh_nodes = true;
-    bool d_normalize_pressure_jump = false;
+    bool d_use_pressure_jump_conditions, d_use_velocity_jump_conditions;
+    bool d_use_l2_lagrange_family;
+    bool d_compute_fluid_traction;
+    bool d_perturb_fe_mesh_nodes;
```
drwells:
```suggestion
    bool d_perturb_fe_mesh_nodes = true;
```

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -547,15 +615,19 @@ class IBFESurfaceMethod : public IBStrategy
...
+    bool d_normalize_pressure_jump;
```
drwells:
```suggestion
    bool d_normalize_pressure_jump = false;
```

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1892,6 +3830,16 @@ IBFESurfaceMethod::commonConstructor(const std::string& object_name,
                                                       use_nodal_quadrature);
     d_default_spread_spec = FEDataManager::SpreadSpec(
         "IB_4", QGAUSS, INVALID_ORDER, use_adaptive_quadrature, point_density, use_nodal_quadrature);
+    d_ghosts = 0;
+    d_use_velocity_jump_conditions = false;
+    d_use_pressure_jump_conditions = false;
+    d_use_l2_lagrange_family = false;
+    d_compute_fluid_traction = false;
+    d_perturb_fe_mesh_nodes = true;
```
drwells:
```suggestion
```

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1892,6 +3830,16 @@ IBFESurfaceMethod::commonConstructor(const std::string& object_name,
...
+    d_normalize_pressure_jump = false;
```
drwells:
```suggestion
```

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -547,15 +615,19 @@ class IBFESurfaceMethod : public IBStrategy
...
     std::vector<libMesh::FEFamily> d_fe_family;
     std::vector<libMesh::Order> d_fe_order;
     std::vector<libMesh::QuadratureType> d_default_quad_type;
     std::vector<libMesh::Order> d_default_quad_order;
-    bool d_use_consistent_mass_matrix = true;
-    bool d_use_direct_forcing = false;
+    bool d_use_consistent_mass_matrix;
```
drwells:
```suggestion
    bool d_use_consistent_mass_matrix = true;
```

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1892,6 +3816,16 @@ IBFESurfaceMethod::commonConstructor(const std::string& object_name,
...
+    d_use_consistent_mass_matrix = true;
```
drwells:
```suggestion
```

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -547,15 +612,20 @@ class IBFESurfaceMethod : public IBStrategy
...
+    bool d_use_direct_forcing;
```
drwells:
```suggestion
    bool d_use_direct_forcing = false;
```

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1892,6 +3816,16 @@ IBFESurfaceMethod::commonConstructor(const std::string& object_name,
...
+    d_use_direct_forcing = false;
```
drwells:
```suggestion
```

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1892,6 +3816,16 @@ IBFESurfaceMethod::commonConstructor(const std::string& object_name,
...
+    d_do_log = false;
```
drwells:
```suggestion
```

drwells: (this should be initialized by the base class IIRC)

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -472,14 +504,29 @@ class IBFESurfaceMethod : public IBStrategy
 
 protected:
     /*!
-     * Impose (pressure) jump conditions.
+     * Impose the jump conditions.
      */
     void imposeJumpConditions(const int f_data_idx,
-                              libMesh::PetscVector<double>& DP_ghost_vec,
+                              libMesh::PetscVector<double>& P_jump_ghost_vec,
+                              boost::array<libMesh::PetscVector<double>*, NDIM>& DU_jump_ghost_vec,
```
drwells:
```suggestion
                              std::array<libMesh::PetscVector<double>*, NDIM>& DU_jump_ghost_vec,
```

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -529,14 +586,22 @@ class IBFESurfaceMethod : public IBStrategy
     const unsigned int d_num_parts = 1;
     std::vector<IBTK::FEDataManager*> d_fe_data_managers;
     SAMRAI::hier::IntVector<NDIM> d_ghosts = 0;
-    std::vector<libMesh::System*> d_X_systems, d_U_systems, d_U_n_systems, d_U_t_systems, d_F_systems, d_DP_systems;
+    std::vector<libMesh::System*> d_X_systems, d_U_systems, d_U_n_systems, d_U_t_systems, d_F_systems, d_P_jump_systems,
+        d_WSS_systems;
+    std::vector<libMesh::System*> d_P_systems, d_TAU_systems;
+    std::vector<boost::array<libMesh::System*, NDIM> > d_DU_jump_systems;
```
drwells:
```suggestion
    std::vector<std::array<libMesh::System*, NDIM> > d_DU_jump_systems;
```

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -529,14 +586,22 @@ class IBFESurfaceMethod : public IBStrategy
...
         d_X_IB_ghost_vecs;
     std::vector<libMesh::PetscVector<double>*> d_U_current_vecs, d_U_new_vecs, d_U_half_vecs;
     std::vector<libMesh::PetscVector<double>*> d_U_n_current_vecs, d_U_n_new_vecs, d_U_n_half_vecs;
     std::vector<libMesh::PetscVector<double>*> d_U_t_current_vecs, d_U_t_new_vecs, d_U_t_half_vecs;
+
     std::vector<libMesh::PetscVector<double>*> d_F_half_vecs, d_F_IB_ghost_vecs;
-    std::vector<libMesh::PetscVector<double>*> d_DP_half_vecs, d_DP_IB_ghost_vecs;
+    std::vector<libMesh::PetscVector<double>*> d_P_jump_half_vecs, d_P_jump_IB_ghost_vecs;
+    std::vector<libMesh::PetscVector<double>*> d_P_half_vecs, d_P_IB_ghost_vecs;
+    std::vector<boost::array<libMesh::PetscVector<double>*, NDIM> > d_DU_jump_half_vecs, d_DU_jump_IB_ghost_vecs;
```
drwells:
```suggestion
    std::vector<std::array<libMesh::PetscVector<double>*, NDIM> > d_DU_jump_half_vecs, d_DU_jump_IB_ghost_vecs;
```

boyceg:
(Note that this branch was created by `git diff`ing 8a3a42742973fbb083a572d06a448fda8a571cd8 against master. The branch has some additional improvements not contained in that branch, however.)

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -3195,37 +3195,21 @@ FEDataManager::collectGhostDOFIndices(std::vector<unsigned int>& ghost_dofs,
...
-                if (node->n_dofs(sys_num, var_num) > 0)
-                {
-                    const unsigned int dof_index = node->dof_number(sys_num, var_num, 0);
-                    if (dof_index < first_local_dof || dof_index >= end_local_dof)
-                    {
-                        ghost_dof_set.insert(dof_index);
-                    }
-                }
-            }
-        }
```
drwells: :+1:

boyceg:
(In turn, 8a3a42742973fbb083a572d06a448fda8a571cd8 was created by merging master into 94b9a30ebb99fdb4382a55b63781421de4c55950 and `clang-format`ting.)

drwells:
There is still a huge amount of code to go through. Fortunately all of the tests still pass.

boyceg:
I have been using Sublime Merge to look at diffs between branches.

drwells:
A few things come to mind that are easy to fix:
1. We need to do a global search and replace for `boost::array` -> `std::array`
2. Changelog entry
3. There are still a ton of calls to `localize` that we need to get rid of if this is going to work in parallel (but a lot of these are already present on `master`)

boyceg:
I just force pushed a commit that changes all `boost::array` to `std::array`.

`localize` works in parallel, it just isn't necessarily a good thing to do. :-)

Also, keep in mind that this is specialized to surface meshes, and so things that are expensive with volumetric meshes may not show up in profiling. The main focus for this branch right now is correctness and avoiding wanton inefficiency.

drwells:
Good point. I didn't mean to imply that we should go back and change the `localize` calls now.

boyceg:
I just changed all instances of `localize` to use `copy_and_synch` instead.

ibtk/src/lagrangian/FEDataManager.cpp
AminKolah: Just a note that this loop for collecting DOFs of the nodes is still present in the master branch...something to keep in mind when doing the merge

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -376,94 +403,233 @@ IBFESurfaceMethod::preprocessIntegrateData(double current_time, double new_time,
...
+            *d_TAU_half_vecs[part] = *d_TAU_systems[part]->solution;
         }
     }
     return;
 } // preprocessIntegrateData
 
+void
+IBFESurfaceMethod::calculateInterfacialFluidForces(double data_time)
+{
+    batch_vec_ghost_update(
```
AminKolah: `d_DU_jump_half_vecs` is missing in the `batch_vec_ghost_update`

boyceg:
> This looks great. I noticed the bugs you were referring to...Looks like this example wasn't being consistently updated while I was making change in IBFESurfaceMethod.cpp. I see that I wasn't careful enough with providing the switch for codim-0 /codim-1. The other version of the flow past a cylinder on gitlab (tailored only for IIM) doesn't have any of these issues though .
> 
> Can we get rid of the U_L1_norm_stream, U_L2_norm_stream and U_max_norm_stream ? It doesn't seem like we are logging anything into those files.

Can you update ex5 in the library and verify that you can reproduce corresponding results?

examples/IBFE/explicit/ex5/input2d.iim
```diff
@@ -0,0 +1,248 @@
...
+   enable_logging      = ENABLE_LOGGING
+}
+
+IBFESurfaceMethod {
+   IB_delta_fcn                = IB_DELTA_FUNCTION
+   split_forces                = SPLIT_FORCES
+   use_pressure_jump_conditions   = USE_PRESSURE_JUMP_CONDITIONS
+   use_velocity_jump_conditions   = USE_VELOCITY_JUMP_CONDITIONS
+   use_l2_lagrange_family = USE_L2_LAGRANGE_FAMILY
+   use_pointwise_jump_imposition = USE_POINTWISE_JUMP_IMPOSITION
```
AminKolah: I should have removed the flag `use_pointwise_jump_imposition = USE_POINTWISE_JUMP_IMPOSITION` since the point wise jump conditions have been removed from the code.

boyceg: OK --- please make a commit that updates to the input files and source code as needed. I will squash together the commits once everything is finished.

AminKolah: Sounds good..Im working on it  now.

examples/IBFE/explicit/ex5/input2d.iim
```diff
@@ -0,0 +1,247 @@
...
+SPLIT_FORCES               = TRUE             // whether to split interior and boundary forces
+USE_PRESSURE_JUMP_CONDITIONS        = TRUE             // whether to impose pressure jumps at fluid-structure interfaces
+USE_VELOCITY_JUMP_CONDITIONS = TRUE			   // whether to impose the velocity jumps at fluid-structure interfaces
+COMPUTE_FLUID_TRACTION 	   = TRUE              // whether to compute the exterior fluid traction
+USE_CONSISTENT_MASS_MATRIX = TRUE              // whether to use a consistent or lumped mass matrix
+
+USE_L2_LAGRANGE_FAMILY = FALSE  				// Whether to use L2_Lagrange family for the calc of jump variables and traction
+
+WSS_CALC_WIDTH 				= 1.05			   // ratio multiplied by the diagonal side of the grid spacing to calculate wss
+P_CALC_WIDTH 			   = 1.5               // ratio multiplied by the diogonal side of the grid spacing to calculate the pressure
```
abarret:
```suggestion
P_CALC_WIDTH 			   = 1.5               // ratio multiplied by the diagonal side of the grid spacing to calculate the pressure
```

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -111,7 +120,14 @@ class IBFESurfaceMethod : public IBStrategy
...
     static const std::string PRESSURE_JUMP_SYSTEM_NAME;
     static const std::string TANGENTIAL_VELOCITY_SYSTEM_NAME;
+    static const std::array<std::string, NDIM> VELOCITY_JUMP_SYSTEM_NAME;
     static const std::string VELOCITY_SYSTEM_NAME;
+    static const std::string WSS_SYSTEM_NAME;
+    static const std::string P_SYSTEM_NAME;
+    static const std::string TAU_SYSTEM_NAME;
+
+    SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > mask_var;
+    int mask_current_idx, mask_new_idx, mask_scratch_idx;
```
abarret:
```suggestion
    int mask_current_idx = IBTK::invalid_index, mask_new_idx = IBTK::invalid_index, mask_scratch_idx = IBTK::invalid_index;
```
Are these actually used anywhere?

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -547,7 +612,10 @@ class IBFESurfaceMethod : public IBStrategy
     IBTK::FEDataManager::SpreadSpec d_default_spread_spec;
     std::vector<IBTK::FEDataManager::InterpSpec> d_interp_spec;
     std::vector<IBTK::FEDataManager::SpreadSpec> d_spread_spec;
-    bool d_use_jump_conditions = false;
+    bool d_use_pressure_jump_conditions, d_use_velocity_jump_conditions;
+    bool d_use_l2_lagrange_family;
+    bool d_compute_fluid_traction;
+    bool d_traction_interior_side;
```
abarret: Do these not get default values?

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -556,6 +624,8 @@ class IBFESurfaceMethod : public IBStrategy
     std::vector<libMesh::Order> d_default_quad_order;
     bool d_use_consistent_mass_matrix = true;
     bool d_use_direct_forcing = false;
+    double d_wss_calc_width, d_p_calc_width;
+    double d_traction_activation_time;
```
abarret: Can we document somewhere what these values do? It's not immediately clear to me what these values do in the computations.

Should they also get default values?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1881,6 +3813,9 @@ IBFESurfaceMethod::commonConstructor(const std::string& object_name,
     // Set some default values.
     const bool use_adaptive_quadrature = true;
     const int point_density = 2.0;
+    d_wss_calc_width = 0.0;
+    d_p_calc_width = 0.0;
+    d_traction_activation_time = 0.0;
```
abarret: Found them. Can these default values go into the header?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1892,6 +3827,11 @@ IBFESurfaceMethod::commonConstructor(const std::string& object_name,
                                                       use_nodal_quadrature);
     d_default_spread_spec = FEDataManager::SpreadSpec(
         "IB_4", QGAUSS, INVALID_ORDER, use_adaptive_quadrature, point_density, use_nodal_quadrature);
+    d_ghosts = 0;
+    d_use_velocity_jump_conditions = false;
+    d_use_pressure_jump_conditions = false;
+    d_use_l2_lagrange_family = false;
+    d_compute_fluid_traction = false;
```
abarret: Can these go into the header?

include/ibamr/IBFESurfaceMethod.h
drwells: I thought we fixed this yesterday.

drwells:
It doesn't look like this will be ready by tomorrow. Will it be done in the next week?

examples/IBFE/explicit/ex5/input2d.iim
```diff
@@ -0,0 +1,247 @@
...
+DX0 = L/N                                      // mesh width on coarsest grid level
+DX  = L/NFINEST                                // mesh width on finest   grid level
+MFAC = 2.0                                     // ratio of Lagrangian mesh width to Cartesian mesh width
+ELEM_TYPE = "TRI3"                             // type of element to use for structure discretization
+PK1_DEV_QUAD_ORDER = "SEVENTH"
+PK1_DIL_QUAD_ORDER = "FIFTH"
+USE_BOUNDARY_MESH = TRUE
+
+// solver parameters
+IB_DELTA_FUNCTION          = "IB_3"            // the type of smoothed delta function to use for Lagrangian-Eulerian interaction
```
abarret:
```suggestion
```

Can we cleanup the input file here? This shouldn't a required input when using jump conditions.

examples/IBFE/explicit/ex5/input2d.iim
AminKolah: @boyceg should we keep these values in the input files? WSS_CALC_WIDTH = 1.05 and P_CALC_WIDTH = 1.5 have worked for all cases, but maybe we can do a little more investigation further along?

include/ibamr/IBFESurfaceMethod.h
AminKolah: Yup I'm using this Eulerian mask variable for the calculation of the pressure in `calculateInterfacialFluidForces`. Maybe there are other ways to go about this too?

include/ibamr/IBFESurfaceMethod.h
AminKolah: Currently we assign some of the default values in `commonConstructor` and some in `IBFESurfaceMethod.h`. Do you suggest to have them all in the .h file? Is there any consistent way to set these values?

examples/IBFE/explicit/ex5/input2d.iim
boyceg: I think we want the input files also to work if all of the jump conditions are turned off, so I would be in favor of keeping this.

include/ibamr/IBFESurfaceMethod.h
boyceg: It does not look like that variable is actually being used as a masking variable (i.e. it is not being used to mark locations on the Cartesian grid). It looks like you are using it as a temporary copy of the pressure. I would suggest cloning the pressure scratch index. There is some infrastructure to do this and automatically setup the data allocation using a new class, `SAMRAIDataCache`.

Please switch over to using `SAMRAIDataCache` and get rid of the mask variable and indices.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: I think this should be fine, but the test suite should catch it if this actually breaks the ghosting indexing in parallel --- right @drwells???? :-)

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -376,94 +403,233 @@ IBFESurfaceMethod::preprocessIntegrateData(double current_time, double new_time,
...
     }
     return;
 } // preprocessIntegrateData
 
+void
+IBFESurfaceMethod::calculateInterfacialFluidForces(double data_time)
+{
+    batch_vec_ghost_update(
+        { d_X_new_vecs, d_U_new_vecs, d_U_n_new_vecs, d_U_t_new_vecs, d_F_half_vecs }, INSERT_VALUES, SCATTER_FORWARD);
+    batch_vec_ghost_update(
```
boyceg: None of these vectors appear to be used in this routine. Please move the ghost updates to the routines where the vectors are actually used.

Also, the idea of `batch_vec_ghost_update` is to do the updates simultaneously, so just put all of the vectors to be updated in the same call.

include/ibamr/IBFESurfaceMethod.h
boyceg: You should set all of the initial values that are consistent between all configurations in the header and get rid of the initialization in the `.cpp` file.

boyceg: (This makes it easier to see what the defaults are.)

boyceg: Also, setup the defaults for IB, not IIM. We don't want to change the default behavior at this point.

examples/IBFE/explicit/ex5/input2d.iim
boyceg: I would suggest setting up the IIM input files for this example to correspond to what was done for the paper.

boyceg:
@AminKolah, what is your availability to finish up the requested changes?

AminKolah:
> @AminKolah, what is your availability to finish up the requested changes?

I will finish this up today

include/ibamr/IBFESurfaceMethod.h
AminKolah: How can I clone the pressure scratch index and use wider ghost width? Don't I need to create and maintain a wider width variable myself? When I tried doing what you recommended here I get the following error:

```
P=00000:hier::VariableDatabase<DIM>::registerVariableAndContext error ...
P=00000:Attempting to to register variable ::P and context INSStaggeredHierarchyIntegrator::SCRATCH with ghost width = (4,4)
P=00000: This variable-context pair is already registered with a different ghost width. 
P=00000:
Abort trap: 6
```

AminKolah: I think that was the reason I used the mask variable with a larger ghost width (6) needed for the calculation of the pressure in the traction. @abarret do you know a way to clone the pressure scratch index with a wider ghost width than 1 and avoid introducing a mask variable?

src/IB/IBFESurfaceMethod.cpp
AminKolah: I moved this to the corresponding routines

examples/IBFE/explicit/ex5/example.cpp
```diff
@@ -280,7 +282,8 @@ run_example(int argc, char** argv)
         const double dx = input_db->getDouble("DX");
         const double ds = input_db->getDouble("MFAC") * dx;
         string elem_type = input_db->getString("ELEM_TYPE");
-        const double R = 0.5;
+        const double R = input_db->getDouble("R");
+
```
boyceg:
```suggestion
```

examples/IBFE/explicit/ex5/example.cpp
```diff
@@ -529,17 +536,15 @@ run_example(int argc, char** argv)
...
-            U_L1_norm_stream.open("U_L1.curve", ios_base::out | ios_base::trunc);
-            U_L2_norm_stream.open("U_L2.curve", ios_base::out | ios_base::trunc);
-            U_max_norm_stream.open("U_max.curve", ios_base::out | ios_base::trunc);
-
-            drag_stream.precision(10);
-            lift_stream.precision(10);
-            U_L1_norm_stream.precision(10);
-            U_L2_norm_stream.precision(10);
-            U_max_norm_stream.precision(10);
+            drag_F_stream.open("C_F_D.curve", ios_base::out | ios_base::trunc);
```
boyceg: Please do not change the default way this example works, e.g. by changing the output streams.

examples/IBFE/explicit/ex5/example.cpp
```diff
@@ -529,17 +536,15 @@ run_example(int argc, char** argv)
...
+            lift_F_stream.open("C_F_L.curve", ios_base::out | ios_base::trunc);
+            drag_TAU_stream.open("C_T_D.curve", ios_base::out | ios_base::trunc);
```
boyceg: These only should be active if IIM is being used.

examples/IBFE/explicit/ex5/example.cpp
```diff
@@ -529,17 +536,15 @@ run_example(int argc, char** argv)
         // velocity.
         if (SAMRAI_MPI::getRank() == 0)
         {
-            drag_stream.open("C_D.curve", ios_base::out | ios_base::trunc);
-            lift_stream.open("C_L.curve", ios_base::out | ios_base::trunc);
-            U_L1_norm_stream.open("U_L1.curve", ios_base::out | ios_base::trunc);
```
boyceg: Please do not remove output streams or change the names of existing output streams.

examples/IBFE/explicit/ex5/example.cpp
```diff
@@ -678,8 +711,8 @@ postprocess_data(Pointer<Database> input_db,
     std::vector<const std::vector<libMesh::VectorValue<double> >*> grad_var_data;
 
     TensorValue<double> FF, FF_inv_trans;
-    boost::multi_array<double, 2> x_node, U_node;
-    VectorValue<double> F, N, U, n, x;
+    boost::multi_array<double, 2> x_node, X0_node, U_node, TAU_node;
```
boyceg: `X0` should just be `X`. `X` indicates reference coordinates --- there is no need to affix a `0`.

examples/IBFE/explicit/ex5/input2d
```diff
@@ -2,7 +2,7 @@
 Re = 200.0
 MU = 1.0/Re
 RHO = 1.0
-
+R = 0.5        											// radius of the cylinder
 // grid spacing parameters
```
boyceg: Please put a gap between physical parameters and numerical ones.

```suggestion

// grid spacing parameters
```

examples/IBFE/explicit/ex5/input2d.iim
```diff
@@ -0,0 +1,245 @@
+// physical parameters
+Re = 200.0
+MU = 1.0/Re
+RHO = 1.0
+R = 0.5        											// radius of the cylinder
+U_MAX = 2.0
+// grid spacing parameters
```
boyceg:
```suggestion

// grid spacing parameters
```

examples/IBFE/explicit/ex5/input2d.iim
```diff
@@ -0,0 +1,245 @@
...
+DX  = L/NFINEST                                // mesh width on finest   grid level
+MFAC = 2.0                                     // ratio of Lagrangian mesh width to Cartesian mesh width
+ELEM_TYPE = "TRI3"                             // type of element to use for structure discretization
+PK1_DEV_QUAD_ORDER = "SEVENTH"
+PK1_DIL_QUAD_ORDER = "FIFTH"
+USE_BOUNDARY_MESH = TRUE
+
+// solver parameters
+IB_DELTA_FUNCTION          = "IB_3"            // the type of smoothed delta function to use for Lagrangian-Eulerian interaction
+SPLIT_FORCES               = TRUE             // whether to split interior and boundary forces
```
boyceg: Please align the comments (and, since they need to be updated anyway, please start each comment with lowercase).

examples/IBFE/explicit/ex5/input2d.test
```diff
@@ -2,7 +2,7 @@
 Re = 200.0
 MU = 1.0/Re
 RHO = 1.0
-
+R = 0.5        											// radius of the cylinder
 // grid spacing parameters
```
boyceg:
```suggestion

// grid spacing parameters
```

examples/IBFE/explicit/ex5/input3d
```diff
@@ -2,7 +2,7 @@
 Re = 400.0
 MU = 1.0/Re
 RHO = 1.0
-
+R = 0.5        											// radius of the cylinder
 // grid spacing parameters
```
boyceg:
```suggestion

// grid spacing parameters
```

examples/IBFE/explicit/ex5/example.cpp
AminKolah: Should C_D and C_L be computed from the exterior traction then? I had output streams of the force coefficients using both forces and fluid traction. Also U_L1 and U_L2 are not being used anywhere in the code. That's why I removed them.

examples/IBFE/explicit/ex5/example.cpp
AminKolah: okay I think this is the answer to my previous question.

examples/IBFE/explicit/ex5/example.cpp
boyceg: OK, if they are not being used in the original example, then it is definitely a good idea to remove them.

I would set this up to use the forces to compute C_D and C_L, regardless of the method configuration. Then, if it is possible to extract the traction accurately, there could be additional streams that are activated.

examples/IBFE/explicit/ex5/example.cpp
```diff
@@ -693,58 +726,87 @@ postprocess_data(Pointer<Database> input_db,
...
             }
         }
-        for (unsigned short int side = 0; side < elem->n_sides(); ++side)
+        if (!use_boundary_mesh)
         {
-            if (elem->neighbor_ptr(side)) continue;
-            fe_face->reinit(elem, side);
-            const unsigned int n_qp_face = qrule_face->n_points();
-            for (unsigned int qp = 0; qp < n_qp_face; ++qp)
+            for (unsigned short int side = 0; side < elem->n_sides(); ++side)
```
boyceg: This is not the correct algorithm for evaluating lift/drag forces for a volumetric mesh. Please revert these changes.

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -101,6 +100,16 @@ namespace IBAMR
  * \brief Class IBFESurfaceMethod is an implementation of the abstract base
  * class IBStrategy that provides functionality required by the IB method with
  * a finite element representation of a surface mesh.
+ * Additionally an immersed interface method (IIM) has been implemented with calculations
+ * of shear stress, pressure and total fluid traction.
+ *
+ * References:
+ * Kolahdouz et al., <A HREF="https://arxiv.org/abs/1812.06840"> An Immersed Interface
+ * Method for Faceted Surfaces
```
boyceg:
```suggestion
 * Method for Discrete Surfaces
```

include/ibamr/IBFESurfaceMethod.h
boyceg: Oh, OK --- in that case, probably the easiest thing to do will be to register a `SCRATCH` variable to store the ghost data.

Please change the name from `mask` to `p` and get rid of the `current` and `new` indices --- just set up a new `scratch` index.

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -295,13 +311,29 @@ class IBFESurfaceMethod : public IBStrategy
...
 
+    /*!
+     * Compute the fluid traction if all jump conditions are applied.
+     */
+    void computeFluidTraction(double current_time, unsigned int part = 0);
+
+    /*!
+     * Compute the pressure used in the calculation of the fluid traction (pressure jump condition needs to be applied).
+     */
+    void interpolatePressureForTraction(int p_data_idx, double data_time, unsigned int part = 0);
```
boyceg: Should this be a `public` function, or should it be a `private` helper function? (Does it make sense for users to call this function, or should it only be called from inside the class?)

If it makes sense for external callers to use these functions, would it be more clear to split them into functions that evaluate the pressure and viscous stresses on the interface (e.g. `evaluatePressureOnInterface()` and `evaluateViscousStressOnInterface()`). `interpolatePressureForTraction()` sounds like this is a helper function.

I would suggest making all of these functions `private` for now, and we can worry about the `public` interface later if needed (e.g. to support better post processing).

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -472,14 +504,29 @@ class IBFESurfaceMethod : public IBStrategy
...
+                              std::array<libMesh::PetscVector<double>*, NDIM>& DU_jump_ghost_vec,
                               libMesh::PetscVector<double>& X_ghost_vec,
                               const double data_time,
                               const unsigned int part);
-
+    /*!
+     * \brief Helper function for checking possible double-counting
+     *  intesection points
+     */
+    void checkDoubleCountingIntersection(int axis,
```
boyceg: If this function does not use any class state, then suggest moving it to `libmesh_utilities` as a static member function along with the other intersection routines.

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -521,6 +568,16 @@ class IBFESurfaceMethod : public IBStrategy
 
     /*
      * FE data associated with this object.
+     * d_X_systems: IB coordinates system
```
boyceg: Probably can drop `IB` here:

```suggestion
     * d_X_systems: coordinates system
```

etc.

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -521,6 +568,16 @@ class IBFESurfaceMethod : public IBStrategy
...
+     * d_F_systems: IB force system
+     * d_U_systems: IB velocity system
+     * d_U_n_systems: IB normal velocity system
+     * d_U_t_sytems: IB tangential velocity system
+     * d_P_jump_systems: IB pressure jump system [[p]]
+     * d_DU_jump_systems:
```
boyceg: Don't forget this one. :-)

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -521,6 +568,16 @@ class IBFESurfaceMethod : public IBStrategy
...
+     * d_WSS_systems: one sided interior/exterior shear stress system
```
boyceg: Add formulae for these comments?

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -529,14 +586,22 @@ class IBFESurfaceMethod : public IBStrategy
     const unsigned int d_num_parts = 1;
     std::vector<IBTK::FEDataManager*> d_fe_data_managers;
     SAMRAI::hier::IntVector<NDIM> d_ghosts = 0;
-    std::vector<libMesh::System*> d_X_systems, d_U_systems, d_U_n_systems, d_U_t_systems, d_F_systems, d_DP_systems;
+    std::vector<libMesh::System*> d_X_systems, d_U_systems, d_U_n_systems, d_U_t_systems, d_F_systems, d_P_jump_systems,
```
boyceg: Please merge this with the next line.

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -529,14 +586,22 @@ class IBFESurfaceMethod : public IBStrategy
...
+    std::vector<std::array<libMesh::System*, NDIM> > d_DU_jump_systems;
     std::vector<libMesh::PetscVector<double>*> d_X_current_vecs, d_X_new_vecs, d_X_half_vecs, d_X0_vecs,
         d_X_IB_ghost_vecs;
     std::vector<libMesh::PetscVector<double>*> d_U_current_vecs, d_U_new_vecs, d_U_half_vecs;
     std::vector<libMesh::PetscVector<double>*> d_U_n_current_vecs, d_U_n_new_vecs, d_U_n_half_vecs;
     std::vector<libMesh::PetscVector<double>*> d_U_t_current_vecs, d_U_t_new_vecs, d_U_t_half_vecs;
+
     std::vector<libMesh::PetscVector<double>*> d_F_half_vecs, d_F_IB_ghost_vecs;
-    std::vector<libMesh::PetscVector<double>*> d_DP_half_vecs, d_DP_IB_ghost_vecs;
+    std::vector<libMesh::PetscVector<double>*> d_P_jump_half_vecs, d_P_jump_IB_ghost_vecs;
```
boyceg: Please keep all of the `P` vectors together.

include/ibamr/IBFESurfaceMethod.h
boyceg: Also, suggest using the ordering:

1. positions
2. velocities
3. forces
4. jump conditions
5. post-processed interfacial quantities.

There are a lot of variables now, and I think it will be helpful for others reading the code if things are listed consistently.

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -1032,7 +1032,6 @@ get_max_edge_length(const libMesh::Elem* const elem, const MultiArray& X_node)
 {
 #ifndef NDEBUG
     TBOX_ASSERT(elem->n_nodes() == X_node.shape()[0]);
-    TBOX_ASSERT(elem->dim() == X_node.shape()[1]);
```
boyceg: @drwells can you take a look at this? Is there a more appropriate assertion to include here?

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -1,6 +1,5 @@
 // Filename: IBFESurfaceMethod.h
 // Created on 19 May 2018 by Boyce Griffith
-//
```
boyceg: Please put this line back in.

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -111,7 +120,14 @@ class IBFESurfaceMethod : public IBStrategy
     static const std::string NORMAL_VELOCITY_SYSTEM_NAME;
```
boyceg: Following the suggestion below, suggest ordering variables into groups: positions; velocities; forces; jump conditions; post-processed quantities (tractions etc).

(Alternatively, please use alphabetical order.)

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1,6 +1,5 @@
 // Filename: IBFESurfaceMethod.cpp
 // Created on 19 May 2018 by Boyce Griffith
-//
```
boyceg: Please do not remove this line.

drwells:
I'll take a look, but it may not be until tonight or tomorrow morning.

ibtk/src/lagrangian/FEDataManager.cpp
drwells: I think so, but it wouldn't hurt to add another test (the IB ghost data is available, albeit indirectly, through `FEDataManager::buildGhostedSolutionVector`)

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -894,38 +1489,77 @@ IBFESurfaceMethod::computeLagrangianForce(const double data_time)
         for (unsigned int d = 0; d < NDIM; ++d)
         {
             TBOX_ASSERT(X_dof_map.variable_type(d) == X_fe_type);
-            TBOX_ASSERT(X_dof_map.variable_type(d) == F_fe_type);
         }
+        TBOX_ASSERT(X_fe_type == F_fe_type);
```
drwells: Why was this changed?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -940,6 +1574,7 @@ IBFESurfaceMethod::computeLagrangianForce(const double data_time)
 
         // Loop over the elements to compute the right-hand side vector.
         boost::multi_array<double, 2> X_node, x_node;
+        double DU[NDIM][NDIM];
```
drwells: We should reduce the scope of this array and define it in the loop.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1107,22 +1812,40 @@ IBFESurfaceMethod::spreadForce(const int f_data_idx,
...
     for (unsigned int part = 0; part < d_num_parts; ++part)
     {
         PetscVector<double>* X_vec = d_X_half_vecs[part];
         PetscVector<double>* X_ghost_vec = d_X_IB_ghost_vecs[part];
         PetscVector<double>* F_vec = d_F_half_vecs[part];
         PetscVector<double>* F_ghost_vec = d_F_IB_ghost_vecs[part];
-        X_vec->localize(*X_ghost_vec);
-        F_vec->localize(*F_ghost_vec);
+        copy_and_synch(*X_vec, *X_ghost_vec);
+        copy_and_synch(*F_vec, *F_ghost_vec);
```
drwells: We should move this communication out of the loop over different parts. In `IBFEMethod::interpolateVelocity` we do
```cpp

    batch_vec_copy(X_vecs, d_X_IB_ghost_vecs);
    batch_vec_ghost_update(d_X_IB_ghost_vecs, INSERT_VALUES, SCATTER_FORWARD);
```
which will have the same effect but work much better with multiple parts.

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1107,22 +1812,40 @@ IBFESurfaceMethod::spreadForce(const int f_data_idx,
...
             f_data_idx, *F_ghost_vec, *X_ghost_vec, FORCE_SYSTEM_NAME, f_phys_bdry_op, data_time);
-        if (d_use_jump_conditions)
+        PetscVector<double>* P_jump_ghost_vec;
+        std::array<PetscVector<double>*, NDIM> DU_jump_ghost_vec;
+        std::array<PetscVector<double>*, NDIM> DU_jump_vec;
+        if (d_use_pressure_jump_conditions)
+        {
+            PetscVector<double>* P_jump_vec = d_P_jump_half_vecs[part];
+            P_jump_ghost_vec = d_P_jump_IB_ghost_vecs[part];
+            copy_and_synch(*P_jump_vec, *P_jump_ghost_vec);
```
drwells: same

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1107,22 +1812,40 @@ IBFESurfaceMethod::spreadForce(const int f_data_idx,
...
+        }
+        if (d_use_velocity_jump_conditions)
+        {
+            DU_jump_ghost_vec = d_DU_jump_IB_ghost_vecs[part];
+            DU_jump_vec = d_DU_jump_half_vecs[part];
+            for (unsigned int d = 0; d < NDIM; ++d)
+            {
+                copy_and_synch(*DU_jump_vec[d], *DU_jump_ghost_vec[d]);
```
drwells: same

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1229,10 +1952,78 @@ IBFESurfaceMethod::initializeFEEquationSystems()
...
+                        // TODO: The value of the FEFamily should just be read in from the input file directly instead
+                        // of using a boolean.
+                        if (d_use_l2_lagrange_family)
+                        {
+                            DU_jump_system[d]->add_variable(system_name, d_fe_order[part], L2_LAGRANGE);
+                        }
+                        else
+                        {
+                            DU_jump_system[d]->add_variable(system_name, d_fe_order[part], d_fe_family[part]);
+                        }
```
drwells: This logic seems overly complicated both here and below. Can we just resolve this TODO now?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1372,113 +2193,786 @@ void IBFESurfaceMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*h
...
+            x_upper_gh[d] = x_upper[d] + (static_cast<double>(p_ghost_num)) * dx[d];
+        }
+
+        double* x_upper_ghost = x_upper_gh.data();
+        double* x_lower_ghost = x_lower_gh.data();
+
+        // Setup vectors to store the values of U and X at the quadrature
+        // points.
+        //
+        // All this loop is doing is computing the total number of quadraturee
```
drwells:
```suggestion
        // All this loop is doing is computing the total number of quadrature
```

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1372,113 +2193,786 @@ void IBFESurfaceMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*h
...
+
+    d_fe_data_managers[part]->computeL2Projection(
+        *P_vec, *P_rhs_vec, P_SYSTEM_NAME, d_default_interp_spec.use_consistent_mass_matrix);
+
+    d_X_half_vecs[part]->close();
+    d_X_current_vecs[part]->close();
+    d_X_new_vecs[part]->close();
+
+    d_P_half_vecs[part]->close();
+    d_X_IB_ghost_vecs[part]->close();
```
drwells: What's going on here? Why are we closing all of these vectors?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -1372,113 +2193,786 @@ void IBFESurfaceMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*h
...
-        const int write_mode = EquationSystems::WRITE_DATA | EquationSystems::WRITE_ADDITIONAL_DATA;
-        d_equation_systems[part]->write(file_name, xdr_mode, write_mode, /*partition_agnostic*/ true);
-    }
-    return;
-}
+    d_WSS_IB_ghost_vecs[part]->close();
 
-/////////////////////////////// PROTECTED ////////////////////////////////////
+    d_P_IB_ghost_vecs[part]->close();
+    d_X_IB_ghost_vecs[part]->close();
```
drwells: Same here: why are we closing the position vectors?

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -161,13 +161,25 @@ libmesh_restart_file_name(const std::string& restart_dump_dirname,
 }
 } // namespace
 
-const std::string IBFESurfaceMethod::COORDS_SYSTEM_NAME = "IB coordinates system";
-const std::string IBFESurfaceMethod::COORD_MAPPING_SYSTEM_NAME = "IB coordinate mapping system";
+const std::string IBFESurfaceMethod::COORDS_SYSTEM_NAME = "coordinates system";
+const std::string IBFESurfaceMethod::COORD_MAPPING_SYSTEM_NAME = "coordinate mapping system";
```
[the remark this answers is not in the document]
AminKolah: Originally these names were assigned based on IB method. With an IIM it makes sense to remove the IB for the coordinates but keep the one for the force. We could also keep "IB" for all of the system as long as it's seen as referring to general "immersed" boundary and not the method.

drwells:
I just chatted with @AminKolah about this branch. We will try to get this done in the next two weeks or so but this might not happen due to the start of the semester.

AminKolah:
@boyceg  @drwells I just did a rebase against master branch and I realized there is no support for Jacobian calculation of line elements in JacobianCalculator. I fixed that in my own branch.. Something to keep in mind once we're ready to do the merge with master

AminKolah:
> @boyceg @drwells I just did a rebase against `master` branch and I realized there is no support for Jacobian calculation of line elements in `JacobianCalculator`. I fixed that in my own branch.. Something to keep in mind once we're ready to do the merge with master

I just made a change  in my branch so it uses Libmesh FEMap for Jacobian calculation whenever `NDIM - mesh.mesh_dimension() == 1`  Otherwise it uses the new built-in Jacobian calculation for volumetric meshes.

drwells:
I would prefer not to go that way - we should keep FEDataManager dimension-independent. This will be fixed in time for the release (see #775).

AminKolah:
> I would prefer not to go that way - we should keep FEDataManager dimension-independent. This will be fixed in time for the release (see #775).

I coded up a 1D line element in the `JacobianCalculator`, but haven't had a chance to do that for 2D patches. If you'd like to add that, that'd be great:-) .. otherwise I don't think bringing in Jacobian calculation from Libmesh for codim-1 problems would be any issue either since the use of that is really minimal. Note that we don't even do any spreading of forces when using the jumps.

drwells:
This works but having these special cases floating around is going to make maintenance much more difficult. The right choice is to fix our own code and keep FEDataManager simpler.

There is something wrong with the history on this branch: @AminKolah did you merge with master? If so, would you please undo the merge and rebase instead?

AminKolah:
> This works but having these special cases floating around is going to make maintenance much more difficult. The right choice is to fix our own code and keep FEDataManager simpler.
> 
> There is something wrong with the history on this branch: @AminKolah did you merge with master? If so, would you please undo the merge and rebase instead?

I have not done any merging. Can you be more specific about the issue with the history of the branch? I rebased against the `master` branch a few days ago. The commit is at 8c8324bfebdf83b58fb3a39763e19b99067d684f

However, it seems like changing in the master branch has been made pretty frequently during the past few days. I don't see a merge to undo but I can rebase again.

AminKolah:
> > This works but having these special cases floating around is going to make maintenance much more difficult. The right choice is to fix our own code and keep FEDataManager simpler.
> > There is something wrong with the history on this branch: @AminKolah did you merge with master? If so, would you please undo the merge and rebase instead?
> 
> I have not done any merging. Can you be more specific about the issue with the history of the branch? I rebased against the `master` branch a few days ago. The commit is at [8c8324b](https://github.com/IBAMR/IBAMR/commit/8c8324bfebdf83b58fb3a39763e19b99067d684f)
> 
> However, it seems like changing in the master branch has been made pretty frequently during the past few days. I don't see a merge to undo but I can rebase again.

W rebasing my recent commits sit on top of the master branch, right? If you look at the history, I maybe have up to 20 commits  after your push on `master` branch on Nov 25 at a014526a244ac5e4ca530078421c6c3536e0b5c4

I can also try to squash some of the commits moving forward.

drwells:
@AminKolah something unintended must have happened when you rebased - take a look at the list of commits on this branch.

AminKolah:
> @AminKolah something unintended must have happened when you rebased - take a look at the list of commits on this branch.

@drwells Quick question while I'm trying to resolve this...have you been also making any changes in `IBFESurfaceMethod.cpp` and the header file along the way?

drwells:
Yes: see #762, #742, #738, #733, and #654.

Do you know how to view this information through GitHub?

drwells:
Those are just the last six months of commits - let me know if you want more and I can show you how to dig them out.

boyceg:
I think that this is a branch, like NWT, that is going to get redone once the code is finalized. I am not sure if it is worth the trouble to save the detailed history.

AminKolah:
> Yes: see #762, #742, #738, #733, and #654.
> 
> Do you know how to view this information through GitHub?

Thanks. Yes I believe I do... Although I was asking particularly for most recent change inside IBFESurfaceMethod.cpp....it should be okay in any case. I'll do another rebase later today and update my branch...I'll just go with making flow past the cylinder and sphere doing the right thing and the code being compatible with changes with other places in the library.

drwells:
Dealing with branches that persist over multiple months is difficult because of all the merge conflicts and fixes that occur - even our basic bug fixes and maintenance have caused problems.

Ultimately, its much faster to merge small chunks as we go along. @AminKolah would it be possible to split out some of this functionality into some smaller PRs so that this becomes more manageable? This would save *everyone* a lot of time.

AminKolah:
> Dealing with branches that persist over multiple months is difficult because of all the merge conflicts and fixes that occur - even our basic bug fixes and maintenance have caused problems.
> 
> Ultimately, its much faster to merge small chunks as we go along. @AminKolah would it be possible to split out some of this functionality into some smaller PRs so that this becomes more manageable? This would save _everyone_ a lot of time.

I can create smaller

AminKolah:
> Dealing with branches that persist over multiple months is difficult because of all the merge conflicts and fixes that occur - even our basic bug fixes and maintenance have caused problems.
> 
> Ultimately, its much faster to merge small chunks as we go along. @AminKolah would it be possible to split out some of this functionality into some smaller PRs so that this becomes more manageable? This would save _everyone_ a lot of time.

I can create new PRs for smaller chunks. They are mostly happening in a few routines n `interpolateVelocity`, `imposeJumpConditions`

> I think that this is a branch, like NWT, that is going to get redone once the code is finalized. I am not sure if it is worth the trouble to save the detailed history.

Would you still prefer for the quantities to be calculated only on one side? In some of the examples that I've done values from both sides are needed, but I remember last time you looked at  this PR you suggested removing one of the sides.

AminKolah:
> Dealing with branches that persist over multiple months is difficult because of all the merge conflicts and fixes that occur - even our basic bug fixes and maintenance have caused problems.
> 
> Ultimately, its much faster to merge small chunks as we go along. @AminKolah would it be possible to split out some of this functionality into some smaller PRs so that this becomes more manageable? This would save _everyone_ a lot of time.

I can create new PRs for smaller chunks but then again it might sound unnecessary if things are gonna be re-done afterall as @boyceg mentioned....Most of the changes are happening in routines like `imposeJumpConditions`, `interpolateVelocity`, `computeLagrangianForce` and `computeFluidTraction`

AminKolah:
> > Dealing with branches that persist over multiple months is difficult because of all the merge conflicts and fixes that occur - even our basic bug fixes and maintenance have caused problems.
> > Ultimately, its much faster to merge small chunks as we go along. @AminKolah would it be possible to split out some of this functionality into some smaller PRs so that this becomes more manageable? This would save _everyone_ a lot of time.
> 
> I can create new PRs for smaller chunks. They are mostly happening in a few routines n `interpolateVelocity`, `imposeJumpConditions`
> 
> > I think that this is a branch, like NWT, that is going to get redone once the code is finalized. I am not sure if it is worth the trouble to save the detailed history.
> 
> Would you still prefer for the quantities to be calculated only on one side? In some of the examples that I've done values from both sides are needed, but I remember last time you looked at this PR you suggested removing one of the sides.

Just one more note, it's true that the values from side A can essentially be calculated from the jump and the value on side B, but then if that is used itself for some kind of calculation, numerically it's not exactly the same as explicitly getting the value from side A itself. There can be an option to calculate both sides to play around with which choice provides better stability depending on the problem, although the difference may not be so significant...Maybe it's not so important to fuss about, or is it

boyceg:
I would suggest having separate fields for extrapolate-from-inner, extrapolate-from-outer, and jump values. Then the user can decide which value to use.

AminKolah:
> I would suggest having separate fields for extrapolate-from-inner, extrapolate-from-outer, and jump values. Then the user can decide which value to use.

Yes that's what you suggested last time, but what if one wants to use wall shear stress from the exterior, but calculate the pressure by adding the pressure jump and the interior value of the pressure. That tends to work better for nearly touching objects.

boyceg:
I think you are making my point for me...

AminKolah:
> I think you are making my point for me...

Okay to clarify more, let's say in a physical problem we'll be dealing either with internal flow (channel flow) or external flow (flow past a cylinder) . In an external flow situation, we will be interested in calculating the exterior traction while in an internal flow problem we will be interested in calculating the interior traction/wss. However, this is separate from how we would want to numerically go about calculating either of the two. For instance, we might want to calculate the fluid traction for **external** flow using **interior** values and the other way around. ...It seems like the best thing to do is to use extrapolate-from-inner, extrapolate-from-outer for the pressure and wss calculation but then have a separate flag for the traction to be calculated from interior and or exterior, if it makes sense.

boyceg:
I am saying just compute both all of the time. Then let the user decide which one to use as appropriate for their problem. I am not sure that "inner" and "outer" are even guaranteed to be consistent for different surface meshes.

boyceg:
Don't use flags. It could be the case that it will be desirable to use one in one place and another in another place, even for a single surface mesh. Leave it to the implementation to decide how to use the values.

I would suggest keeping the viscous stress and the pressure in separate fields. So there would be a total of 6 variables, 3 scalar-valued and 3 vector-valued:

1. inner/outer pressures + pressure jump
2. inner/outer shear stress + shear stress jump

AminKolah:
> I am saying just compute both all of the time. Then let the user decide which one to use as appropriate for their problem. I am not sure that "inner" and "outer" are even guaranteed to be consistent for different surface meshes.

If I'm calculating both sides then this is already resolved. My initial understanding was that we wanted to reduce the number variables by not assigning them as interior and exterior but like a general thing that can be calculated differently based on the flag....The kind of interior/exterior that I was talking about is based on the normal on the surface mesh but I agree that depending on the problem this may not match the physical interior/exterior, like the 2D channel case.

boyceg:
I don't think there is a need to reduce the number of variables. I just think we want to make sure that the variables are clearly/consistently named.

AminKolah:
> Don't use flags. It could be the case that it will be desirable to use one in one place and another in another place, even for a single surface mesh. Leave it to the implementation to decide how to use the values.
> 
> I would suggest keeping the viscous stress and the pressure in separate fields. So there would be a total of 6 variables, 3 scalar-valued and 3 vector-valued:
> 
> 1. inner/outer pressures + pressure jump
> 2. inner/outer shear stress + shear stress jump

Okay this makes perfect sense now...I initially had them calculated on both sides but from the first round of review I got the impression that you want me to keep only one side and add a flag instead... Sorry for the confusion. I'll fix this now.

AminKolah:
> > Don't use flags. It could be the case that it will be desirable to use one in one place and another in another place, even for a single surface mesh. Leave it to the implementation to decide how to use the values.
> > I would suggest keeping the viscous stress and the pressure in separate fields. So there would be a total of 6 variables, 3 scalar-valued and 3 vector-valued:
> > 
> > 1. inner/outer pressures + pressure jump
> > 2. inner/outer shear stress + shear stress jump
> 
> Okay this makes perfect sense now...I initially had them calculated on both sides but from the first round of review I got the impression that you want me to keep only one side and add a flag instead... Sorry for the confusion. I'll fix this now.

I think it might have been because I had to define separate systems of equations for each side, but that's kind of unavoidable if there is the extra L2_projection involved.

boyceg:
@AminKolah, here is my suggestion for how to break this work out into separate PRs:

- [ ] impose jumps in the viscous stress
- [ ] interpolate the corrected velocity
- [ ] extrapolate one-sided pressures
- [ ] extrapolate one-sided viscous stresses
- [ ] update examples and tests

boyceg:
> > > Don't use flags. It could be the case that it will be desirable to use one in one place and another in another place, even for a single surface mesh. Leave it to the implementation to decide how to use the values.
> > > I would suggest keeping the viscous stress and the pressure in separate fields. So there would be a total of 6 variables, 3 scalar-valued and 3 vector-valued:
> > > 
> > > 1. inner/outer pressures + pressure jump
> > > 2. inner/outer shear stress + shear stress jump
> > 
> > 
> > Okay this makes perfect sense now...I initially had them calculated on both sides but from the first round of review I got the impression that you want me to keep only one side and add a flag instead... Sorry for the confusion. I'll fix this now.
> 
> I think it might have been because I had to define separate systems of equations for each side, but that's kind of unavoidable if there is the extra L2_projection involved.

I am not sure that is true. You could have a single pressure jump system that has vectors that correspond to the inner, outer, and jump values. (@drwells, do you have an opinion?)

AminKolah:
oops sorry I might have mistakenly closed this

AminKolah:
oops sorry I might have mistakenly closed this...

AminKolah:
They shouldn't put the "close" bottom right next to "comment" :l

AminKolah:
> > > > Don't use flags. It could be the case that it will be desirable to use one in one place and another in another place, even for a single surface mesh. Leave it to the implementation to decide how to use the values.
> > > > I would suggest keeping the viscous stress and the pressure in separate fields. So there would be a total of 6 variables, 3 scalar-valued and 3 vector-valued:
> > > > 
> > > > 1. inner/outer pressures + pressure jump
> > > > 2. inner/outer shear stress + shear stress jump
> > > 
> > > 
> > > Okay this makes perfect sense now...I initially had them calculated on both sides but from the first round of review I got the impression that you want me to keep only one side and add a flag instead... Sorry for the confusion. I'll fix this now.
> > 
> > 
> > I think it might have been because I had to define separate systems of equations for each side, but that's kind of unavoidable if there is the extra L2_projection involved.
> 
> I am not sure that is true. You could have a single pressure jump system that has vectors that correspond to the inner, outer, and jump values. (@drwells, do you have an opinion?)

That's true...I guess that's how we're doing that for velocity and the components.

drwells:
I don't know enough about the problem to really comment on whether or not its worthwhile to break it up into three different systems: my best guess is yes since those are conceptually three different things.

boyceg:
Yeah, I think that the right thing is to have different systems for interior/exterior/jump. If nothing else, it will make it automatic for the three fields to show up in the visualization files.

AminKolah:
> Yeah, I think that the right thing is to have different systems for interior/exterior/jump. If nothing else, it will make it automatic for the three fields to show up in the visualization files.

I'll resume working on this tomorrow evening...One other thing is that I c

> @AminKolah, here is my suggestion for how to break this work out into separate PRs:
> 
> * [ ]  impose jumps in the viscous stress
> * [ ]  interpolate the corrected velocity
> * [ ]  extrapolate one-sided pressures
> * [ ]  extrapolate one-sided viscous stresses
> * [ ]  update examples and tests

I would add extrapolate on-sided fluid traction as well?  That's being calculated in a separate function using the pressure, the normal, and the viscous stresses.

AminKolah:
> Yeah, I think that the right thing is to have different systems for interior/exterior/jump. If nothing else, it will make it automatic for the three fields to show up in the visualization files.

I'll resume working on this tomorrow evening and try to finish the tests by midnight.

AminKolah:
@boyceg @drwells  can you check out the changes? I added separate interior/exterior variables and fixed some of the previous warning.  It's tested with the 2D/3D flow past a cylinder/sphere (example 5 of IBFE) to be run with input2d.iim and input3d.iim respectively. I can also create separate pull request for different functionalities if it seems hard to check all changes.

boyceg:
It would be easier to review smaller patches.

AminKolah:
> It would be easier to review smaller patches.

Okay, also how are we going to add other ILE examples? Or are we adding any? Should they be in a separate folder?

boyceg:
Let’s add those later...

drwells:
There is too much to review in this patch right now; its just not a good use of anyone's time to review it until we get rid of the problems with the history and also the merge conflicts. Please split off the changes into smaller PRs.

drwells:
Superceded by #828.

PR #638, opened 2019-07-01
Use a more stable partitioner.

### LINKED ISSUE #636: Different RHS vectors in IBFEMethod::interpolateVelocity
I am getting slightly different RHS vectors in parallel when the only thing I change is the version of PETSc (in this case 3.7.7 vs. 3.11.0). The differences are pretty large: the L2 norm of the RHS vector changes in the fifth significant digit, which is way larger than it should be if the issue were purely from roundoff error accumulation.

This causes the maximum pointwise error to vary in the second significant digit, which is bad w.r.t. the test suite.

drwells:
The default partitioner for libMesh, metis, does a good job, but the way it is seeded changed in libMesh 1.4.0: i.e., we get different partitionings with different versions of libMesh. This results in (for reasons I don't fully understand) slightly different solutions to the velocity projection problem.

Get around the issue by using libMesh's linear partitioner, which is much more consistent than metis (and always available).

Fixes #636. I don't like that we can get different answers with different partitioners but this gives us the same numbers across different libMesh versions.

### DIFF
--- a/tests/IBFE/interpolate_velocity_01_2d.a.mpirun=4.output
+++ b/tests/IBFE/interpolate_velocity_01_2d.a.mpirun=4.output
@@ -4,4 +4,4 @@ IBFEMethod: mesh part 0 is using FIRST order LAGRANGE finite elements.
 
 IBHierarchyIntegrator::initializePatchHierarchy(): tag_buffer = 1 3
 max vertex distance: 0.0236627
-max norm errors: 0.00017567228356929476263   0.00026357631227225120085
+max norm errors: 0.0001756721356397372702   0.00021896759573336588289
--- a/tests/IBFE/interpolate_velocity_01.cpp
+++ b/tests/IBFE/interpolate_velocity_01.cpp
@@ -47,6 +47,7 @@
 // Headers for basic libMesh objects
 #include <libmesh/boundary_info.h>
 #include <libmesh/equation_systems.h>
+#include <libmesh/linear_partitioner.h>
 #include <libmesh/mesh.h>
 #include <libmesh/mesh_generation.h>
 #include <libmesh/mesh_triangle_interface.h>
@@ -179,6 +180,12 @@ main(int argc, char** argv)
         const int n_refinements = int(std::log2(R / dx));
         MeshTools::Generation::build_sphere(mesh, R, n_refinements, Utility::string_to_enum<ElemType>(elem_type), 10);
         mesh.prepare_for_use();
+        // metis does a good job partitioning, but the partitioning relies on
+        // random numbers: the seed changed in libMesh commit
+        // 98cede90ca8837688ee13aac5e299a3765f083da (between 1.3.1 and
+        // 1.4.0). Hence, to achieve consistent partitioning, use a simpler partitioning scheme:
+        LinearPartitioner partitioner;
+        partitioner.partition(mesh);
 
         plog << "Number of elements: " << mesh.n_active_elem() << std::endl;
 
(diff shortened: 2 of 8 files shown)


boyceg:
Hmmmm. We also need to figure out why the partitioning matters. It should not make any difference.

Can we switch to unghosted assembly for these RHS vectors to see if it makes any difference?

drwells:
Thats an interesting point: I didn't realize it before, but the tests that use an explicit accumulation strategy are not susceptible to this problem. I will investigate.

boyceg:
Should we have a test that verifies that the RHS norms are equal for different accumulation strategies? (They should be!)

drwells:
No, I am wrong: changing the partitioner fixes those tests, but only by coincidence.

drwells:
> Should we have a test that verifies that the RHS norms are equal for different accumulation strategies? (They should be!)

We do, at least, get the same output files for both accumulation strategies (in fact they are symbolic links to the same file)

boyceg:
I wonder if this is a bug in the code that determines which ghost elements are needed on a given processor and/or the mapping between elements and grid patches.

Although you say that differences also show up in serial?

drwells:
I can get different results in serial with my spack-compiled copy of libmesh 1.3.1 vs. my self-compiled versions of libMesh 1.1.0 - 1.4.1 (which are all consistent). I think thats unrelated to this issue.

boyceg:
OK, got it.

What happens if you tighten the linear solver tolerance for the L2 projection solve, or switch to a direct solver (e.g. SuperLU in parallel)? If different partitioners give different DOF numberings / assignments for the L2 projection solve, then small differences in the solution could result simply from iterative linear algebra.

drwells:
I set the tolerances to something really small already in the test:
```

    PetscOptionsSetValue(nullptr, "-ksp_rtol", "1e-16");
    PetscOptionsSetValue(nullptr, "-ksp_atol", "1e-16");
```

The issue, though, isn't small differences: the L-infinity error goes from `0.00022` to `0.00026`.

boyceg:
Are node and/or element IDs independent of the partitioner? If so, then one thing to do is to verify that each SAMRAI `Patch` gets the same libMesh node and element IDs for both partitioners.

boyceg:
If the IDs change, then we could at least check that the _number_ of nodes and elements is consistent.

drwells:
We do check that the global number of elements is consistent: we print it out. I'll check the others.

boyceg:
> If the IDs change, then we could at least check that the _number_ of nodes and elements is consistent.

(I mean --- the number of nodes / elements per SAMRAI grid patch.)

drwells:
I think that the remaining suggestions are not immediately related to the current state of the PR: we should add more tests, but since this PR fixes test output across several libMesh versions I think we should merge it as-is.

tests/IBFE/interpolate_velocity_01_2d.a.mpirun=4.output
```diff
@@ -4,4 +4,4 @@ IBFEMethod: mesh part 0 is using FIRST order LAGRANGE finite elements.
 
 IBHierarchyIntegrator::initializePatchHierarchy(): tag_buffer = 1 3
 max vertex distance: 0.0236627
-max norm errors: 0.00017567228356929476263   0.00026357631227225120085
+max norm errors: 0.0001756721356397372702   0.00021896759573336588289
```
boyceg: At least the errors go down. :-D

PR #641, opened 2019-07-05
Add and use a macro for setting rpath.

### LINKED ISSUE #639: Missing rpath arguments
It looks like, when we have external dependencies that are shared object files, we don't always set rpath correctly. This came up on the mailing list recently. I think that we don't usually run into this problem because PETSc and libMesh add rpath to their exported link flags, which we also use: for that reason I can only reproduce this problem with my shared object copy of silo.

For reasons I don't fully understand, I can compile the examples without linkage problems, but (like on the mailing list) I cannot successfully link against silo with the provided makefile.

drwells:
Fixes #639. I also verified that this works with the provided external-project makefile.

This fix still uses `AC_CHECK_LIB` to verify that the library exists and contains a known symbol. I added a new macro that will add `-Wl,-rpath,/path/to/library/` to `LDFLAGS` so that we can link against shared object libraries that are not in the standard search path.


boyceg:
How robust is this? This seems to assume that `-Wl,-rpath,` is the right flag, but it is not the right flag for all compilers. The original approach that was removed could actually figure out the right flags to use to do shared linking.

boyceg:
It looks like `AC_LIB_HAVE_LINKFLAGS` is now part of Autoconf, and so it _may_ be possible to drop `lib-ld.m4`, `lib-link.m4`, and `libtool.m4` from the `m4` subdirectory.

boyceg:
(Although I also feel like I may have hacked some of those `m4` files to try to avoid redundant/confusing `configure` options.)

drwells:
This change still uses the `gnulib` macro to figure out what the correct way to set `rpath` is, so it does support multiple compilers (although I seriously doubt anyone has compiled IBAMR with a SUN compiler in the last decade) [1].

I don't really like `AC_LIB_HAVE_LINKFLAGS` because it doesn't use the `-lname` syntax: AFAICT it just dumps a path to the library, which I find really bizarre. However, since you want it, I'll take a second look.

[1] I read through libMesh's rpath code and they set it to `-Wl,-rpath` and change it to `-Wl,-R` if we are on solaris, so we are still ahead of our dependencies in this regard.

boyceg:
Ok, that sounds fine.

drwells:
@boyceg I want to check before merging: is this solution (and the reasons for using our own rpath macro) ok? The old approach *did* work, even if its weird.

boyceg:
If it works it is OK with me!

drwells:
Okay. I wanted to check that this is indeed what your last comment meant.

PR #646, opened 2019-07-12
Fix some inconsistencies with copyrights/licenses.

drwells:
While a lot of lines changed, this was a pretty quick fix to apply.

This commit fixes some inconsistencies in how we handle copyright and software license declarations in our headers and source files. There is still more to do after this commit:
1. We need to do something similar to the build system (i.e., automake files and our own m4 macros)
2. We need to add copyright notices to the compilation stubs for the examples and old test programs.

@nishantnangia Before we merge this we need your consent for licensing all of this source code under the BSD-3 license.

I would prefer to change the way we handle copyright so that we clearly state the rights of contributors: it would be best if contributors provided their source without some statement of "I reserve copyright on this". I strongly recommend (we discussed this briefly by email) following the Linux Foundation's guidelines, which is what deal.II does:

https://github.com/dealii/dealii/blob/master/LICENSE.md

However, making this change will require all contributors to revoke their individual claims of copyright on files in IBAMR, so I recommend we make ourselves internally consistent (i.e., label everything with BSD-3) before proceeding. Getting in touch with everyone will also take some time.


drwells:
To be clear, since this is a really big change set, the main fixes are:
1. Replace NYU with UNC
2. Fix some typos in licenses
3. Add missing license declarations to source files that already contain authorship info

drwells:
I received confirmation by email that @nishantnangia is willing to license his code under BSD-3.

For the record: the context of this PR is that everything else is licensed under BSD-3, as is about half of the code provided by @nishantnangia. This PR does not really change anything, it just clarifies our licensing situation.

drwells:
I read through the changes and fixed one typo. Since @boyceg is otherwise occupied this week I will merge this myself once the testers pass (all changes are to comments and we need it for the release).

PR #648, opened 2019-07-17
Fix elem_hmax_02's link libraries.

drwells:
This was caught by @abarret. Fortunately, since the test doesn't use SAMRAI and dimensionality is a run-time parameter for libMesh, the error is harmless.


drwells:
Since this only involves the test suite I will merge as-is.

PR #649, opened 2019-07-18
Added tests for extrapolation at physical boundaries.

abarret:
This adds some tests to ensure the extrapolation at physical boundaries can interpolate the correct fields.

Currently, the quadratic operation does not interpolate a quadratic function correctly. @boyceg, is this by design?


boyceg:
> Currently, the quadratic operation does not interpolate a quadratic function correctly. @boyceg, is this by design?

I don't think so. Sounds like a bug or blunder. There are some old tests of this in IBTK. I would suggest including tests for both extrapolation and prescribed boundary conditions.

drwells:
Twelve new tests and only one executable makes me happy.

drwells:
I like that you added the tests even though they don't yet work.

Would you be willing to split off the non-quadratic parts into another PR so that we can merge them right now?

abarret:
I didn't know they didn't work until I added the test :). I also wanted to add the physical boundary tests in this PR, but if you want I can make another PR for that.

abarret:
After I do the prescribed boundary conditions, this will fully replace the `IBTK/phys_boundary_ops` test.

drwells:
Since #650 is now merged we can merge this once we fix the warnings.

abarret:
I removed the warnings, as well as fixed a warning in the test.

PR #650, opened 2019-07-18
Fix quadratic extrapolation.

abarret:
Part of #649. This fixes the quadratic extrapolation at codim 1 boundaries. There was a missing minus sign.

This does not fix the problem at other types of boundaries, i.e. corner boundaries in two spatial dimensions. Since our quadratic interpolation is dimension by dimension, we do not match _all_ quadratic functions. Currently, the code reduces to linear interpolation in these cases. I updated the header file to clarify this. I also fixed some other random spacing inconsistencies in the header.

### DIFF
--- a/ibtk/include/ibtk/CartExtrapPhysBdryOp.h
+++ b/ibtk/include/ibtk/CartExtrapPhysBdryOp.h
@@ -86,9 +86,8 @@ public:
      *
      * \param patch_data_index  Patch data index requiring ghost cell filling.
      * \param extrap_type       Type of extrapolation to perform: \p "CONSTANT" specifies
-     *constant
-     *extrapolation, \p "LINEAR" specifies linear extrapolation, and \p "QUADRATIC" specifies
-     *quadratic extrapolation.
+     * constant extrapolation, \p "LINEAR" specifies linear extrapolation, and \p "QUADRATIC" specifies
+     * quadratic extrapolation. At codim 2 or 3 boundaries, "QUADRATIC" reduces to "LINEAR" extrapolation.
      */
     CartExtrapPhysBdryOp(int patch_data_index, const std::string& extrap_type = "CONSTANT");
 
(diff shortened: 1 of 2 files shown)


drwells:
Does this fix the tests?

abarret:
Except for the corner cases, yes. I would need to update the outputs of the tests since we don't do quadratic interpolation for boundaries that are not codim 1.

boyceg:
What is the corner accuracy?

abarret:
It is linear at corners.

abarret:
We could use the same quadratic interpolation scheme at corners. I think that would be bilinear accurate, but I would need to test that. Currently corner cases just use the linear interpolation scheme.

boyceg:
Would higher order accuracy at corners help with the complex fluid code?

Also, we probably should make it easier to specify these options at runtime.

abarret:
I would be surprised if it did. The code shouldn't be using extrapolation at the problem corners. I can check it out though. My guess is that the culprit is in the prescribed boundary condition code which is next on my list.

abarret:
@boyceg, should we wait on merging this for more testing on accuracy, or go ahead and merge for 0.5?

drwells:
If we think we can fix the corner's quickly then I suggest we wait.

boyceg:
I am not sure whether quadratic extrapolation is actually used in the library currently. At present, the way that this is specified is terrible. There are strings with names like `BDRY_EXTRAP_TYPE` that are defined at the top of the implementation files that determine whether extrapolation is `CONSTANT` or `LINEAR` or `QUADRATIC`. A quick search of the code suggests that master does not use `QUADRATIC`. OTOH, user code could in principal use different extrapolation types.

I do not know why these are defined per-class, nor why they are not settable at runtime. Clearly I was being lazy.

abarret:
> I am not sure whether quadratic extrapolation is actually used in the library currently. At present, the way that this is specified is terrible. There are strings with names like `BDRY_EXTRAP_TYPE` that are defined at the top of the implementation files that determine whether extrapolation is `CONSTANT` or `LINEAR` or `QUADRATIC`. A quick search of the code suggests that master does not use `QUADRATIC`. OTOH, user code could in principal use different extrapolation types.
> 
> I do not know why these are defined per-class, nor why they are not settable at runtime. Clearly I was being lazy.

It looks like about half the classes set extrapolation values at runtime by reading in values from the input file. This is done in both the `HierarchyIntegrator` objects (except for `AdvDiffHierarchyIntegrator` for some reason) and the convective operators. But most of the solvers stick to`'LINEAR` extrapolation.

drwells:
@abarret What is the status with this branch? Can we reasonably obtain quadratic extrapolation at corners?

abarret:
I don't think it's worth the trouble. Doing a full quadratic reconstruction at the corners would require significant refactoring of the code to deal with the expanded stencil. Plus, from what I can tell, the corner ghost cells are not used in most if not all of our finite difference stencils (including the complex fluid code).

My proposal is to leave corner extrapolation as `LINEAR` and make a note about it in case someone needs to change it later.

drwells:
I agree (I don't think we support any stencils that even use such points).

abarret:
In that case, this PR is good to go as is. I will push a change to the tests branch to update the tests soon.

boyceg:
I am not certain, but I think that these corner points show up in some of the space-time nonlinear convective discretizations. These are used by the space-time advection-diffusion solver and the cell-centered incompressible Navier-Stokes solver, which is built on top of that advection-diffusion solver.

I suspect that there are not very many current users of those solvers.

PR #652, opened 2019-07-23
Suppress a new GCC warning.

drwells:
Implicitly-generated copy constructors where, in some circumstances, deprecated in C++11, but we are a long way from this being actually removed from C++.

Lets put this in 0.5 and then get ready to tag a release.


drwells:
This is pretty dull so I will merge now so that I can put up a release candidate.

PR #653, opened 2019-07-23
Collate changelog entries for 0.5.0

drwells:


drwells:
Since this is just a documentation change I will merge as-is.

Issue #647, opened 2019-07-12
Refactor FEDataManager to not expose Eulerian data


drwells:
This comes from some thoughts I have had while working with FEDataManager and IBFEMethod this week.

We use FEDataManager in two different ways: we use it to access libMesh data and also to do some FSI stuff (i.e., the loop that collects IB points). It would be easier to avoid shooting ourselves in the foot if we separated these a bit better: users should really not be calling `interpWeighted` directly. We could reimplement what we have as a base class that contains all the libMesh stuff and then an inheriting class with all the more internal stuff (and SAMRAI data) to be used by IBFEMethod.

I am not sure if this is reasonable but I want to at least write down the idea for future consideration somewhere.

drwells:
I thought about this some more in the context of #654's primary issue (dealing with postprocessors). The fundamental issue is that a postprocessor should not need to know about the scratch hierarchy that IBFEMethod needs for performance reasons: in fact, the postprocessor doesn't really need to do any IB calculations at all so it might want to deal with Eulerian data in a completely different way. We can do this correctly if we use the same set of libMesh data structures for the postprocessor and IBFEMethod but each object maintains its own "FEDataManager" object which only reads libMesh data and keeps its own set of patch-to-element maps, SAMRAIDataCache, etc.

drwells:
This changes would not be backwards compatible but they would make future work with these classes a lot simpler.

boyceg:
Sounds good to me.

drwells:
We more-or-less have this done now in a way that doesn't massively break user codes, so I think that now that we have #704 this is done.

PR #654, opened 2019-07-25
Add ib interaction hierarchy

drwells:
Super WIP (the history is presently nonsense) but I figured I would put this up to show what I have been up to for the last two weeks.

@boyceg and I discussed dealing with load imbalancing with IBFE at a previous group meeting. We came to the conclusion that the best way to do this would be to use a 'scratch hierarchy' whose patches are partitioned by the number of interaction (quadrature) points, rather than the number of (Eulerian) cells.

We can now do this! I am working on some scaling experiments now, but the tentative results look great: all processors have, within about 0.1%, the same number of quadrature points on the scratch hierarchy and (I didn't count them explicitly) an approximately equal number of Eulerian cells on the real hierarchy.

There are a number of problems with the current implementation:
- [x] The relationship between FEDataManager and IBFEMethod is *really* complicated. This patch makes this problem substantially worse by adding a ton of checks all over the place to see if the scratch hierarchy is set up. The main issue is that `IBFEMethod::InitializeLevelData` (inherited from `SAMRAI::StandardTagAndInitialize::initializeLevelData`) should set up all data on a given level but doesn't: we violate the Liskov substitution principle and its possible to get `FEDataManager` in an invalid state pretty easily this way. See also #647.
- [x] Related: `IBFEPostProcessor` also knows too much about the internals of `FEDataManager`. This branch completely breaks this class: I have to comment out all postprocessor usage to get the examples (and the heart model) to work.
- [x] We should refactor the way we store communication schedules. The current implementation works but it involves a lot of copying and pasting.
- [x] I think we might be able to delete `CopyAsRefineOperator` and instead provide a `nullptr`: I think SAMRAI will just avoid prolongation/restriction if we do that but I didn't try. (update: SAMRAI can copy data perfectly well on its own so supplying a `nullptr` suffices).
- [x] This isn't entirely our fault, but there is a lot of sketchy memory allocation and deallocation going on with the tag index.
- [x] I left a ton of `#if 1` and `#if 0` statements in the code; those need to go.
- [x] I hardcoded IBFEMethod's load balancer settings. Those should be set up in the same way I set up the IBFEMethod-specific GriddingAlgorithm. update: we can now configure these load balancing settings in the input database.
- [x] Similarly, initializing a `StandardTagAndInitialize` without a database creates a warning. We should set up a fake database for it.
- [x] A more serious problem is that, for reasons I don't understand, every time we regrid the volume occupied by the finest level grows in all directions by some constant amount (until the entire grid is covered with fine level patches). This doesn't slow down the computation, but since all of these patches have workload zero they usually all end up on the same processor which is really bad for memory usage. *Update*: I partially fixed this by completely rebuilding the scratch hierarchy at each time step.

### DIFF
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -684,6 +684,11 @@ public:
      * Set integer tags to "one" in cells where refinement of the given level
      * should occur according to user-supplied feature detection criteria.
      *
+     * The name here is misleading, but SAMRAI expects us to use one of two
+     * tagging methods to refine the grid, and IBAMR consistently uses
+     * gradient detection: hence this function has the same name but tags
+     * cells in a different way.
+     *
      * \see SAMRAI::mesh::StandardTagAndInitStrategy::applyGradientDetector
      */
     void applyGradientDetector(SAMRAI::tbox::Pointer<SAMRAI::hier::BasePatchHierarchy<NDIM> > hierarchy,
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -1207,6 +1207,13 @@ private:
      * Update the caches of IB-ghosted vectors.
      */
     void updateCachedIBGhostedVectors();
+
+    /*!
+     * At the present time this class and FEDataManager assume that the finite
+     * element mesh is always on the finest grid level. This function
+     * explicitly asserts that this condition is met.
+     */
+    void assertStructureOnFinestLevel() const;
 };
 } // namespace IBAMR
 
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -658,8 +658,16 @@ public:
                                SAMRAI::tbox::Pointer<SAMRAI::mesh::GriddingAlgorithm<NDIM> > gridding_alg) override;
 
     /*!
-     * Initialize data on a new level after it is inserted into an AMR patch
-     * hierarchy by the gridding algorithm.
+     * this function only exists for compatibility with the base class and
+     * does nothing: data reinitialization is handled by
+     * endDataRedistribution() instead.
+     *
+     * The reasoning is this: since this class stores data only on particular
+     * levels (at the present time, the structure is always on the finest
+     * level) setting up level data is nontrivial when generating the initial
+     * grid (i.e., when tagging cells that contain interaction points for
+     * refinement). In a sense there is no level data to compute until we are
+     * done regridding.
      *
      * \see SAMRAI::mesh::StandardTagAndInitStrategy::initializeLevelData
      */
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -791,7 +791,7 @@ protected:
      * PatchHierarchy object, and so different scratch data caching objects are
      * needed for the regular and scratch patch hierarchies.
      */
-    std::shared_ptr<IBTK::SAMRAIDataCache> d_eulerian_data_cache, d_scratch_eulerian_data_cache;
+    std::shared_ptr<IBTK::SAMRAIDataCache> d_primary_eulerian_data_cache, d_scratch_eulerian_data_cache;
 
     /*
      * The current time step interval.
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -121,7 +121,7 @@ namespace IBAMR
  * from <code>GHOSTED</code>, the default, to <code>CACHE</code>, which will
  * use PETSc's VecCache object to distribute data.
  *
- * <h2>Options Controlling Partitioning</h2>
+ * <h2>Options Controlling libMesh Partitioning</h2>
  *
  * This class can repartition libMesh data in a way that matches SAMRAI's
  * distribution of patches; put another way, if a certain region of space on
@@ -160,6 +160,68 @@ namespace IBAMR
  * <code>AUTOMATIC</code>. The intent of these choices is to automatically use
  * the fairest (that is, partitioning based on workload estimation)
  * partitioner.
+ *
+ * <h2>Options Controlling IB Data Partitioning</h2>
+ *
+ * The main computational expenses of this class are
+ * IBFEMethod::interpolateVelocity() and IBFEMethod::spreadForce(). These two
+ * methods compute at IB points placed inside the patches owned on the current
+ * processor: i.e., they use the Eulerian partitioning of the domain. This
+ * partitioning scales very poorly at higher processor counts with some
+ * Lagrangian geometries since the Eulerian partitioning places equal number
+ * of cells, which do not necessarily coincide with IB points, on different
+ * processors: i.e., some processors will have a large number of IB points and
+ * some may have zero.
+ *
+ * To get around this, this class can optionally work with a different
+ * partitioning of the Eulerian data that is partitioned so that each
+ * processor has roughly the same number of IB points, or some more elaborate
+ * partitioning scheme that takes into account the number of mesh nodes as
+ * well. This class will set up this scratch hierarchy and manage its state
+ * (see IBFEMethod::d_scratch_hierarchy). The scratch hierarchy can be set up
+ * by adding the following parameters to the input database:
+ *
+ * @code
+ * use_scratch_hierarchy = TRUE
+ * workload_quad_point_weight = 1.0
+ *
+ * GriddingAlgorithm
+ * {
+ *     max_levels = MAX_LEVELS
+ *     ratio_to_coarser
+ *     {
+ *         level_1 = REF_RATIO,REF_RATIO
+ *         level_2 = REF_RATIO,REF_RATIO
+ *     }
+ *
+ *     largest_patch_size
+ *     {
+ *         level_0 = 512,512
+ *     }
+ *
+ *     smallest_patch_size
+ *     {
+ *         level_0 = 8,8
+ *     }
+ *
+ *     efficiency_tolerance = 0.80e0
+ *     combine_efficiency   = 0.80e0
+ * }
+ *
+ * LoadBalancer
+ * {
+ *    bin_pack_method     = "SPATIAL"
+ *    max_workload_factor = 1
+ * }
+ * @endcode
+ *
+ * i.e., providing <code>use_scratch_hierarchy = TRUE</code> (the default is
+ * <code>FALSE</code>) turns on the scratch hierarchy and the remaining
+ * parameters determine how patches are generated and load balanced. The
+ * parameter <code>workload_quad_point_weight</code> is the multiplier
+ * assigned to an IB point when calculating the work per processor: in the
+ * future additional weights, such as <code>workload_node_point_weight</code>
+ * will also be added.
  */
 class IBFEMethod : public IBStrategy
 {
@@ -716,6 +778,12 @@ public:
      */
     void writeFEDataToRestartFile(const std::string& restart_dump_dirname, unsigned int time_step_number);
 
+    /*!
+     * Return a pointer to the scratch hierarchy used by this object. See the
+     * main documentation of this class for more information.
+     */
+    SAMRAI::tbox::Pointer<SAMRAI::hier::BasePatchHierarchy<NDIM> > getScratchHierarchy();
+
 protected:
     /*!
      * \brief Compute the stress normalization field Phi.
@@ -776,6 +844,32 @@ protected:
      */
     void initializeVelocity(unsigned int part);
 
+    /*!
+     * Get the transfer schedule from the primary hierarchy to the scratch
+     * hierarchy associated with the given level and index. If necessary the
+     * schedule is created and stored in a map.
+     *
+     * If needed, a SAMRAI::xfer::RefinePatchStrategy object can be provided
+     * for filling ghost data at physical boundaries.
+     */
+    SAMRAI::xfer::RefineSchedule<NDIM>&
+    getPrimaryToScratchSchedule(const int level_number,
+                                const int data_idx,
+                                SAMRAI::xfer::RefinePatchStrategy<NDIM>* patch_strategy = nullptr);
+
+    /*!
+     * Get the transfer schedule from the scratch hierarchy to the primary
+     * hierarchy associated with the given level and index. If necessary the
+     * schedule is created and stored in a map.
+     *
+     * If needed, a SAMRAI::xfer::RefinePatchStrategy object can be provided
+     * for filling ghost data at physical boundaries.
+     */
+    SAMRAI::xfer::RefineSchedule<NDIM>&
+    getScratchToPrimarySchedule(const int level_number,
+                                const int data_idx,
+                                SAMRAI::xfer::RefinePatchStrategy<NDIM>* patch_strategy = nullptr);
+
     /*
      * Indicates whether the integrator should output logging messages.
      */
@@ -805,6 +899,43 @@ protected:
      * needed for the regular and scratch patch hierarchies.
      */
     std::shared_ptr<IBTK::SAMRAIDataCache> d_primary_eulerian_data_cache, d_scratch_eulerian_data_cache;
+
+    /*
+     * Pointer to the scratch patch hierarchy (which is only used for the
+     * evaluation of IB terms, i.e., in IBFEMethod::interpolateVelocity(),
+     * IBFEMethod::spreadForce(), and IBFEMethod::spreadFluidSource()).
+     */
+    SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > d_scratch_hierarchy;
+
+    int d_lagrangian_workload_current_idx = IBTK::invalid_index;
+    int d_lagrangian_workload_new_idx = IBTK::invalid_index;
+    int d_lagrangian_workload_scratch_idx = IBTK::invalid_index;
+
+    SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > d_lagrangian_workload_var;
+    const std::string d_lagrangian_workload_coarsen_type = "CONSERVATIVE_COARSEN";
+    const std::string d_lagrangian_workload_refine_type = "CONSERVATIVE_LINEAR_REFINE";
+
+    /*
+     * Refinement schedules for transferring data from d_hierarchy to
+     * d_scratch_hierarchy. The keys are the level number and data index (in
+     * that order).
+     *
+     * @note this function assumes that only data on the finest level needs to
+     * be transferred.
+     */
+    std::map<std::pair<int, int>, SAMRAI::tbox::Pointer<SAMRAI::xfer::RefineSchedule<NDIM> > >
+        d_scratch_transfer_forward_schedules;
+
+    /*
+     * Refinement schedules for transferring data from d_scratch_hierarchy to
+     * d_hierarchy. The keys are the level number and data index (in
+     * that order).
+     *
+     * @note this function assumes that only data on the finest level needs to
+     * be transferred.
+     */
+    std::map<std::pair<int, int>, SAMRAI::tbox::Pointer<SAMRAI::xfer::RefineSchedule<NDIM> > >
+        d_scratch_transfer_backward_schedules;
 
     /*
      * The current time step interval.
@@ -1002,6 +1133,16 @@ protected:
      * Restart file type for libMesh equation systems (e.g. xda or xdr).
      */
     std::string d_libmesh_restart_file_extension;
+
+    /**
+     * database for the GriddingAlgorithm used with the scratch hierarchy.
+     */
+    SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> d_gridding_algorithm_db;
+
+    /**
+     * database for the LoadBalancer used with the scratch hierarchy.
+     */
+    SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> d_load_balancer_db;
 
 private:
     /*!
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1607,8 +1607,6 @@ IBFEMethod::initializeLevelData(Pointer<BasePatchHierarchy<NDIM> > hierarchy,
     {
         d_primary_fe_data_managers[part]->setPatchHierarchy(hierarchy);
         d_primary_fe_data_managers[part]->setPatchLevels(0, finest_hier_level);
-        d_primary_fe_data_managers[part]->initializeLevelData(
-            hierarchy, level_number, init_data_time, can_be_refined, initial_time, old_level, allocate_data);
     }
     return;
 } // initializeLevelData
@@ -1623,7 +1621,6 @@ IBFEMethod::resetHierarchyConfiguration(Pointer<BasePatchHierarchy<NDIM> > hiera
     {
         d_primary_fe_data_managers[part]->setPatchHierarchy(hierarchy);
         d_primary_fe_data_managers[part]->setPatchLevels(0, hierarchy->getFinestLevelNumber());
-        d_primary_fe_data_managers[part]->resetHierarchyConfiguration(hierarchy, coarsest_level, finest_hier_level);
     }
     return;
 } // resetHierarchyConfiguration
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -51,6 +51,7 @@
 
 #include "BasePatchHierarchy.h"
 #include "BasePatchLevel.h"
+#include "BergerRigoutsos.h"
 #include "Box.h"
 #include "CartesianPatchGeometry.h"
 #include "CellIndex.h"
@@ -64,8 +65,10 @@
 #include "Patch.h"
 #include "PatchHierarchy.h"
 #include "PatchLevel.h"
+#include "RefineOperator.h"
 #include "SideData.h"
 #include "SideIndex.h"
+#include "StandardTagAndInitialize.h"
 #include "Variable.h"
 #include "VariableDatabase.h"
 #include "tbox/Array.h"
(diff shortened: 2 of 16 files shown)


boyceg:
> * A more serious problem is that, for reasons I don't understand, every time we regrid the volume occupied by the finest level grows in all directions by some constant amount (until the entire grid is covered with fine level patches). This doesn't slow down the computation, but since all of these patches have workload zero they usually all end up on the same processor which is really bad for memory usage.

I feel like I have encountered this before. You might need to explicitly clear the tags before retagging. Otherwise, I think that the tag buffer will cause this kind of behavior.

drwells:
I tried that but it didn't seem to help. Let me recompile everything again to triple check.

knepley:
How are you currently dealing with communication schedules?

drwells:
I store the schedules in maps inside IBFEMethod:
```cpp
    /*
     * Refinement schedules for transferring data from d_hierarchy to
     * d_scratch_hierarchy.
     */
    std::map<int, SAMRAI::tbox::Pointer<SAMRAI::xfer::RefineSchedule<NDIM> > > d_scratch_transfer_forward_schedules;

    /*
     * Refinement schedules for transferring data from d_scratch_hierarchy to
     * d_hierarchy.
     */
    std::map<int, SAMRAI::tbox::Pointer<SAMRAI::xfer::RefineSchedule<NDIM> > > d_scratch_transfer_backward_schedules;
```
The key is a data index. I clear the maps when I regrid and then recompute the schedules when they are first needed. Some data transfers, like velocity interpolation, only go one way: others, like force spreading, go both ways.

knepley:
Oh, are these just local transfers between grids?

drwells:
I'm not sure that I understand the question: I think the answer is 'no' since they involve communication.

Both the normal hierarchy and the scratch hierarchy completely cover the structure with patches on the finest level. These schedules describe transferring data between the two different partitioning of the (approximately same) sets of patches.

knepley:
So SAMRAI makes MPI structures from the map above? I was wondering what kind of persistent MPI structs were being used.

drwells:
No, I just put SAMRAI's stuff in the maps. From IBFEMethod::interpolateVelocity:

```cpp
    if (d_scratch_transfer_forward_schedules.count(u_data_idx) == 0)
    {
        const int ln = d_hierarchy->getFinestLevelNumber();
        Pointer<PatchLevel<NDIM> > level = d_hierarchy->getPatchLevel(ln);
        Pointer<PatchLevel<NDIM> > scratch_level = d_scratch_hierarchy->getPatchLevel(ln);
        scratch_level->allocatePatchData(u_data_idx, 0.0);
        Pointer<RefineAlgorithm<NDIM> > refine_algorithm = new RefineAlgorithm<NDIM>();
        Pointer<RefineOperator<NDIM> > refine_op_u = new CopyAsRefineOperator<NDIM>();
        refine_algorithm->registerRefine(u_data_idx, u_data_idx, u_data_idx, refine_op_u);
        d_scratch_transfer_forward_schedules[u_data_idx] =
            refine_algorithm->createSchedule("DEFAULT_FILL", scratch_level, level, nullptr, false, nullptr);
    }
    d_scratch_transfer_forward_schedules[u_data_idx]->fillData(0.0);
```

boyceg:
> So SAMRAI makes MPI structures from the map above? I was wondering what kind of persistent MPI structs were being used.

I can point you to the SAMRAI data transfer code if you want to check it out.

drwells:
I have some nice plots now for demonstrating where we spend our time in the heart model (FYI @cpuelz). I plotted the maximum time across all processors for each section, including the barrier after velocity interpolation (but before the linear solve). This is why the barrier and velocity interpolation times match in the first case.

without this patch set:
non-lb

and with this patch:
lb

knepley:
That is great data @drwells . Can you get the volume of communication in the velocity interpolation? That would allow us to see if it is latency or bandwidth dominated. I would be really surprised if its latency since there is only one round of communication, and we expect good load balance.

drwells:
Full disclosure: I did not put timers around the code that actually transfers data. These timers are for computing, e.g., the load vector. I will add that: I really have no idea how much time we spend communicating with this approach, nevermind latency.

I found some example MPI latency measuring code online but I'm not sure if its what you had in mind. Is there a relevant PETSc tutorial or code snippet that you know of which shows where to start?

knepley:
You are right that timing latency is hard. I am proposing that we do it in an indirect way at first. We just time the communication operation, and also record how many bytes are sent. PETSc does this automatically by wrapping the comm operations in the library, and IBAMR could do a similar thing. We can easily measure comm bandwidth by sending some large messages. Once we have bandwidth, we can look at the time and bytes transferred to see how close we are to the bandwidth limit, which should tell us how much latency we have.

drwells:
It took a large chunk of the last two weeks but I have now resolved the problems with the postprocessors: i.e., this PR is finally backwards-compatible.

I still need to do some more clean up and add some more tests but this is getting close!

drwells:
Its a good thing I added the new test: for reasons I haven't been able to figure out yet using the scratch hierarchy for velocity interpolation produces wrong results when there are lag body forces (i.e., things attached via something like `ib_method_ops->registerLagBodyForceFunction(block_tether_force_data, 0)`). Otherwise everything seems to be in order.

I've been stuck on this for most of today. The problem occurs when we use a single processor so it cannot be due to a mismatch somewhere in parallel partitionings. @boyceg any idea why this might happen?

boyceg:
> Its a good thing I added the new test: for reasons I haven't been able to figure out yet using the scratch hierarchy for velocity interpolation produces wrong results when there are lag body forces (i.e., things attached via something like `ib_method_ops->registerLagBodyForceFunction(block_tether_force_data, 0)`). Otherwise everything seems to be in order.

So including Lagrangian forces messes up velocity interpolation --- is that right?

drwells:
Yup. I have no idea why - my best guess is that something gets cached or modified in some non-obvious way.

drwells:
OK, I think I figured it out: its all a weird coincidence (like such bugs commonly are). Turning that force on caused the position of the structure to actually impact the flow. The position of the structure was wrong because the computed (projected onto the structure) velocity was wrong.

It looks like the current copy strategy doesn't correctly update boundary values: if I shrink the structures in this example to be far away from the boundary then the computed values are correct. Here is the modification to example 8:
```cpp
MeshTools::Generation::build_square(block1_mesh,
                                    ceil(0.5 / ds_block),
                                    ceil(0.5 / ds_block),
                                    0.0 + 2*dx,
                                    0.5,
                                    0.0 + 2*dx,
                                    0.5,
                                    Utility::string_to_enum<ElemType>(block_elem_type));
Mesh block2_mesh(init.comm(), NDIM);
MeshTools::Generation::build_square(block2_mesh,
                                    ceil(0.5 / ds_block),
                                    ceil(0.5 / ds_block),
                                    1.5,
                                    2.0 - 2*dx,
                                    0.0 + 2*dx,
                                    0.5,
                                    Utility::string_to_enum<ElemType>(block_elem_type));
```

There should be a way to tell SAMRAI that yes, I do really want to copy any extrapolated or ghost or whatever values. Let me take a look.

boyceg:
> There should be a way to tell SAMRAI that yes, I do really want to copy any extrapolated or ghost or whatever values. Let me take a look.

If you are using a `RefineAlgorithm`/`RefineSchedule` to transfer the data, then you will also need to set up the ghost filling algorithms.

For now I would suggest doing this in two phases: first transfer data from one hierarchy to the other, and then fill ghost values on the new hierarchy. This should ultimately use the same routines to fill data as are currently in use.

Once this works, it should be possible to merge those two operations (to avoid performing unnecessary copys).

boyceg:
AFAIK, there is no way to get `RefineAlgorithm`/`RefineSchedule` to transfer ghost data from the source to the destination, but maybe there is. For this to be reliable, however, the levels that comprise the two hierarchies would need to cover the same regions of physical space. It seems better to refill the ghost values without worrying about this.

drwells:
I think that the current implementation (according to the documentation of `RefineSchedule`) will fill ghost cells: I guess that this does not include boundary values. It looks like I need to provide a `RefinePatchStrategy` at some point in the pipeline to correctly set boundary values.

boyceg:
Fill ghost values --- not copy them.

boyceg:
(At the physical boundaries, that is. At interior patch boundaries values are copied.)

drwells:
@boyceg quick question before I head down a rabbit hole (if you happen to know the answer): refinement operations need extra objects to specify how boundary values are filled. Do coarsening operations copy boundary values by default?

drwells:
Hm. I'll try switching all of my refines to coarsens; perhaps that will automatically fix the ghost data.

boyceg:
Coarsening will not fill ghost values...

drwells:
You're right. I misread the coarsen documentation.

I think the way to solve this is to somehow extract the refinement strategy objects and provide them to the scratch hierarchy refinement algorithms.

drwells:
It worked!

Getting the `RefinePatchStrategy` out in this way involves some weird changes to `HierarchyIntegrator`: we have to make more stuff public.

boyceg:
It might be worth thinking about packaging up the BC specification stuff to make it easier to pass around a full specification. The actual BC values are determined both by BC coefficients and also the extrapolation routines that actually fill the ghost values -- and those routines use different extrapolation orders.

Boundary conditions are the worst.

drwells:
I agree.

This is a part of SAMRAI that I don't know that well: does the `RefinePatchStrategy` know how to fill boundary values correctly in addition to filling ghost values that are outside of the domain?

> Boundary conditions are the worst.

Yes, and managing them in parallel with AMR and finite differences is the worst combination that I can think of, at least until we get a fourth order MAC code working :)

boyceg:
> This is a part of SAMRAI that I don't know that well: does the `RefinePatchStrategy` know how to fill boundary values correctly in addition to filling ghost values that are outside of the domain?

The`RefinePatchStrategy` interface just allows for user-specified operations that happen after the ghost values are filled. None of the standard SAMRAI refine objects will touch physical boundary ghost data. `RefinePatchStrategy` provides a way to say what happens at physical boundaries. But you also could use it to do other things (like filtering interpolated data).

IBAMR has specialized `RefinePatchStrategy` classes (derived from `RobinPhysBdryPatchStrategy`) that are specifically for setting physical BC values. We (errr, I) have systematically abused the notion of "Robin boundary conditions" and probably should spend some time rationalizing the design of these classes.

drwells:
That should be all we need to do.

Lets finish #733 before we merge this since both of them add code for the scratch partitioning.

include/ibamr/IBFEMethod.h
```diff
@@ -683,6 +754,11 @@ class IBFEMethod : public IBStrategy
      * Set integer tags to "one" in cells where refinement of the given level
      * should occur according to user-supplied feature detection criteria.
      *
+     * The name here is misleading, but SAMRAI expects us to use one of three
```
boyceg:
```suggestion
     * The name here is misleading, but SAMRAI expects us to use one of two
```

include/ibamr/IBFEMethod.h
```diff
@@ -702,6 +778,12 @@ class IBFEMethod : public IBStrategy
      */
     void writeFEDataToRestartFile(const std::string& restart_dump_dirname, unsigned int time_step_number);
 
+    /*!
+     * Return a pointer to the scratch hierarchy used by this object. See the
+     * main documentation of this class for more information.
+     */
+    SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > getScratchHierarchy();
```
boyceg: `hier::BasePatchHierarchy`?

include/ibamr/IBFEMethod.h
```diff
@@ -762,6 +844,32 @@ class IBFEMethod : public IBStrategy
...
+     * scratch hierarchy) associated with the given level and index. If
+     * necessary the schedule is created and stored in a map.
+     *
+     * If needed, a SAMRAI::xfer::RefinePatchStrategy object can be provided
+     * for filling ghost data at physical boundaries.
+     */
+    SAMRAI::xfer::RefineSchedule<NDIM>&
+    getForwardScratchSchedule(const int level_number,
+                              const int data_idx,
+                              SAMRAI::xfer::RefinePatchStrategy<NDIM>* rps = nullptr);
```
boyceg:
```suggestion
                              SAMRAI::xfer::RefinePatchStrategy<NDIM>* patch_strategy = nullptr);
```
or `refine_patch_strategy`

include/ibamr/IBFEMethod.h
```diff
@@ -762,6 +844,32 @@ class IBFEMethod : public IBStrategy
...
+     * the main hierarchy) associated with the given level and index. If
+     * necessary the schedule is created and stored in a map.
+     *
+     * If needed, a SAMRAI::xfer::RefinePatchStrategy object can be provided
+     * for filling ghost data at physical boundaries.
+     */
+    SAMRAI::xfer::RefineSchedule<NDIM>&
+    getBackwardScratchSchedule(const int level_number,
+                               const int data_idx,
+                               SAMRAI::xfer::RefinePatchStrategy<NDIM>* rps = nullptr);
```
boyceg:
```suggestion
                               SAMRAI::xfer::RefinePatchStrategy<NDIM>* patch_strategy = nullptr);
```
or `refine_patch_strategy`

include/ibamr/IBFEMethod.h
boyceg: Instead of "forward" and "backward", can we try to make these function names include something like "to scratch hierarchy" and "from scratch hierarchy"?

Another alternative would be to have a single function that has a mode that could be either `FORWARD` or `REVERSE` (similar to PETSc data transfer functions).

include/ibamr/IBFEMethod.h
```diff
@@ -781,6 +889,46 @@ class IBFEMethod : public IBStrategy
...
+     */
+    SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > d_scratch_hierarchy;
+
+    int d_lagrangian_workload_current_idx = IBTK::invalid_index;
+    int d_lagrangian_workload_new_idx = IBTK::invalid_index;
+    int d_lagrangian_workload_scratch_idx = IBTK::invalid_index;
+
+    // TODO this tag index is static and known to the gridding algorithm. Is
+    // there a better way to look it up?
+    int d_tag_index = IBTK::invalid_index;
```
boyceg: Do we need to cache this?

src/IB/IBFEMethod.cpp
```diff
@@ -319,6 +322,58 @@ libmesh_restart_file_name(const std::string& restart_dump_dirname,
                      << std::setfill('0') << std::right << time_step_number << "." << extension;
     return file_name_prefix.str();
 }
+
+template <int DIM>
+class CopyAsRefineOperator : public xfer::RefineOperator<DIM>
```
boyceg: What is this meant to do?

src/IB/IBFEMethod.cpp
```diff
@@ -319,6 +322,58 @@ libmesh_restart_file_name(const std::string& restart_dump_dirname,
...
+        // TBOX_ASSERT(fine->getDepth() == fine_data->getDepth()); // TODO implement depth check
+
+        // TODO: implement non-cell overlaps
+        tbox::Array<Box<DIM> > boxes(2);
+        boxes[0] = fine_box;
+        boxes[1] = coarse.getBox();
+        BoxList<DIM> boxes_list(boxes);
+        CellOverlap<DIM> cell_overlap(boxes_list, IntVector<DIM>(0));
+
+        fine_data->copy(*coarse_data, cell_overlap);
```
boyceg: This should not work unless the levels are already in the same index space --- and in that case SAMRAI should not need to use the refine operator --- it should just copy.

src/IB/IBFEMethod.cpp
```diff
@@ -744,6 +799,21 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
...
 {
+    // Communicate ghost data.
+    for (const auto& u_ghost_fill_sched : u_ghost_fill_scheds)
+    {
+        if (u_ghost_fill_sched) u_ghost_fill_sched->fillData(data_time);
+    }
+
+    if (d_use_scratch_hierarchy)
+    {
+        // TODO: this assumes the structure is always on the finest level
```
boyceg: Can we add an assertion...?

src/IB/IBFEMethod.cpp
```diff
@@ -787,6 +852,10 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
                                                         /*close_F*/ false,
                                                         /*close_X*/ false);
     }
+
+    // note that FEDataManager only reads (no writes) Eulerian info so no data
```
boyceg: Do we spread to the scratch hierarchy?

src/IB/IBFEMethod.cpp
```diff
@@ -929,6 +998,13 @@ IBFEMethod::spreadForce(const int f_data_idx,
     batch_vec_copy({ d_X_half_vecs, d_F_half_vecs }, { d_X_IB_ghost_vecs, d_F_IB_ghost_vecs });
     batch_vec_ghost_update({ d_X_IB_ghost_vecs, d_F_IB_ghost_vecs }, INSERT_VALUES, SCATTER_FORWARD);
 
+    if (d_use_scratch_hierarchy)
+    {
+        // TODO: this assumes the structure is always on the finest level
+        // TODO: we don't need a RefinePatchStrategy here, right?
```
boyceg: You do if you want physical boundary values filled.

src/IB/IBFEMethod.cpp
```diff
@@ -929,6 +998,13 @@ IBFEMethod::spreadForce(const int f_data_idx,
     batch_vec_copy({ d_X_half_vecs, d_F_half_vecs }, { d_X_IB_ghost_vecs, d_F_IB_ghost_vecs });
     batch_vec_ghost_update({ d_X_IB_ghost_vecs, d_F_IB_ghost_vecs }, INSERT_VALUES, SCATTER_FORWARD);
 
+    if (d_use_scratch_hierarchy)
+    {
+        // TODO: this assumes the structure is always on the finest level
```
boyceg: Can we add an assertion...?

src/IB/IBFEMethod.cpp
```diff
@@ -944,6 +1020,13 @@ IBFEMethod::spreadForce(const int f_data_idx,
                                                 /*close_X*/ false);
     }
 
+    if (d_use_scratch_hierarchy)
+    {
+        // TODO: this assumes the structure is always on the finest level
```
boyceg: Can we add an assertion...?

src/IB/IBFEMethod.cpp
```diff
@@ -944,6 +1020,13 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+        // TODO: we don't need a RefinePatchStrategy here, right?
```
boyceg: If it is just a straightforward copy, then no.

This could be a good place to do the accumulation from the ghost region into interior, however, and could use a patch strategy to accomplish that.

src/IB/IBFEMethod.cpp
```diff
@@ -1076,6 +1159,13 @@ IBFEMethod::spreadFluidSource(const int q_data_idx,
     TBOX_ASSERT(MathUtilities<double>::equalEps(data_time, d_half_time));
     batch_vec_copy({ d_X_half_vecs, d_Q_half_vecs }, { d_X_IB_ghost_vecs, d_Q_IB_ghost_vecs });
     batch_vec_ghost_update({ d_X_IB_ghost_vecs, d_Q_IB_ghost_vecs }, INSERT_VALUES, SCATTER_FORWARD);
+
+    if (d_use_scratch_hierarchy)
+    {
+        // TODO: this assumes the structure is always on the finest level
```
boyceg: Can we add an assertion...?

src/IB/IBFEMethod.cpp
```diff
@@ -1088,6 +1178,13 @@ IBFEMethod::spreadFluidSource(const int q_data_idx,
                                                 /*close_Q*/ false,
                                                 /*close_X*/ false);
     }
+
+    if (d_use_scratch_hierarchy)
+    {
+        // TODO: this assumes the structure is always on the finest level
```
boyceg: Can we add an assertion...?

src/IB/IBFEMethod.cpp
```diff
@@ -1454,11 +1560,25 @@ IBFEMethod::initializePatchHierarchy(Pointer<PatchHierarchy<NDIM> > hierarchy,
...
+
     // Initialize the FE data managers.
     for (unsigned int part = 0; part < d_num_parts; ++part)
     {
         d_primary_fe_data_managers[part]->reinitElementMappings();
-        if (d_use_scratch_hierarchy) d_scratch_fe_data_managers[part]->reinitElementMappings();
+        if (d_use_scratch_hierarchy)
+        {
+            d_scratch_fe_data_managers[part]->setPatchHierarchy(d_scratch_hierarchy);
+            // TODO: this assumes the structure lives on the finest grid level.
```
boyceg: Can we add an assertion...?

src/IB/IBFEMethod.cpp
```diff
@@ -1573,48 +1791,77 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
...
 
+        if (d_use_scratch_hierarchy)
+        {
+            // FEDataManager needs
+            // 1. velocity
+            // 2. force
+            // 3. workload or some quadrature point per cell count
+            // 4. tagging
+            //
+            // patches in the scratch hierarchy. However, the data index is not
```
boyceg:
```suggestion
            // patch data in the scratch hierarchy. However, the data index is not
```

src/IB/IBFEMethod.cpp
```diff
@@ -1573,48 +1791,77 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
...
+            // known until the call to interpolateVelocity or spreadForce, so
+            // clear existing data but do not allocate yet.
+            d_scratch_transfer_forward_schedules.clear();
+            d_scratch_transfer_backward_schedules.clear();
+
+            for (unsigned int part = 0; part < d_num_parts; ++part)
+            {
+                d_scratch_fe_data_managers[part]->setPatchHierarchy(d_scratch_hierarchy);
+                // TODO: this assumes the structure is always on the finest level
```
boyceg: Can we add an assertion...?

src/IB/IBFEMethod.cpp
```diff
@@ -1627,13 +1874,32 @@ IBFEMethod::applyGradientDetector(Pointer<BasePatchHierarchy<NDIM> > base_hierar
                                   bool initial_time,
                                   bool uses_richardson_extrapolation_too)
 {
+    // TODO: this technically works since SAMRAI only lets us ever use one tag index (its static)
```
boyceg: Do we really need to cache this value?

src/IB/IBFEMethod.cpp
```diff
@@ -1627,13 +1874,32 @@ IBFEMethod::applyGradientDetector(Pointer<BasePatchHierarchy<NDIM> > base_hierar
...
+    d_tag_index = tag_index;
     Pointer<PatchHierarchy<NDIM> > hierarchy = base_hierarchy;
     TBOX_ASSERT(hierarchy);
     TBOX_ASSERT((level_number >= 0) && (level_number <= hierarchy->getFinestLevelNumber()));
     TBOX_ASSERT(hierarchy->getPatchLevel(level_number));
+
+    // TODO SAMRAI should allocate this, I think
```
boyceg: Yes.

include/ibamr/IBFEMethod.h
drwells: I went with `getPrimaryToScratchSchedule` and `getScratchToPrimarySchedule`.

include/ibamr/IBFEMethod.h
drwells: Nice catch: no, we never read this.

include/ibamr/IBFEMethod.h
drwells: We do this in a few places and I don't understand why. What is the advantage of returning a pointer to the base class when we already know it is a derived class?

src/IB/IBFEMethod.cpp
drwells: I should clarify this comment: I meant `FEDataManager only reads (no writes) Eulerian data during interpolation`.

include/ibamr/IBFEMethod.h
boyceg: For that day in the distant future when we turn on full support for multiblock domains?

src/IB/IBFEMethod.cpp
drwells: No.

include/ibamr/IBFEMethod.h
drwells: Ah. I will make the change.

src/IB/IBFEMethod.cpp
drwells: We don't actually need this class at all so I deleted it.

src/IB/IBFEMethod.cpp
drwells: Illustrate my lack of SAMRAI knowledge :)

src/IB/IBFEMethod.cpp
drwells: Yes, but this is never relevant with computed forces, right?

src/IB/IBFEMethod.cpp
drwells: This would require a significant change to the way we do this: we currently accumulate inside `FEDataManager`. It would be a lot more efficient to do the
1. setup temporary force data index
2. spread force
3. accumulate force
4. add temporary force to given index

inside IBFEMethod but this changes behavior.

Let me open an issue for this so that we can address it later: I would prefer to leave it out of this PR since it is only tangentially related to the scratch hierarchy.

drwells:
I think that should be all we need aside from the remaining question I have for @boyceg.

src/IB/IBFEMethod.cpp
boyceg: For spreading as currently implemented, the `FEDataManager::spread()` routine needs to know what the physical BCs are (to accumulate from the physical boundary ghost values back into the interior), but I don't think that needs to be used to transfer from the scratch hierarchy back to the primary hierarchy.

drwells: I think my question is, more exactly, since there are no boundary conditions for forces we do not need to do this when moving data to the scratch hierarchy, right?

boyceg: Ah. It depends.

If for some reason the forces need to be interpolated from coarser levels to fill in the fine grid level, and if that interpolation needs to happen near the boundary, then you will need some ghost values...

However, I think that we want to try to ensure that the scratch hierarchy is contained in the corresponding fine level of the primary hierarchy, so that data transfer is just straight-up copying in the overlap regions.

drwells: The scratch hierarchy should always occupy the same region of space as the primary hierarchy so I think this is fine.

drwells:
That should do it so lets merge!

PR #655, opened 2019-07-30
Removed solve for advection diffusion solver with zero diffusion coefficient

### LINKED ISSUE #573: Advection diffusion solver iterates when there is no diffusion.
The advection diffusion solver will iterate even if the diffusion coefficient is set to 0. Since we are using explicit time stepping for advection, we should skip the ksp solve in this case.

abarret:
This removes the solve in `AdvDiffSemiImplicitHierarchyIntegrator` if the diffusion coefficient is zero. This does not change the error norms for the examples. Fixes #573.

### DIFF
--- a/src/adv_diff/AdvDiffSemiImplicitHierarchyIntegrator.cpp
+++ b/src/adv_diff/AdvDiffSemiImplicitHierarchyIntegrator.cpp
@@ -819,20 +819,29 @@ AdvDiffSemiImplicitHierarchyIntegrator::integrateHierarchy(const double current_
             d_hier_cc_data_ops->axpy(Q_rhs_scratch_idx, 1.0, F_scratch_idx, Q_rhs_scratch_idx);
         }
 
-        // Solve for Q(n+1).
-        Pointer<PoissonSolver> helmholtz_solver = d_helmholtz_solvers[l];
-        helmholtz_solver->solveSystem(*d_sol_vecs[l], *d_rhs_vecs[l]);
-        d_hier_cc_data_ops->copyData(Q_new_idx, Q_scratch_idx);
-        if (d_enable_logging && d_enable_logging_solver_iterations)
-            plog << d_object_name << "::integrateHierarchy(): diffusion solve number of iterations = "
-                 << helmholtz_solver->getNumIterations() << "\n";
-        if (d_enable_logging)
-            plog << d_object_name << "::integrateHierarchy(): diffusion solve residual norm        = "
-                 << helmholtz_solver->getResidualNorm() << "\n";
-        if (helmholtz_solver->getNumIterations() == helmholtz_solver->getMaxIterations())
+        if (isDiffusionCoefficientVariable(Q_var) || (d_Q_diffusion_coef[Q_var] != 0.0))
         {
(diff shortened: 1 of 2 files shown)


abarret:
Force pushed to add a changelog entry.

src/adv_diff/AdvDiffSemiImplicitHierarchyIntegrator.cpp
```diff
@@ -819,20 +819,29 @@ AdvDiffSemiImplicitHierarchyIntegrator::integrateHierarchy(const double current_
...
-        helmholtz_solver->solveSystem(*d_sol_vecs[l], *d_rhs_vecs[l]);
-        d_hier_cc_data_ops->copyData(Q_new_idx, Q_scratch_idx);
-        if (d_enable_logging && d_enable_logging_solver_iterations)
-            plog << d_object_name << "::integrateHierarchy(): diffusion solve number of iterations = "
-                 << helmholtz_solver->getNumIterations() << "\n";
-        if (d_enable_logging)
-            plog << d_object_name << "::integrateHierarchy(): diffusion solve residual norm        = "
-                 << helmholtz_solver->getResidualNorm() << "\n";
-        if (helmholtz_solver->getNumIterations() == helmholtz_solver->getMaxIterations())
+        if (isDiffusionCoefficientVariable(Q_var) || !(MathUtilities<double>::equalEps(d_Q_diffusion_coef[Q_var], 0.0)))
```
drwells: I think it would be better to be conservative and directly check that `eps != 0.0`. I sometimes see people use 1e-20 for the diffusion coefficient and in that case we should still iterate.

src/adv_diff/AdvDiffSemiImplicitHierarchyIntegrator.cpp
```diff
@@ -819,20 +819,29 @@ AdvDiffSemiImplicitHierarchyIntegrator::integrateHierarchy(const double current_
...
+                pout << d_object_name << "::integrateHierarchy():"
+                     << "  WARNING: linear solver iterations == max iterations\n";
+            }
+        }
+        else
         {
-            pout << d_object_name << "::integrateHierarchy():"
-                 << "  WARNING: linear solver iterations == max iterations\n";
+            // No solve needed for Q(n+1)
+            d_hier_cc_data_ops->scale(Q_new_idx, dt, Q_rhs_scratch_idx);
```
drwells: I'm not sure how the rest of this code works, but should we be assuming here that we can do forward Euler?

abarret: The patch data stored by `Q_rhs_scratch_idx` already accounts for the time stepping scheme. Essentially, we're solving an identity system scaled by 1/dt for `Q_new_idx`.

drwells: Ah. Sounds good :+1:

src/adv_diff/AdvDiffSemiImplicitHierarchyIntegrator.cpp
abarret: Good point.

PR #665, opened 2019-08-29
Add a quadratic interpolation option at codim 1 physical boundaries.

abarret:
This adds a quadratic interpolation option for cell- and side-centered quantities at codim 1 boundaries. To be second order, certain operations require this interpolation. This reduces to linear extrapolation at codim 2 and 3 boundaries, similar to the quadratic extrapolation code.

The interpolation is similar to the linear interpolation option. We form a quadratic polynomial `p` between the ghost cell, it's mirror interior cell, and the cell adjacent to the mirror image. We then solve for the ghost cell value in the equation
`a p(x_g) + b p'(x_g) = u_g`

An option is also added to `HierarchyGhostCellInterpolation` to allow for quadratic interpolation at the boundary. It defaults to `LINEAR`. This also adds 16 tests to ensure that the linear and quadratic interpolation interpolates the correct quantities.

### DIFF
--- a/ibtk/include/ibtk/CartCellRobinPhysBdryOp.h
+++ b/ibtk/include/ibtk/CartCellRobinPhysBdryOp.h
@@ -107,39 +107,48 @@ public:
      * \param patch_data_index  Patch data index requiring ghost cell filling.
      * \param bc_coef           Robin boundary conditions to use with this class.
      * \param homogeneous_bc    Whether to employ the homogeneous form of the boundary
-     *conditions.
+     * conditions.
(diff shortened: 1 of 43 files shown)


abarret:
@drwells Most of the line additions here are the "incorrect" values being detected at codim 2 and 3 boundaries in 3 spatial dimensions for the test suite (where we expect them to be wrong). Would it be worth it to ignore these type of boundaries when we test?

abarret:
@drwells, don't answer that. I'm going to change it to fix the output.

abarret:
There. How about that?

drwells:
I ran the new tests on my machine and they pass.

PR #666, opened 2019-08-29
Silence a bogus boost warning.

drwells:
Some versions of GCC complain about U_node.shape()[0] being uninitialized. There is nothing we can do about it in this function so ignore the warning.


drwells:
Since all that this patch does is silence a warning and it passes the CI build I will merge.

PR #667, opened 2019-08-30
Rename a test to be much shorter.

drwells:
The test name is long enough that the console output becomes much wider than 80 characters. The test prefix `physical_boundary/physical_boundary` is redundant, anyway, so just remove the second part.


drwells:
@abarret are you okay with the rename? If so I will merge.

abarret:
One thought, how will this affect running specific tests using `attest -R`?

abarret:
Is there some way to run all the tests in a specific directory?

drwells:
`attest -R` is broken in that it doesn't take into account the directory. I still need to fix that.

drwells:
To be clear: `attest -R '^foo/'` should run all the tests in the `foo/` directory but this currently does not work: the searching is done on file names, not complete test names.

drwells:
The problem raised by @abarret was addressed in #670 so lets merge.

PR #669, opened 2019-09-04
Higher order element support in get_max_edge_length

### LINKED ISSUE #668: Implement max_edge_length for higher order elements
From our meeting today: we should implement this by linearizing element edges with line segments and calculating the sum of their lengths.

drwells:
Fixes #668.

The test output changed because instead of the old algorithm (where we find the maximum distance between two nodes with a common edge) we find the lengths of both line segments and add them together.


boyceg:
Does this affect performance any?

drwells:
It should help a lot. I'll check.

drwells:
On my little benchmark (IBFE/explicit/ex4/main3d with `Hex27` elements and 5 time steps) we go from 687 million instructions in `getQuadratureKey` to 110 million instructions. Its a bit less than 1% of the total instruction count.

drwells:
@boyceg are you happy with this? If so lets merge.

PR #670, opened 2019-09-04
Update attest

drwells:
Addresses a comment in #667: `attest` now uses the full test name (i.e., including its subdirectory) when filtering tests by regular expressions.


drwells:
This is just for the test suite so I will merge myself.

Issue #643, opened 2019-07-09
Pair opening istreams with checks


drwells:
From our group meeting today: `std::istream` (or, more relevantly, `std::ifstream`) can stream in junk if the file name we gave it does not correspond to a file that actually exists. This is pretty easy to check: we just need to do
```cpp
std::ifstream input("filename");
TBOX_ASSERT(input);
```
to verify that the stream is good.

boyceg:
Can we close this issue?

drwells:
Yup.

PR #673, opened 2019-09-09
Consistently check ifstream states.

### LINKED ISSUE #413: IBStandardInitializer should fail when file exists but not readable
`IBStandardInitializer` quietly ignores files that can't be opened. It should either quit or at least emit a helpful warning message.

C++ seems to be ill equipped to deal with this. As @drwells pointed out, we can either use C++17's file system library (which requires C++17), boost's version of the file system library (which requires a compiled boost build), or stat (which is non-standard). None of these options are ideal, but stat seems to be the least intrusive option. We could also just leave this as a "feature" of IBAMR. @boyceg, do you have any opinions?

drwells:
Ensure that we actually opened the file in all cases.

Fixes #413 and #643.

### DIFF
--- a/doc/news/changes/minor/20190919DavidWells
+++ b/doc/news/changes/minor/20190919DavidWells
@@ -1,0 +1,4 @@
+Fixed: Fixed a problem where <code>rpath</code> arguments to the linker were
+incorrectly formatted, resulting in missing libraries at run time.
+<br>
+(David Wells, 2019/09/19)
--- a/doc/news/changes/minor/20190920DavidWells
+++ b/doc/news/changes/minor/20190920DavidWells
@@ -1,4 +1,4 @@
-Fixed: Fixed a problem where <code>rpath</code> arguments to the linker were
-incorrectly formatted, resulting in missing libraries at run time.
+Improved: IBStandardInitializer will now log a message explaining what happened
+when optional input files (e.g., rod input files) are not available.
 <br>
-(David Wells, 2019/09/19)
+(David Wells, 2019/09/20)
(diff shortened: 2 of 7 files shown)


drwells:
I think the CI infrastructure is down: I get 404s when I click on the `Details` link.

drwells:
This approach won't work: `IBStandardInitializer` loads files based on whether or not they exist. This patch causes the multiphase tests to fail since they only load vertex files and ignore the rest.

We can resolve this by consistently checking that the file stream was opened successfully, but any other checks will break peoples codes.

boyceg:
> This approach won't work: `IBStandardInitializer` loads files based on whether or not they exist. This patch causes the multiphase tests to fail since they only load vertex files and ignore the rest.

Yes, that is right --- there is one required file and several optional files that are used by `IBStandardInitializer`. If the optional files are not present, then they are skipped (with a log message).

drwells:
I don't think we consistently log that files are skipped. Should we?

boyceg:
I think that we log that certain things are "skipped". Maybe we don't?

drwells:
We do not: e.g., if the spring files do not exist we silently continue. Lets add it.

drwells:
I fixed the merge conflict.

drwells:
Thats a tricky question to answer. I think that users are more likely to look at the logfile than the slurm output so I think the current option (plog) is the right one.

drwells:
@boyceg are you okay with not making further changes? If so lets merge.

drwells:
@boyceg can we merge?

boyceg:
Oh, sorry, I thought I already approved.

drwells:
You did but I wanted to be sure you were OK with the bit on plog vs pout.

PR #674, opened 2019-09-10
Refactor fe data manager 1

drwells:
Part of #647: as discussed in that issue, we should split FEDataManager into two parts: something that keeps track of all the libMesh objects and something (building on top of the first) that keeps track of the parts dependent on the Eulerian grid.

This PR just moves all of the libMesh data structures that do not need to know anything about Eulerian data (the IB-ghosted vectors are still in FEDataManager) into a new base class `FEData`. I am not in love with the name but it works.


drwells:
This passes the test suite and moves things around in a compatible way so lets merge.

PR #677, opened 2019-09-15
update setup_compile_commands.sh to set specific compiler

boyceg:
Set a specific C++ compiler to make it easier to get `clangd` to work with the generated `compile_commands.json`.


boyceg:
I am going to go ahead and merge because we don't have tests for this anyway.

Issue #675, opened 2019-09-12
Our FECache code doesn't work with subdivision elements


drwells:
@xxjaehoxx This is the issue we talked about today :)

This is not so much a problem with FECache, but the way we use it:
```cpp
FEBase& X_fe = X_fe_cache[key];
QBase& qrule = d_quadrature_cache[key];

// See the note in interpWeighted to explain why we override
// libMesh's reinit logic here
if (used_quadratures.find(key) == used_quadratures.end())
{
    X_fe.get_phi();
    X_fe.reinit(elem);
    used_quadratures.insert(key);
}
const std::vector<std::vector<double> >& X_phi = X_fe.get_phi();
```
i.e., we get the (possibly uninitialized) object from the cache and then only initialize it if we have a new quadrature rule. This won't work with subdivision elements since they have, in general, a different number of degrees of freedom per element (i.e., we can keep the same quadrature rule but the `phi` array will still be different).

I think that the correct way to fix this is to instruct FECache to also initialize the FE objects it stores appropriately: e.g., if we have subdivision elements then we always need to recompute shape functions, but if we have Lagrange elements we only need to do it once.

boyceg:
Can we close this issue?

xxjaehoxx:
Sounds good to me!

drwells:
The fix in #679 should be sufficient; we can always open this again if we want to do more work with subdivision elements.

PR #679, opened 2019-09-16
Redo fe cache

drwells:
Most of the fix for #675: we should now reinitialize the FE objects correctly with subdivision elements inside FEDataManager.

@xxjaehoxx I don't actually have any test code for subdivision elements but we should add some to this PR. Do you have the L2 projection code done? If so, send it to me and I will modify it to use `FECache` to check that our implementation actually works.

### DIFF
--- a/ibtk/include/ibtk/FECache.h
+++ b/ibtk/include/ibtk/FECache.h
@@ -115,43 +115,43 @@ protected:
     /**
      * Dimension of the FE mesh.
      */
-    const unsigned int dim;
+    const unsigned int d_dim;
 
     /**
      * Object describing the finite element type.
      */
-    const libMesh::FEType fe_type;
+    const libMesh::FEType d_fe_type;
 
     /**
      * Managed libMesh::Quadrature objects. These are attached to the FE
      * objects.
      */
-    QuadratureCache quadrature_cache;
+    QuadratureCache d_quadrature_cache;
 
     /**
      * Managed libMesh::FE objects of specified dimension and family.
      */
-    std::map<key_type, std::unique_ptr<libMesh::FEBase> > fes;
+    std::map<key_type, std::unique_ptr<libMesh::FEBase> > d_fes;
 };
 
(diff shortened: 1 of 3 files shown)


ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1669,25 +1657,7 @@ FEDataManager::interpWeighted(const int f_data_idx,
...
-                // long as we have called it on this FE object at least once.
-                if (used_X_quadratures.find(key) == used_X_quadratures.end())
-                {
-                    // get_phi changes the state of the object if called
-                    // before reinit
-                    X_fe.get_phi();
-                    X_fe.reinit(elem);
-                    used_X_quadratures.insert(key);
-                }
-
```
drwells: A nice thing about this patch is that I got to delete a previous round of hacks, now that I know how to implement `FECache` correctly.

ibtk/include/ibtk/FECache.h
```diff
@@ -54,6 +54,22 @@ namespace IBTK
...
+ * libMesh::FEBase objects stored by FECache. Multiple flags can be enabled
+ * with bitwise or operations, e.g.,
+ *
+ * @code
+ * const IBTK::FEUpdateFlags update_flags = IBTK::update_phi | IBTK::update_dphi;
+ * @endcode
+ */
+enum FEUpdateFlags
+{
+    update_default = 0,
```
boyceg: What is computed for `default`?

drwells: Nothing. I should go back and label these.

drwells:
I checked and there is no performance regression with this patch: in fact, since we get rid of the ad-hoc initialization logic, things are about 0.03% faster :tada:

drwells:
This PR doesn't fix all of our problems with subdivision elements, but its a good step forward. Since these changes have been approved I am going to merge to avoid getting more merge conflicts as work on FEDataManager continues.

PR #680, opened 2019-09-17
Make FEDataManager no longer inherit from FEData.

drwells:
This is necessary for some upcoming work that will require multiple `FEDataManager` objects to use the same Lagrangian data but different Eulerian data: i.e., multiple `FEDataManager` objects will share the same `FEData` object. In particular: both `IBFEPostProcessor` and `IBFEMethod` need to do some Lagrangian-Eulerian interaction, but really need different `FEDataManager` objects since the former should use the immediately available hierarchy and the later should use the scratch hierarchy.

Part of #647.


drwells:
Its not immediately clear, but this patch is backwards compatible and just moves data around.

PR #681, opened 2019-09-19
zero IB forces in the right place

boyceg:
This should fix the issue @jabrown893 was encountering, in which he "inactivated" a structure but it still applied forces.

### DIFF
--- a/doc/news/changes/major/20190919BoyceGriffith
+++ b/doc/news/changes/major/20190919BoyceGriffith
@@ -1,0 +1,4 @@
+Fixed: LDataManager properly zeros structural forces associated with
+inactivated structures.
+<br>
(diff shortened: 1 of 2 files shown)


boyceg:
Perhaps @jabrown893 could make a unit test by modifying `IB/explicit/ex1` to inactivate the structure? The exact solution should be u = v = p = 0.

I'd like to suggest that we go ahead and merge this and get the unit test in afterward.

drwells:
Hm. I prefer to bundle tests with fixes but since this is an annoying bug lets just open an issue for the test. I will hold @jabrown893 to writing one :)

@boyceg would you mind adding a changelog entry?

jabrown893:
Thank you!

PR #684, opened 2019-09-19
IBFE updates

boyceg:


drwells:
I thought we used some of this overlap code in the heart model (which would explain a lot). Let me take a closer look.

boyceg:
The overlapping stuff does not work AFAIK.

boyceg:
(It is easy to drop that commit if we want to, though.)

boyceg:
@drwells a difference between the heart model and the IBFE examples is that the heart model heavily uses `FEDataInterpolation` to evaluate field variables in stress functions etc. I am looking to see if there is something screwy in that class.

drwells:
Cleaning up `FEDataInterpolation` has been on my todo list for months. This might be a good time to spend a few days working through it.

drwells:
@boyceg would you mind if I moved a951f1d5ebd09fd2a0a29c658e1b08ac09087832 into another PR? Since that just chops stuff out we can merge it immediately.

boyceg:
> @boyceg would you mind if I moved [a951f1d](https://github.com/IBAMR/IBAMR/commit/a951f1d5ebd09fd2a0a29c658e1b08ac09087832) into another PR? Since that just chops stuff out we can merge it immediately.

OK with me.

drwells:
@boyceg would it be worthwhile for us to step back for a moment and consider redesigning `FEDataInterpolation`? It is quite slow since it does not use any of the caching infrastructure.

boyceg:
> @boyceg would it be worthwhile for us to step back for a moment and consider redesigning `FEDataInterpolation`? It is quite slow since it does not use any of the caching infrastructure.

Sounds good to me.

drwells:
Using the caches would require completely rewriting this class and then replacing the way we use it at all call sites: I think it would be better to try and fix the memory leak first and then worry about this separately.

The rest of the patches in this PR are all pretty uncontroversial; I would be in favor of merging after we make sure @cpuelz is okay with #685.

boyceg:
@drwells shall we merge this?

drwells:
All commits look reasonable and the test suite passes so :+1:

PR #685, opened 2019-09-19
remove overlapping mesh forces in IBFEMethod

drwells:
Taken from #684.

@cpuelz The heart model currently uses this function but the only input files I see in the repository that actually use it are in the `old_input_files` directory.


drwells:
@cpuelz Is it okay if we remove this?

doc/news/changes/incompatibilities/20190919BoyceGriffith
```diff
@@ -0,0 +1,6 @@
+Removed: The overlap code in IBFEMethod
+(IBFEMethod::computeOverlapConstraintForceDensity(),
+IBFEMethod::resetOverlapNodalValues(), and related functions) has been removed
+due to several issues with bugs and memory leaks.
```
boyceg:
```suggestion
because the algorithm generates spurious motion.
```

cpuelz:
> @cpuelz Is it okay if we remove this?

yeah no problem with me.

PR #687, opened 2019-09-19
Fix a path fed to rpath.

### LINKED ISSUE #686: Weird flags in LDFLAGS
We are getting flags of the form ` -Wl,-rpath `in `LDFLAGS` inside `make.inc`: this doesn't mean anything and might be a bug. We correctly generate rpath commands in `LIBS`.

drwells:
Fixes #686. I need to do some more tests before we merge but this appears to add the correct linker flags. @jabrown893 if you have some time it would help a lot if you could run this on longleaf.

We should not have the -L prefix here, so we should pass the directory instead.

### DIFF
--- a/configure
+++ b/configure
@@ -26955,7 +26955,9 @@ fi
 
 PETSC_LDFLAGS=""
 if test -d "${PETSC_DIR}/${PETSC_ARCH}/lib" ; then
-    PETSC_LDFLAGS="-L${PETSC_DIR}/${PETSC_ARCH}/lib"
+  PETSC_LDFLAGS="-L${PETSC_DIR}/${PETSC_ARCH}/lib"
+else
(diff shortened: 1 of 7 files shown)


drwells:
@jabrown893 To answer your question from last night: you can get a branch from another fork by running

```
> git checkout master
> git checkout -b fix-rpath-2
> git pull https://github.com/drwells/ibamr fix-rpath-2
```

In general, you can pull changes from any complete path, e.g., sometimes it is useful to do
```
git pull $(pwd)/../other-ibamr/ other-branch
```
which pulls from another directory on the current machine.

jabrown893:
This seems to work! 👍

drwells:
@jabrown893 this is now merged so `master` should work for you on Longleaf again.

jabrown893:
@drwells I will check it and let you know!

PR #688, opened 2019-09-20
Add a document describing the release process.

drwells:
#686 demonstrates that we need to do a better job testing releases: all of the machines I run on have a system copy of HDF5 (or I compile PETSc with HDF5 support), which caused me to miss this configuration bug. To get around this I wrote up a checklist for putting out a release inspired by

https://github.com/dealii/release-papers/blob/master/release-tasks

which will remind us to actually check this configuration at release time.

### DIFF
--- a/doc/release-tasks.md
+++ b/doc/release-tasks.md
@@ -1,0 +1,75 @@
+# IBAMR release tasks
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


drwells:
I am the primary user of this document, so lets merge as-is.

PR #689, opened 2019-09-23
Remove some debugging messages from configure.

drwells:
These are from an m4 macro I modified but did not commit.


drwells:
This is quite minor and only involves calls to `echo` done by configure, so lets merge.

PR #696, opened 2019-09-24
Silence a warning from Intel's compiler.

drwells:
This is something I noticed with @jabrown893 yesterday.


drwells:
This is simple enough (and all the tests pass) that I will merge myself.

Issue #683, opened 2019-09-19
post 681: test for deactivated structures


drwells:
We should add a little test to make sure that deactivated structures don't move once #681 is merged. We don't need anything fancy, just an IB example where one structure is deactivated (and we can compute that its velocity is zero and that the forces are correct).

jabrown893:
Sounds good! I have to grade calc tests this weekend, but I will try to get on this next week.

PR #699, opened 2019-09-25
Create new test with inactivated structure

jabrown893:
This commit should close #681 and #683 by adding a test that checks whether a structure is truly inactivated.


drwells:
This test (correctly) fails without the previous patch and passes on master.

PR #700, opened 2019-09-25
Add restart test

drwells:
It looks like I messed up FEDataManager's restart settings as part of #647. This PR fixes its restart code and also adds some tests that verify we can successfully restart and then compute correct values.


drwells:
Since this PR fixes a rather important regression (and just puts code back where it was in the 0.5.1 release) and I need it to continue work on other stuff I will go ahead and merge myself.

PR #701, opened 2019-09-27
Actually fix the restart test.

drwells:
This should have been committed with the previous PR.


drwells:
I triple-checked that this actually works so I will merge.

PR #703, opened 2019-10-01
Add test for IBFECentroidPostProcessor.

drwells:
Part of #654: we need to be able to test the postprocessor before we can implement the 'use multiple `FEDataManager`s' logic.

This PR just adds one new test for the postprocessor: I will add more tests once we can use multiple `FEDataManager`s with the same Lagrangian data.


drwells:
This just modifies tests and I need it for the next patch in the sequence so I will merge as-is without further review.

PR #704, opened 2019-10-02
Allow FEDataManagers to share one FEData object.

drwells:
Part of #647: we can now successfully create `IBFECentroidPostProcessor`s that use an `FEDataManager` object that is external to `IBFEMethod`. While this itself is a bit clumsy, it verifies that the planned approach (`IBFEMethod` maintaining two different sets of `FEDataManager`s with different Eulerian partitionings) will work.

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -215,6 +215,15 @@ protected:
     std::string d_object_name;
     bool d_registered_for_restart;
 
+    /*!
+     * Name of the coordinates system.
+     *
+     * @note For backwards compatibility reasons this string may be reassigned
+     * to another value by assignment to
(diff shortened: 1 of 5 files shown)


ibtk/include/ibtk/FEDataManager.h
```diff
@@ -215,6 +215,15 @@ class FEData : public SAMRAI::tbox::Serializable
...
     bool d_registered_for_restart;
 
+    /*!
+     * Name of the coordinates system.
+     *
+     * @note For backwards compatibility reasons this string may be reassigned
+     * to another value by assignment to
+     * FEDataManager::COORDINATES_SYSTEM_NAME.
+     */
+    std::string d_coordinates_system_name = "coordinates system";
```
drwells: I only caught this issue because of the test suite.

PR #706, opened 2019-10-02
Include a python script Aaron wrote.

drwells:
This should also now work with python 3.

@jabrown893 Two things:
1. Would you please try out this version of the script with both python 2 and python 3 to make sure it works with both?
2. I didn't read the script very carefully. Is the summary of it in the changelog entry correct?


jabrown893:
@drwells 
1. It appears to work with both python 2 & python 3.

jabrown893:
@drwells 
2. I think the description is pretty much right. Maybe add that it lists all the data files inside the files named dumps.visit and lag_data.visit.

drwells:
Thanks. I also added a documentation string to the script to record how it works.

drwells:
Since this PR only adds a script that is never executed by travis I will merge as-is without waiting on the CI build.

PR #707, opened 2019-10-03
added documentation for the jacobian function in libmesh_utilities.h

marshallrdavey:


ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -585,6 +585,25 @@ interpolate(libMesh::TypeVector<double>& U,
     return;
 } // interpolate
 
+/**
+ * Compute the jacobian with respect to the initial configuration in the deformed configuration
```
drwells: We should add a formula defining the Jacobian here. I think we can do this with MathJax.

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -585,6 +585,25 @@ interpolate(libMesh::TypeVector<double>& U,
...
+ * @p X_node at quadrature point number @qp.
+ *
+ * @param[out] dX_ds Tensor containing the output of this function of size 3x3.
```
drwells: Similarly, we should write down the formula defining this.

ibtk/include/ibtk/libmesh_utilities.h
marshallrdavey: I don't know how to use mathjax, but I'm willing to write out the equations

drwells:
@marshallrdavey Slight correction: our current setup is to prerender formulas (not use MathJax), though its not relevant to this PR.

Try writing, e.g.,
```
\f[
    \vec{F}_l = \kappa M
\f]
```
and using LaTeX syntax inside the `\f[ ... \f]` brackets.

marshallrdavey:
@drwells I've added the formula for finding the jacobian at a given quadrature point. I can modify it if you want to use different parameter names/need me to add something else for clarity.

marshallrdavey:
e.g. \xi(qp) = \sum_{i = 1}^n \xi_i \phi_i(qp)

PR #708, opened 2019-10-03
attest: allow -j (without a number) for all cores.

drwells:
A small missing feature.


drwells:
This only impacts the test runner (which is not executed by Jenkins) so lets merge as-is.

PR #709, opened 2019-10-03
Fix some pylint complaints about attest.

drwells:


drwells:
This commit only pertains to attest so lets merge as-is.

PR #711, opened 2019-10-04
Add a pointer check to FEDataManager.

drwells:
Peeled off from #654.

These pointers have to point to the same patch hierarchy so explicitly assert it.


ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2531,6 +2531,7 @@ FEDataManager::applyGradientDetector(const Pointer<BasePatchHierarchy<NDIM> > hi
     TBOX_ASSERT(hierarchy);
     TBOX_ASSERT((level_number >= 0) && (level_number <= hierarchy->getFinestLevelNumber()));
     TBOX_ASSERT(hierarchy->getPatchLevel(level_number));
+    TBOX_ASSERT(hierarchy.getPointer() == d_hierarchy.getPointer());
```
boyceg:
```suggestion
    TBOX_ASSERT(hierarchy == d_hierarchy);
```
?

drwells: I am surprised that this worked since the two objects have different types, but it does. I'll make the suggested change and merge myself.

PR #713, opened 2019-10-05
Add optional argument to not draw convected quantities or source terms

abarret:
This adds an optional argument when registering variables with the advection diffusion integrators to turn off drawing of the variables.

### DIFF
--- a/include/ibamr/AdvDiffHierarchyIntegrator.h
+++ b/include/ibamr/AdvDiffHierarchyIntegrator.h
@@ -182,12 +182,13 @@ public:
     getAdvectionVelocityFunction(SAMRAI::tbox::Pointer<SAMRAI::pdat::FaceVariable<NDIM, double> > u_var) const;
 
     /*!
-     * Register a cell-centered source term.
+     * Register a cell-centered source term. Can optionally turn off outputting the source.
      *
      * Data management for the registered source term will be handled by the
      * hierarchy integrator.
      */
-    virtual void registerSourceTerm(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > F_var);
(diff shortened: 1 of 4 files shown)


abarret:
I knew 'draw' sounded silly for some reason...

The issue with separating registration from outputting is that the decision must be made _before_ `initializeHierarchyIntegrator` is called. At this point the variables are registered with the visit data writer, and as far as I can tell, variables can not be removed from the data writer.

boyceg:
Could you have a separate `registerVariableForPlotting()` routine that is called before `initializeHierarchyIntegrator()`?

Although I understand the appeal of not having to make the separate calls. (@drwells, do you have an opinions?)

drwells:
Is there an easy way to get the index? I would much rather write, e.g.,
```cpp
if (uses_visit)
{
    visit_data_writer->registerPlotQuantity("advect_q", "SCALAR", time_integrator->getQDataIndex());
}
```
or something similar.

boyceg:
This might not be a terrible time to revisit the way that plotting and other data export is handled.

abarret:
This PR follows @drwells suggestion while leaving the default behavior intact.

The current behavior is that the advection diffusion integrator registers all variables with the visit data writer when `registerVariables()` is called, e.g. in `initializeHierarchyIntegrator()`. The way this PR is currently set up would allow users to forego that option and maintain the output themselves.

By the way, @drwells, you can get the patch data index via the variable context from the hierarchy integrator `getCurrentContext()`.

abarret:
Force pushed an update to rename draw to output.

drwells:
This seems fine to me: I think any other approach would not be backwards compatible.

@boyceg are you happy with the current state of this PR?

PR #714, opened 2019-10-05
WIP: overhaul implicit solver

boyceg:
This overhauls the implicit IB solver framework to use things that look like fixed-point iteration on the position DOFs `X`.

To simplify the implementation, it temporarily removes direct support for Eulerian approaches that used a specialized MG solver for the IB equations. The fully Eulerian approach does not appear to be practical at this time, but we probably should keep it around for experimentation.

### DIFF
--- a/src/IB/IBImplicitHierarchyIntegrator.cpp
+++ b/src/IB/IBImplicitHierarchyIntegrator.cpp
@@ -241,7 +241,7 @@ IBImplicitHierarchyIntegrator::integrateHierarchy(const double current_time, con
 
     if (d_use_fixed_LE_operators) d_ib_implicit_ops->updateFixedLEOperators();
 
-    if (d_implicit_algorithm == "AITKEN" || d_implicit_algorithm == "ANDERSON")
+    if (d_implicit_algorithm == "AITKEN" || d_implicit_algorithm == "ANDERSON" || d_implicit_algorithm == "IRONS-TUCK")
     {
         // Use the crossed secant method (Aitken extrapolation) or the alternate secant method (Anderson extrapolation
         // with M = 1).
@@ -285,6 +285,9 @@ IBImplicitHierarchyIntegrator::integrateHierarchy(const double current_time, con
         ierr = VecWAXPY(DX_nm1, -1.0, X_nm1, GX_nm1);
         IBTK_CHKERRQ(ierr);
 
+        // Initialize omega_{n-1}:
+        double omega_nm1 = std::min(d_omega_n, d_omega_max);
+
         for (k = 0; k < max_its && !converged; ++k)
         {
             // G(X_{n}):
@@ -308,10 +311,10 @@ IBImplicitHierarchyIntegrator::integrateHierarchy(const double current_time, con
             {
                 // X_{n+1} := G(X_{n}) - Q DX_{n}
                 //
-                // with Q = dot(G(X_{n}) - G(X_{n-1}, DX_{n} - DX_{n-1})/||DX_{n} - DX_{n-1}||^2
-                //        = dot(DGX_{n}, DDX_{n})/||DDX_{n}||^2
+                // with Q = dot(DX_{n-1}, DX_{n} - DX_{n-1})/||DX_{n} - DX_{n-1}||^2
+                //        = dot(DX_{n}, DDX_{n})/||DDX_{n}||^2
                 double Q1;
-                ierr = VecDot(DGX_n, DDX_n, &Q1);
+                ierr = VecDot(DX_n, DDX_n, &Q1);
                 IBTK_CHKERRQ(ierr);
                 double Q2;
                 ierr = VecDot(DDX_n, DDX_n, &Q2);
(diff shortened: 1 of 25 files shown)


src/IB/IBImplicitHierarchyIntegrator.cpp
```diff
@@ -304,160 +307,63 @@ IBImplicitHierarchyIntegrator::integrateHierarchy(const double current_time, con
...
+                ierr = VecDot(DDX_n, DDX_n, &Q2);
+                IBTK_CHKERRQ(ierr);
 
-            const double Q = Q1 / Q2;
-            ierr = VecWAXPY(X_np1, -Q, DGX_n, GX_n);
-            IBTK_CHKERRQ(ierr);
+                d_omega_n = -omega_nm1 * Q1 / Q2;
+                ierr = VecAXPBYPCZ(X_np1, d_omega_n, 1.0 - d_omega_n, 0.0, GX_n, X_n);
+                IBTK_CHKERRQ(ierr);
+                omega_nm1 = d_omega_n;
```
JianhuaQin: Maybe for k==0 we only need to use the one from previous time step like the following code.

  ```
  else if (d_implicit_algorithm == "IRONS-TUCK")
            {
                if(k>0)
                {
                // X_{n+1} := omega_{n} G(X_n) + (1.0-omega_{n}) X_n
                //
                // with omega_{n} = -omega_{n-1} dot(DX_{n-1}, DX_{n} - DX_{n-1})/||DX_{n} - DX_{n-1}||^2
                //                = -omega_{n-1} dot(DX_{n-1}, DDX_{n})/||DDX_{n}||^2
                double Q1;
                ierr = VecDot(DX_nm1, DDX_n, &Q1);
                IBTK_CHKERRQ(ierr);
                double Q2;
                ierr = VecDot(DDX_n, DDX_n, &Q2);
                IBTK_CHKERRQ(ierr);
            const double Q = Q1 / Q2;
            ierr = VecWAXPY(X_np1, -Q, DGX_n, GX_n);
            IBTK_CHKERRQ(ierr);
                d_omega_n = -omega_nm1 * Q1 / Q2;
                ierr = VecAXPBYPCZ(X_np1, d_omega_n, 1.0 - d_omega_n, 0.0, GX_n, X_n);
                IBTK_CHKERRQ(ierr);
                omega_nm1 = d_omega_n;
                }else
               {
                d_omega_n=omega_nm1; 
                }
```

Comment (boyceg) [reply src/IB/IBImplicitHierarchyIntegrator.cpp]: It is pretty cheap to compute the updated value of `omega`, so I think we might as well do it unless this improves convergence.

Comment (JianhuaQin) [reply src/IB/IBImplicitHierarchyIntegrator.cpp]: I think we should put the following code into integrateHiearchy
```
    Vec X_old_time_step;
   d_ib_implicit_ops->createSolutionVec(X_old_time_step);
   d_ib_implicit_ops->getUpdatedPosition(X_old_time_step);
    if (d_use_structure_predictor)
    {
        if (d_enable_logging)
            plog << d_object_name << "::preprocessIntegrateHierarchy(): performing Lagrangian forward Euler step\n";
        d_ib_implicit_ops->forwardEulerStep(current_time, new_time);
    }

```
Also we need to delete the predictor in the 
```

IBImplicitHierarchyIntegrator::preprocessIntegrateHierarchy(const double current_time,
                                                            const double new_time,
                                                            const int num_cycles)

```

boyceg: This is a little opaque in the interface, but the IB implementation keeps track of `X^{n}` and `X^{n+1}` internally and always maintains `X^{n+1/2} = (X^{n+1}+X^{n})/2` for the current values of `X^{n}` and `X^{n+1}`.

`setUpdatedPosition(X_updated)` will set the current approximation to `X^{n+1}` used in the IB implementation to the provided position vector `X_updated`: `X^{n+1} := X_updated`.

`getUpdatedPosition(X_updated)` does the opposite; it will get : `X_updated := X^{n+1}`.

`iterateSolution(X_updated)` will do a basic momentum equation solve using the provided value of `X_updated` as the approximation to `X^{n+1}` for the purposes of evaluating the structural forces, and for spreading and interpolating forces. Then, the very last thing that `iterateSolution` does is to use a discrete approximation to `dX/dt = U(X,t)` to update the internal approximation to `X^{n+1}`.

So in each call to `iterateSolution()`, there is an initial approximation to `X^{n+1}` and a final approximation to `X^{n+1}`. We declare convergence if the difference between these configurations is relatively small.

boyceg: Note also that the internal vector storing `X^{n}` in the IB implementation is not reset until `IBImplicitHierarchyIntegrator::postprocessIntegrateHierarchy()`.

JianhuaQin: That's right! Storing X^{n} until the subiteration converges is what we want.
Thank you for clarifying this, I am more familiar with the code now.

drwells:
Nice work getting this going again.

Could we meet at some point to talk about the interface? I'd like to avoid interfaces which take `Vec`s or `Vec *`s as not everything inheriting from `IBStrategy` (or possibly `IBStrategy`) uses PETSc (notably IFEDMethod). It might be possible to express this in a higher-level way which avoids this.

boyceg:
Unfortunately I don't have a workable solver approach at this time --- let's worry about the interface if/when we have something like a scalable solver.

PR #718, opened 2019-10-07
Update test output following #673.

drwells:
That change added some new logfile output.


drwells:
This only effects a test so lets merge.

PR #719, opened 2019-10-08
Add a parallel postprocessor test.

drwells:
This test will help verify backwards compatibility of the scratch hierarchy branch: i.e., this test won't work there at the moment since I broke the postprocessor in parallel.


drwells:
Since this just adds another test I will merge as-is.

PR #720, opened 2019-10-08
Rearrange things to permit the scratch hierarchy.

drwells:
IBFEMethod will need to, in the future, handle multiple FEDataManagers which are associated with different patch hierarchies. This patch adds support for multiple FEDataManager sets: future work will actually turn the scratch hierarchy on.

This patch doesn't really do anything: the default and only choice is to use the 'primary' FEDataManager, which is associated with the main patch hierarchy.

### DIFF
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -768,6 +768,12 @@ protected:
     bool d_do_log = false;
 
     /*
+     * Boolean controlling whether or not the scratch hierarchy should be
+     * used.
+     */
+    bool d_use_scratch_hierarchy = false;
+
+    /*
      * Pointers to the patch hierarchy and gridding algorithm objects associated
      * with this object.
      */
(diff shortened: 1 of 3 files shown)


drwells:
Part of #654.

include/ibamr/IBFEMethod.h
```diff
@@ -767,6 +767,12 @@ class IBFEMethod : public IBStrategy
      */
     bool d_do_log = false;
 
+    /*
+     * Boolean controlling whether or not the scratch hierarchy should be
+     * used.
+     */
+    bool d_use_scratch_hierarchy = false;
```
boyceg: Could this just be determined by whether the scratch hierarchy is null/non-null?

drwells: Unfortunately no - we cannot set up the scratch hierarchy until partway through the initialization process, so we need something else to indicate whether or not to use it. This boolean is actually set in #654 from a value in the input file.

PR #724, opened 2019-10-14
Use a robust partitioner in some libmesh tests.

drwells:
Same as 8f7dc5ebf9, but for the two files I did not get at that point.


drwells:
The different partitioning has a small effect on the final result, but since we log the number of quadrature points on each processor the test will fail since, e.g., the difference between 326 and 327 is not roundoff.

PR #727, opened 2019-10-25
Add some tests based on IBFE/explicit/ex8.

drwells:
These use multiple FEDataManagers and work in a different way than ex4.

I want to add these before proceeding with #654 to verify that things still work correctly with multiple parts.


drwells:
Since this only impacts the test suite and not the library I will merge myself.

PR #729, opened 2019-10-26
WIP: numerical wave tank

boyceg:
This is WIP to clean up the NWT branch in preparation for merging it into master. I will try to do a bit more clean up later tonight. **This branch is not intended to be merged.**

@amneetb, this branch currently removes the `regridHierarchy()` and `regridProjection()` implementations that have been added in your NWT branch. The methods that you and @nishantnangia added are things that are intended to be handled using implementations of `regridHierarchyBeginSpecialized()` and `regridHierarchyEndSpecialized()`. Instead of adding alternative implementations of `regridHierarchy()`, I would like to fix the `HierarchyIntegrator` interface.

I also do not want `regridProjection()` to be public at this point.


boyceg:
By the way, my proposal for finishing this up is to get this branch in "final form" and then to apply the changes directly to master in a small number of follow-up PRs. This will not allow us to preserve the full history but seems like the best approach at this point.

boyceg:
@drwells, there were a number of merge conflicts related to your updates to the license and copyright statements. I think that I resolved them correctly, but it might be worth running through this branch to make sure that everything is consistent with master.

examples/IBLevelSet/ex0/FlowGravityForcing.h
```diff
@@ -0,0 +1,119 @@
...
+    FlowGravityForcing(const std::string& object_name,
+                       SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db,
+                       Pointer<AdvDiffHierarchyIntegrator> adv_diff_hierarchy_integrator,
+                       SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > ls_gas_var,
+                       std::vector<double> grav_const);
+
+    /*!
+     * \brief Empty destructor.
+     */
+    ~FlowGravityForcing();
```
boyceg:
```suggestion
    ~FlowGravityForcing() = default;
```

examples/IBLevelSet/ex0/FlowGravityForcing.h
```diff
@@ -0,0 +1,119 @@
...
+                        Pointer<hier::Variable<NDIM> > var,
+                        Pointer<Patch<NDIM> > patch,
+                        const double data_time,
+                        const bool initial_time = false,
+                        Pointer<PatchLevel<NDIM> > patch_level = Pointer<PatchLevel<NDIM> >(NULL));
+
+    //\}
+
+private:
+    FlowGravityForcing();
```
boyceg:
```suggestion
    FlowGravityForcing() = delete;
```

examples/IBLevelSet/ex0/FlowGravityForcing.h
```diff
@@ -0,0 +1,119 @@
...
+
+    FlowGravityForcing(const FlowGravityForcing& from);
```
boyceg:
```suggestion
    FlowGravityForcing(const FlowGravityForcing& from) = delete;
```

examples/IBLevelSet/ex0/FlowGravityForcing.h
```diff
@@ -0,0 +1,119 @@
...
+
+    FlowGravityForcing& operator=(const FlowGravityForcing& that);
```
boyceg:
```suggestion
    FlowGravityForcing& operator=(const FlowGravityForcing& that) = delete;
```

examples/IBLevelSet/ex0/GravityForcing.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+#if !defined(NDEBUG)
+    TBOX_ASSERT(rho_ins_idx >= 0);
+#endif
+    for (int ln = coarsest_ln; ln <= finest_ln; ++ln)
+    {
+        Pointer<PatchLevel<NDIM> > level = hierarchy->getPatchLevel(ln);
+        for (PatchLevel<NDIM>::Iterator p(level); p; p++)
+        {
+            Pointer<Patch<NDIM> > patch = level->getPatch(p());
```
boyceg: @amneetb, does this need to be here, or could it go in `setDataOnPatch()`?

examples/IBLevelSet/ex0/GravityForcing.h
```diff
@@ -0,0 +1,112 @@
...
+     * \brief Class constructor.
+     */
+    GravityForcing(const std::string& object_name,
+                   Pointer<INSVCStaggeredHierarchyIntegrator> ins_hierarchy_integrator,
+                   std::vector<double> grav_const);
+
+    /*!
+     * \brief Empty destructor.
+     */
+    ~GravityForcing();
```
boyceg:
```suggestion
    ~GravityForcing() = default;
```

examples/IBLevelSet/ex0/GravityForcing.h
```diff
@@ -0,0 +1,112 @@
...
+                        Pointer<hier::Variable<NDIM> > var,
+                        Pointer<Patch<NDIM> > patch,
+                        const double data_time,
+                        const bool initial_time = false,
+                        Pointer<PatchLevel<NDIM> > patch_level = Pointer<PatchLevel<NDIM> >(NULL));
+
+    //\}
+
+private:
+    GravityForcing();
```
boyceg:
```suggestion
    GravityForcing() = delete;
```

examples/IBLevelSet/ex0/GravityForcing.h
```diff
@@ -0,0 +1,112 @@
...
+
+    GravityForcing(const GravityForcing& from);
```
boyceg:
```suggestion
    GravityForcing(const GravityForcing& from) = delete;
```

examples/IBLevelSet/ex0/GravityForcing.h
```diff
@@ -0,0 +1,112 @@
...
+
+    GravityForcing& operator=(const GravityForcing& that);
```
boyceg:
```suggestion
    GravityForcing& operator=(const GravityForcing& that) = delete;
```

examples/IBLevelSet/ex0/SetFluidGasSolidViscosity.h
```diff
@@ -0,0 +1,129 @@
...
+ * Pre processing call back function to be hooked into IBAMR::VCINSStaggeredHierarchyIntegrator class.
+ *
+ * \param rho_idx a patch data index for the current density variable maintained by the integrator.
+ * \param ctx is the pointer to SetFluidGasSolidViscosity class object.
+ *
+ * \TODO: Let's move this out of the global namespace and use "snake case" for static function names.
+ */
+void callSetFluidGasSolidViscosityCallbackFunction(int mu_idx,
+                                                   SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > mu_var,
+                                                   SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps> hier_math_ops,
```
boyceg: @amneetb, is there a reason to prefer passing around the `HierarchyMathOps` object instead of the `PatchHierarchy`?

examples/IBLevelSet/ex0/SetFluidGasSolidDensity.cpp
```diff
@@ -0,0 +1,403 @@
...
+      d_ls_reinit_interval(ls_reinit_interval),
+      d_num_solid_interface_cells(num_solid_interface_cells),
+      d_num_gas_interface_cells(num_gas_interface_cells)
+{
+    // intentionally left blank
+    return;
+} // SetFluidGasSolidDensity
+
+void
+SetFluidGasSolidDensity::setDensityPatchData(int rho_idx,
```
boyceg: @amneetb, can we change this to `setDensityOnPatchHierarchy`?

examples/IBLevelSet/ex0/SetLSProperties.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+SetLSProperties::SetLSProperties(const std::string& object_name, Pointer<LSInitStrategy> ls_gas_ops)
+    : d_object_name(object_name), d_ls_gas_ops(ls_gas_ops)
+{
+    // intentionally left blank
+    return;
+} // SetLSProperties
+
+void
+SetLSProperties::setLSGasPatchData(int ls_gas_idx,
```
boyceg: @amneetb, can we change this to something like `setGasPhaseLSOnPatchHierarchy`?

boyceg: A follow-up question -- is there a reason that we cannot use the `CartGridFunction` interface to set these values?

examples/wave_tank/ex0/GravityForcing.cpp
```diff
@@ -0,0 +1,125 @@
...
+    const int coarsest_ln = (coarsest_ln_in == -1 ? 0 : coarsest_ln_in);
+    const int finest_ln = (finest_ln_in == -1 ? hierarchy->getFinestLevelNumber() : finest_ln_in);
+
+    // Get interpolated density variable
+    const int rho_ins_idx = d_ins_hierarchy_integrator->getLinearOperatorRhoPatchDataIndex();
+
+#if !defined(NDEBUG)
+    TBOX_ASSERT(rho_ins_idx >= 0);
+#endif
+    for (int ln = coarsest_ln; ln <= finest_ln; ++ln)
```
boyceg: @amneetb same question as above -- is there a reason that this needs to be in `setDataOnPatchHierarchy` instead of `setDataOnPatch`?

amneetb: The reason is we need ghost cell information for suitable interpolation to faces etc. By default CartGridFunction sends data with no ghost cell, and assumes the grid function is an explicit and a known function of (x,y,z,t)

examples/IBLevelSet/ex0/SetFluidGasSolidViscosity.h
amneetb: The preference is if one needs to find rho(phi, grad(phi)) etc, HierachyMathOps will have canned functions to do these ops.

examples/IBLevelSet/ex0/GravityForcing.cpp
amneetb: Since we need to fill ghost cells of level set, we do this on the entire hierarchy directly.

boyceg:
There is a good amount of fairly lengthy and nearly identical code in these examples. Is there core functionality that should be migrated out of the examples and into the core library?

amneetb:
I believe only regridProjection() being a public member is the only overlapping difference between NWT and master. I think it is safe to remove it. This is only going to affect IBLevelSet/ex0.

On a related note: can we add initial projection() before start of while loop integration internally. Currently, the projection happens when initial hierarchy is  initialized. But at that time we do not have solid level set info to set the right density to do a density-weighted projection.

boyceg:
1) Why does the LS example need to call regridProjection()?

2) There also are the new implementations of regridHierarchy() that need to be sorted out. Can you remind me what is not working correctly with regridHierarchy() in master with the multiphase solvers?

boyceg:
> I believe only regridProjection() being a public member is the only overlapping difference between NWT and master. I think it is safe to remove it. This is only going to affect IBLevelSet/ex0.

It looks like it only calls `regridProjection()` before the first time step. Should we just set up `INSHierarchyIntegrator()` to have the option to call `regridProjection()` before the first time step?

include/ibamr/WaveDampingStrategy.h
```diff
@@ -0,0 +1,119 @@
...
+ * The instance of a WaveDampingStrategy and the particular damping function
+ * should be registered with an appropriate hierarchy integrator via
+ * registerPostprocessIntegrateHierarchyCallback.
+ *
+ * \param ctx is the pointer to WaveDampingStrategy struct.
+ */
+
+namespace WaveDampingFunctions
+{
+void callRelaxationZoneCallbackFunction(double current_time,
```
boyceg: snake_case for static functions.

include/ibamr/WaveDampingStrategy.h
```diff
@@ -0,0 +1,119 @@
...
+    /*
+     * Pointers to the fluid and advection-diffusion integrators.
+     */
+    SAMRAI::tbox::Pointer<IBAMR::INSVCStaggeredHierarchyIntegrator> d_ins_hier_integrator;
+    SAMRAI::tbox::Pointer<IBAMR::AdvDiffHierarchyIntegrator> d_adv_diff_hier_integrator;
+
+    /*
+     * Pointer to the level set variable representing the wave interface.
+     */
+    SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > d_phi_var;
```
boyceg: Does this also need the context?

include/ibamr/WaveGenerationStrategy.h
```diff
@@ -0,0 +1,220 @@
...
+ * The instance of a WaveGenerationStrategy and the particular generation function
+ * should be registered with an appropriate hierarchy integrator via
+ * registerPostprocessIntegrateHierarchyCallback.
+ *
+ * \param ctx is the pointer to WaveGenerationStrategy struct.
+ */
+
+namespace WaveGenerationFunctions
+{
+void callStokesWaveRelaxationCallbackFunction(double current_time,
```
boyceg: snake_case for static functions.

include/ibamr/WaveGenerationStrategy.h
```diff
@@ -0,0 +1,220 @@
...
+    /*
+     * Pointers to the fluid and advection-diffusion integrators.
+     */
+    SAMRAI::tbox::Pointer<IBAMR::INSVCStaggeredHierarchyIntegrator> d_ins_hier_integrator;
+    SAMRAI::tbox::Pointer<IBAMR::AdvDiffHierarchyIntegrator> d_adv_diff_hier_integrator;
+
+    /*
+     * Pointer to the level set variable representing the wave interface.
+     */
+    SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > d_phi_var;
```
boyceg: Also needs the `VariableContext`?

include/ibamr/WaveGenerationStrategy.h
```diff
@@ -0,0 +1,220 @@
...
+                                              double new_time,
+                                              bool skip_synchronize_new_state_data,
+                                              int num_cycles,
+                                              void* ctx);
+} // namespace WaveGenerationFunctions
+
+/*! Struct for generating waves based on relaxation method.
+ *
+ */
+struct WaveGenerationStrategy
```
boyceg: Technically this is not a "strategy" class (it does not do anything). Possibly it should be an abstract base class, or a base class with default constructors and destructors.

@amneetb, how general is this meant to be? Is this something that really only works for Stokes waves, or are there other things that you could imaging using this for?

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -64,7 +64,38 @@ static const int GLEVELSETG = 1;
...
-} // namespace
+
+inline int
+sign(const double X)
+{
+    return ((X > 0) ? 1 : ((X < 0) ? -1 : 0));
+} // sign
+
+inline void
+get_physical_coordinate(IBTK::Vector3d& side_coord, Pointer<Patch<NDIM> > patch, const SideIndex<NDIM>& side_idx)
```
boyceg: @amneetb it looks like the same code is showing up in several different places. Should this go into, e.g., `IndexUtilities`?

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -0,0 +1,1303 @@
...
+        contains_vert0 = contains_vert0 && (c - dx <= vert0(d) && vert0(d) <= c + dx);
+        contains_vert1 = contains_vert1 && (c - dx <= vert1(d) && vert1(d) <= c + dx);
+        contains_vert2 = contains_vert2 && (c - dx <= vert1(d) && vert2(d) <= c + dx);
+    }
+    if (contains_vert0 && contains_vert1 && contains_vert2)
+    {
+        return true;
+    }
+
+    // Use the separating axis theorem to test the overlap between the tri element and the box
```
boyceg: @amneetb, is there a reference that can be cited here for the mathematical expressions?

src/navier_stokes/INSVCStaggeredConservativeHierarchyIntegrator.cpp
```diff
@@ -470,11 +458,11 @@ INSVCStaggeredConservativeHierarchyIntegrator::preprocessIntegrateHierarchy(cons
                                 current_time,
                                 d_mu_vc_interp_type);
 
+
```
boyceg:
```suggestion
```

examples/IBLevelSet/ex0/GravityForcing.cpp
boyceg: I think we should make it possible to have the `CartGridFunction` base class fill ghost cells (in another PR).

boyceg:
@drwells, can you update the copyright info in this branch?

include/ibamr/WaveGenerationStrategy.h
amneetb: It is algorithm dependent. I do this as a post-processing callback, and just use new_context directly.

include/ibamr/WaveGenerationStrategy.h
amneetb: We will add more wave generation from other theories like: cnoidal, irregular etc as we go along. This should be easily extendable.

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
amneetb: 👍

src/level_set/FESurfaceDistanceEvaluator.cpp
amneetb: I can add that.

include/ibamr/WaveGenerationStrategy.h
boyceg: OK. Should the virtual functions be moved into the base class (`WaveGenerationStrategy`)?

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
boyceg: Actually that method is already in `IndexUtilities` but with a fixed return type. I will template on the return type.

include/ibamr/WaveGenerationStrategy.h
boyceg: Should this class record the relevant contexts? Or if it should always be "new", could we put it in? I don't like the fact that users have to have extra knowledge about which context to use to actually access the data.

include/ibamr/WaveGenerationStrategy.h
boyceg: Hmm, looks like this is not done consistently in the library at this point. :-(

examples/IBLevelSet/ex0/SetFluidGasSolidViscosity.h
boyceg: I would prefer to use the `CartGridFunction` interface for setting grid data (and to improve that interface if needed).

amneetb:
> 1) Why does the LS example need to call regridProjection()? 

I think I have explained this in the past, but let me try it one more time. The multiphase flow solvers need a density weighted projection. Solid density is not known, unless we know the solid level set. To find solid level set, e.g., by FESurfaceDistance class, we need a full hierarchy first. The initial regidProjection() should be called just before time loop. Also the initial VisIt plotting should also reflect the initial solid level set found by FESurfaceDistance. If everything is done internally, I am not sure how the order can be enforced. The correct order should be :

1) Generate patch hierarchy.
2) Compute solid (+ liquid/gas) level set field.
3) Set density field using all the level set fields (+ some more user specific calculations).
4) Do initial projection.
5) Plot everything
6) Enter time while loop in main.cpp 

2) There also are the new implementations of regridHierarchy() that need to be sorted out. Can you remind me what is not working correctly with regridHierarchy() in master with the multiphase solvers?

I don't recall now. Here was my comment half a year ago

`Here are the files to review

    src/wave_tank/*.cpp
    src/IB/BrinkmanPenalization*.cpp
    src/IB/IB{LevelSet,Interpolant}Method.cpp
    src/IB/IBInterpolantHierarchyIntegrator.cpp
    src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
    src/level_set/FESurfaceDistanceEvaluator.cpp

There are associated .h files in the include directory.

======
Things changed in the IBAMR library (hardly 10 lines of code)

1]    nishant/regrid-hierarchy that I merged into NWT
https://github.com/IBAMR/IBAMR/commit/998405dffd43cc1c0e0679220ced7c61c9a08eeb


2]  Always regrid before integrating timestep # 0.   
 https://github.com/IBAMR/IBAMR/blob/amneetb/numerical_wave_tank/ibtk/src/utilities/HierarchyIntegrator.cpp#L1285-L1288

=== `

boyceg:
> > I believe only regridProjection() being a public member is the only overlapping difference between NWT and master. I think it is safe to remove it. This is only going to affect IBLevelSet/ex0.
> 
> It looks like it only calls `regridProjection()` before the first time step. Should we just set up `INSHierarchyIntegrator()` to have the option to call `regridProjection()` before the first time step?

In fact, that appears to be how `INSCollocatedHierarchyIntegrator`, `INSStaggeredHierarchyIntegrator`, and `INSVCStaggeredHierarchyIntegrator` are set up already.

amneetb:
> > 1. Why does the LS example need to call regridProjection()?
> 
> I think I have explained this in the past, but let me try it one more time. The multiphase flow solvers need a density weighted projection. Solid density is not known, unless we know the solid level set. To find solid level set, e.g., by FESurfaceDistance class, we need a full hierarchy first. The initial regidProjection() should be called just before time loop. Also the initial VisIt plotting should also reflect the initial solid level set found by FESurfaceDistance. If everything is done internally, I am not sure how the order can be enforced. The correct order should be :
> 
>     1. Generate patch hierarchy.
> 
>     2. Compute solid (+ liquid/gas) level set field.
> 
>     3. Set density field using all the level set fields (+ some more user specific calculations).
> 
>     4. Do initial projection.
> 
>     5. Plot everything
> 
>     6. Enter time while loop in main.cpp
> 
>     7. There also are the new implementations of regridHierarchy() that need to be sorted out. Can you remind me what is not working correctly with regridHierarchy() in master with the multiphase solvers?
> 
> 
> I don't recall now. Here was my comment half a year ago
> 
> `Here are the files to review
> 
> ```
> src/wave_tank/*.cpp
> src/IB/BrinkmanPenalization*.cpp
> src/IB/IB{LevelSet,Interpolant}Method.cpp
> src/IB/IBInterpolantHierarchyIntegrator.cpp
> src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
> src/level_set/FESurfaceDistanceEvaluator.cpp
> ```
> 
> There are associated .h files in the include directory.
> 
> ======
> Things changed in the IBAMR library (hardly 10 lines of code)
> 
> 1] nishant/regrid-hierarchy that I merged into NWT
> [998405d](https://github.com/IBAMR/IBAMR/commit/998405dffd43cc1c0e0679220ced7c61c9a08eeb)
> 
> 2] Always regrid before integrating timestep # 0.
> https://github.com/IBAMR/IBAMR/blob/amneetb/numerical_wave_tank/ibtk/src/utilities/HierarchyIntegrator.cpp#L1285-L1288
> 
> === `

include/ibamr/WaveGenerationStrategy.h
amneetb: We can, but I think not all wave theories can readily provide surfaceElevation() and/or velocity. Hence, I kept this to StokesWaveClass only. Stokes-[I-IV] can give provide this estimate easily. `WaveGenerationStrategy` is meant to hold required data structures that all of the wave generation strategies will need.

amneetb:
> > > I believe only regridProjection() being a public member is the only overlapping difference between NWT and master. I think it is safe to remove it. This is only going to affect IBLevelSet/ex0.
> > 
> > 
> > It looks like it only calls `regridProjection()` before the first time step. Should we just set up `INSHierarchyIntegrator()` to have the option to call `regridProjection()` before the first time step?
> 
> In fact, that appears to be how `INSCollocatedHierarchyIntegrator`, `INSStaggeredHierarchyIntegrator`, and `INSVCStaggeredHierarchyIntegrator` are set up already.

It is called right after ParentHI::initializePatchHierarchy(). The user does not get a chance to find solid LS and to set density field, so that correct density weighted projection could be applied.

include/ibamr/WaveGenerationStrategy.h
amneetb: It is not always "new". For example one can generate waves before the beginning of timestep instead of at the end. Former will use "current" context and latter will use new context. It is a timestepping choice. I just do it at the end of the timestep.

amneetb: One can also do both.

boyceg:
> > > > I believe only regridProjection() being a public member is the only overlapping difference between NWT and master. I think it is safe to remove it. This is only going to affect IBLevelSet/ex0.
> > > 
> > > 
> > > It looks like it only calls `regridProjection()` before the first time step. Should we just set up `INSHierarchyIntegrator()` to have the option to call `regridProjection()` before the first time step?
> > 
> > 
> > In fact, that appears to be how `INSCollocatedHierarchyIntegrator`, `INSStaggeredHierarchyIntegrator`, and `INSVCStaggeredHierarchyIntegrator` are set up already.
> 
> It is called right after ParentHI::initializePatchHierarchy(). The user does not get a chance to find solid LS and to set density field, so that correct density weighted projection could be applied.

What prevents this from being set up in `initializePatchHierarchy()`?

amneetb:
Sorry, I had accidentally closed it.

amneetb:
> > > > > I believe only regridProjection() being a public member is the only overlapping difference between NWT and master. I think it is safe to remove it. This is only going to affect IBLevelSet/ex0.
> > > > 
> > > > 
> > > > It looks like it only calls `regridProjection()` before the first time step. Should we just set up `INSHierarchyIntegrator()` to have the option to call `regridProjection()` before the first time step?
> > > 
> > > 
> > > In fact, that appears to be how `INSCollocatedHierarchyIntegrator`, `INSStaggeredHierarchyIntegrator`, and `INSVCStaggeredHierarchyIntegrator` are set up already.
> > 
> > 
> > It is called right after ParentHI::initializePatchHierarchy(). The user does not get a chance to find solid LS and to set density field, so that correct density weighted projection could be applied.
> 
> What prevents this from being set up in `initializePatchHierarchy()`?

Application specificity. Solid LS can be computed in multiple ways. More importantly, is the hierarchy fully complete and all other classes like IBFEMethod constructed such FESurfaceDistance class be be used at this point? To me doing everything at the same time is inevitably per-determining an order of setup. This is not flexible design, I think.

include/ibamr/WaveGenerationStrategy.h
boyceg: OK, well, at present `WaveGenerationStrategy` is not really a [strategy](https://sourcemaking.com/design_patterns/strategy) -- so it should get a new name.

amneetb: Agreed. Is there a design pattern name where the base class provides just data structures, rather than methods? I used `strategy` in a sense to hint that it not useful directly, as it provides no functionality.

include/ibamr/WaveGenerationStrategy.h
boyceg: I would suggest getting rid of this base class for now and merging the data members into the derived class.

Where are the functions that actually set the data? I think that they should tell the caller where to set the data values (either by passing in a patch data index or a (var,ctx) pair).

amneetb: Since it is a `struct` with public data members, I just set it in main.cpp directly. It is a lightweight structure utility that will be common for ALL wave theory implementations. Removing it will replicate these data members for all kinds of waves. Prefer doing that? Either way is fine by me.

include/ibamr/WaveGenerationStrategy.h
boyceg: Usually we want to use inheritance / class hierarchies to specify the interface in terms of a set of virtual functions. If there is not a common set of functions that will be implemented, my inclination is not to put the classes in the same hierarchy.

include/ibamr/WaveGenerationStrategy.h
boyceg: I would suggest to make a class like `WaveGenerationData` to store data members --- not to accomplish this through inheritance.

amneetb: Ah, yes, that sounds like the right design pattern.

include/ibamr/WaveDampingStrategy.h
amneetb: I will also rename this to `WaveDampingData`

amneetb: `callConservedWaveAbsorbingCallbackFunction()` uses both current and new to solve a nonlinear mass-balance equation. It really is an implementation dependent thing :-(

boyceg:
> > > > I believe only regridProjection() being a public member is the only overlapping difference between NWT and master. I think it is safe to remove it. This is only going to affect IBLevelSet/ex0.
> > > 
> > > 
> > > It looks like it only calls `regridProjection()` before the first time step. Should we just set up `INSHierarchyIntegrator()` to have the option to call `regridProjection()` before the first time step?
> > 
> > 
> > In fact, that appears to be how `INSCollocatedHierarchyIntegrator`, `INSStaggeredHierarchyIntegrator`, and `INSVCStaggeredHierarchyIntegrator` are set up already.
> 
> It is called right after ParentHI::initializePatchHierarchy(). The user does not get a chance to find solid LS and to set density field, so that correct density weighted projection could be applied.

It looks like right now these regrid projections happen in `initializePatchHierarchy()`.

What about doing the regrid projection in `initializeCompositeHierarchyData()`/`initializeCompositeHierarchyDataSpecialized()`?

Note that `initializeCompositeHierarchyData()` is a public member function, and so it should be easy for users to call it in the appropriate order. The main thing that we would presumably want to do is to make sure that it is "safe" to call multiple times.

include/ibamr/WaveDampingStrategy.h
boyceg: But the class that calls those functions presumably knows where it wants the results stored? This is what I want to make sure is codified in the API: where do the results go?

boyceg: (And similarly knows when it wants the results computed?)

amneetb: I see. In practice, I do it as postprocessing and it works great. I will just add new_context pointer.

amneetb:
> initializeCompositeHierarchyData()

Aha! This also sounds like the right name --- since regridProjection op is for a composite grid.

boyceg:
`regridHierarchy()` is also public, and so a user code could call it before the initial viz dump to ensure that the initial configuration in the viz file really is the one used in the initial time step.

amneetb:
Hmmn, sounds good. The current NWT branch makes sure of this here.
https://github.com/IBAMR/IBAMR/blob/amneetb/numerical_wave_tank/ibtk/src/utilities/HierarchyIntegrator.cpp#L1285-L1288

Interestingly, IBHierarchyIntegrator does this by overriding HI::regridHierarchy, but this should be done for all HI's including INS too. I noticed this anomaly when I was doing 2-phase flows and tagged cells were not doing any finer level generation in the beginning of the simulation.

boyceg:
> Hmmn, sounds good. The current NWT branch makes sure of this here.
> https://github.com/IBAMR/IBAMR/blob/amneetb/numerical_wave_tank/ibtk/src/utilities/HierarchyIntegrator.cpp#L1285-L1288

Yes, a version of this is in this branch also.

> Interestingly, IBHierarchyIntegrator does this by overriding HI::regridHierarchy, but this should be done for all HI's including INS too. I noticed this anomaly when I was doing 2-phase flows and tagged cells were not doing any finer level generation in the beginning of the simulation.

`IBHierarchyIntegrator` overrides `regridHierarchyBeginSpecialized()` and `regridHierarchyEndSpecialized()`. The intention is that derived classes will implement those functions, and will not reimplement `regridHierarchy()` itself.

I think that the tagging issue is the reason why you reimplement `regridHierarchy()` in each of the `HierarchyIntegrator` implementations in your branch. Does that sound right? We _should_ be able to fix that in the `HierarchyIntegrator` base class.

amneetb:
> > Hmmn, sounds good. The current NWT branch makes sure of this here.
> > https://github.com/IBAMR/IBAMR/blob/amneetb/numerical_wave_tank/ibtk/src/utilities/HierarchyIntegrator.cpp#L1285-L1288
> 
> Yes, a version of this is in this branch also.
> 
> > Interestingly, IBHierarchyIntegrator does this by overriding HI::regridHierarchy, but this should be done for all HI's including INS too. I noticed this anomaly when I was doing 2-phase flows and tagged cells were not doing any finer level generation in the beginning of the simulation.
> 
> `IBHierarchyIntegrator` overrides `regridHierarchyBeginSpecialized()` and `regridHierarchyEndSpecialized()`. The intention is that derived classes will implement those functions, and will not reimplement `regridHierarchy()` itself.
> 
> I think that the tagging issue is the reason why you reimplement `regridHierarchy()` in each of the `HierarchyIntegrator` implementations in your branch. Does that sound right? We _should_ be able to fix that in the `HierarchyIntegrator` base class.

I believe so. To be 100% sure please take a look at what specifically/exactly we did in these commits:
1] nishant/regrid-hierarchy that I merged into NWT
998405d

2] Always regrid before integrating timestep # 0.
https://github.com/IBAMR/IBAMR/blob/amneetb/numerical_wave_tank/ibtk/src/utilities/HierarchyIntegrator.cpp#L1285-L1288 

These are the only changes we made for tagging to work properly. I also recall  David's split regridHierarchyBeginSpecialized() and regridHierarchyEndSpecialized() had achieved what nishant/regrid-hierarchy aimed to achieve. Can you please confirm that?

boyceg:
OK, please take a look at the branch and see if it appears to work. If, I will work on merging once the other changes discussed above are done.

amneetb:
Thanks, I will make changes and do some tests in the coming days. I will leave setLSFLuidGassSolidProperties as is. I realize that in the current master we also have some code (but just one implementation and all tests use the same code). So, I do agree that we should have a core library implementation of this function, perhaps in the src/level_set directory. We can change that in a separate PR though. That will simulateously change examples/multiphase_flow and tests/multiphase_flow.

amneetb:
@boyceg: The branch is working. I updated and tested IBLevelSet/ex0 and the cylinder fall velocity is close to analytical solution. This is a very coarse grid example.  

Comparison

Please check the update for the example. I am working on other suggestions now.

amneetb:
For some reason my NWT branch answer is little off. This branch is giving more accurate result ;-). Mysterious are the ways in which Lord works!

amneetb:
@boyceg: Everything except moving setFluidProperties code remains.

boyceg:
> For some reason my NWT branch answer is little off. This branch is giving more accurate result ;-). Mysterious are the ways in which Lord works!

Could be wrong, but my guess is that this is related to `projectVelocity()`. When this is called inside the class, there is extra synchronization that is performed that was not being done when it was called from outside the class.

include/ibamr/WaveGenerationStrategy.h
boyceg: Looks good. Can you split into two headers --- one for `WaveGenerationData` and the other for `StokesWaveGenerator`?

If you wanted to indicate that the base class is an interface and not a concrete class, you could consider renaming `StokesWaveGenerator` to `StokesWaveGeneratorStrategy` since it is indeed specifying a generic interface for Stokes waves. OTOH, the current name is also fine.

include/ibamr/WaveDampingStrategy.h
```diff
@@ -0,0 +1,125 @@
...
+    /*
+     * Start and end coordinates of the damping zone, water depth, and damping coefficient.
+     */
+    double d_x_zone_start, d_x_zone_end, d_depth;
+    double d_alpha;
+    int d_sign_gas_phase = 1;
+};
+
+/*!
+ * A struct holding the required information used by the variable alpha damping method
```
boyceg: Can you add a brief comment saying what `operator()` does here?

Also, is there a reason to prefer functors over plain-old-functions + `Data` object? It seems like most things are being specified through static functions + `ctx`.

(I am OK with functors but also prefer that users interact with the classes in a consistent way when possible.)

include/ibamr/WaveGenerationStrategy.h
```diff
@@ -0,0 +1,233 @@
...
+     *  \brief Stokes wave generator data members.
+     */
+    /*!
+     * Get wave parameters from input db.
+     */
+    void getFromInput(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db);
+
+}; // StokesWaveGenerator
+
+class FifthOrderStokesWaveGenerator : public StokesWaveGenerator
```
boyceg: Usually we would have each of these classes in their own headers, so that users do not have to guess which header provides which classes. Can we split out the Stokes wave generator?

include/ibamr/WaveDampingStrategy.h
amneetb: We are using a functor as required by boost's Newton-Ralphson root finding interface
https://www.boost.org/doc/libs/1_62_0/libs/math/doc/html/math_toolkit/roots/roots_deriv.html

include/ibamr/WaveGenerationStrategy.h
amneetb: I think we put all Stokes wave classes (1st, 2nd and 5th) in a single header. Other kinds of waves will be in different header. Does that sound good?

include/ibamr/WaveDampingStrategy.h
boyceg: Got it. Does this need to be part of the "public" interface.

include/ibamr/WaveGenerationStrategy.h
boyceg: I think it is better to split them out into individual headers. But, sounds OK.

include/ibamr/WaveDampingStrategy.h
amneetb: I don't think so. I think we can move this `struct` to anonymous namespace of `WaveDampingStrategy.cpp` file. Do you agree?

boyceg: Sounds great!

include/ibamr/WaveGenerationStrategy.h
boyceg: Upon further reflection ... I would prefer to keep with convention that classes that are intended to be used by users go into their own headers.

boyceg:
@amneetb are there any other pending changes?

amneetb:
I think I did all the suggested changes.

boyceg:
OK. I will try to take another look over the weekend.

My plan is to make a new branch that has a cleaned-up history and to merge that new branch.

@drwells, would you be willing to apply your copyright fixes either directly to this branch, or to make a PR to merge those changes into this branch?

amneetb:
Something is not right with this branch. Some of my application codes (which are not on github) are not working correctly. The coding is running but the WEC is not doing any dynamics even when there are incoming waves...

amneetb:
I reverted as back as https://github.com/IBAMR/IBAMR/pull/729/commits/9d3e2bffceacf302104cd7d5fc3b410ea55a9c98, but it did not fix the problem. Sigh.

boyceg:
OK -- will need to figure out where things diverged ...

drwells:
Doing the copyright update will be really easy now that I have scripts for it: I think it would be better to hold off on that until right before we merge.

I have another patch for this PR that updates test output files since we now print an additional logging message.

drwells:
> Something is not right with this branch. Some of my application codes (which are not on github) are not working correctly.

Once we get everything working again we need to convert some of these programs into tests so that these stay fixed!

amneetb:
Ok, I tried going back to this branch before master was merged till this commit https://github.com/IBAMR/IBAMR/commit/be771f42984c57bbff1c66d8c4031a8cce13562b. However, there are many files missing like `FESurfaceDistanceEvaluator` and `IBInterpolantHierarchyIntegrator`. So I can not test what particular commit is causing problem.

amneetb:
ok, the merge is messed up. Right off the bat I see two BUGS

Comparing IBHydrodynamicSurfaceForceEvaluator in my branch and in boyceg branch:

++ ===> Indicate buggy code added in boyceg branch. 

`
@@ -364,6 +345,8 @@ IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForceTorque(IBTK::Vector
                     pressure_force += (-pn * dS);
                     pressure_torque += (r_vec.cross(-pn)) * dS;
 
+                    pressure_torque += r_vec.cross(-pn) * dS;
+

[This adds pressure torque twice]
                     // Add up the viscous forces n.(mu*(grad U + grad U)^T)dS

`

2) 

                     // Get the relative coordinate from X0
-                    get_physical_coordinate(r_vec, patch, s_i);
+                    const IBTK::VectorNd side_center = IBTK::IndexUtilities::getSideCenter(*patch, s_i);
+                    IBTK::Vector3d r_vec = IBTK::Vector3d::Zero();
+                    {
+                        r_vec[d] = side_center[d] - X0[d];
+                    }
+
+                    // Get the relative coordinate from X0
+                    r_vec = IBTK::IndexUtilities::getSideCenter<IBTK::Vector3d>(*patch, s_i);
                     r_vec -= X0;

amneetb:
Third BUG

+    SAMRAI_MPI::sumReduction(pressure_force.data(), NDIM);
+    SAMRAI_MPI::sumReduction(viscous_force.data(), NDIM);
+    SAMRAI_MPI::sumReduction(pressure_torque.data(), NDIM);
+    SAMRAI_MPI::sumReduction(viscous_torque.data(), NDIM);

The torque vectors have three components even in 2D. The torque never gets added in z direction for 2D rotation. 

I recall David did the same thing before by converting my Vector3d code to VectorNd code.

drwells:
Yup, the NDIM mistake is in #607. Let me put up a PR for that.

I am perplexed that the tests pass but those values are wrong.

boyceg:
Are these all things that are fixed in master that are undone in this branch?

In working on this, I took Amneet’s NWT branch to be the gold standard, and so for conflicts, I would usually use what was in NWT — or at least that was what I intended to do.

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -291,16 +292,7 @@ IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForceTorque(IBTK::Vector
...
-                    IBTK::Vector3d r_vec = IBTK::Vector3d::Zero();
-                    for (unsigned int d = 0; d < NDIM; ++d)
-                    {
-                        r_vec[d] = side_center[d] - X0[d];
-                    }
-
-                    // Get the relative coordinate from X0
-                    r_vec = IBTK::IndexUtilities::getSideCenter<IBTK::Vector3d>(*patch, s_i);
-                    r_vec -= X0;
+                    const auto r_vec = IBTK::IndexUtilities::getSideCenter<IBTK::Vector3d>(*patch, s_i) - X0;
```
amneetb: Would `IBTK::IndexUtilities::getSideCenter<IBTK::Vector3d>(*patch, s_i)` guarantee that r_vec[2] = 0.0 when `NDIM == 2`? This is necessary for the cross product to work properly in 2D.

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -356,10 +346,10 @@ IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForceTorque(IBTK::Vector
...
     }
     // Sum the net force and torque across processors.
-    SAMRAI_MPI::sumReduction(pressure_force.data(), NDIM);
-    SAMRAI_MPI::sumReduction(viscous_force.data(), NDIM);
-    SAMRAI_MPI::sumReduction(pressure_torque.data(), NDIM);
-    SAMRAI_MPI::sumReduction(viscous_torque.data(), NDIM);
+    SAMRAI_MPI::sumReduction(pressure_force.data(), 3);
+    SAMRAI_MPI::sumReduction(viscous_force.data(), 3);
+    SAMRAI_MPI::sumReduction(pressure_torque.data(), 3);
+    SAMRAI_MPI::sumReduction(viscous_torque.data(), 3);
```
amneetb: This was last one. Looks good.

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
```diff
@@ -345,8 +337,6 @@ IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForceTorque(IBTK::Vector
                     pressure_force += (-pn * dS);
                     pressure_torque += (r_vec.cross(-pn)) * dS;
 
-                    pressure_torque += r_vec.cross(-pn) * dS;
-
                     // Add up the viscous forces n.(mu*(grad U + grad U)^T)dS
```
amneetb: Looks good.

src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
boyceg: It didn't, but now it does.

boyceg:
I am going to delete this branch. If we need to get it back, it was f871fea82007905b53d690f16deb211a97db8a06.

PR #730, opened 2019-10-26
open and close files each time we write meter data

cpuelz:


src/IB/IBFEInstrumentPanel.cpp
```diff
@@ -1110,8 +1110,11 @@ IBFEInstrumentPanel::getMeterRadius(const int meter_mesh_number)
 void
 IBFEInstrumentPanel::outputData(const double data_time)
 {
-    if (SAMRAI_MPI::getRank() == 0)
+    static const int mpi_root = 0;
+    if (SAMRAI_MPI::getRank() == mpi_root)
```
drwells: It seems weird that we have the constant in one place and not the other. I would just revert this to be `== 0`.

boyceg: I have a slight preference for naming the constant in these cases, but can go either way.

cpuelz: I deferred to Boyce's preference... sorry David :-).

drwells: My first preference is consistency, so I will still call it a win :)

src/IB/IBFEInstrumentPanel.cpp
```diff
@@ -317,10 +317,11 @@ IBFEInstrumentPanel::IBFEInstrumentPanel(SAMRAI::tbox::Pointer<SAMRAI::tbox::Dat
...
 
     // set up file streams
-    if (SAMRAI_MPI::getRank() == 0)
+    static const int mpi_root = 0;
+    if (SAMRAI_MPI::getRank() == mpi_root)
     {
-        d_mean_pressure_stream.open(d_plot_directory_name + "/mean_pressure.dat");
-        d_flux_stream.open(d_plot_directory_name + "/flux.dat");
+        d_mean_pressure_stream.open(d_plot_directory_name + "/mean_pressure.dat", std::ofstream::out);
+        d_flux_stream.open(d_plot_directory_name + "/flux.dat", std::ofstream::out);
```
drwells: I don't think this is right: this code will now open the streams here and then open them a second time (without closing) below.

I believe the way to fix this is to just get rid of these calls to open at this point.

src/IB/IBFEInstrumentPanel.cpp
```diff
@@ -1110,8 +1101,13 @@ IBFEInstrumentPanel::getMeterRadius(const int meter_mesh_number)
 void
 IBFEInstrumentPanel::outputData(const double data_time)
 {
-    if (SAMRAI_MPI::getRank() == 0)
+    static const int mpi_root = 0;
+    if (SAMRAI_MPI::getRank() == mpi_root)
     {
+        d_mean_pressure_stream.open(d_plot_directory_name + "/mean_pressure.dat", std::ofstream::app);
+        d_flux_stream.open(d_plot_directory_name + "/flux.dat", std::ofstream::app);
+        d_mean_pressure_stream.precision(15)
```
drwells: @cpuelz you forgot a semicolon. Please add it and squash.

cpuelz:
@drwells did I correctly rebase?

drwells:
It looks like Jenkins is happy now so lets merge.

PR #733, opened 2019-10-30
reuse SAMRAIDataCache objects between FEDataManager instances

boyceg:
Allow callers to share one (or more) `SAMRAIDataCache` objects between multiple `FEDataManager`s.

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -421,6 +421,7 @@ public:
                const SpreadSpec& default_spread_spec,
                const WorkloadSpec& default_workload_spec,
                const SAMRAI::hier::IntVector<NDIM>& min_ghost_width = SAMRAI::hier::IntVector<NDIM>(0),
+               std::shared_ptr<SAMRAIDataCache> eulerian_data_cache = nullptr,
                bool register_for_restart = true);
 
     /*!
@@ -441,6 +442,7 @@ public:
                const SpreadSpec& default_spread_spec,
                const WorkloadSpec& default_workload_spec,
                const SAMRAI::hier::IntVector<NDIM>& min_ghost_width = SAMRAI::hier::IntVector<NDIM>(0),
+               std::shared_ptr<SAMRAIDataCache> eulerian_data_cache = nullptr,
                bool register_for_restart = true);
 
     /*!
@@ -961,6 +963,7 @@ protected:
                   SpreadSpec default_spread_spec,
                   WorkloadSpec default_workload_spec,
                   SAMRAI::hier::IntVector<NDIM> ghost_width,
+                  std::shared_ptr<SAMRAIDataCache> eulerian_data_cache,
                   bool register_for_restart = true);
 
     /*!
@@ -973,6 +976,7 @@ protected:
                   SpreadSpec default_spread_spec,
                   WorkloadSpec default_workload_spec,
                   SAMRAI::hier::IntVector<NDIM> ghost_width,
+                  std::shared_ptr<SAMRAIDataCache> eulerian_data_cache,
                   bool register_for_restart = true);
 
     /*!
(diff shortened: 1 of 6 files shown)


boyceg:
Addresses #733.

drwells:
We might want to just break the signature of that function: I really doubt its being called in user codes.

I want to wrap up some work on #654 before reviewing this carefully.

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -454,6 +456,7 @@ class FEDataManager : public SAMRAI::tbox::Serializable, public SAMRAI::mesh::St
                const InterpSpec& default_interp_spec,
                const SpreadSpec& default_spread_spec,
                const SAMRAI::hier::IntVector<NDIM>& min_ghost_width = SAMRAI::hier::IntVector<NDIM>(0),
+               std::shared_ptr<SAMRAIDataCache> eulerian_data_cache = nullptr,
                bool register_for_restart = true);
```
drwells: This is the backwards compatible overload; I think that we should keep this one as-is and add the argument to the two newer ones instead.

drwells: The two other overloads are the ones that IBFEMethod uses anyway, not this one.

tests/IBFE/explicit_ex4.cpp
```diff
@@ -366,11 +369,15 @@ main(int argc, char** argv)
...
+            IntVector<NDIM> min_ghost_width = 0;
+            std::shared_ptr<IBTK::SAMRAIDataCache> eulerian_data_cache(new IBTK::SAMRAIDataCache());
             other_manager = FEDataManager::getManager(fe_data_manager->getFEData(),
                                                       "cloned_fe_data_manager",
                                                       fe_data_manager->getDefaultInterpSpec(),
                                                       fe_data_manager->getDefaultSpreadSpec(),
-                                                      spec);
+                                                      spec,
+                                                      min_ghost_width,
+                                                      eulerian_data_cache);
```
drwells: Is modifying this test necessary? Would it work in its current form?

src/IB/IBFEMethod.cpp
```diff
@@ -1143,19 +1143,28 @@ IBFEMethod::initializeFEEquationSystems()
...
         {
+            if (!d_scratch_eulerian_data_cache) d_scratch_eulerian_data_cache.reset(new SAMRAIDataCache());
             d_scratch_fe_data_managers[part] = FEDataManager::getManager(d_primary_fe_data_managers[part]->getFEData(),
                                                                          manager_name + "::scratch",
                                                                          d_interp_spec[part],
                                                                          d_spread_spec[part],
-                                                                         d_workload_spec[part]);
+                                                                         d_workload_spec[part],
+                                                                         min_ghost_width,
+                                                                         d_scratch_eulerian_data_cache);
```
drwells: :heavy_check_mark:

include/ibamr/IBFEMethod.h
```diff
@@ -781,6 +782,11 @@ class IBFEMethod : public IBStrategy
     SAMRAI::tbox::Pointer<SAMRAI::mesh::GriddingAlgorithm<NDIM> > d_gridding_alg;
     bool d_is_initialized = false;
 
+    /*
+     * Scratch data caching objects.
```
drwells: Could you add some more comments explaining that these are used by different sets of `FEDataManager`s?

tests/IBFE/explicit_ex4.cpp
boyceg: Nope, not necessary. I will revert.

boyceg: It is not a necessary change. I will revert.

include/ibamr/IBFEMethod.h
boyceg: Sure.

PR #735, opened 2019-10-31
Remove CIBFEMethod

### LINKED ISSUE #721: CIBFEMethod doesn't work
I tried to write a test for CIBFEMethod based on `examples/CIBFE/ex0/example.cpp` but then realized that this class has some serious problems: there will always be a segmentation fault in the constructor since it tries to add a system to the `EquationSystems` object which is not set up until we actually call `initializeFEEquationSystems`.

I suspect that this is already known and is why we did #193. We should either fix this class or remove it.

drwells:
Fixes #721. I'll put up a branch that reverts this commit once this is merged so that its clear what the state of the implementation is.


drwells:
It looks like our copyright fix made this totally out of date. @amneetb are you okay with deleting this and then putting the code on a new branch on GitHub? If so I'll go ahead and do it :)

amneetb:
Yes, go ahead. You can also put `CIBStochasticMethod` (#523) there.

drwells:
Sounds good. I fixed the merge conflicts. Once this is merged I will put up a new branch `add-CIBFEMethod` that reverts the relevant commit.

drwells:
This conflicts with #772 - lets merge that first.

drwells:
The branch is
https://github.com/IBAMR/IBAMR/tree/add-CIBFEMethod

which just reverts this PR.

PR #736, opened 2019-11-04
Convert ibfe ex0 test

drwells:
I noticed on Friday that the gtest test no longer worked. Fortunately, this is because the fix in #552 improves the accuracy and lowers the error slightly in the velocity and pressure.

Part of #569.


drwells:
This PR just adds a test and doesn't change the library so I will merge as-is.

Issue #604, opened 2019-05-30
Clean up copyright notices


drwells:
We mark authorship and years somewhat inconsistently in headers. We should add a script that cleans up the headers and puts in the correct years.

Furthermore, though I am not an expert on the subject, the more common route for attribution is to have the header text simply state 'copyright (C) 2000 - 2019 by the IBAMR authors' and not list individual contributors. I think explicitly assigning ourselves copyright can avoid some legal problems. I am not sure how much of this UNC owns too.

boyceg:
> We mark authorship and years somewhat inconsistently in headers. We should add a script that cleans up the headers and puts in the correct years.

Yes.

> Furthermore, though I am not an expert on the subject, the more common route for attribution is to have the header text simply state 'copyright (C) 2000 - 2019 by the IBAMR authors' and not list individual contributors.

Yes.

> I am not sure how much of this UNC owns too.

That is a question that I would rather not delve into. :-D

drwells:
At other universities I have had to fill out intellectual property forms yearly. I won't tell if you don't :speak_no_evil:

drwells:
On a related note: we should include a LICENSE file, which we do not currently have. I am *pretty sure* that we stated in a funding source that we should release under the LGPL. Is this correct?

boyceg:
I think we use the [3-clause BSD license](https://opensource.org/licenses/BSD-3-Clause) --- see the top of any source code file in the library.

drwells:
I thought that looked familiar. Thanks.

drwells:
This was partially addressed by #646.

I spent some time this morning reading up on our copyright situation. AFAICT IBAMR is a non-directed work and therefore UNC does not make any claim to copyright: see

https://unc.policystat.com/policy/6197184/latest/
and
https://policy.uncg.edu/university-policies/copyright/
the second one is not UNC-CH, but our policy does not explicitly define some terms. I assume that there are no significant differences in IP management between different UNCs.

It is not clear to me if NYU claims copyright over older versions of IBAMR: see
https://www.nyu.edu/content/dam/nyu/compliance/documents/IPPolicy.pdf
see page 18 (section XI, subsection A1) and page 19 (section XI, subsection B1). I am pretty sure that the answer is 'no' but it depends on how @boyceg and others were funded at that time (and upon OIL, NYU's relevant legal team).

[...]

PR #738, opened 2019-11-05
Source copyright

drwells:
First half of #604: I'll put up a followup PR later that implements the same fix for automake etc. files.


boyceg:
Is there a script to do this stuff? I would like to use it to update copyright in the pending NWT branch.

drwells:
I wrote a python script for it but it currently isn't smart enough to not try and 'fix' files whose copyright statement has already been updated. I'll post a fixed version here in a bit.

drwells:
Try this for source and header files:
```python
#!/usr/bin/env python2

import os
import subprocess

def main():
    top_directories = ["./ibtk/include", "./ibtk/src", "./ibtk/examples",
                       "./src", "./include", "./examples", "./tests"]
    files = list()
    for top_directory in top_directories:
        for root, _, file_names in os.walk(top_directory):
            for file_name in file_names:
                if file_name.endswith('.h') or file_name.endswith('.cpp'):
                    files.append(root + os.sep + file_name)

    for file_name in files:
        contents = list()
        found_end_of_copyright = False
        file_already_updated = False
        with open(file_name, 'r') as file_handle:
            for line in file_handle.readlines():
                if found_end_of_copyright:
                    contents.append(line)
                else:
                    if "SUCH DAMAGE." in line:
                        found_end_of_copyright = True
                    elif "by the IBAMR developers" in line:
                        file_already_updated = True
                        break

        if not file_already_updated:
            # output is 2019\n2018\n etc.
            years = subprocess.check_output(
                ["git", "log", "--format=format:%cd", "--date=format:%Y",
                file_name])
            # TODO: this string handling is probably not compatible with python3
            copyright_start = str(years[-4:])
            copyright_end = str(years[:4])

            message = \
"""// ---------------------------------------------------------------------
//
// Copyright (c) {} - {} by the IBAMR developers
// All rights reserved.
//
// This file is part of IBAMR.
//
// IBAMR is free software and is distributed under the 3-clause BSD
// license. The full text of the license can be found in the file
// COPYRIGHT at the top level directory of IBAMR.
//
""".format(copyright_start, copyright_end)

            # some files don't have a copyright statement at all, so dump everything
            # into their contents
            if not found_end_of_copyright:
                with open(file_name, 'r') as file_handle:
                    for line in file_handle.readlines():
                        contents.append(line)

            if contents[0] != "\n":
                contents = ["\n"] + contents

            with open(file_name, 'w') as file_handle:
                file_handle.write(message)

                for line in contents:
                    file_handle.write(line)

if __name__ == '__main__':
    main()
```

drwells:
Lets hold off on merging this until #744 is done: we don't need multiple commits for updating copyrights.

drwells:
#744 is in so lets merge this now.

There will always be a few files which are out of sync (at least until we get all active branches merged) so I also added 'rerun the copyright script' to the release tasks document.

PR #739, opened 2019-11-05
Fix other copyright

drwells:
Second half of #604.


drwells:
Lets hold off on merging this until #744 is done: we don't need multiple commits for updating automake copyrights.

drwells:
This has been updated to account for changes on NWT and also fix copyright with matlab scripts.

PR #740, opened 2019-11-05
Fix test output.

drwells:
1. Use the linear partitioner to get stable partitionings
2. avoid printing the input database


drwells:
Since this just relates to tests I will merge as-is.

PR #741, opened 2019-11-05
Avoid calculating an end iterator in a loop.

drwells:
This is expensive with libMesh: it currently occupies about half the run time of IBTK::get_global_active_element_bounding_boxes.

I checked and we do this correctly everywhere else.

### DIFF
--- a/ibtk/src/utilities/libmesh_utilities.cpp
+++ b/ibtk/src/utilities/libmesh_utilities.cpp
@@ -241,9 +241,11 @@ get_local_active_element_bounding_boxes(const libMesh::MeshBase& mesh, const lib
 
     std::vector<libMesh::dof_id_type> dof_indices;
     std::vector<double> X_node;
-    for (auto it = mesh.active_local_elements_begin(); it != mesh.active_local_elements_end(); ++it)
+    const auto el_begin = mesh.active_local_elements_begin();
+    const auto el_end = mesh.active_local_elements_end();
+    for (auto el_it = el_begin; el_it != el_end; ++el_it)
     {
-        const libMesh::Elem* const elem = *it;
+        const libMesh::Elem* const elem = *el_it;
         const unsigned int n_nodes = elem->n_nodes();
         bboxes.emplace_back();
         libMesh::BoundingBox& box = bboxes.back();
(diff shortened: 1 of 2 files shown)


ibtk/src/utilities/libmesh_utilities.cpp
```diff
@@ -241,7 +241,8 @@ get_local_active_element_bounding_boxes(const libMesh::MeshBase& mesh, const lib
 
     std::vector<libMesh::dof_id_type> dof_indices;
     std::vector<double> X_node;
-    for (auto it = mesh.active_local_elements_begin(); it != mesh.active_local_elements_end(); ++it)
+    const auto endit = mesh.active_local_elements_end();
```
boyceg:
```suggestion
    auto it = mesh.active_local_elements_begin();
    const auto endit = mesh.active_local_elements_end();
```

ibtk/src/utilities/libmesh_utilities.cpp
```diff
@@ -241,7 +241,8 @@ get_local_active_element_bounding_boxes(const libMesh::MeshBase& mesh, const lib
...
+    for (auto it = mesh.active_local_elements_begin(); it != endit; ++it)
```
boyceg:
```suggestion
    for (; it != endit; ++it)
```

boyceg: Take it or leave it, but I like also to extract the begin iterator.

ibtk/src/utilities/libmesh_utilities.cpp
```diff
@@ -295,7 +296,8 @@ get_global_active_element_bounding_boxes(const libMesh::MeshBase& mesh, const li
     const std::size_t n_elem = mesh.n_active_elem();
     std::vector<double> flattened_bboxes(2 * LIBMESH_DIM * n_elem);
     std::size_t elem_n = 0;
-    for (auto it = mesh.active_local_elements_begin(); it != mesh.active_local_elements_end(); ++it)
+    const auto endit = mesh.active_local_elements_end();
+    for (auto it = mesh.active_local_elements_begin(); it != endit; ++it)
```
boyceg: Same as above, FWIW.

drwells:
I unified the style with that used in FEDataManager and IBFEMethod.

drwells:
Since @boyceg has already approved and the tests pass I will merge.

PR #742, opened 2019-11-08
Use array sizes for reductions.

drwells:
This avoids some confusion where we are using 3D arrays in 2D; i.e., `NDIM` is wrong. I made this mistake back in #607.


drwells:
The tests added in #607 pass with or without this test; i.e., it didn't matter that the torque was totally wrong. @amneetb do we have more test code?

PR #743, opened 2019-11-10
NWT-final

boyceg:
A version of #729 that cleans up the history.


boyceg:
Hmmm, it looks like I did something to mess up history. I am going to close this, get rid of the "clean" branch, and try again.

PR #744, opened 2019-11-10
NWT-final

boyceg:
Version of #729 that cleans up the history.


boyceg:
@drwells, @amneetb, I think that this is ready to merge. Please take a look when you get a chance. I think we all would love to get this done. :-)

drwells:
Thanks a lot for cleaning up the history. I am nervous about the lack of unit tests but the honest answer is that if no one has managed to write them this year then it simply isn't going to get done.

I would still prefer to merge #742 before this.

drwells:
Quick comments:
1. We still have the namespace conflict between recent versions of Eigen and SAMRAI: we should always write `hier::Index<NDIM>`. This is in WaveGenerationFunctions.cpp.
2. `FifthOrderStokesWaveGenerator.h` has extra tokens after its `#endif`
3. The changes to regridding break all tests that depend on logfiles since there is an extra line now printed. I have a patch that fixes this.

@boyceg do you mind if I amend these changes onto your commits and force push?

boyceg:
> @boyceg do you mind if I amend these changes onto your commits and force push?

Not at all. Please make any changes you'd like to make.

amneetb:
Thanks guys! Once you push it, I will do a final test run on my ISWEC application and check for its accuracy. We should then merge it and have a little celebration 😉

boyceg:
@amneetb Can you make a unit test or two so that we do not immediately break it? :-D

drwells:
This PR is still pretty big. I am putting all the independent commits in other PRs to speed this up.

amneetb:
I need to think about it. So-far I have hefty codes using this branch.

boyceg:
> I need to think about it. So-far I have hefty codes using this branch.

I think it is safe to say that we have hefty codes using all of the stuff that is currently under active development, but we still want to have tests so that coding errors are less likely to make it into master.

amneetb:
> > I need to think about it. So-far I have hefty codes using this branch.
> 
> I think it is safe to say that we have hefty codes using all of the stuff that is currently under active development, but we still want to have tests so that coding errors are less likely to make it into master.

I now have a test in my mind that would not involve committing FE mesh as a part of the test. I will add it in the coming days. Right now I am re-running ISWEC app with this branch.

amneetb:
The results are quite different with this branch and are definitely wrong

NWT-final:
``
0.000500000000  7.806500000000  0.621500000000  0.000000000000  0.000000000000  0.000000000000  -0.000350524766 0.000000000000
0.001000000000  7.806500000000  0.621500000000  -0.000000363120 0.000000000000  0.000000000000  -0.000746352030 -0.751430894115
0.001500000000  7.806500000000  0.621500000000  -0.000000934922 0.000000000000  0.000000000000  -0.001159265220 -0.794504109593
0.002000000000  7.806500000000  0.621500000000  -0.000001721558 0.000000000000  0.000000000000  -0.001585582071 -0.828014493127
0.002500000000  7.806500000000  0.621500000000  -0.000002727929 0.000000000000  0.000000000000  -0.002022475257 -0.854318710850

``
NWT -old

0.000500000000  7.806500000000  0.621500000000  0.000000000000  0.000000000000  0.000000000000  -0.000060624627 0.000000000000
0.001000000000  7.806500000000  0.621500000000  -0.000000058077 0.000000000000  0.000000000000  -0.000112964467 -0.111058632821
0.001500000000  7.806500000000  0.621500000000  -0.000000140861 0.000000000000  0.000000000000  -0.000163438973 -0.105205803587
0.002000000000  7.806500000000  0.621500000000  -0.000000247904 0.000000000000  0.000000000000  -0.000212648503 -0.101293665480
0.002500000000  7.806500000000  0.621500000000  -0.000000378889 0.000000000000  0.000000000000  -0.000260989210 -0.098645636335

``

image

amneetb:
ok, I found the the source but I don't know why your zeroing out of side_coordinate does not give right answer:

git diff
diff --git a/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp b/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
index e09307b..cfdc669 100644
--- a/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
+++ b/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
@@ -292,7 +292,15 @@ IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForceTorque(IBTK::Vector
                     n(axis) = signof(phi_upper - phi_lower);
 
                     // Get the relative coordinate from X0
-                    const auto r_vec = IBTK::IndexUtilities::getSideCenter<IBTK::Vector3d>(*patch, s_i) - X0;
+                    const IBTK::VectorNd side_center = IBTK::IndexUtilities::getSideCenter(*patch, s_i);
+                    IBTK::Vector3d r_vec = IBTK::Vector3d::Zero();
+                    {
+                        r_vec[d] = side_center[d] - X0[d];
+                    }
+

boyceg:
Is there anything in the 3rd component of `X0`? `X0` is a 3-component vector, but the old code only shifted the first `NDIM` components.

amneetb:
The third component of X0 is 0 for 2D. This is what I have been using in `amneetb/numerical_wave_tank`. See this code 

https://github.com/IBAMR/IBAMR/blob/amneetb/numerical_wave_tank/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp#L319-L320

which is quite similar to new `boyceg/numerical_wave_tank`

https://github.com/IBAMR/IBAMR/blob/boyceg/numerical_wave_tank-final/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp#L295

It looks like some peculiarity with returned value of `IndexUtilities::getSideCenter`

drwells:
@boyceg if it helps, this PR also currently breaks the `check_hydro_force` test: it should print zeros but now gives a non-zero answer on this branch.

amneetb:
@drwells : If you switch the code from 

`const auto r_vec = IBTK::IndexUtilities::getSideCenter<IBTK::Vector3d>(*patch, s_i) - X0;`

to 

const IBTK::Vector3d r_vec = IBTK::IndexUtilities::getSideCenter<IBTK::Vector3d>(*patch, s_i) - X0;

given here
https://github.com/IBAMR/IBAMR/blob/boyceg/numerical_wave_tank-final/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp#L295

I think it will fix all the issues.

boyceg:
Those two functions should be identical. The second is implemented using the first.

boyceg:
Hmmm, it does look like Eigen is using some weird type here -- maybe because of the subtraction. I have made the type explicit now --- @amneetb, see if this fixes things for you.

boyceg:
Evidently Eigen does not play well with `auto`: https://eigen.tuxfamily.org/dox/TopicPitfalls.html.

boyceg:
History is now screwed up. Commits show up twice.

amneetb:
I did ‘git pull —force origin boyceg/numerical_wave_tank-final’. What is the right way of updating a local branch when the remote has rewritten the history? I thought —force pull is the right way. Apparently not 😥

boyceg:
I would fetch (e.g. `git fetch --all`) and then reset (`git reset --hard origin/branch`).

boyceg:
I got rid of the duplicated commits.

boyceg:
@amneetb perhaps you would be interested in https://www.sublimemerge.com. :-D

boyceg:
This dropped your comment. We probably should put a note about `auto` in the developer documentation instead of hoping that people see it in this one random source file.

Related: should we drop the `IBTK::VectorXX` `typedef`s and just use `Eigen::VectorXX`, so that it is clear that we are really using Eigen (and, consequently, should do things like add `EIGEN_MAKE_ALIGNED_OPERATOR_NEW` as needed)?

amneetb:
I agree. We should just use Eigen::Vector. There are too many Vectors (from libMesh, Eigen, IBAMR) and as many potential APIs.

drwells:
We should avoid using `libMesh::Point` et al when we don't directly interface with libMesh.

It might be best to retire the IBTK aliases of Eigen vectors to make it clearer that they really are the same thing.

amneetb:
The results are agreeing well with this branch now. I will try to add a unit test today

image

amneetb:
`asbhalla@Amneets-MBP-2:~/Softwares/IBAMR-Git/objs-opt$ ./attest  -R nwt
1/1 starting test wave_tank/nwt_cylinder.mpirun=8.input
    wave_tank/nwt_cylinder.mpirun=8.input --------------------------- PASSED

1/1 (100.0%) tests passed`

boyceg:
@drwells, can we push the button?

drwells:
Let me take a look quickly at the test suite. I would also like to merge #746 first to make the history slightly cleaner (i.e., so that tests always pass when we bisect).

drwells:
The test doesn't compile with modern versions of Eigen since it uses `Index` and not `hier::Index` - please fix this.

amneetb:
Can you point me where?

drwells:
line 153.

More importantly, this test immediately experiences a segmentation fault on my workstation. I'm investigating.

drwells:
of course the test works in debug mode :(

amneetb:
To be fair, I ran it in opt mode as can be seen in my path 😉

asbhalla@Amneets-MBP-2:~/Softwares/IBAMR-Git/objs-opt$

drwells:
Thats even more perplexing: it works in debug mode but fails inconsistently in optimized mode.

drwells:
The test runs in debug mode but the relative errors I compute are quite high:
```
[drwells@auricle wave_tank]$ numdiff -a 1e-10 -r 1e-10 nwt_cylinder.mpirun\=8.output output
----------------
##3       #:6   <== -0.005529762324
##3       #:6   ==> -0.005529693669
@ Absolute error = 6.8655000000e-8, Relative error = 1.2415696802e-5
----------------
##4       #:6   <== -0.004957595293
##4       #:6   ==> -0.004957609065
@ Absolute error = 1.3772000000e-8, Relative error = 2.7779597135e-6
----------------
##5       #:6   <== -0.004693105627
##5       #:6   ==> -0.004693122034
@ Absolute error = 1.6407000000e-8, Relative error = 3.4959792734e-6
----------------
##6       #:6   <== -0.004540591465
##6       #:6   ==> -0.004540593057
@ Absolute error = 1.5920000000e-9, Relative error = 3.5061511529e-7
----------------
##7       #:6   <== -0.004430466073
##7       #:6   ==> -0.004430463823
@ Absolute error = 2.2500000000e-9, Relative error = 5.0784750534e-7
----------------
##8       #:6   <== -0.004338105366
##8       #:6   ==> -0.004338104945
@ Absolute error = 4.2100000000e-10, Relative error = 9.7046983726e-8
+++  File "nwt_cylinder.mpirun=8.output" differs from file "output"
```
so the test ends up failing.

drwells:
We currently run numdiff with `-r 1e-6`, so only the first three lines are problematic in the test runner.

amneetb:
The numbers are matching up to 6th decimal place, but not beyond that. I used PETSc 3.10 while generating the output file. The far decimal points are quite sensitive to linear solvers.

drwells:
Good point - I set unreasonably strict solver tolerances elsewhere in the test suite for exactly this reason. Lets try that.

drwells:
@amneetb Would you please try adding something like

```
    if (NDIM == 2)
    {
        PetscOptionsSetValue(nullptr, "-stokes_ksp_atol", "1e-14");
        PetscOptionsSetValue(nullptr, "-stokes_ksp_rtol", "1e-14");
    }
    else
    {
        PetscOptionsSetValue(nullptr, "-stokes_ksp_atol", "1e-12");
        PetscOptionsSetValue(nullptr, "-stokes_ksp_rtol", "1e-12");
    }
```
(or whatever prefix the relevant solvers have in this example) to the test source, recomputing the output, and posting it here? That might be enough to fix the problem.

drwells:
Another thing we need to do is run clang-format again on all of the source files and headers - the `#include`s at least are not sorted correctly in some headers.

amneetb:
> Good point - I set unreasonably strict solver tolerances elsewhere in the test suite for exactly this reason. Lets try that.

Note that the input file has the solver settings I used to generate the `output`
https://github.com/IBAMR/IBAMR/blob/boyceg/numerical_wave_tank-final/tests/wave_tank/nwt_cylinder.mpirun%3D8.input#L331-L336

If you run `$mpiexec -np 8 <exe> <input>` it should pick up the same solver settings I used.

amneetb:
> @amneetb Would you please try adding something like
> 
> ```
>     if (NDIM == 2)
>     {
>         PetscOptionsSetValue(nullptr, "-stokes_ksp_atol", "1e-14");
>         PetscOptionsSetValue(nullptr, "-stokes_ksp_rtol", "1e-14");
>     }
>     else
>     {
>         PetscOptionsSetValue(nullptr, "-stokes_ksp_atol", "1e-12");
>         PetscOptionsSetValue(nullptr, "-stokes_ksp_rtol", "1e-12");
>     }
> ```
> 
> (or whatever prefix the relevant solvers have in this example) to the test source, recomputing the output, and posting it here? That might be enough to fix the problem.

But hardcoding solver settings is against the design philosophy of IBAMR 😂. You can set it through input file in here:
https://github.com/IBAMR/IBAMR/blob/boyceg/numerical_wave_tank-final/tests/wave_tank/nwt_cylinder.mpirun%3D8.input#L331-L336

drwells:
Setting it there is definitely a better option.

> But hardcoding solver settings is against the design philosophy of IBAMR

I know but its a test so it doesn't have to be done well, it just has to be done :)

drwells:
I found the source of the segmentation fault: there are several places where we use `std::vector<Eigen::Matrix3d>` or similar in the code, which causes undefined behavior when we enable machine-specific optimizations (e.g., my desktop computer hangs forever on the first time step and my workstation segmentation faults).

We need to go back and change all of these: see also #428.

drwells:
I've left a lot of comments that are now mixed in with some discussion. Here's the list of things that need to happen before we merge:
- [ ] avoid using `std::vector` without a custom allocator with Eigen objects (see #428)
- [ ] run clang-format again
- [ ] set solver tolerances to be much lower in the test. We should be getting the same output (up to a relative error of 1e-8 or so; the threshold in the test suite is currently 1e-6) on multiple machines.
- [ ] Make sure there are no unqualified `Index<>` objects (i.e., always write `hier::Index<>`)

amneetb:
Attached is the output with rtol/atol 1e-14. You can `numdiff` it with what you have on your machine. 

[output.zip](https://github.com/IBAMR/IBAMR/files/3848582/output.zip)

Any reason for running the test at such a high tolerance? Can you run the test as is with the solver tolerance given in the input file of 1e-8.

drwells:
I am not sure I understand the questions - let me try to answer.

> Any reason for running the test at such a high tolerance?
It tends to produce more stable (between compilers, versions of dependencies, etc) output since the exact details of the residual calculation can vary a lot.
> Can you run the test as is with the solver tolerance given in the input file of 1e-8.
I'm not sure I understand - the answer is yes, in debug mode.

If I turn the stokes relative residual tolerance down to 1e-14 then I get (within default attest numerical tolerances) the same output as in the output file you sent me.

amneetb:
> I am not sure I understand the questions - let me try to answer.
> 
> > Any reason for running the test at such a high tolerance?
> 
> It tends to produce more stable (between compilers, versions of dependencies, etc) output since the exact details of the residual calculation can vary a lot.
> 
> > Can you run the test as is with the solver tolerance given in the input file of 1e-8.
> 
> I'm not sure I understand - the answer is yes, in debug mode.
> 
> If I turn the stokes relative residual tolerance down to 1e-14 then I get (within default attest numerical tolerances) the same output as in the output file you sent me.



> I am not sure I understand the questions - let me try to answer.
> 
> > Any reason for running the test at such a high tolerance?
> 
> It tends to produce more stable (between compilers, versions of dependencies, etc) output since the exact details of the residual calculation can vary a lot.
> 
> > Can you run the test as is with the solver tolerance given in the input file of 1e-8.
> 
> I'm not sure I understand - the answer is yes, in debug mode.
> 
> If I turn the stokes relative residual tolerance down to 1e-14 then I get (within default attest numerical tolerances) the same output as in the output file you sent me.

Ok, great! You can commit that output file and modify the input file for the solver setting as:

```
stokes_solver_db {
      options_prefix    = "stokes_"
      ksp_type          = "fgmres"
      rel_residual_tol  = 1e-14
      abs_residual_tol  = 1e-14
      max_iterations    = 1000
   } 
```

boyceg:
Is all that is left to make sure that we properly prefix `Index<NDIM>`?

Should we always write `Index<NDIM>` as `hier::Index<NDIM>`? Would be easy to do but would be a big patch.

boyceg:
> Should we always write `Index<NDIM>` as `hier::Index<NDIM>`? Would be easy to do but would be a big patch.

(I.e. easy to a search-and-replace across the codebase.)

amneetb:
These silent forced pushes (Github does not send any notifications for them to alert others) are very dangerous. When I was committing this new code, the origin rejected it and showed conflict with the pull. Had to unroll the commit, force update my local branch and re-do all the work.

drwells:
I have been rewriting history a lot on this branch to fix various issues (e.g., eigen alignment). I was not aware that we were still adding more features to it. I will refrain from force pushing to prevent this problem.

drwells:
I would prefer that new features go in other branches so that we can merge this one ASAP :)

drwells:
@boyceg Yes, we should always qualify it to prevent the conflict with Eigen.

boyceg:
@amneetb, let us know if there are other things you are planning to push to this branch.

Once everything is merged, @drwells, can you qualify `Index<NDIM>`, or do you want me to do it?

drwells:
I'll let you do it :)

boyceg:
OK --- @amneetb let me know if you want me to do this. It will be one big messy patch. I think it is the last thing needed before merging?

drwells:
I just force pushed so that we are up to date with master (I rebased).

drwells:
The NWT test is too expensive now that we cranked down the tolerances: running
```
./attest -j8 -R nwt
```
takes 143 seconds. Running
```
./attest -j8 -E nwt
```
(i.e., all tests *except* the test added here) takes 199 seconds - that makes this test nearly as expensive as the rest of the test suite combined. @amneetb would you please find a way to make this test run quicker (e.g., coarser grids)?

drwells:
Once we resolve that we should be done :+1:

drwells:
I would also prefer longer tests to use fewer MPI processes. Running this test on my laptop takes something like 18 minutes (!!!) since I only have four logical cores. The other `mpirun=8` tests are much quicker.

amneetb:
I can generate the output by lowering the grid resolution.

amneetb:
> @amneetb, let us know if there are other things you are planning to push to this branch.

Nothing till December 1st week. 

> 
> Once everything is merged, @drwells, can you qualify `Index<NDIM>`, or do you want me to do it?

boyceg:
> > @amneetb, let us know if there are other things you are planning to push to this branch.
> 
> Nothing till December 1st week.

Does that mean that you do have other work you've been doing on this branch, or you are not planning on doing anything else until December?

If everything is currently done, let's finish it up and merge, so that further work can be in a new branch off `master`.

amneetb:
> > > @amneetb, let us know if there are other things you are planning to push to this branch.
> > 
> > 
> > Nothing till December 1st week.
> 
> Does that mean that you do have other work you've been doing on this branch, or you are not planning on doing anything else until December?

Latter. Right now I just using this branch to run some simulations. I have already updated my apps based on this branch, and I am not using the original NWT branch. 
> 
> If everything is currently done, let's finish it up and merge, so that further work can be in a new branch off `master`.

Agreed.

amneetb:
Updated input file and output with lower grid resolution.

[new_output.zip](https://github.com/IBAMR/IBAMR/files/3865844/new_output.zip)

drwells:
I didn't mean that we should go through the whole library on this patch and fix things. Nevertheless let me check it and then I will merge.

boyceg:
Oh.

If you want to revert and do this more selectively, have at it.

drwells:
I prefer to do that so we can finally merge this PR. Since things compile without that patch its secondary.

The `hier::Index` patch also needs some work - we still have, e.g., `std::vector<Index<NDIM>>`.

boyceg:
OK. Could you please revert and then hit the button? :-)

drwells:
We're done!

PR #746, opened 2019-11-11
HierarchyIntegrator now always regrids before the first time step

drwells:
Another independent commit from #744. Once we merge this we can just delete the commit there.

### DIFF
--- a/ibtk/src/utilities/HierarchyIntegrator.cpp
+++ b/ibtk/src/utilities/HierarchyIntegrator.cpp
@@ -1297,7 +1297,12 @@ HierarchyIntegrator::atRegridPointSpecialized() const
     }
     else
     {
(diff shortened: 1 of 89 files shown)


drwells:
Hm - not all tests pass this way, even though they passed with the other branch. There must be another change lurking around.

boyceg:
There are some changes in the INS integrators related to where regrid projections are implemented, although nothing that I would expect to change output -- I think the operations should be the same as before.

drwells:
It changes the output in a very minor way - we print an additional logging message stating where (i.e., in which function) we regrid. None of the computed output actually changes.

I merged the two commits together so that we don't have a commit in our history where half the tests no longer pass.

boyceg:
@drwells can we merge?

drwells:
Yup. I will rebase #744 afterwards.

PR #753, opened 2019-11-14
Add an excluding regex option to attest.

drwells:
A minor new feature for `attest`: `attest -E regex` will skip tests whose name matches `regex`.

@abarret This is the patch I mentioned a day or two ago.


drwells:
There is no reason to run the CI on this patch so lets merge.

PR #755, opened 2019-11-14
Disable OpenMP threading in the test suite.

drwells:
This leads to catastrophically bad performance since we expect all parallelization to be done with MPI in IBAMR.

This is largely necessary for hypre, which will try to parallelize itself with OpenMP when it is configured to do so (which happens to be the default on my desktop).


drwells:
Since this only effects the test runner I will merge it myself.

PR #759, opened 2019-11-19
Allow IBAMR to be built with bundled boost when libMesh is enabled with no boost support.

abarret:
Currently, IBAMR _must_ be compiled with an external boost if libMesh is present. This allows IBAMR to use the bundled boost if libMesh is configured without boost support.

@drwells, I'm not very familiar with the buildsystem. Can you double check this?

### DIFF
--- a/ibtk/configure
+++ b/ibtk/configure
@@ -24238,6 +24238,9 @@ $as_echo "$as_me: using Eigen library bundled with libMesh" >&6;}
 
   if test -e "$LIBMESH_DIR/include/boost" ; then
     as_fn_error $? "libMesh appears to be configured to use a bundled Boost library, but when IBAMR is also configured to use libMesh, IBAMR and libMesh both must use the same (external) Boost library" "$LINENO" 5
+
+$as_echo "#define LIBMESH_HAS_BOOST 1" >>confdefs.h
+
   fi
 
   CPPFLAGS="$LIBMESH_CPPFLAGS $CPPFLAGS"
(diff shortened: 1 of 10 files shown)


m4/configure_boost.m4
```diff
@@ -14,7 +14,7 @@ AC_ARG_VAR(BOOST_ROOT,[the location of the Boost installation that is to be used
 #$as_unset boost_cv_version
 #BOOST_REQUIRE([1.57.0],[AC_MSG_WARN([could not find system Boost library, using bundled Boost library])])
 if test x"$USING_BUNDLED_BOOST" = xyes ; then
-  if test "$LIBMESH_ENABLED" = yes ; then
+  if test "$LIBMESH_HAS_BOOST" = yes ; then
     AC_MSG_ERROR([IBAMR is configured to use a bundled Boost library, but when IBAMR is also configured to use libMesh, IBAMR and libMesh both must use the same (external) Boost library])
```
drwells: I don't think this error message is right: we support using libMesh's bundled copy of boost, no?

drwells:
I'm surprised that I have not run into this configuration problem before given that I tried to test all permutations of our dependencies. Thanks for the fix :)

drwells:
Also, please add a changelog entry indicating that we fixed this bug.

m4/configure_boost.m4
boyceg: At least at some point we used stuff from boost not provided by libMesh's bundled copy. I do not know if that is currently the case.

drwells:
I think its reasonable to support using the following combinations:
1. libMesh without boost, our bundled copy of boost
2. libMesh without boost, external boost
3. libMesh with external boost, external boost

i.e., we should just ban using libMesh's bundled boost with IBAMR. In addition, we should recommend that libMesh be compiled without boost - they only use it in one spot where they have an alternative implementation anyway.

abarret:
While we're talking about the build system, is there any reason some of the IBTK .m4 configure scripts link to the IBAMR .m4 scripts but others don't?

abarret:
Added changelog.

drwells:
> While we're talking about the build system, is there any reason some of the IBTK .m4 configure scripts link to the IBAMR .m4 scripts but others don't?

It doesn't make a ton of sense in retrospect, but every time I had to modify one `m4` script or the other I replaced the IBTK m4 scripts with symbolic links to the IBAMR ones. If I haven't had to edit the script I haven't replaced it with a link yet.

m4/configure_boost.m4
drwells: I think this PR is ready once this error message is clarified.

abarret:
I've updated the error message.

ibtk/configure
```diff
@@ -27225,7 +27228,7 @@ CONTRIB_SRCDIR="$ABSOLUTE_SRCDIR/contrib/boost"
 #$as_unset boost_cv_version
 #BOOST_REQUIRE([1.57.0],[AC_MSG_WARN([could not find system Boost library, using bundled Boost library])])
 if test x"$USING_BUNDLED_BOOST" = xyes ; then
-  if test "$LIBMESH_ENABLED" = yes ; then
+  if test "$LIBMESH_HAS_BOOST" = yes ; then
     as_fn_error $? "IBAMR is configured to use a bundled Boost library, but when IBAMR is also configured to use libMesh, IBAMR and libMesh both must use the same (external) Boost library" "$LINENO" 5
```
drwells: I think you need to rerun `autoreconf`.

abarret: Whoops. Fixed.

PR #761, opened 2019-11-19
Fix compatibility with libMesh 1.1.0.

drwells:
`libMesh::BoundingBox` was `libMesh::MeshTools::BoundingBox` until 1.2.0.

### DIFF
--- a/ibtk/include/ibtk/libmesh_utilities.h
+++ b/ibtk/include/ibtk/libmesh_utilities.h
@@ -40,7 +40,11 @@
 
 #include "tbox/Utilities.h"
 
+#if 1 <= LIBMESH_MAJOR_VERSION && 2 <= LIBMESH_MINOR_VERSION
 #include "libmesh/bounding_box.h"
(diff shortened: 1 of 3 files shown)


drwells:
Since we only use this in a small number of places I think I would rather do this than implement it with a new 'libmesh compatibility' header.

boyceg:
What about dropping support for 1.1.0? I think that is pretty old now...

drwells:
Its only 3 years old. I have no idea if anyone actually uses it, but I like to keep things around for 4-5 years.

drwells:
I did kill compatibility with 1.0 earlier this year since 1.1.0 introduced an important mesh base class. Since its much easier to keep compatibility with 1.1.0 I don't want to axe it yet.

drwells:
Since this is approved and I need it for something else (I rewrite of how we determine which elements should be used for spreading and interpolation) I am going to merge. We can come up with a prettier fix later.

PR #762, opened 2019-11-20
use hier::Index<NDIM> instead of Index<NDIM>

drwells:
@boyceg take a look :)

### DIFF
--- a/ibtk/src/math/PETScMatUtilities.cpp
+++ b/ibtk/src/math/PETScMatUtilities.cpp
@@ -93,7 +93,7 @@ namespace IBTK
 
 namespace
 {
-bool inline is_cf_bdry_idx(const Index<NDIM>& idx, const std::vector<Box<NDIM> >& cf_bdry_boxes)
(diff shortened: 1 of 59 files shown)


ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -162,7 +162,7 @@ PETScMatUtilities::constructPatchLevelCCLaplaceOp(Mat& mat,
 
     // Setup the finite difference stencil.
     static const int stencil_sz = 2 * NDIM + 1;
-    std::vector<Index<NDIM> > stencil(stencil_sz, Index<NDIM>(0));
+    std::vector<Index<NDIM> > stencil(stencil_sz, hier::Index<NDIM>(0));
```
drwells: here, etc.

drwells:
Based on an in-person discussion I will finish this patch.

drwells:
Since @boyceg wrote the majority of this patch and I checked it (and added a few missing namespace qualifications) I will consider this approved and merge as-is.

PR #766, opened 2019-11-21
Add more bbox

drwells:
I talked about this with @boyceg yesterday - one reason that the scratch partitioning scheme is not much cheaper than the default one is that, to load balance well, we have to make more patches. If we make more patches we do a ton more work per patch since we use a very conservative scheme to determine whether or not an element is relevant to a given patch.

This PR adds a function that computes bounding boxes of quadrature points of deformed elements - the intent is that we can replace the current logic in `FEDataManager` with something that just checks element intersections with a tolerance: i.e., add the element if its bounding box intersects the patch's bounding box plus ghost region (plus some additional layer of cells that allow us to not regrid after every time step).


ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -1335,6 +1335,18 @@ void write_elem_partitioning(const std::string& file_name, const libMesh::System
  */
 void write_node_partitioning(const std::string& file_name, const libMesh::System& position_system);
 
+/**
+ * Compute bounding boxes based on where an elements quadrature points
+ * are. See getQuadratureKey for descriptions of the last five arguments.
+ */
+std::vector<libMesh::BoundingBox> get_local_active_element_bounding_boxes(const libMesh::MeshBase& mesh,
```
drwells: whoops - this needs the `#ifdef` guards for compatibility with libMesh 1.1.0.

tests/IBTK/bounding_boxes_01.cpp
```diff
@@ -0,0 +1,205 @@
...
+        {
+            TBOX_ASSERT((box_2.min() - box_3.min()).norm() <
+                        std::max(1.0, box_2.min().norm())*1e-16);
+            TBOX_ASSERT((box_2.max() - box_3.max()).norm() <
+                        std::max(1.0, box_2.max().norm())*1e-16);
+        }
+
+        // in both cases check that we are inside the libMesh box: it claims
+        // to be a loose fit
+        BoundingBox box_union(box_3);
```
drwells: same

tests/IBTK/bounding_boxes_01.cpp
```diff
@@ -0,0 +1,205 @@
...
+            TBOX_ASSERT(box_2.signed_distance(box_1.min()) <= 0.0);
+            TBOX_ASSERT(box_2.signed_distance(box_1.max()) <= 0.0);
+#endif
+        }
+        out << box_2.first << ", " << box_2.second << std::endl;
+
+        // Since X is just the initial coordinates of the mesh, we should
+        // match the bounding box which is computed directly from the
+        // element too.
+        libMesh::BoundingBox box_3 = (*el_it)->loose_bounding_box();
```
drwells: same

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -49,6 +49,18 @@ IBTK_ENABLE_EXTRA_WARNINGS
 
 namespace IBTK
 {
+namespace libMeshWrappers
```
boyceg: 👍

PR #767, opened 2019-11-22
Convert the Navier-Stokes gtest.

drwells:
Last test for #569.


drwells:
Since this just adds a test and does not modify the library I will merge as-is.

PR #768, opened 2019-11-25
Add the developer certificate of origin.

### LINKED ISSUE #604: Clean up copyright notices

drwells:
This document clarifies what expectations we have for copyrights. Fixes #604.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,53 @@
+Contributions to IBAMR
+----------------------
+
+IBAMR is collectively developed by a community at multiple universities and is
+happy to receive patches of any size for consideration. Contributions should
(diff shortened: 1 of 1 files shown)


drwells:
This is an already agreed-upon fix to the copyright situation (we already discussed doing exactly this) so I will merge as-is without waiting for review.

PR #769, opened 2019-11-25
Add two more notes to release tasks.

drwells:
These also need to happen.


drwells:
Again, since its just a minor documentation change, I'll merge without waiting for review.

PR #771, opened 2019-11-25
Reformat some files to use Unix line endings.

drwells:


drwells:
I need this to finish #772 so I will merge as-is.

PR #777, opened 2019-12-05
Add another release task

drwells:
At some point we should add this task to a linter CI job, but for now it suffices to spend a few minutes before each release checking.


drwells:
Its just a release task, and I added it to the issue tracking the 0.6 release - I'll just merge this as-is.

PR #779, opened 2019-12-06
Update bundled eigen

### LINKED ISSUE #754: Install modern copy of Eigen for CI
New versions of Eigen defined an `Index<>` template, which conflicts (when we include all namespaces with SAMRAI's `hier::Index<>` template. We should set up one of the CI builds to use a new copy of Eigen so that this fails correctly.

drwells:
Fixes #754. This is necessary to fix a bug in the 2x2 eigenvector calculation in #725.


drwells:
I included all of `Eigen/`. If there is stuff we can neglect then let me know.

drwells:
The testers don't have a configuration that verifies the bundled eigen. It works on my machine so I will merge.

PR #781, opened 2019-12-07
Increment FEDataManager::FE_DATA_MANAGER_VERSION.

### LINKED ISSUE #691: Improve restart version numbers in FEDataManager
@jabrown893 wants to use some old restart data, but I broke restart data with recent changes to FEDataManager. We should increase the version number in FEDataManager and also teach it to handle old restart data.

drwells:
Fixes #691.


drwells:
Yup - see #674. We completely changed the way we read and write restart data to permit multiple FEDataManager objects to share one set of libMesh data structures.

PR #782, opened 2019-12-09
included <memory> for linux build

Kaustubh-Khedkar:


drwells:
Yup: we use `std::unique_ptr` below so we need the header. Thanks a lot for the fix!

PR #785, opened 2019-12-10
Fix test partitioning

### LINKED ISSUE #773: Get the test suite working on my desktop
Subj.

For whatever reason, I get different results for some tests on my workstation than I do on my personal desktop. Some of the differences (like for linear spreading and interpolation) are quite large. Some are very small: e.g., one processor will have one more quadrature point and another will have one less. I will get this sorted out before the release.

drwells:
This is enough of a fix for #773 for now. It would be nice to solve the root problem and get SAMRAI to deterministically initially partition patches but I doubt that this is possible from inside IBAMR - hopefully it has been fixed in a more recent release.

Fixes #773.

### DIFF
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -256,6 +256,19 @@ namespace IBAMR
  * assigned to an IB point when calculating the work per processor: in the
  * future additional weights, such as <code>workload_node_point_weight</code>
  * will also be added.
+ *
+ * <h2>Options Controlling Logging</h2>
+ * The logging options set by this class are propagated to the owned
+ * FEDataManager objects.
+ * <ol>
+ *   <li><code>enable_logging</code>: set to <code>TRUE</code> to enable logging.</code>.
+ *   Defaults to <code>false</code>.</li>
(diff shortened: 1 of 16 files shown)


drwells:
@abarret if you have time could you rerun the testsuite with this patch? This should fix the IBFE tests.

abarret:
This fixes the `IBFE/explicit_ex4_2d.postprocessor.scratch_hier.mpirun=4.input` test for me.

boyceg:
Does this show up for both `GREEDY` and `SPATIAL` bin packing algorithms?

drwells:
Yup.

boyceg:
Only with AMR?

drwells:
AFAICT yes - if we just have a single nonrefined grid then each processor gets an equally-sized chunk of that grid regardless of the machine used.

drwells:
Additionally, those chunks are distributed in the same way - e.g., processor 2 always gets (8, 8) - (15, 15).

drwells:
@boyceg This patch isn't perfect but its good enough for the release. Is there anything else you would like to change?

PR #788, opened 2019-12-16
Fix copyright

drwells:
Part of #749, the release process.


boyceg:
I created such scripts long ago, but I would have to dig them up.

drwells:
Probably? I didn't look that closely at the scripts: a few of them seemed to deal with the build system so they can probably be removed (i.e., the predate the autotools build system).

Lets open an issue for it and tackle it later.

PR #792, opened 2019-12-17
Get rid of some warnings from hypre headers.

drwells:
These are new in hypre 2.18 (or some other recent release) that added some CUDA functions that we don't use.


drwells:
Part of #749.

drwells:
Since this just moves headers around I am going to merge without waiting for review.

PR #796, opened 2019-12-17
Remove some unused variables.

drwells:
Found by cppcheck. Part of #749 

I am not going to fix everything cppcheck found: some of the issues, e.g., use of virtual functions in ctors and dtors, require care. I am going to put up patches with some of the really obvious fixes (like this).


drwells:

PR #797, opened 2019-12-17
Reduce the scope of some variables.

drwells:


drwells:
Part of #749. Also found by cppcheck.

drwells:

ibtk/src/math/PETScVecUtilities.cpp
```diff
@@ -623,10 +623,9 @@ PETScVecUtilities::constructPatchLevelAO_cell(AO& ao,
                                               Pointer<PatchLevel<NDIM> > patch_level,
                                               const int ao_offset)
 {
-    int ierr;
     if (ao)
     {
-        ierr = AODestroy(&ao);
+        const int ierr = AODestroy(&ao);
```
boyceg: Not a big deal, but I don't like to make these `const`, because you then have to go back and remove the qualifier if you add in more PETSc calls.

boyceg: (BTW, this is the same reason that `ierr` is scoped the way that it is.)

src/IB/IBFEDirectForcingKinematics.cpp
```diff
@@ -223,9 +223,6 @@ IBFEDirectForcingKinematics::forwardEulerStep(double current_time,
     // Rotate the body with current rotational velocity about center of mass
     // and translate the body to predicted position.
     Eigen::Vector3d dr = Eigen::Vector3d::Zero();
```
boyceg: Move this down to around line 264 (and skip initializing it)?

src/IB/IBInterpolantMethod.cpp
```diff
@@ -523,7 +523,6 @@ IBInterpolantMethod::updateMeshPosition(double current_time,
             double* const X_new = &X_new_array[local_idx][0];
             const double* const X0 = &X0_array[local_idx][0];
             Eigen::Vector3d dr = Eigen::Vector3d::Zero();
```
boyceg: Same as above.

src/IB/IBFEDirectForcingKinematics.cpp
drwells: We need to keep initializing it so things work in 2D (or put in a much longer check; I'd prefer to just zero it) but we can move it down.

drwells:
I un-consted the integers and moved the `dr` initializations around (though I kept the `= Zero()` bit).

boyceg:
> I un-consted the integers and moved the `dr` initializations around (though I kept the `= Zero()` bit).

Sounds good.

Also, if cppcheck really wants these things `const`ed, then that is OK with me too.

drwells:
In this case it complained about scope. If it complains about constness I'll fix that later too.

PR #798, opened 2019-12-17
Use std::expm1 where appropriate.

drwells:
Part of #749. Also found by cppcheck.

@amneetb this should make the wave calculations ever so slightly better :)


drwells:

PR #799, opened 2019-12-17
Cppcheck 4

drwells:
A few more small things caught by cppcheck.

### DIFF
--- a/ibtk/include/ibtk/FEDataInterpolation.h
+++ b/ibtk/include/ibtk/FEDataInterpolation.h
@@ -340,9 +340,9 @@ private:
 
     // Data associated with the current element.
     const libMesh::Elem* d_current_elem = nullptr;
-    unsigned int d_current_side;
+    unsigned int d_current_side = std::numeric_limits<unsigned int>::max();
     std::vector<boost::multi_array<double, 2> > d_system_elem_data;
-    unsigned int d_n_qp;
+    unsigned int d_n_qp = std::numeric_limits<unsigned int>::max();
 };
 } // namespace IBTK
 
--- a/src/IB/GeneralizedIBMethod.cpp
+++ b/src/IB/GeneralizedIBMethod.cpp
@@ -264,6 +264,7 @@ GeneralizedIBMethod::interpolateVelocity(const int u_data_idx,
     {
         W_data = &d_W_new_data;
     }
+    TBOX_ASSERT(W_data);
 
     Pointer<Variable<NDIM> > u_var = d_ib_solver->getVelocityVariable();
     Pointer<CellVariable<NDIM, double> > u_cc_var = u_var;
@@ -512,6 +513,8 @@ GeneralizedIBMethod::spreadForce(const int f_data_idx,
         N_data = &d_N_new_data;
         N_needs_ghost_fill = &d_N_new_needs_ghost_fill;
     }
+    TBOX_ASSERT(N_data);
+    TBOX_ASSERT(N_needs_ghost_fill);
 
     std::vector<Pointer<LData> >* X_LE_data;
     bool* X_LE_needs_ghost_fill;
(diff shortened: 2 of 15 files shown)


drwells:

ibtk/include/ibtk/FEDataInterpolation.h
```diff
@@ -340,9 +340,9 @@ class FEDataInterpolation
 
     // Data associated with the current element.
     const libMesh::Elem* d_current_elem = nullptr;
-    unsigned int d_current_side;
+    unsigned int d_current_side = std::numeric_limits<unsigned int>::max();
```
boyceg: Is the goal to provide initial/default values for all member variables? I gave up on doing that with C++03, but it is a lot easier now with C++11.

src/IB/GeneralizedIBMethod.cpp
```diff
@@ -266,6 +266,7 @@ GeneralizedIBMethod::interpolateVelocity(const int u_data_idx,
     {
         W_data = &d_W_new_data;
     }
+    TBOX_ASSERT(W_data != nullptr);
```
boyceg:
```suggestion
    TBOX_ASSERT(W_data);
```

src/IB/GeneralizedIBMethod.cpp
```diff
@@ -513,6 +514,8 @@ GeneralizedIBMethod::spreadForce(const int f_data_idx,
         N_data = &d_N_new_data;
         N_needs_ghost_fill = &d_N_new_needs_ghost_fill;
     }
+    TBOX_ASSERT(N_data != nullptr);
```
boyceg:
```suggestion
    TBOX_ASSERT(N_data);
```

src/IB/GeneralizedIBMethod.cpp
```diff
@@ -513,6 +514,8 @@ GeneralizedIBMethod::spreadForce(const int f_data_idx,
...
+    TBOX_ASSERT(N_needs_ghost_fill != nullptr);
```
boyceg:
```suggestion
    TBOX_ASSERT(N_needs_ghost_fill);
```

ibtk/include/ibtk/FEDataInterpolation.h
drwells: Yes - it would be great to get everything initialized in the constructor to a deterministic value.

src/IB/IBStandardForceGen.cpp
```diff
@@ -759,23 +759,23 @@ IBStandardForceGen::initializeSpringLevelData(std::set<int>& nonlocal_petsc_idx_
     const int num_local_nodes = static_cast<int>(local_nodes.size());
 
     // Determine how many springs are associated with the present MPI process.
-    unsigned int num_springs = 0;
+    unsigned int total_num_springs = 0;
```
boyceg: Also prefix other "total" numbers of things with `total`?

drwells: Is there something else in this function which could be renamed in this way? I don't see it.

boyceg: Is this change because `num_springs` is used elsewhere in this function? Otherwise, I would prefer to leave it as `num_springs`.

boyceg: For instance, lower down in this file, should it be `total_num_beams` instead of `num_beams`?

I don't care if this is prefixed by `total` or not, but I do care that names are consistent. E.g., `total_num_springs` and `total_num_beams`, or `num_springs` and `num_beams`, but not `total_num_springs` and `num_beams`.

drwells: Good idea - I'll get those too.

boyceg: There would also be `total_num_target_points`.

drwells: Got it :)

PR #800, opened 2019-12-17
edited git.md docs for cleaning up history

marshallrdavey:
added the force push step for editing git history


drwells:
thanks a lot for the quick fix!

PR #801, opened 2019-12-18
Add a hint for a release task.

drwells:
It is easier to use grep -L here.


drwells:
This doesn't immediately pertain to the library so I will merge as-is.

PR #803, opened 2019-12-18
Fix clang warnings

drwells:
I found at least one real problem here and silenced a bunch of warnings generated by clang.

Part of #749.


drwells:
This is the last patch we need for the release. Since everyone else is otherwise occupied I will merge it myself.

PR #809, opened 2019-12-19
Kaustubhk/irregular-wave-gen

amneetb:
1) Added irregular wave generation zone code to mitigate wave reflection from structures at the wave maker boundary.
2) Fixed 3D bugs in FESurfaceDistanceEvaluator and BrinkmanPenalization classes.

### DIFF
--- a/include/ibamr/IrregularWaveGenerator.h
+++ b/include/ibamr/IrregularWaveGenerator.h
@@ -1,0 +1,108 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2019 - 2019 by the IBAMR developers
+// All rights reserved.
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+/////////////////////// INCLUDE GUARD ////////////////////////////////////
+
+#ifndef included_IrregularWaveGenerator
+#define included_IrregularWaveGenerator
+
+///////////////////////////// INCLUDES ///////////////////////////////////
+
+#include "ibamr/StokesWaveGeneratorStrategy.h"
+
+#include <fstream>
+
+namespace IBAMR
+{
+/*!
+ * \brief Class for generating Irregular waves.
+ */
+class IrregularWaveGenerator : public StokesWaveGeneratorStrategy
+{
+public:
+    IrregularWaveGenerator(const std::string& object_name, SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db);
+
+    /*!
+     * Get surface elevation at a specified horizontal position and time.
+     */
+    double getSurfaceElevation(double x, double time) const;
+
+    /*!
+     * Get velocity component at a specified position and time.
+     */
+    double getVelocity(double x, double z_plus_d, double time, int comp_idx) const;
+
+    /*!
+     * Print the wave data.
+     */
+    void printWaveData(std::ofstream& ostream) const;
+
+private:
+    /*!
+     * Get wave parameters from input db.
+     */
+    void getFromInput(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db);
+
+    ///
+    /// Number of component waves with random phases to be generated (default = 50).
+    ///
+    int d_num_waves = 50;
+
+    ///
+    /// Significant wave height [length].
+    ///
+    double d_Hs = std::numeric_limits<double>::quiet_NaN();
+
+    ///
+    /// Significant wave period [time].
+    ///
+    double d_Ts = std::numeric_limits<double>::quiet_NaN();
+
+    ///
+    /// Lowest angular frequency in the spectrum [rad/time].
+    ///
+    double d_omega_begin = std::numeric_limits<double>::quiet_NaN();
+
+    ///
+    /// Highest angular frequency in the spectrum [rad/time].
+    ///
+    double d_omega_end = std::numeric_limits<double>::quiet_NaN();
+
+    ///
+    /// JONSWAP/Bretschneider wave spectrum.
+    ///
+    std::string d_wave_spectrum;
+
+    ///
+    /// Angular frequencies of component waves [rad/time].
+    ///
+    std::vector<double> d_omega;
+
+    ///
+    /// Wave number of component waves [2\f$ \pi \f$/length].
+    ///
+    std::vector<double> d_wave_number;
+
+    ///
+    /// Amplitude of component waves [length].
+    ///
+    std::vector<double> d_amplitude;
+
+    ///
+    /// Phase (random) of component waves [rad].
+    ///
+    std::vector<double> d_phase;
+};
+
+} // namespace IBAMR
+#endif
--- a/src/wave_generation/IrregularWaveGenerator.cpp
+++ b/src/wave_generation/IrregularWaveGenerator.cpp
@@ -1,0 +1,192 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2019 - 2019 by the IBAMR developers
+// All rights reserved.
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include "ibamr/IrregularWaveGenerator.h"
+#include "ibamr/RNG.h"
+#include "ibamr/app_namespaces.h"
+
+#include "ibtk/ibtk_utilities.h"
+
+#include "tbox/Database.h"
+
+namespace IBAMR
+{
+/////////////////////////////// STATIC ///////////////////////////////////////
+namespace
+{
+static const unsigned SEED = 1234567;
+}
+/////////////////////////////// PUBLIC ///////////////////////////////////////
+
+IrregularWaveGenerator::IrregularWaveGenerator(const std::string& object_name, Pointer<Database> input_db)
+    : StokesWaveGeneratorStrategy(object_name, input_db)
+{
+    // Get wave parameters.
+    getFromInput(input_db);
+
+    // Resize vectors according to number of wave components
+    d_amplitude.resize(d_num_waves);
+    d_wave_number.resize(d_num_waves);
+    d_omega.resize(d_num_waves);
+    d_phase.resize(d_num_waves);
+
+    double delta_omega = std::abs(d_omega_end - d_omega_begin) / (d_num_waves - 1);
+    double omega_s = 2 * M_PI / d_Ts;
+
+    RNG::srandgen(SEED);
+
+    // Calculating component waves
+    for (int i = 0; i < d_num_waves; i++)
+    {
+        double rn;
+        RNG::genrand(&rn);
+        d_phase[i] = 2 * M_PI * rn;
+
+        d_omega[i] = d_omega_begin + i * delta_omega;
+
+        // Using an approximate formula for the dispersion relationship, calculate the wave number.
+        // See Eqn. (5.4.22) in WAVES IN OCEANIC AND COASTAL WATERS by LEO H. HOLTHUIJSEN.
+        const double alpha = std::pow(d_omega[i], 2) * d_depth / d_gravity;
+        const double beta = alpha * std::pow(tanh(alpha), -0.5);
+        d_wave_number[i] = (alpha + std::pow(beta, 2) * std::pow(cosh(beta), -2)) /
+                           (d_depth * (tanh(beta) + beta * std::pow(cosh(beta), -2)));
+
+        double spectral_density = std::numeric_limits<double>::quiet_NaN();
+
+        if (d_wave_spectrum == "JONSWAP")
+        {
+            double sigma, A, gamma = 3.3;
+
+            sigma = d_omega[i] <= omega_s ? 0.07 : 0.09;
+
+            A = std::exp(-std::pow((d_omega[i] / omega_s - 1) / (sigma * std::sqrt(2)), 2));
+
+            spectral_density = 320 * std::pow(d_Hs, 2) / (std::pow(d_Ts, 4) * std::pow(d_omega[i], 5)) *
+                               std::exp(-1950.0 / (std::pow(d_Ts * d_omega[i], 4))) * std::pow(gamma, A);
+        }
+        else if (d_wave_spectrum == "BRETSCHNEIDER")
+        {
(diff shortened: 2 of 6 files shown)


src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -1254,23 +1254,22 @@ FESurfaceDistanceEvaluator::computeSignedDistanceSurfaceMesh(int n_idx, int d_id
...
+#if !defined(NDEBUG)
+                    TBOX_ASSERT(n_nodes == NDIM);
 #endif
+                    for (unsigned int i = 0; i < n_nodes; ++i)
+                    {
+                        for (int d = 0; d < NDIM; ++d)
+                        {
+                            X_node[i][d] = (elem->point(i))(d);
+                        }
+                    }
```
amneetb: @drwells : You would like this 😛

drwells: This is much nicer!

drwells: There is similar code circa line 1190 that should also be generalized to an arbitrary number of nodes.

drwells:
Could you move the bug fixes into another PR? I want to tag the release today and it would be nice to include those fixes.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -1254,23 +1254,22 @@ FESurfaceDistanceEvaluator::computeSignedDistanceSurfaceMesh(int n_idx, int d_id
...
-
-#if (NDIM == 3)
-                    const libMesh::Point& n2 = elem->point(2);
-                    X_node[2][0] = n2(0);
-                    X_node[2][1] = n2(1);
-                    X_node[2][2] = n2(2);
+                    const unsigned int n_nodes = elem->n_nodes();
+#if !defined(NDEBUG)
+                    TBOX_ASSERT(n_nodes == NDIM);
 #endif
```
drwells: This doesn't seem right. Do we really only support TRI3 and EDGE2?

drwells:
I feel like we are just kicking the can down the road with FESurfaceDistanceEvaluator: there are surely other problems with this class (like the one I noted). I think we should stop here and write some unit tests for this class before proceeding so that we know it really works as advertised. Could @amneetb and @Kaustubh-Khedkar please find time to do this so we can shake out any remaining bugs?

src/level_set/FESurfaceDistanceEvaluator.cpp
amneetb: Yes. We check this condition  in the constructor and map intersection routine.

The computational geometry work done in this class is quite expensive to carry out. It has been optimized to do it for EDGE2 and TRI3, the two most common elements for engineering geometries.

src/level_set/FESurfaceDistanceEvaluator.cpp
amneetb: That one is intentional.

amneetb:
> Could you move the bug fixes into another PR? I want to tag the release today and it would be nice to include those fixes.

Sure. What is the recommended git workflow to do this?

drwells:
I think the easiest way would be to do a soft reset. The following should work:
```
git checkout Kaustubhk/irregular-wave-gen
git reset --soft master
git checkout -b amneetb/fix-brinkman
git rebase --autostash master
git add src/IB/BrinkmanPenalizationRigidBodyDynamics.cpp
git commit -m "fix brinkman"
```
that should put you on a new branch with just the changes to Brinkman committed. The rest will be staged but not committed.

amneetb:
> I think we should stop here and write some unit tests for this class before proceeding so that we know it really works as advertised. Could @amneetb and @Kaustubh-Khedkar please find time to do this so we can shake out any remaining bugs?

I agree there should be exclusive tests for this functionality. However, it is going to take us quite some time to test out the code for real and complicated geometries. We are doing it as we go along in our research. The boring ones like circles, spheres etc have been tested extensively and they give desired second order accuracy. We have tested simple cars and wave energy devices. We aim to test more complex ships and Stanford bunny in the future.

amneetb:
> I think the easiest way would be to do a soft reset. The following should work:
> 
> ```
> git checkout Kaustubhk/irregular-wave-gen
> git reset --soft master
> git checkout -b amneetb/fix-brinkman
> git rebase --autostash master
> git add src/IB/BrinkmanPenalizationRigidBodyDynamics.cpp
> git commit -m "fix brinkman"
> ```
> 
> that should put you on a new branch with just the changes to Brinkman committed. The rest will be staged but not committed.



> I think the easiest way would be to do a soft reset. The following should work:
> 
> ```
> git checkout Kaustubhk/irregular-wave-gen
> git reset --soft master
> git checkout -b amneetb/fix-brinkman
> git rebase --autostash master
> git add src/IB/BrinkmanPenalizationRigidBodyDynamics.cpp
> git commit -m "fix brinkman"
> ```
> 
> that should put you on a new branch with just the changes to Brinkman committed. The rest will be staged but not committed.

What does 

```
> git reset --soft master
```
and 

```
> > git rebase --autostash master
```

do?

drwells:
`git reset --soft master` will undo all commits between the current commit and whereever `master` is. The changes themselves are kept (they are staged); the only thing removed is the history.

`git rebase --autostash master` 'replays' the commits on the current branch on top of `master` (i.e., it changes the parent from `Kaustubhk/irregular-wave-gen` to `master`). the `--autostash` option means that uncommitted changes are ignored (they are automatically hidden via `git stash` and then automatically brought by via `git stash pop`).

drwells:
> The boring ones like circles, spheres etc have been tested extensively and they give desired second order accuracy.

Could we add these to the library?

amneetb:
Yes, it would be a nice practice my students (sometime during the next semester)

drwells:
Sounds good - I'll open an issue to track it.

drwells:
In case its not clear: @amneetb and I moved the bugfix commits off of this branch. The only thing remaining is the irregular wave generation contribution.

src/wave_generation/IrregularWaveGenerator.cpp
```diff
@@ -0,0 +1,198 @@
...
+#include "ibtk/ibtk_utilities.h"
+
+#include "tbox/Database.h"
+
+namespace IBAMR
+{
+/////////////////////////////// STATIC ///////////////////////////////////////
+namespace
+{
+static const unsigned SEED = 1234567;
```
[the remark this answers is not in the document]
amneetb: I think the right approach is to have a SEED defined in the RNG class. Codes that need to generate `same` set of random numbers at different locations can use the SEED from there. Right now we have hard-coded the same SEED at two places, this file and `IrregularWaveBcCoef.cpp`. It was one of the non-discussed issues in the prior software calls  😉

amneetb:
@Kaustubh-Khedkar : Please make other changes as suggested by @drwells.

Kaustubh-Khedkar:
Yes, I will get it done.

src/wave_generation/IrregularWaveGenerator.cpp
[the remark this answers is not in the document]
drwells: I would really, really prefer to deprecate our own RNG and just use the C++11 `<random>` library to completely avoid these issues and do exactly what you want (generate the same set of random numbers at different locations). The (major) downside is that `<random>` is complicated. Would you be willing to do this? I can show you how if you are!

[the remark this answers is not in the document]
amneetb: We do have a parallel version of RNG, that seems pretty useful. Why not enhance our own RNG class, which is quite simple and easy to use for Gaussian distributions? C+11 version is pretty exhaustive though, but for this app I don’t require those features and distributions.

In any case, we can switch to C++11. Can you send us some example?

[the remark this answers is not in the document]
amneetb: We do have a parallel version of RNG, that seems pretty useful. Why not enhance our own RNG class, which is quite simple and easy to use for Gaussian distributions? C+11 version is pretty exhaustive though, but for this app I don’t require those features and distributions.

In any case, we can switch to C++11. Can you send us some example?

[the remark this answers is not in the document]
boyceg: Yes, note that the current RNG does attempt to handle random number generation properly in parallel --- see `RNG::parallel_seed()`.

include/ibamr/IrregularWaveGenerator.h
```diff
@@ -0,0 +1,71 @@
...
+ */
+class IrregularWaveGenerator : public StokesWaveGeneratorStrategy
+{
+public:
+    IrregularWaveGenerator(const std::string& object_name, SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db);
+
+    /*!
+     * Get surface elevation at a specified horizontal position and time.
+     */
+    double getSurfaceElevation(double x, double time) const;
```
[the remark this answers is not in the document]
boyceg: Because this is pass by value, `const` is an implementation decision that should go into the implementation file, not the header.

include/ibamr/IrregularWaveGenerator.h
```diff
@@ -0,0 +1,71 @@
...
+
+    /*!
+     * Get velocity component at a specified position and time.
+     */
+    double getVelocity(double x, double z_plus_d, double time, int comp_idx) const;
```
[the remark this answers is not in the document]
boyceg: Because this is pass by value, const is an implementation decision that should go into the implementation file, not the header.

src/wave_generation/IrregularWaveGenerator.cpp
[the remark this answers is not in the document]
drwells: Sure. Here is an example of a uniform random int distribution with `<random>`:
```cpp
// minimal example of <random>

#include <random>
#include <iostream>

int main()
{
    std::mt19937 std_seq(12345u);
    std::uniform_int_distribution<int> die(1, 6);

    for (std::size_t i = 0; i < 10u; ++i)
    {
        std::cout << "dice roll: " << die(std_seq) << std::endl;
    }
}
```
The standard library provides both a source of pseudo-random numbers (here the standard marsenne twister, mt19937) as well as a distribution: the distribution takes the source of pseudo-randomness and gives us a random number in the distribution. This avoids the problem of shared seeds (and hence shared random state) between objects.

The library RNG function does have the advantage that we can easily generate different seeds for each process, though I don't think the process is deterministic (we call `std::time` at some point).

drwells:
Please do not update branches by merging: it makes the history of the repository difficult to understand. Could you clean this up by rebasing instead?

include/ibamr/IrregularWaveGenerator.h
```diff
@@ -0,0 +1,85 @@
...
+     * \brief Wave Spectrum parameters.
+     *
+     * \param d_omega_begin     Lowest angular frequency in the spectrum [rad/s]
+     * \param d_omega_end       Highest angular frequency in the spectrum [rad/s]
+     * \param d_wave_spectrum   JONSWAP/Bretschneider wave spectrum.
+     * \param d_omega           Angular frequencies of component waves [rad/s]
+     * \param d_wave_number     Wave number of component waves [2\f$ \pi \f$/m]
+     * \param d_amplitude       Amplitude of component waves [m]
+     * \param d_phase           Phase (random) of component waves [rad]
+     */
```
drwells: This comment block has the same issue: this will not show up in doxygen correctly. Would you please split this so that there is a single comment block for each class member that needs to be documented?

amneetb: Kaustubh can you send the SAMRAI doxygen that you followed.

Kaustubh-Khedkar:
Check the pictures from the header file and from SAMRAI doxygen. I have foillowed from here. But, I will split them individually.

include/ibamr/IrregularWaveGenerator.h
boyceg: I am not 100% sure that you cannot make this work for member variables, but I think that Doxygen does want separate annotations for each member variable. I think that you can use `///` so that these can be made one-line annotations, e.g.,
```
    /// Amplitude of component waves [m]
    std::vector<double> d_amplitude;
```
Also, does the implementation require the use of a specific system of units? If not, I would suggest omitting specific units wherever possible. So, e.g.,
```
    /// Amplitude of component waves [length]
    std::vector<double> d_amplitude;
```

amneetb: > I am not 100% sure that you cannot make this work for member variables, but I think that Doxygen does want separate annotations for each member variable. I think that you can use 
 
Ok, let's do one-line annotation then. But it would be nice to confirm this somehow for future usage.  
`///` so that these can be made one-line annotations, e.g.,
> 
> ```
>     /// Amplitude of component waves [m]
>     std::vector<double> d_amplitude;
> ```
> 
> Also, does the implementation require the use of a specific system of units? If not, I would suggest omitting specific units wherever possible. So, e.g.,
> 
> ```
>     /// Amplitude of component waves [length]
>     std::vector<double> d_amplitude;
> ```

Yeah that sounds right. It should be dimensions [length, time] rather than units [m,s].

amneetb:
> Check the pictures from the header file and from SAMRAI doxygen. I have foillowed from here. But, I will split them individually. 

Better to post the "said" picture 😉

include/ibamr/IrregularWaveGenerator.h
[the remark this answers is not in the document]
drwells: I keep forgetting that C++ rule.

drwells:
This looks mostly done aside from adding a unit test and also dealing with the RNG.

I think we should avoid using our own RNG class here - IIRC that class uses its own global state, so if we initialize objects in a different order they will produce different results. Its a much better idea to give this class its own RNG and take a seed as an optional input so that we can get reproducible results without this issue. Ultimately, though, the current approach generates random waves so its not a show stopper. @amneetb @Kaustubh-Khedkar what is your decision?

amneetb:
@drwells : Unit test is not needed. We have verified this code with analytical solutions. That case is in wave_tank/ex0. 

We are happy with RNG and static value of the seed. There is no real need of having a seed value through input file, because this is an arbitrary choice. That is why we selected a fun seed value `1234567` for the class to highlight its arbitrariness 😉

amneetb:
> This looks mostly done aside from adding a unit test and also dealing with the RNG.
> 
> I think we should avoid using our own RNG class here - IIRC that class uses its own global state, so if we initialize objects in a different order they will produce different results. 

I don't think so. If we initialize it with the same seed value, it is guaranteed to generate the same set of random numbers. We have verified it and this is also necesssary for the irregular waves  to work properly. We need the same random phases with `IrregularWaveBcCoef` and `IrregularWaveGenerator` class to produce sensible results.  

Its a much better idea to give this class its own RNG and take a seed as an optional input so that we can get reproducible results without this issue. Ultimately, though, the current approach generates random waves so its not a show stopper. @amneetb @Kaustubh-Khedkar what is your decision?

PR #805, opened 2019-12-18
added restart capabilities to IBFE examples

marshallrdavey:
edited examples to include IBFE restart stuff

PR #810, opened 2019-12-19
Fix IBFESurfaceMethod's restart load.

### LINKED ISSUE #806: restart doesn't work for IBFESurfaceMethod
It looks like #805 revealed that we need to apply a fix similar to #456 to IBFESurfaceMethod: i.e., restart doesn't work with multiple MPI processes because, AFAICT, the vector type is set to PARALLEL and not GHOSTED. I would like to include this and some relevant tests in the release.

drwells:
Fixes #806. Part of #749.

Same as 5ff31c0060 but for IBFESurfaceMethod.

### DIFF
--- a/tests/IBFE/explicit_ex5_2d.restart=40.mpirun=2.output
+++ b/tests/IBFE/explicit_ex5_2d.restart=40.mpirun=2.output
@@ -1,0 +1,261 @@
+
+IBFEMethod: mesh part 0 is using SECOND order LAGRANGE finite elements.
+
+IBHierarchyIntegrator::initializePatchHierarchy(): tag_buffer = 1 3
+
++++++++++++++++++++++++++++++++++++++++++++++++++++
+At beginning of timestep # 40
+Simulation time is 0.025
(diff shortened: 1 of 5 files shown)


tests/IBFE/explicit_ex5_2d.restart=40.mpirun=2.output
```diff
@@ -0,0 +1,261 @@
...
+IBHierarchyIntegrator::postprocessIntegrateHierarchy(): estimated upper bound on IB point displacement since last regrid = 5.875598288e-05
+IBHierarchyIntegrator::advanceHierarchy(): synchronizing updated data
+IBHierarchyIntegrator::advanceHierarchy(): resetting time dependent data
+
+At end       of timestep # 40
+Simulation time is 0.025625
++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+0.025625 0.0158658812
+0.025625 0.00968107699
```
drwells: These values are the same (up to the last digit) as those computed without doing the restart.

drwells:
This bugfix is necessary for the release; its also the same solution that @marshallrdavey wrote in #805. I'm going to merge so that we can get on with the release process.

PR #812, opened 2019-12-19
Fixing 3D bug for finding distance from surface mesh and 3D rotation …

amneetb:
…in Brinkman penalization method


drwells:
@amneetb and I are in a video call right now and he explained the need for these bug fixes.

drwells:
All the tests pass (albeit we don't have any tests for 3D FESurfaceDistanceEvaluator yet) so lets merge.

PR #814, opened 2019-12-20
Collate changelog entries for 0.6.0.

drwells:


drwells:
Part of #749 
Since this is just a documentation fix I will merge as-is.

Issue #784, opened 2019-12-10
Fix tests with undefined behavior


drwells:
Spinoff from #773.

The following tests fail due to memory corruption problems (i.e., valgrind complains):
- [x] spread/spread_01_3d.discontinuous_linear.input
- [x] ~spread/spread_01_3d.piecewise_constant.input~ interpolate/interpolate_01_3d.piecewise_constant.input (the later test shows the problem on master with valgrind)
- [x] adv_diff/adv_diff_01_2d.input

I think these might also have memory problems, but valgrind does not complain on my desktop:
- [x] IBFE/interpolate_velocity_01_2d.b.discontinuous-linear.input
- [x] IBFE/interpolate_velocity_01_2d.b.piecewise-constant.input

drwells:
I don't have time to fix these prior to the release - unless someone volunteers to do it its not going to get done.

drwells:
The problem caught by valgrind in the interpolation test is completely internal to SAMRAI - since we can't fix it lets close this issue.

PR #819, opened 2019-12-23
Remove tests that are known to fail.

drwells:
These correspond to memory problems: see #784. We should remove them from the release candidate to make testing easier (i.e., all tests should pass on supported platforms).


drwells:
Part of #749.

Issue #811, opened 2019-12-19
Add unit tests for FESurfaceDistanceEvaluator


drwells:
@amneetb mentioned that his group already has some tests for this - it would be great if we could add them to IBAMR's test suite. I'm putting this on the next release.

I'll let @amneetb assign this issue to one of his students :)

drwells:
Most of what we need is in IBLevelSet/ex0.

drwells:
Fixed by #827.

PR #827, opened 2020-01-11
Implemented angle-weighted averaging for surface normal

Raamkrishnan:


include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -97,6 +97,11 @@ class FESurfaceDistanceEvaluator
      */
     void mapIntersections();
 
+    /*!
+     * Data structure to map the face normal of the elements, common vertices and edges.
```
amneetb: This comment should be 
`\brief Compute the face normal of the surface elements.`

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -188,14 +185,15 @@ class FESurfaceDistanceEvaluator
     void collectNeighboringPatchElements(int level_number);
 
     /*!
-     * Compute signed distance from volume extracted boundary mesh.
+     * Store the side elements in a set for a volume mesh.
      */
-    void computeSignedDistanceVolExtractedBdryMesh(int n_idx, int d_idx);
+    void sideElemVolExtractedBdryMesh();
```
amneetb: `sideElemVolExtractedBdryMesh()` ==> `extractSideElementsOfVolumeMesh()`

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -188,14 +185,15 @@ class FESurfaceDistanceEvaluator
...
 
     /*!
-     * Compute signed distance from surface mesh.
+     * Get the closest point on a triangle for a given point P and the angle-weighted pseudo normal.
      */
```
amneetb: `\brief Compute the closest point on a triangle for a given Eulerian point P and obtain its corresponding angle-weighted pseudo-normal from the surface mesh elements.`.

Add `\brief` at the beginning of each function description.

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -254,6 +252,25 @@ class FESurfaceDistanceEvaluator
...
     std::map<SAMRAI::pdat::CellIndex<NDIM>, std::set<libMesh::Elem*>, IBTK::CellIndexFortranOrder>
         d_cell_elem_neighbor_map;
+
+    /*!
+     * Map the node and the set of elements which comprise this node
```
amneetb: `comprise` ==> `share`

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -254,6 +252,25 @@ class FESurfaceDistanceEvaluator
...
+     */
+    std::map<std::pair<libMesh::Node*, libMesh::Node*>, std::set<libMesh::Elem*> > d_edge_to_elem;
+
+    /*!
+     * Map the element and its face normal
+     */
+    std::map<libMesh::Elem*, IBTK::VectorNd> d_elem_face_normal;
+
+    /*!
+     * In case of volume mesh, store the side_elements, triangle(edge) for 3D(2D), in a set
```
amneetb: `In case of volume mesh, store the side elements as TRI3(EDGE2) for three(two) dimensions.`

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -254,6 +252,25 @@ class FESurfaceDistanceEvaluator
...
+     */
+    std::set<libMesh::UniquePtr<libMesh::Elem> > d_side_elem_vol_extracted_bdry_msh;
 };
```
amneetb: Lets use `std::vector` here instead of a `std::set` as we do not care about sorting. `std::set` requires more work than a `vector`.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -279,7 +288,7 @@ FESurfaceDistanceEvaluator::mapIntersections()
                                 IndexUtilities::getCellIndex(r_tr.data(), grid_geom, level_ratio));
             if (!ghost_box.intersects(struct_box)) continue;
 
-            // Prepare the required vectors
+                // Prepare the required vectors
 #if (NDIM == 2)
```
amneetb: Is this space generated by clang-format?

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +358,297 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+                    P[d] = patch_X_lower[d] + patch_dx[d] * (static_cast<double>(ci(d) - patch_lower_index(d)) + 0.5);
+                }
+                double min_dist = std::numeric_limits<double>::max();
+
+                // Create a pair to take care of projection and pseudo normals for cells equidistant to
+                // multiple elements.
+                std::vector<std::pair<IBTK::VectorNd, IBTK::VectorNd> > vec_equidistant_pair;
+
+                // clear out the data
+                vec_equidistant_pair.clear();
```
amneetb: Is there a need to call  `vec_equidistant_pair.clear()` here as the `vector` is just created?

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +358,297 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+                double sgn = 0.0;
+                const size_t vec_length = vec_equidistant_pair.size();
+                TBOX_ASSERT(vec_length > 0);
+                IBTK::VectorNd avg_unit_normal, avg_proj;
+                avg_unit_normal.setZero();
+                avg_proj.setZero();
+                for (const auto& elem_vec_pair : vec_equidistant_pair)
+                {
+                    const IBTK::VectorNd proj = elem_vec_pair.first;
+                    ;
```
amneetb: Extraneous `;`

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +358,297 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+                    avg_proj += proj;
+
+                    const IBTK::VectorNd N = elem_vec_pair.second;
+                    avg_unit_normal += N;
+                }
+
+                // Take the average normal and normalize it
+                avg_unit_normal /= (double)vec_length;
```
amneetb: This should be 
` avg_unit_normal /= static_cast<double>(vec_length);`

I forgot to convert to C++ casting from C casting here.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +358,297 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+                avg_unit_normal /= avg_unit_normal.norm();
+
+                // Average the proj point
+                avg_proj /= (double)vec_length;
+
```
amneetb: Use C++ casting style.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +358,297 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
-        computeSignedDistanceVolExtractedBdryMesh(n_idx, d_idx);
+        for (const auto& itr : d_side_elem_vol_extracted_bdry_msh)
+        {
+            Elem* elem = itr.get();
+
+#if !defined(NDEBUG)
+            TBOX_ASSERT(elem->n_nodes() == 3);
+            TBOX_ASSERT(elem->n_edges() == 3);
+#endif
+
```
amneetb: These assert statements should be put in the routine which is generating side elements from volume mesh.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +358,297 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+            std::array<VectorValue<double>, 2> Edge;
+            const libMesh::Point& n0 = elem->point(0);
+            const libMesh::Point& n1 = elem->point(1);
+            Edge[0] = VectorValue<double>((n1(0) - n0(0)), (n1(1) - n0(1)), (n1(2) - n0(2)));
+
```
amneetb: This should be 

`Edge[0] = VectorValue<double>(MAKE_EDGE(n0,n1));` 

Check for correctness.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +358,297 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+            if (NDIM == 2)
+            {
+                Edge[1] = VectorValue<double>(0.0, 0.0, 1.0);
+            }
+
+#if (NDIM == 3)
+            const libMesh::Point& n2 = elem->point(2);
+            Edge[1] = VectorValue<double>((n2(0) - n0(0)), (n2(1) - n0(1)), (n2(2) - n0(2)));
+#endif
```
amneetb: Same here.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +358,297 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+        // loop over each element
+        for (MeshBase::const_element_iterator itr = d_mesh.elements_begin(); itr != d_mesh.elements_end(); ++itr)
+        {
+            Elem* elem = *itr;
+
+            std::array<VectorValue<double>, 2> Edge;
+            const libMesh::Point& n0 = elem->point(0);
+            const libMesh::Point& n1 = elem->point(1);
+            Edge[0] = VectorValue<double>((n1(0) - n0(0)), (n1(1) - n0(1)), (n1(2) - n0(2)));
+
```
amneetb: Use 'MAKE_EDGE` here also.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +358,297 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+            if (NDIM == 2)
+            {
+                Edge[1] = VectorValue<double>(0.0, 0.0, 1.0);
+            }
+
+#if (NDIM == 3)
+            const libMesh::Point& n2 = elem->point(2);
+            Edge[1] = VectorValue<double>((n2(0) - n0(0)), (n2(1) - n0(1)), (n2(2) - n0(2)));
+#endif
```
amneetb: And here too.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +358,297 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+            d_elem_face_normal[elem] = normal;
+
+#if (NDIM == 3)
 
+#if !defined(NDEBUG)
+            TBOX_ASSERT(elem->n_nodes() == 3);
+            TBOX_ASSERT(elem->n_edges() == 3);
+#endif
+            // for each node, store the current element in the node to element map.
```
amneetb: Do we need assertion here?

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -404,6 +684,7 @@ FESurfaceDistanceEvaluator::updateSignAwayFromInterface(int D_idx,
     int n_global_updates = 1;
     while (n_global_updates > 0)
     {
+        printf("n_global_updates = %d\n", n_global_updates);
         ghost_fill_sched->fillData(/*time*/ 0.0);
```
amneetb: Remove the debugging print statement here and at other places.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -686,18 +967,26 @@ FESurfaceDistanceEvaluator::checkIntersection3D(const IBTK::Vector3d& box_center
...
-    E1 << vert2(0) - B(0), vert2(1) - B(1), vert2(2) - B(2);
-    D = B - P;
+    const libMesh::Point& n0 = elem->point(0);
+    const libMesh::Point& n1 = elem->point(1);
+    const libMesh::Point& n2 = elem->point(2);
+
+    const IBTK::Vector3d vert0(n0(0), n0(1), n0(2));
+    const IBTK::Vector3d vert1(n1(0), n1(1), n1(2));
+    const IBTK::Vector3d vert2(n2(0), n2(1), n2(2));
+
```
amneetb: There is no real need for `vert0`, `vert1`, and `vert2` here.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -686,18 +967,26 @@ FESurfaceDistanceEvaluator::checkIntersection3D(const IBTK::Vector3d& box_center
...
+    const IBTK::Vector3d B(vert0(0), vert0(1), vert0(2));
+    const IBTK::Vector3d D = B - P;
```
amneetb: Create `B` from `n0` directly.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -686,18 +967,26 @@ FESurfaceDistanceEvaluator::checkIntersection3D(const IBTK::Vector3d& box_center
...
+
+    const IBTK::Vector3d E0((vert1(0) - B(0)), (vert1(1) - B(1)), (vert1(2) - B(2)));
```
amneetb: Use `const IBTK::Vector3d E0(MAKE_EDGE(B,n1));`

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -686,18 +967,26 @@ FESurfaceDistanceEvaluator::checkIntersection3D(const IBTK::Vector3d& box_center
...
+    const IBTK::Vector3d E1((vert2(0) - B(0)), (vert2(1) - B(1)), (vert2(2) - B(2)));
```
amneetb: Same here.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -709,6 +998,21 @@ FESurfaceDistanceEvaluator::getClosestPoint3D(const IBTK::Vector3d& P,
...
+        EDGE_0,
+        EDGE_1,
+        EDGE_2,
+        VERTEX_0,
+        VERTEX_1,
+        VERTEX_2,
+        UNKNOWN_FEATURE = -1
+    };
+
+    TRIANGLEFEATURE which_feature = UNKNOWN_FEATURE;
```
amneetb: Let's replace `which_feature` with just `feature`.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -877,21 +1228,175 @@ FESurfaceDistanceEvaluator::getClosestPoint3D(const IBTK::Vector3d& P,
...
+            Elem* nbr_elem = itr;
+            angle_weighted_normal += d_elem_face_normal[nbr_elem];
+        }
+        angle_weighted_normal /= angle_weighted_normal.norm();
+        break;
+    case EDGE_2:
+        edge_nodes[0] = elem->node_ptr(2);
+        edge_nodes[1] = elem->node_ptr(0);
+        std::sort(edge_nodes.begin(), edge_nodes.end());
+        std::sort(edge_nodes.begin(), edge_nodes.end());
```
amneetb: Extaneous sorting is getting done here.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -922,422 +1427,131 @@ FESurfaceDistanceEvaluator::collectNeighboringPatchElements(int level_number)
...
+    {
+        Elem* elem = *itr;
+        // Loop over the sides of the element. If it has no
+        // neighbors on a side, then it MUST live on the boundary
+        // of the mesh
+        for (unsigned int s = 0; s < elem->n_sides(); ++s)
+        {
+            if (elem->neighbor_ptr(s) != nullptr) continue;
+            UniquePtr<Elem> side_elem = elem->build_side_ptr(s, /*proxy*/ false);
+            d_side_elem_vol_extracted_bdry_msh.insert(std::move(side_elem));
```
amneetb: Use assertions related to number of nodes and edges here for 2D  and 3D appropriately.

src/level_set/FESurfaceDistanceEvaluator.cpp
amneetb: There is an extra ; at line 489 also.

drwells:
@Raamkrishnan Let us know if you have any questions or need help finishing this :)

amneetb:
@drwells , @Raamkrishnan  is working on writing units tests for this class including order of convergence studies. One thing he can actually use your help is in figuring out how to create a sphere with TET4 elements using libMesh constructs (using TETGEN interface within libMesh). This will avoid having us to commit mesh files for the test.

Backdrop for TET4 is that it will help to extract TRI3 surface elements, which is what this class really wants. Currently libMesh only allows for  HEX8 elements, which produces QUAD4 surface elements.

 @drwells : Could you provide us some code for generating a TET4 sphere?

knepley:
@amneetb For this test, you could use https://www.mcs.anl.gov/petsc/petsc-master/docs/manualpages/DMPLEX/DMPlexCreateSphereMesh.html which should now be convertible to a valid LibMesh. You can refine it any number of times with DMRefine(), and fill in the ball with DMGenerate (if CTetGen is installed).

amneetb:
@knepley Is there an example that we can take a look at? I have never used `DMPlex` before, but I have heard great things about it ;-) 

BTW, does it generate a surface mesh or volumetric mesh for the sphere?

drwells:
> This will avoid having us to commit mesh files for the test.

Its good to avoid this in principle but the alternatives are going to take a lot longer to implement. Could we generate an input file that is just O(1000) lines? How big of a mesh do we need?

drwells:
@amneetb Do you only need a surface mesh? If so, we have a much easier solution available:
1. generate the HEX8 volumetric mesh
2. extract the QUAD4 surface mesh
3. use `MeshTools::modification::all_tri` to convert quadrilaterals to triangles

I am pretty sure that this will work if you only need the volume mesh and it is even easier than the other options written here.

amneetb:
> @amneetb Do you only need a surface mesh? If so, we have a much easier solution available:
> 
>     1. generate the HEX8 volumetric mesh
>

``
Mesh solid_mesh(init.comm(), NDIM);
MeshTools::Generation::build_sphere(solid_mesh, R, num_segments, HEX8);

``


>     2. extract the QUAD4 surface mesh
>
        Using something like this?

``
        BoundaryMesh boundary_mesh(solid_mesh.comm(), solid_mesh.mesh_dimension() - 1);
        solid_mesh.boundary_info->sync(boundary_mesh);
        boundary_mesh.prepare_for_use();

``
 
 
>     3. use `MeshTools::modification::all_tri` to convert quadrilaterals to triangles
> 

Do you mean using something like:

``
MeshTools::Modification::all_tri(boundary_mesh);	

``

 
> I am pretty sure that this will work if you only need the ~volume~surface mesh and it is even easier than the other options written here.
> 
> edit: fix word

drwells:
Yup! As long as libMesh performs as advertised that should do it.

knepley:
@amneetb It sounds like you have it sorted in LibMesh, but Plex can make the surface or volume mesh, whichever you need. The surface is one line, and then whatever you need to convert it to LibMesh.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +348,229 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+        Pointer<CellData<NDIM, double> > d_data = patch->getPatchData(d_idx);
+
+        // Note that we only work with cells that satisfy the intersecting criteria.
+        for (Box<NDIM>::Iterator it(patch_box); it; it++)
+        {
+            CellIndex<NDIM> ci(it());
+            const bool contains_key = (d_cell_elem_neighbor_map.find(ci) != d_cell_elem_neighbor_map.end());
+            if (contains_key)
+            {
+                std::set<Elem*> elem_set = d_cell_elem_neighbor_map[ci];
```
amneetb: This should be a reference
`std::set<Elem*>& elem_set = d_cell_elem_neighbor_map[ci];`

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -709,6 +913,21 @@ FESurfaceDistanceEvaluator::getClosestPoint3D(const IBTK::Vector3d& P,
     double s = b * e - c * d;
     double t = b * d - a * e;
```
amneetb: Lines 

https://github.com/IBAMR/IBAMR/blob/ramakrishnan/fe-surface-distance-evaluator/src/level_set/FESurfaceDistanceEvaluator.cpp#L897-L900

We should use `N0, N1, N2` instead of `n0,n1,n2` here:

```
    const libMesh::Point& N0 = elem->point(0);
    const libMesh::Point& N1 = elem->point(1);
    const libMesh::Point& N2 = elem->point(2);
    const IBTK::Vector3d B(N0(0), N0(1), N0(2));
    const IBTK::Vector3d D = B - P;

    const IBTK::Vector3d E0(MAKE_EDGE(B, N1));
    const IBTK::Vector3d E1(MAKE_EDGE(B, N2));
```

as we use `n0,n1,n2` later down to compute nodes of `nbr_elem` nodal points.

boyceg: How are `N0`, `N1`, `N2` related to `n0`, `n1`, `n2`?

In general I would recommend avoiding having variable names differ only by capitalization; it becomes easy to make mistakes. One exception is for cases like nonlinear mechanics where you might indicate corresponding quantities in the reference and current configurations by switching between uppercase/lowercase.

amneetb: `N0`, `N1`, `N2` are current element's node pointers and  `n0`, `n1`, `n2` are its neighboring element's pointer. Earlier both used the same variable name, which could also lead to a mistake in the future. 

How about using `elem_n0`, `elem_n1`, `elem_n2` instead?

boyceg: Sure. Maybe prefix the "neighbor" ones instead of the "local" ones?

amneetb: Agreed. @Raamkrishnan : Let's use `nbr_n0`,`nbr_n1`, `nbr_n1` at that place along with the prior suggestion.

Raamkrishnan: Hi, I incorporated the changes and updated it.

amneetb:
The log_file_name indicated in the input file for the tests should not be the same as what is provided as a reference file to compare the solution against. 

I would use the log_file_name as “FESurfaceDistanceTest.log” for all the cases.

drwells:
Would you please move the test to `level_set` instead? I would prefer that we more-or-less mirror the layout of the source directory unless there is something (e.g., spreading) for which we will write dozens of tests.

drwells:
Is there a reason to require 16 processors for a test? If it doesn't demonstrate some bug or other issue dependent on the number of processors I would prefer to use fewer so that we can still run the test suite on laptops.

Raamkrishnan:
> Would you please move the test to `level_set` instead? I would prefer that we more-or-less mirror the layout of the source directory unless there is something (e.g., spreading) for which we will write dozens of tests.

Done

Raamkrishnan:
> Is there a reason to require 16 processors for a test? If it doesn't demonstrate some bug or other issue dependent on the number of processors I would prefer to use fewer so that we can still run the test suite on laptops.

Now I have every test with single and 8 processors.

amneetb:
@drwells --- ping

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -877,21 +1143,174 @@ FESurfaceDistanceEvaluator::getClosestPoint3D(const IBTK::Vector3d& P,
...
+                nbr_e0 << MAKE_EDGE(nbr_n2, nbr_n0);
+                nbr_e1 << MAKE_EDGE(nbr_n2, nbr_n1);
+            }
+            const double nbr_e0_norm = nbr_e0.norm();
+            const double nbr_e1_norm = nbr_e1.norm();
+            const double angle = acos(nbr_e0.dot(nbr_e1) / (nbr_e0_norm * nbr_e1_norm));
+            angle_weighted_normal += (angle * d_elem_face_normal[nbr_elem]);
+        }
+        angle_weighted_normal /= angle_weighted_normal.norm();
+        break;
```
drwells:
```suggestion
        break;
    case UNKNOWN_FEATURE:
    default:
        TBOX_ERROR("Unknown geometric feature");
```

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -357,18 +348,229 @@ FESurfaceDistanceEvaluator::computeSignedDistance(int n_idx, int d_idx)
...
+                {
+                    IBTK::VectorNd v, w, proj;
+                    double dist = std::numeric_limits<double>::max();
+
+                    // Get the nodes.
+                    const libMesh::Point& n0 = elem->point(0);
+                    const libMesh::Point& n1 = elem->point(1);
+#if (NDIM == 3)
+                    const libMesh::Point& n2 = elem->point(2);
+#endif
```
drwells: n0, n1, and n2 aren't used in 3D. Could you remove n2 and then move n1 and n2 inside the `#ifdef` block below?

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -106,25 +109,22 @@ line_equation(const IBTK::Vector3d& coord, const libMesh::Point& n0, const libMe
     return (y1 - y0) * coord(0) + (x0 - x1) * coord(1) + (x1 * y0 - x0 * y1);
 } // line_equation
 
+#define MAKE_EDGE(n0, n1) n1(0) - n0(0), n1(1) - n0(1), n1(2) - n0(2)
```
drwells: Why is this a macro? It would make more sense for this to be an inline function in an anonymous namespace.

tests/level_set/fe_surface_distance.cpp
```diff
@@ -0,0 +1,432 @@
...
+
+        // Parse command line options, set some standard options from the input
+        // file, initialize the restart database (if this is a restarted run),
+        // and enable file logging.
+        Pointer<AppInitializer> app_initializer = new AppInitializer(argc, argv, "FESurfaceDistance.log");
+        Pointer<Database> input_db = app_initializer->getInputDatabase();
+
+        // Get various standard options set in the input file.
+        const bool dump_viz_data = app_initializer->dumpVizData();
+        const int viz_dump_interval = app_initializer->getVizDumpInterval();
```
drwells: viz_dump_interval, uses_visit, uses_exodus, dump_timer_data, and timer_dump_interval are all unused. Would you please remove them?

src/level_set/FESurfaceDistanceEvaluator.cpp
drwells: There might be a better error message, but its quite risky to not cover all cases and have a `default:` at the end.

boyceg: `switch` statements should always have a `default` case.

drwells:
Fixes #811.

src/level_set/FESurfaceDistanceEvaluator.cpp
Raamkrishnan: I make use of macro expansion for both IBTK::Vector3d and the libMesh::VectorValue. 

''' const IBTK::Vector3d E0(MAKE_EDGE(B, n1)); '''

''' Edge[0] = VectorValue<double>(MAKE_EDGE(n0, n1)); '''

For this, I have to write two separate inline functions.

Raamkrishnan: I use single MAKE_EDGE macro for IBTK::Vector3d, libMesh::VectorValue and libMesh::Point.

Code snippet is here:
```
const IBTK::Vector3d E0(MAKE_EDGE(B, n1));
Edge[0] = VectorValue<double>(MAKE_EDGE(n0, n1));
nbr_e0 << MAKE_EDGE(nbr_n0, nbr_n1); //nbr_n0 and nbr_n1 are libmesh::Point
```
In this case, I may have to write separate inline functions.

amneetb: Macro is fine.

src/level_set/FESurfaceDistanceEvaluator.cpp
Raamkrishnan: I added the default case.

amneetb: We have a safety catch of `TBOX_ASSERT(feature != UNKNOWN_FEATURE)` upstream of `switch`.

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -877,21 +1151,177 @@ FESurfaceDistanceEvaluator::getClosestPoint3D(const IBTK::Vector3d& P,
...
                 }
             }
         }
     }
 
-    return B + s * E0 + t * E1;
-} // getClosestPoint3D
+#if !defined(NDEBUG)
+    TBOX_ASSERT(feature != UNKNOWN_FEATURE);
+    TBOX_ASSERT(s >= 0.0 && s <= 1.0);
```
amneetb: Here.

src/level_set/FESurfaceDistanceEvaluator.cpp
drwells: I would prefer to still use a template function rather than this but its not a big deal; lets leave it if @amneetb is happy.

src/level_set/FESurfaceDistanceEvaluator.cpp
drwells: Sure, but its still a compiler warning to not define all switch cases.

amneetb: I see. Now all 8 switch cases are defined, so hopefully there are no more compiler warnings for this.

src/level_set/FESurfaceDistanceEvaluator.cpp
boyceg: We should prefer using functions instead of macros. In this case a templated function should work and is not much more code.

src/level_set/FESurfaceDistanceEvaluator.cpp
boyceg: I think it is good practice to always include a `default` case. Reasons include addressing the situation in which items are added to the `enum` at some point in the future, and in which case statements are _removed_ (either accidentally or on purpose). It is easier to debug these things if there is always a defined default behavior other than silently passing through the `switch`.

src/level_set/FESurfaceDistanceEvaluator.cpp
Raamkrishnan: I replaced the macro with the function template and updated the code.

src/level_set/FESurfaceDistanceEvaluator.cpp
drwells: "It's not paranoia if they are actually out to get you"

amneetb: The template function made it look even uglier than the original in-place substitution of `n0-n1`. Hence, I had recommended using a macro. 

But there may be other advantages to switching to a function than a macro.

drwells:
Is this done?

amneetb:
I think so. @Raamkrishnan ?

Raamkrishnan:
Yes. I incorporated all the requested changes in the code and pushed it here. I did the auto test as well.

drwells:
If @amneetb is happy with the present implementation then lets merge :)

amneetb:
Since it is Friday, let's merge it ;-)

drwells:
Want to hit 'approve' first? There is still a requested change from you.

PR #828, opened 2020-01-13
Amin kolah/add surface jumps

AminKolah:
This is a successor to [#637](https://github.com/IBAMR/IBAMR/pull/637) which by itself was a successor to [#240](https://github.com/IBAMR/IBAMR/pull/240).


AminKolah:
There are two additional input files ( input2d.iim and input3d.iim) for ex5 of IBFE/explicit. These input files are exclusively used for solving the codim-1 version of the flow past a cylinder using IBFESurfaceMethod. The options can be 1) spreading the full force using regularized Dirac delta function or 2) imposing the pressure and viscous jump or 3) all the combinations between the two. If all jump are computed, there is an option for computing the force coefficients using the exterior traction.

examples/IBFE/explicit/ex5/example.cpp
```diff
@@ -117,6 +117,9 @@ struct TetherData
     }
 };
 
+bool use_boundary_mesh 			  = false;
+bool compute_fluid_traction 	  = false;
```
drwells: Please run clang-format on your code before we continue the review.

boyceg:
@AminKolah do you have the git `clang-format` pre-commit hook setup? That would only apply clang-format to the changed lines, which is often better than running it on everything. However, it would mean re-doing the commits. Basically you would do a soft reset and then re-add the changes.

abarret:
> @AminKolah do you have the git `clang-format` pre-commit hook setup? That would only apply clang-format to the changed lines, which is often better than running it on everything. However, it would mean re-doing the commits. Basically you would do a soft reset and then re-add the changes.

@AminKolah https://stackoverflow.com/questions/43974371/run-git-clang-format-on-series-of-git-commits

AminKolah:
Yes I've been using  git clang-format-6.0 on ubuntu  but it seems like the changes are only made to *.cpp and *.h files. Would I need to do something special for the input files?

examples/IBFE/explicit/ex5/example.cpp
AminKolah: Do you see these spaces in the *.h and *.cpp files too? I think the version of clang-format I used did not fix the input files. Does the pre-commit hook automatically do that for all files?

abarret:
> Yes I've been using git clang-format-6.0 on ubuntu but it seems like the changes are only made to *.cpp and *.h files. Would I need to do something special for the input files?

Clang format won't fix the input files. The rule for formatting of input files that I've been using is "do your best."

boyceg:
> Yes I've been using git clang-format-6.0 on ubuntu but it seems like the changes are only made to *.cpp and *.h files. Would I need to do something special for the input files?

It looks like there may be something fishy with your setup --- the lines that @drwells highlighted are in a `.cpp` file.

AminKolah:
> > Yes I've been using git clang-format-6.0 on ubuntu but it seems like the changes are only made to *.cpp and *.h files. Would I need to do something special for the input files?
> 
> It looks like there may be something fishy with your setup --- the lines that @drwells highlighted are in a `.cpp` file.

I think I just fixed it...@

> > Yes I've been using git clang-format-6.0 on ubuntu but it seems like the changes are only made to *.cpp and *.h files. Would I need to do something special for the input files?
> 
> Clang format won't fix the input files. The rule for formatting of input files that I've been using is "do your best."

Thanks Aaron. That was a relief..good to know that I don't have to bang my head around any more trying to find a way to make the clang-format work for the input files

AminKolah:
> > Yes I've been using git clang-format-6.0 on ubuntu but it seems like the changes are only made to *.cpp and *.h files. Would I need to do something special for the input files?
> 
> It looks like there may be something fishy with your setup --- the lines that @drwells highlighted are in a `.cpp` file.

I think I just fixed this?  @drwells can you double-check?

drwells:
Yup, it looks like you ran it. Would you please now remove the `.patch` file?

AminKolah:
> Yup, it looks like you ran it. Would you please now remove the `.patch` file?

Ahh sorry, just did.

tests/IBFE/explicit_ex5_2d.mpirun=2.input
```diff
@@ -2,6 +2,7 @@
 Re = 200.0
 MU = 1.0/Re
 RHO = 1.0
+R = 0.5        								   // radius of the cylinder
```
drwells: Why was this file modified? This file corresponds to `tests/IBFE/explicit_ex5.cpp` which has not been changed.

tests/IBFE/explicit_ex5_2d.restart=40.mpirun=2.input
```diff
@@ -2,6 +2,7 @@
 Re = 200.0
 MU = 1.0/Re
 RHO = 1.0
+R = 0.5        								   // radius of the cylinder
```
drwells: same

tests/IBFE/explicit_ex5_3d.mpirun=2.input
```diff
@@ -2,7 +2,7 @@
 Re = 400.0
 MU = 1.0/Re
 RHO = 1.0
-
+R = 0.5        											// radius of the cylinder
```
drwells: same

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2888,17 +2888,74 @@ FEDataManager::updateQuadPointCountData(const int coarsest_ln, const int finest_
 std::vector<std::pair<Point, Point> >*
 FEDataManager::computeActiveElementBoundingBoxes()
 {
+    // Get the necessary FE data.
```
drwells: Why was this method reverted to the old implementation?

tests/IBFE/explicit_ex5_2d.mpirun=2.input
AminKolah: interesting..I actually didn't manually change this. How are these tests generated? 
In the other two iim input files there is an option for prescribing the radius, and there is a call for it in the code. We can set the default value to 0.5 in example.cpp

ibtk/src/lagrangian/FEDataManager.cpp
AminKolah: The only difference between the file here and the `master` version is in `computeActiveElementBoundingBoxes` function...I didn't dig too deep into this but clearly if I use the version on the master branch I get wrong results. So I think the new changes made in `FEDataManager.cpp ` on the `master` branch might have to be reviewed to make sure it works for surface mesh as well.

drwells: This might be fixed by some bugs I addressed in #829. If not could you please put this in a separate PR?

tests/IBFE/explicit_ex5_2d.mpirun=2.input
drwells: These tests are based on the examples but have their own source files, (here we have https://github.com/IBAMR/IBAMR/blob/master/tests/IBFE/explicit_ex5.cpp). See https://github.com/IBAMR/IBAMR/wiki/Tests-for-IBAMR for more information.

ibtk/src/lagrangian/FEDataManager.cpp
AminKolah: I just tried it with the newest version on master but the problem persists. It's only an issue when running in parallel btw...yup, I can create a separate PR for it.

drwells: Thanks.

I think we can fix the problem with either #829 or another fix. I'm surprised that ex5 works at all in parallel given this issue. I'll spend some time today trying to get to the bottom of this.

tests/IBFE/explicit_ex5_2d.mpirun=2.input
AminKolah: Oh interesting. I do not remember going in there to change anything. It makes sense to revert them back to the `master` version.

drwells:
@AminKolah we can use `git diff` again later, but it would be easier for now if we rebased against master instead of merged to deal with conflicts. Do you know how to do this?

drwells:
At a high level I have three main concerns about this PR:
1. Having to manually deal with our own collections of vectors for each relevant `libMesh::System` is bad. We've talked about this in the past and this PR makes the situation much worse. Its going to be hard to maintain and modify this code when we have to reinitialize ten different `System`s manually. I suggest we take the time to figure out how to systematically deal with this issue now instead of writing
```cpp
    d_TAU_out_systems.resize(d_num_parts);
    d_TAU_out_half_vecs.resize(d_num_parts);
    d_TAU_out_IB_ghost_vecs.resize(d_num_parts);
```
a dozen different times in subtly different ways.

2. The functions this PR changes are now huge and hard to understand. In particular, `IBFESurfaceMethod::interpolateVelocity` is now 500 lines long (as opposed to 250 lines on master). Could we divide these functions into free functions (i.e., functions not in a class) called by IBFESurfaceMethod?

3. Related to 2: there are no unit tests for this new functionality which is, like 1., going to make maintenance harder. For example: at some point we need to get rid of all the `localize()` calls in `IBFESurfaceMethod` and we are making that much harder by not adding any new tests.

AminKolah:
> @AminKolah we can use `git diff` again later, but it would be easier for now if we rebased against master instead of merged to deal with conflicts. Do you know how to do this?

I can revert the last commit and do a `git rebase` instead. I've had some issues with doing git rebase though (maybe the way I do it is wrong). Normally what happens most of the times is that the rebase doesn't pick up all the changes from `master` and when I try `git rebase -Xours` to prioritize master vs. feature branch, the history seems to get messed up. I'd be curious to know if you know an easy way to avoid both of these issues. Thanks.

AminKolah:
> At a high level I have three main concerns about this PR:
> 
> 1. Having to manually deal with our own collections of vectors for each relevant `libMesh::System` is bad. We've talked about this in the past and this PR makes the situation much worse. Its going to be hard to maintain and modify this code when we have to reinitialize ten different `System`s manually. I suggest we take the time to figure out how to systematically deal with this issue now instead of writing
> 
> ```c++
>     d_TAU_out_systems.resize(d_num_parts);
>     d_TAU_out_half_vecs.resize(d_num_parts);
>     d_TAU_out_IB_ghost_vecs.resize(d_num_parts);
> ```
> 
> a dozen different times in subtly different ways.
> 
> 1. The functions this PR changes are now huge and hard to understand. In particular, `IBFESurfaceMethod::interpolateVelocity` is now 500 lines long (as opposed to 250 lines on master). Could we divide these functions into free functions (i.e., functions not in a class) called by IBFESurfaceMethod?
> 2. Related to 2: there are no unit tests for this new functionality which is, like 1., going to make maintenance harder. For example: at some point we need to get rid of all the `localize()` calls in `IBFESurfaceMethod` and we are making that much harder by not adding any new tests.

I can work on some of these (at least the ones that are readily comprehensible for me to do). One very important thing to keep in mind when making large changes to classes like this is to make sure the code is going to be as robust and work perfectly and exactly for ALL the cases that it's been tested before.

drwells:
I agree: that is exactly why I listed 3.

drwells:
@AminKolah we don't have to revert anything, its just that in future it would be easier to update this against master by running `git rebase master`. I recommend resolving conflicts manually since its difficult to know which change is correct automatically.

AminKolah:
> @AminKolah we don't have to revert anything, its just that in future it would be easier to update this against master by running `git rebase master`. I recommend resolving conflicts manually since its difficult to know which change is correct automatically.

I'll have to show you what I mean sometime on a screen. I know `git rebase` is supposed to make life easy not scary:)

drwells:
@AminKolah We should eventually add the scratch hierarchy to this class but its not related to the original purpose of this PR - would you please revert that changeset?

AminKolah:
I imagined it shouldn't make it harder to review since the most recent changes to this class have made it very similar to the `IBFEMethod` class? The only places to look for new changes would be related to the jump systems/calculations.

drwells:
Its going to be too hard to figure out if any new problems are due to the new features for surface jump or issues due to the scratch hierarchy - we really need those to be in separate PRs. Additionally, since the logic for the scratch hierarchy is basically the same for both classes, it really needs to be in `IBFEMethodStrategy` or whatever we end up calling the common base class of both of these classes - i.e., it should not be added to `IBFESurfaceMethod`.

AminKolah:
> At a high level I have three main concerns about this PR:
> 
> 1. Having to manually deal with our own collections of vectors for each relevant `libMesh::System` is bad. We've talked about this in the past and this PR makes the situation much worse. Its going to be hard to maintain and modify this code when we have to reinitialize ten different `System`s manually. I suggest we take the time to figure out how to systematically deal with this issue now instead of writing
> 
> ```c++
>     d_TAU_out_systems.resize(d_num_parts);
>     d_TAU_out_half_vecs.resize(d_num_parts);
>     d_TAU_out_IB_ghost_vecs.resize(d_num_parts);
> ```
> 
> a dozen different times in subtly different ways.
> 
> 1. The functions this PR changes are now huge and hard to understand. In particular, `IBFESurfaceMethod::interpolateVelocity` is now 500 lines long (as opposed to 250 lines on master). Could we divide these functions into free functions (i.e., functions not in a class) called by IBFESurfaceMethod?
> 2. Related to 2: there are no unit tests for this new functionality which is, like 1., going to make maintenance harder. For example: at some point we need to get rid of all the `localize()` calls in `IBFESurfaceMethod` and we are making that much harder by not adding any new tests.

I'm going to try to address number 2 and 3 here. Let me know if you or @boyceg can come up with a way to not have to write out the re-initialization for "ten different `systems`". I personally think this is not a priority, at least for now?

AminKolah:
> Its going to be too hard to figure out if any new problems are due to the new features for surface jump or issues due to the scratch hierarchy - we really need those to be in separate PRs. Additionally, since the logic for the scratch hierarchy is basically the same for both classes, it really needs to be in `IBFEMethodStrategy` or whatever we end up calling the common base class of both of these classes - i.e., it should not be added to `IBFESurfaceMethod`.

I agree that they should eventually be in a `IBFEMethodStrategy`. In fact I was going to discuss that with you and Boyce. I have already tested the current implementation for both stationary and prescribed motions. Specific examples for testing smaller components like the interpolation is also a good idea though.

I am currently working on a few more branches off this branch but they none of them are essential for the `merge` with `master.` So I can revert this PR but I would rather keep the new changes in those new sub-branches since I know they are working correctly. Do you have a suggestion how to go about accomplishing that?

I don't mind if we go back to the previous version for the purpose of just checking the IIM related code. I just thought I would be making your life easier by making the code more compatible with your implementation in `IBFEMethod.cpp`, otherwise there is nothing new in terms of the actual validation since last time we looked at this :)

drwells:
> I just thought I would be making your life easier by making the code more compatible with your implementation in IBFEMethod.cpp

I do appreciate the gesture but the main thing that will make this easier to understand is lowering the changed line count, not increasing it :)

drwells:
> I am currently working on a few more branches off this branch but they none of them are essential for the merge with master. So I can revert this PR but I would rather keep the new changes in those new sub-branches since I know they are working correctly. Do you have a suggestion how to go about accomplishing that?

The way I usually save such changes is by creating a branch off of a branch
> git checkout AminKolah/add-surface-jumps
> git checkout AminKolah/add-surface-jumps-with-scratch-hierarchy
or whatever you want to call it.

drwells:
Finally: yes, 2 and 3 are the most important so we can sacrifice 1 to achieve those.

AminKolah:
I decided to keep the new commits on this branch just in case , and instead of reverting create a new branch from the commit that we were last discussing here. On that new branch under `AminKolah/merge-surface-jumps` I also did a git rebase with `master `and the history seems pretty clean now including the most recent changes on the master branch...I will open a new pull request for that branch to continue discussing the merge.

PR #829, opened 2020-01-13
Redo collect active elem

drwells:
Followup to #826: I replaced the recursive algorithm for computing which elements lie on a given patch with one based on bounding boxes. The bounding boxes are computed from the locations of the displaced nodes and quadrature points of each element.

This is a bit more conservative than the old approach since we use bounding boxes instead of excluding elements that do not currently have any quadrature points on the current element.


ibtk/src/utilities/libmesh_utilities.cpp
```diff
@@ -301,7 +301,6 @@ std::vector<libMeshWrappers::BoundingBox>
 get_local_active_element_bounding_boxes(const libMesh::MeshBase& mesh, const libMesh::System& X_system)
 {
     const unsigned int dim = mesh.mesh_dimension();
```
drwells: @AminKolah here's at least one bug with the surface mesh code: `mesh_dimension()` is the topological dimension but for bounding boxes we want the spatial dimension. I'll fix this.

drwells:
I have no idea how the ex5 test managed to ever work when things were this broken: I'll check that next.

Either way I fixed a bunch of bugs regarding the partitioning of surface meshes: things should now work correctly. I added a bunch of new tests and also checked things visually. Here is a nice picture of our IB partitioning for a circle

visit0000

and another in 3D just for fun :)
visit0001

drwells:
We now have over 200 tests :tada:

ibtk/src/utilities/libmesh_utilities.cpp
```diff
@@ -300,8 +301,8 @@ get_local_active_element_bounding_boxes(const libMesh::MeshBase& mesh,
 std::vector<libMeshWrappers::BoundingBox>
 get_local_active_element_bounding_boxes(const libMesh::MeshBase& mesh, const libMesh::System& X_system)
 {
-    const unsigned int dim = mesh.mesh_dimension();
-    TBOX_ASSERT(dim <= LIBMESH_DIM);
+    const unsigned int spacedim = mesh.spatial_dimension();
```
boyceg: `mesh_dim` ?

drwells: I'm not sure I understand the question: what do you have in mind?

boyceg: The way that I think of it, at least, is that `NDIM` is the "spatial dimension". So I am suggesting:
```suggestion
    const unsigned int mesh_dim = mesh.spatial_dimension();
```

drwells: I agree with the first part but not the second. `mesh.mesh_dimension()` is the topological dimension of the mesh (which is not what we want): `mesh_dim` looks too much like that so I would prefer to write `spacedim` or `space_dimension`

AminKolah:
> I have no idea how the ex5 test managed to ever work when things were this broken: I'll check that next.
> 
> Either way I fixed a bunch of bugs regarding the partitioning of surface meshes: things should now work correctly. I added a bunch of new tests and also checked things visually. Here is a nice picture of our IB partitioning for a circle
> 
> visit0000
> 
> and another in 3D just for fun :)
> visit0001

Nice! ..I never checked the codim-0 version of the ex5 with IBFEMethod (at least not recently), but the one with surface mesh was definitely working as expected on my branch...I did realize that in some parts of the libmesh_utilities.h it was assumed that NDIM the mesh dimensions for both solid and the fluid mesh (which would be only true for volumetric mesh)

drwells:
@boyceg approved so I will merge so that @AminKolah can get on with his life :)

Issue #830, opened 2020-01-16
Reimplement force spreading accumulation from ghost cells


drwells:
Background: when we do force spreading, on a given patch we calculate contributions from points outside the patch that can spread into the patch: i.e., if a point can spread into the patch interior we compute its contribution. A long time ago IBAMR instead had each patch compute its own contributions (including values spread into ghost cells) and then did a parallel reduction to get the correct values on each locally owned patch interior.

I talked about this with @boyceg yesterday and, after mulling it over this morning, I am convinced that we should switch back to something like the old algorithm. A fundamental problem with IBFE is that we end up with a huge number of elements on more than one processor: a mesh may have 200k elements, but since elements will show up on multiple patches we will have (e.g., in ex4 in 3D) 400k elements to loop over when we do force spreading. Its very difficult to load balance this correctly since SAMRAI does not have a notion of additional work done at interfaces, only workload per cell.

One way to think of this is is to note that if we have a ghost width of 3, then 16 x 16 x 16 = 4096 becomes 22 x 22 x 22 = 10648, but with a ghost width of 1 we have 18 x 18 x 18 = 5832. That is, we do nearly twice as much work when we consider the full ghost region as when we consider a ghost region with one adjacent cell.

The reason this was (rightly) removed previously was that the summation was done in an arbitrary order - values were not reproducible between runs since values were accumulated in an order dependent on traffic. AFAICT SAMRAI does not implement the algorithm we need, which is roughly as follows:
1. As every processor has a copy of the complete set of boxes, each processor can determine which processes can contribute values into its own patch interiors and also determine where it needs to send its own ghost data.
2. By growing boxes and computing intersections each processor can compute the regions in index space where there are overlaps: e.g., processor 0 can determine which processors have relevant ghost values to it and where its own ghost values should go.
3. Pass messages with `MPI_Isend` and `MPI_Irecv` into dedicated buffers per processor. Do an MPI_Wait before proceeding.
4. Sum passed ghost values into array interiors in a predetermined order independent of traffic.

I don't know what equivalent algorithm SAMRAI implements, but since SAMRAI doesn't use `MPI_Isend` or `MPI_Irecv` in any places relevant to solving this problem I think that we need to implement this on our own.

drwells:
To get good load balancing with this example we need nearly 6x as many IB elements as we have active elements.

knepley:
[...]

PR #832, opened 2020-01-17
Merging loadbalancer

drwells:
Still WIP - most of the changes in this PR are hacks to get visualization working (and to ignore a bunch of stuff in my working directories) but the new class `MergingLoadBalancer` is mostly written.

One problem with the scratch hierarchy is that, in order to get good load balancing, we have to add more patches. To get more patches we split up given patches: the results are assigned to processors in the usual workload-balancing way. This tends to give suboptimal results because most of the children of a split up patch will end up on the same processor, which gives us a ton of elements in patch ghost regions for force spreading.

Fortunately this is pretty easy to avoid - we just have to implement our own `LoadBalancer` class that postprocesses the usual `LoadBalancer` and combines adjacent patches.
Before:
visit0001
number of active elements: 19416
total IB elements on the primary hierarchy: 23749
total IB elements on the scratch hierarchy: 29986

After:
visit0000
number of active elements: 19416
total IB elements on the primary hierarchy: 23749
total IB elements on the scratch hierarchy: 24990

In 3D the pictures are so busy they are hard to understand, but here are the equivalent numbers (also with 8 processors):
Before:
number of active elements: 229376
total IB elements on the primary hierarchy: 426005
total IB elements on the scratch hierarchy: 612800

After:
number of active elements: 229376
total IB elements on the primary hierarchy: 426005
total IB elements on the scratch hierarchy: 509723

Here 'total IB elements' means the total number of elements on which we will do IB calculations. Due to the large ghost regions this value is far larger than the number of active elements. Since, when we don't load balance by IB points, some processors end up with few or no IB elements the total number of IB elements is going to be lower on the primary hierarchy.

this *improves* the situation, but addressing #830 will get the number of IB elements down significantly more.


drwells:
This is now done.

IBFEMethod defaults to using this with the scratch hierarchy; its not a compatible change, but I don't think anyone will notice but me since the scratch hierarchy is still a bit WIP.

examples/IBFE/explicit/ex4/input2d
```diff
@@ -4,7 +4,7 @@ RHO = 1.0
 L   = 1.0
 
 // grid spacing parameters
-MAX_LEVELS = 1                                      // maximum number of levels in locally refined grid
+MAX_LEVELS = 2                                      // maximum number of levels in locally refined grid
```
boyceg: Should we avoid changing the behavior of the examples?

include/ibamr/IBFEMethod.h
```diff
@@ -225,6 +225,8 @@ namespace IBAMR
  *
  *     largest_patch_size
  *     {
+ *         // We recommend using very large values here: large patches
```
boyceg: Should we also increase the min box size?

examples/IBFE/explicit/ex4/input2d
drwells: That was an accident. I will revert it.

include/ibamr/IBFEMethod.h
drwells: Good question - I would say no because a large minimum patch size makes load balancing harder. I added a bunch of comments explaining the phenomenon.

drwells:
I fixed the example input file - since this is approved I will merge.

PR #834, opened 2020-01-20
Fix compilation with libMesh 1.1.

drwells:
This function was renamed and deprecated in libMesh 1.2.


drwells:
Yup - they did keep the old name around for compatibility, though, but it will be removed eventually so we should just make the change now.

PR #837, opened 2020-01-22
Combine some translation units to improve inlining.

drwells:
This makes the 3D BSPLINE_3 kernels for both interpolation and spreading take about 15% less time since the delta functions are inlined.

autotools doesn't generate dependency information for Fortran - AFAICT it cannot correctly track the file dependency created by the `include` statement. Hence we have to manually specify this dependency in the Makefile.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -186,8 +186,8 @@ am__DIST_COMMON = $(srcdir)/Makefile.in \
 	$(top_srcdir)/config/install-sh $(top_srcdir)/config/ltmain.sh \
 	$(top_srcdir)/config/make.inc.in $(top_srcdir)/config/missing \
 	config/compile config/config.guess config/config.rpath \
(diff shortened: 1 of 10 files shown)


drwells:
The examples *really* don't like this change. Let me take a look.

drwells:
Fixed!

boyceg:
The advection routines in IBAMR could possibly benefit from similar inlining.

drwells:
Do you mean `advect_predictors2d.f` and `advect_predictors3d.f`?

boyceg:
Yes.

drwells:
Odd - `median` and `minmod` are in the same source file but don't get inlined. I'll take a look.

drwells:
Would you be OK with opening another issue to investigate inlining in that case? I don't think that we can solve it in the same way we solved this one.

PR #838, opened 2020-01-22
Add some more files to .gitignore.

drwells:
Not too important, but I have dozens of files in my root directory that match these patterns that should not be tracked by git.


drwells:
The testers don't check things like this so lets merge.

PR #841, opened 2020-01-27
Fix a libMesh version check and clean up the rest.

drwells:
The version checks in FEDataManager are wrong - they use `LIBMESH_MAJOR_VERSION` twice. Ooops :)

Fix the problem more permanently by using the convenience macro provided by libMesh for doing version checks.


boyceg:
Oops.

The convenience macro is provided by all of the versions that we aim to support?

drwells:
Yup: It's in 1.1 and newer.

PR #843, opened 2020-01-28
Fix compilation with clang

drwells:
It looks new versions of clang are more pedantic than my installation of GCC - the functor objects provided for sorting inside `std::set` and `std::map` must have a const calling operator, i.e., `operator()(const T&a, const T&b) const`.

I meant to delete some logging statements in this code before merging #832; I put those changes in this PR too.


drwells:
@jabrown893 Would you please check that this patch fixes things for you?

jabrown893:
@drwells Seems to fix things!

drwells:
uh oh - the CI failed with `no space left on device`.

drwells:
Lets try the easy thing first:

Jenkins retest this please

Issue #840, opened 2020-01-27
Inconsistent quad/hex partitioning with libMesh


drwells:
I get different results with different versions of libMesh for the new surface mesh partitioning tests. The graphical outputs all look fine - I think the issue just comes from libMesh generating slightly different meshes between versions.

This isn't too important but we should fix it by the release.

PR #846, opened 2020-01-29
Add some more surface mesh partitioning tests.

drwells:
Part of #840 - add some more tests to improve our element coverage.


drwells:

PR #847, opened 2020-01-29
Index tests

drwells:
I'm working on #830 and I wrote some tests on the way as I learned how the code works.


drwells:

Issue #579, opened 2019-05-10
Use std::unique_ptr to express ownership


drwells:
From #570: we should use `std::unique_ptr` when moving pointer ownership around in `HierarchyIntegrator`.

I think its okay if we break this interface for the sake of progress.

drwells:
I think we can preserve compatibility and make the semantics clearer by overloading the remaining functions, i.e., have both
```cpp
void registerGhostfillRefineAlgorithm(const std::string& ,
                                      SAMRAI::tbox::Pointer<SAMRAI::xfer::RefineAlgorithm<NDIM> > ,
                                      std::unique_ptr<SAMRAI::xfer::RefinePatchStrategy<NDIM> );

void registerGhostfillRefineAlgorithm(const std::string& ,
                                      SAMRAI::tbox::Pointer<SAMRAI::xfer::RefineAlgorithm<NDIM> > );
```
since we always assume ownership of the `RefinePatchStrategy` object anyway.

drwells:
@marshallrdavey Are you willing to take a look at this?

marshallrdavey:
@drwells I'm running into issues in IBHierachyInegrator and others that pass a pointer into registerGhostfillRefineAlgorithm instead of unique_ptr. What would you like to do in those cases?

drwells:
If the class inherits from `HierarchyIntegrator` then the parent class is responsible for calling the destructor. IBHierarchyIntegrator stores a pointer to the object itself.

We can probably do the following:
```cpp
std::unique_ptr<RefinePatchStrategy<NDIM> > rps(d_u_phys_bdry_op);
registerGhostfillRefineAlgorithm(d_object_name + "::u", d_u_ghostfill_alg, std::move(rps));
```
we should also update the documentation to clarify that `d_u_phys_bdry_op` is managed by the parent class since it is registered as a ghostfill refine algorithm.

drwells:
This works only because we create the object (via `new`) in the inheriting class but delete it (by putting it in a `unique_ptr`) in the base class.

drwells:
Fixed by #849.

PR #849, opened 2020-01-31
unique_ptr for HierarchyIntegrator

marshallrdavey:
see #579

### DIFF
--- a/src/IB/IBHierarchyIntegrator.cpp
+++ b/src/IB/IBHierarchyIntegrator.cpp
@@ -286,7 +286,8 @@ IBHierarchyIntegrator::initializeHierarchyIntegrator(Pointer<PatchHierarchy<NDIM
     d_u_ghostfill_alg = new RefineAlgorithm<NDIM>();
     d_u_ghostfill_op = nullptr;
     d_u_ghostfill_alg->registerRefine(d_u_idx, d_u_idx, d_u_idx, d_u_ghostfill_op);
-    registerGhostfillRefineAlgorithm(d_object_name + "::u", d_u_ghostfill_alg, d_u_phys_bdry_op);
+    std::unique_ptr<RefinePatchStrategy<NDIM> > u_phys_bdry_op_unique(d_u_phys_bdry_op);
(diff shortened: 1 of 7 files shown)


src/IB/IBHierarchyIntegrator.cpp
```diff
@@ -317,7 +318,8 @@ IBHierarchyIntegrator::initializeHierarchyIntegrator(Pointer<PatchHierarchy<NDIM
         d_p_ghostfill_alg = new RefineAlgorithm<NDIM>();
         d_p_ghostfill_op = nullptr;
         d_p_ghostfill_alg->registerRefine(d_p_idx, d_p_idx, d_p_idx, d_p_ghostfill_op);
-        registerGhostfillRefineAlgorithm(d_object_name + "::p", d_p_ghostfill_alg, d_p_phys_bdry_op);
+        std::unique_ptr<RefinePatchStrategy<NDIM> > d_p_phys_bdry_op_unique(d_p_phys_bdry_op);
```
drwells: We shouldn't prefix these with `d_` - that is reserved for member variables.

marshallrdavey:
@drwells not sure about the documentation you wanted added, but the changes compile and run

drwells:
We should document `d_u_phys_bdry_op` and `d_p_phys_bdry_op` to state that the base class is responsible for the objects since they are placed in `d_ghostfill_strategies`.

marshallrdavey:
@drwells should this documentation be added to IBHierarchyIntegrator.h above where the variables are defined? I don't know how to edit changelog.

drwells:
To edit the changelog just add a new file under `doc/news/changes/incompatibilities/` - you can look at some entries in `doc/news/changes/major` to get an idea of what to write.

marshallrdavey:
@drwells I'll collate the commits once everything is good to go

drwells:
This does break the interface slightly, but since that interface is not used in any of the examples I don't think its a big deal.

We really do need to change this interface since passing a non-unique pointer implies that we do not assume ownership, but we do - we are inviting double-delete errors with the current implementation.

Issue #850, opened 2020-01-31
NWT test takes too long


drwells:
@amneetb @Raamkrishnan Before #827 the `wave_tank/nwt_cylinder.mpirun=8.input` test only took perhaps 30 seconds to run - its now been running on my workstation for 15 minutes and its still going. Would you please take a look to see what has happened?

drwells:
Fixed by #851.

PR #851, opened 2020-02-01
Fixed a bug in numerical wave tank test

Raamkrishnan:
Hi @drwells,

I fixed the bug. Could you please rerun the test with this updated code?

### DIFF
--- a/tests/wave_tank/nwt_cylinder.cpp
+++ b/tests/wave_tank/nwt_cylinder.cpp
@@ -992,6 +992,9 @@ main(int argc, char* argv[])
         pout << "Started mapping intersections" << std::endl;
         surface_distance_eval->mapIntersections();
         pout << "Finished mapping intersections" << std::endl;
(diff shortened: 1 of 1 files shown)


drwells:
Do we need to make the same change in the examples? It looks like this class only now works if we make one more function call.

Raamkrishnan:
In examples, only IBLevelSet/ex0 uses FESurfaceDistanceEvaluator class. I already included this function call and checked it. However, I forgot to include in tests/wave_tank.

drwells:
I'd like @amneetb to take a look since this means we have to call the class in a different way now.

drwells:
Fixes #850.

amneetb:
Looks good. The routine is broken for efficiency reasons. (For pure translation, surface normals would remain the same, so no need to recompute them)

drwells:
@amneetb is happy so lets merge!

PR #852, opened 2020-02-01
Fix IB5

drwells:
Like the other wider IB stencils (IB-4 and up) IB-5 inlines the computation of the kernel and computes the full stencil regardless of where we are in the box. However, it indexes the computed stencil from 0 regardless of whether or not we actually use the full stencil - this results in using the wrong coefficients when we don't use the full stencil.

I'm not sure why but this doesn't seem to matter in 2D - I tried a bunch of different input files with 1x1 or 2x2 patches and I wasn't able to get different results. It does produce different results in 3D - I updated the test output accordingly.


drwells:
I stared at this for a long time and I'm pretty sure I am right but someone should look this over carefully to make sure this was indeed a mistake - its quite tricky!

boyceg:
You might be able to use IBFE ex0 with either a single patch or with multiple patches to test that we get the same error norms (up to some reasonable tolerance), independent of the patch layout. I would expect that something bad should happen with the current implementation of IB_5 for structures that get close to patch boundaries.

boyceg:
Also, FWIW, I go ahead and explicitly evaluate the tensor product of the weights for IB_6. I don't know if this is worth doing. Here is some relevant code:
```
c
c     Compute the tensor product of the interpolation weights.
c
         do i2 = 0,5
            wz = w2(i2)
            do i1 = 0,5
               wyz = w1(i1)*wz
               do i0 = 0,5
                  w(i0,i1,i2) = w0(i0)*wyz
               enddo
            enddo
         enddo
```

drwells:
> I would expect that something bad should happen with the current implementation of IB_5 for structures that get close to patch boundaries.

Good idea.

> Also, FWIW, I go ahead and explicitly evaluate the tensor product of the weights for IB_6. I don't know if this is worth doing. Here is some relevant code:

I don't remember the details, but I recall trying this in the past and determining (via callgrind) that we execute more instructions when we first compute the tensor product then when we just do it inline so I would prefer to not change that at the moment. Exhaustively reviewing these choices should wait until we get rid of all variable loop bounds after #830 is done.

boyceg:
I just tried out IBFE ex1 with 1 patch and with 16 patches and get identical results (up to solver tolerances), so I think that this is working correctly.

However, master also seems to generate the same results. So I think that this might not actually be a bug fix.

drwells:
Hm. Did you try in 2D or 3D?

boyceg:
2D, not 3D.

drwells:
That's what I saw too. I still recommend we proceed with the patch without worrying too much about 2D since the code is still not right.

boyceg:
Sure.

Issue #615, opened 2019-06-17
Improve ghost region handling in kernels


drwells:
From #598 (there is additional discussion in that thread):

To simplify these implementation (and hopefully obtain better vectorization), we could do the following:

1. Filter out any Lagrangian interaction points that cannot interact with the interior of the patch data. (I think that there is code in LEInteractor that does something like this already.)
2. Allocate a temporary buffer (if needed) that includes a ghost region that is large enough to avoid truncating the kernel stencil for any interaction point that is being processed.

I think that it would suffice to exclude Lagrangian points that are more than kernel_size / 2 away from a patch boundary, and to extend the ghost region to kernel_size. This would greatly simplify the implementation of the actual numerical routines, and hopefully allow for better vectorization.

We could probably do the same thing in interpolation --- filter out Lagrangian interaction points that are outside of the patch, and require that the ghost region be large enough to avoid worrying about the support of the kernel.

We won't be able to vectorize these kernels if we have to truncate stencils at boundaries: we can't do branches inside vectorized code since we have to execute the same instructions for all points. Its not clear to me how we can do this with spreading, but it should be possible with interpolation if we have enough ghost data available.

boyceg:
This sounds good. I would possibly extend the ghost region to `kernel_size + 1` to avoid having to worry about rounding-related corner cases.

It seems like we should be able to avoid branches in both spreading and interpolation as long as we prune the correct collections of points and make the buffer sizes large enough.

drwells:
This will take some effort but could make these kernels run 4x faster so we should investigate. I am still going to go ahead with some other simple performance patches in the mean time (i.e., this week, so that they can be in 0.5).

We will also need to teach SAMRAI how to allocate all of our data in an aligned manner (i.e., via `posix_memalign`).

boyceg:
If we are going to do this anyway, it might make sense to think about automatically generating these kernels. I think that @knepley has some ideas about how to do this.

drwells:
I had another idea on this (now that I am reading the proposal again): some profiling with callgrind indicates that we spend most of our time evaluating the kernel with symmetric stencils (i.e., points are on the interior). We could partition our IB points into two bins: one for boundary points (where each point can have a different stencil) and one for interior points (where the stencils are identical). We can then vectorize the common case easily without having to worry about providing more ghost data.

Another possible way to do this would be to copy data to a scratch hierarchy (like we are going to do with IBFEMethod) which has stronger requirements on ghost data availability.

[...]

PR #854, opened 2020-02-01
Clean up most of the IB kernels.

drwells:
I have been working on this patch off-and-on for the last two months - its now done ~except for the high-order IB kernels (those are unlike the rest of the kernels in that they compute the complete stencil regardless of the position of the point in the box: i.e., they compute stencil values that are never used if we are at a patch boundary).~ Clarification: once #615 is done we will always need the complete stencil for every point so converting the kernels that already unconditionally compute the stencil (constant, RT0, and high order IB) is not worth doing quite yet.

This mostly addresses the goals of #615 - once we change things to sum values spread into ghost regions into other patch interiors then we can fix the remaining issues (we can avoid the awkward partial stencil calculations).

For all but the constant, linear, and high-order IB kernels:
1. use the same convenience arrays
2. loop over the dimension when assembling stencils
3. implement macros for unrolling spreading loops
4. consistently use the interpolation and spreading macros for unrolling

The excluded kernels are unusual in that they unconditionally compute the stencil at all points and use a different indexing scheme.

### DIFF
--- a/ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
+++ b/ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
@@ -47,6 +47,31 @@ define(INTERPOLATE_2D_SPECIALIZE_FIXED_WIDTH,
            INTERPOLATE_INNER_2D($1, $2,
                                 $3, $4)
          endif')dnl
+define(SPREAD_INNER_2D,
+          ` do d = 0,depth-1
+               do ic1 = $1,$2
+                  do ic0 = $3,$4
+                     u(ic0,ic1,d) = u(ic0,ic1,d) + (
+     &                    +w(0,ic0-$3)
+     &                    *w(1,ic1-$1)
+     &                    *V(d,s)/(dx(0)*dx(1)))
+                  enddo
+               enddo
+            enddo')
+dnl Same arguments as before, but the seventh argument is the width of the
+dnl stencil (e.g., 3 for bspline 3). The first branch is a hotter code path
+dnl since when we are not at a boundary the number of inner loop iterations
+dnl is known. Exposing this to the compiler helps generate code which speeds
+dnl up the subroutine by about 25%.
+define(SPREAD_2D_SPECIALIZE_FIXED_WIDTH,
+`   if ($2 - $1 == ($5 - 1) .and.
+     &       $4 - $3 == ($5 - 1)) then
+           SPREAD_INNER_2D($1, ($1 + $5 - 1),
+                           $3, ($3 + $5 - 1))
+         else
+           SPREAD_INNER_2D($1, $2,
+                           $3, $4)
+         endif')dnl
 include(SAMRAI_FORTDIR/pdat_m4arrdim2d.i)dnl
 
 c     this is a Fortran include, not an m4 include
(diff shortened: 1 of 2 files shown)


drwells:
I didn't do any profiling but other changes in #620 lowered the cost of the kernels by about 25% - these should be about the same.

ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
```diff
@@ -712,72 +737,71 @@ c
...
-     &           lagrangian_piecewise_cubic_delta(
-     &           (X(0,s)+Xshift(0,l)-X_cell(0))/dx(0))
-         enddo
-
-         do ic1 = ic_lower(1),ic_upper(1)
-            X_cell(1) = x_lower(1)+(dble(ic1-ilower1)+0.5d0)*dx(1)
-            w(1,ic1-ic_lower(1)) =
-     &           lagrangian_piecewise_cubic_delta(
-     &           (X(1,s)+Xshift(1,l)-X_cell(1))/dx(1))
+            do ic0 = ic_lower(d),ic_upper(d)
```
boyceg: Could we change `ic0` to `ic` here? I think that I pretty consistently use `ic0` to mean a "cell index in the 0 axis".

ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
```diff
@@ -829,80 +853,70 @@ c
...
```
boyceg: Same comment as above.

ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
```diff
@@ -953,67 +967,65 @@ c
...
-     &           lagrangian_ib_3_delta(
-     &           (X(0,s)+Xshift(0,l)-X_cell(0))/dx(0))
-         enddo
-
-         do ic1 = ic_lower(1),ic_upper(1)
-            X_cell(1) = x_lower(1)+(dble(ic1-ilower1)+0.5d0)*dx(1)
-            w(1,ic1-ic_lower(1)) =
-     &           lagrangian_ib_3_delta(
-     &           (X(1,s)+Xshift(1,l)-X_cell(1))/dx(1))
+            do ic0 = ic_lower(d),ic_upper(d)
```
boyceg: Same.

ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
```diff
@@ -1065,75 +1077,65 @@ c
...
```
boyceg: Same.

ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
```diff
@@ -2316,67 +2318,65 @@ c
...
-     &           lagrangian_bspline_3_delta(
-     &           (X(0,s)+Xshift(0,l)-X_cell(0))/dx(0))
-         enddo
-
-         do ic1 = ic_lower(1),ic_upper(1)
-            X_cell(1) = x_lower(1)+(dble(ic1-ilower1)+0.5d0)*dx(1)
-            w(1,ic1-ic_lower(1)) =
-     &           lagrangian_bspline_3_delta(
-     &           (X(1,s)+Xshift(1,l)-X_cell(1))/dx(1))
+            do ic0 = ic_lower(d),ic_upper(d)
```
boyceg: Same.

ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
```diff
@@ -2428,75 +2428,65 @@ c
...
```
boyceg: You get the picture.

boyceg:
Also, could/should we m4-ize some of the boiler plate setup?

drwells:
I fixed the `ic` bit.

> Also, could/should we m4-ize some of the boiler plate setup?

It would be nice to eventually do this. However, I would prefer to hold off until we can assume that we use the full stencil at each point; i.e., we will make substantial modifications to these in the near future so it seems best to not worry about further generalizations at this point in time.

drwells:
The new tests pass on this branch too so lets merge :)

PR #856, opened 2020-02-05
Fix clang warnings 2

drwells:
Fix a variety of minor warnings generated by my copy of clang.

### DIFF
--- a/src/IB/IBInstrumentPanel.cpp
+++ b/src/IB/IBInstrumentPanel.cpp
@@ -1233,7 +1233,7 @@ IBInstrumentPanel::writePlotData(const int timestep_num, const double simulation
     // Output the web data on the available MPI processes.
     for (unsigned int meter = 0; meter < d_num_meters; ++meter)
     {
-        if (meter % mpi_nodes == mpi_rank)
+        if (static_cast<int>(meter) % mpi_nodes == mpi_rank)
         {
             std::string dirname = d_instrument_names[meter];
             if (DBMkDir(dbfile, dirname.c_str()) == -1)
--- a/src/complex_fluids/CFRelaxationOperator.cpp
+++ b/src/complex_fluids/CFRelaxationOperator.cpp
@@ -47,18 +47,14 @@ CFRelaxationOperator::convertToConformation(const MatrixNd& mat)
     {
     case SQUARE_ROOT:
         return mat * mat;
-        break;
     case LOGARITHM:
         return mat.exp();
-        break;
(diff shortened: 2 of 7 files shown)


src/IB/IBInstrumentPanel.cpp
```diff
@@ -1233,7 +1233,7 @@ IBInstrumentPanel::writePlotData(const int timestep_num, const double simulation
     // Output the web data on the available MPI processes.
     for (unsigned int meter = 0; meter < d_num_meters; ++meter)
     {
-        if (meter % mpi_nodes == mpi_rank)
+        if (int(meter) % mpi_nodes == mpi_rank)
```
boyceg: Do we have a guideline about when to use a construction like `int(meter)` vs. `static_cast<int>(meter)`? I typically use `static_cast` to make it easier to make it easy to find these kinds of things later on using simple search expressions.

src/complex_fluids/CFRelaxationOperator.cpp
```diff
@@ -47,18 +47,14 @@ CFRelaxationOperator::convertToConformation(const MatrixNd& mat)
     {
     case SQUARE_ROOT:
         return mat * mat;
-        break;
```
boyceg: Does clang really complain about this? I would rather keep these `break` statements.

drwells: It does when very pedantic warning flags are on.

boyceg: Yuck.

boyceg: Would it be better to capture the output and have a single `return` statement?

src/IB/IBInstrumentPanel.cpp
drwells: Lets stick with `static_cast` - they do the same thing but that is the more common style in the codebase.

src/complex_fluids/CFRelaxationOperator.cpp
drwells: The proposed implementation is the simplest - we just should immediately return rather than setting up temporary variables.

Nevertheless, I'm mostly interested in getting rid of the warning, so if that is your preference then lets do it.

boyceg: Let's leave it as is and get rid of the `break` statements to appease clang.

drwells:
That should do it!

Issue #737, opened 2019-11-04
Improve spreading in IBFEMethod/FEDataManager


drwells:
From #654: during review we pointed out that the way we handle force spreading in IBFEMethod is pretty inefficient (with multiple parts). It would be better to resolve the accumulation of forces spread outside the domain once inside IBFEMethod instead for each part separately in FEDataManager.

drwells:
Another idea: since we add spread data we do not need to communicate from the primary hierarchy to the scratch hierarchy: right now FEDataManager immediately swaps whatever force data is put in and then swaps back and adds. We could get around this by calculating the new force contribution on the scratch hierarchy, communicating that to a scratch index, and then adding that to the existing force values.

boyceg:
> Another idea: since we add spread data we do not need to communicate from the primary hierarchy to the scratch hierarchy: right now FEDataManager immediately swaps whatever force data is put in and then swaps back and adds. We could get around this by calculating the new force contribution on the scratch hierarchy, communicating that to a scratch index, and then adding that to the existing force values.

Yes, exactly.

For spreading, just communicate Eulerian force data from the scratch hierarchy to the primary hierarchy.

For interpolation, just communicate Eulerian velocities from the primary hierarchy to the scratch hierarchy.

drwells:
From an email @boyceg sent me earlier (I don't want this to get lost in my inbox):

```cpp
// Communicate ghost data.
for (const auto& u_ghost_fill_sched : u_ghost_fill_scheds)
{
    if (u_ghost_fill_sched) u_ghost_fill_sched->fillData(data_time);
}

if (d_use_scratch_hierarchy)
{
    assertStructureOnFinestLevel();
    getPrimaryToScratchSchedule(
        d_hierarchy->getFinestLevelNumber(),
        u_data_idx,
        d_ib_solver->getVelocityPhysBdryOp())
        .fillData(data_time);
}
```

Probably can be simplified to either do one or the other based on whether the scratch hierarchy is used or not.

Specifically, if the scratch hierarchy is not used, then ghost data should be communicated using the provided `u_ghost_fill_sched`.

On the other hand, there is no need to execute the `u_ghost_fill_sched` if data are being moved from the primary hierarchy to the scratch hierarchy, because those ghost values on the primary hierarchy will never be used in the interpolation routine.

In `IBFEMethod::spreadForce()`:
```cpp
if (d_use_scratch_hierarchy)
{
    assertStructureOnFinestLevel();
[...]

PR #857, opened 2020-02-05
Redo force spreading 1

drwells:
Most of #737 - I will address the parts pertinent to FEDataManager in another PR.

First part of #830 - now that we have the forces spread onto their own index we can go ahead and redo the way we accumulate data in parallel without worrying about other forces.

I encountered a few bugs along the way while writing this patch - to aid in future debugging I added a new test and modified an existing test so that we have an IBFE-based simulation that executes multiple regrids and gives the same answers with and without the scratch hierarchy.

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -948,11 +948,20 @@ IBFEMethod::spreadForce(const int f_data_idx,
         d_active_fe_data_managers[part]->spread(f_scratch_data_idx,
                                                 *F_ghost_vec,
                                                 *X_ghost_vec,
-                                                FORCE_SYSTEM_NAME,
-                                                f_phys_bdry_op,
-                                                data_time,
-                                                /*close_F*/ false,
-                                                /*close_X*/ false);
+                                                FORCE_SYSTEM_NAME);
+    }
+
+    // Deal with force values spread outside the physical domain.
+    if (f_phys_bdry_op)
+    {
+        f_phys_bdry_op->setPatchDataIndex(f_scratch_data_idx);
+        Pointer<PatchLevel<NDIM> > level = hierarchy->getPatchLevel(ln);
+        for (PatchLevel<NDIM>::Iterator p(level); p; p++)
+        {
+            const Pointer<Patch<NDIM> > patch = level->getPatch(p());
+            Pointer<PatchData<NDIM> > f_data = patch->getPatchData(f_scratch_data_idx);
+            f_phys_bdry_op->accumulateFromPhysicalBoundaryData(*patch, data_time, f_data->getGhostCellWidth());
+        }
     }
 
     if (d_use_scratch_hierarchy)
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -741,7 +741,7 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
     {
         assertStructureOnFinestLevel();
         getPrimaryToScratchSchedule(
-            d_hierarchy->getFinestLevelNumber(), u_data_idx, d_ib_solver->getVelocityPhysBdryOp())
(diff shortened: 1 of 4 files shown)


src/IB/IBFEMethod.cpp
```diff
@@ -929,33 +929,67 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+        // harder case: we have a scratch index on the scratch hierarchy but
+        // no corresponding index on the primary hierarchy.
+        //
+        // unlike the other data ops, this is always on the primary hierarchy
+        auto f_primary_data_ops = HierarchyDataOpsManager<NDIM>::getManager()->getOperationsDouble(f_var, d_hierarchy);
+        f_primary_data_ops->resetLevels(ln, ln);
+        const auto f_primary_scratch_data_idx = d_primary_eulerian_data_cache->getCachedPatchDataIndex(f_data_idx);
+        // we have to zero everything here since the scratch to primary
+        // communication does not touch ghost cells, which may have junk
+        f_primary_data_ops->setToScalar(f_primary_scratch_data_idx, 0.0, /*interior_only*/ false);
```
drwells: @boyceg this is what I was missing an hour ago - oops.

drwells:
@boyceg could you take a quick look at this soon? I need this PR merged to proceed with the force spreading redo.

src/IB/IBFEMethod.cpp
boyceg: OK, so accumulation from the ghost region to the interior is done separately for each structure?

drwells:
We have a test based on ex8 - I should check again but I believe those structures are right at the boundary.

It would be wise to add some more spreading tests that explicitly test spreading when we have a structure on the boundary - I'll add that to the relevant issue.

src/IB/IBFEMethod.cpp
drwells: No - see circa line 950. We do the accumulation after looping over all parts.

boyceg:
The thing I would suggest is to have a case that has two structures that are close to each other that are right at the boundary. They even could be clones of each other.

src/IB/IBFEMethod.cpp
boyceg: Do different parts have different scratch hierarchies?

drwells: No, but they do presently use different scratch patch data indices - this PR was written before the overloads #858 were added to avoid using a scratch force data index inside `FEDataManager::spread`.

boyceg: Can you remind me --- is there one scratch hierarchy, or are there separate scratch hierarchies for each part?

boyceg: (As you can probably guess, I am worried about losing the ghost values.)

drwells: There is just one scratch hierarchy managed by IBFEMethod. The FEDataManagers in `d_scratch_fe_data_managers` all store `Pointer`s to that hierarchy instead of the 'normal' one.

drwells: That's a really good point. Even though we have some tests based on `ex8` in the test suite which this passes, lets be paranoid and add some more spreading tests (as described below) before proceeding.

drwells:
I fixed the merge conflict on my own machine but now the `ex8` tests (an IBFE example with a structure on the boundary) fail. At least we know that the test suite is good enough to find a bug with this implementation :)

drwells:
I found the problem - we don't properly handle ghost data in the overload of FEDataManager::spread() that takes an optional boundary op object and the `close` booleans. The new version of this PR works because it uses a different overload.

I'll put up another PR that fixes that bug in a minute.

drwells:
@boyceg you approved this 10 days ago - the changes on top of those are minor (its just the test addition and the use of a different `FEDataManager::spread()` overload). Would you please take a quick look at this when you have a minute?

PR #858, opened 2020-02-05
Add more overloads of FEDataManager::spread().

drwells:
These do not copy data - a nice feature for IBFEMethod, which will copy data on its own anyway.

This is the FEDataManager part of #737. Once this and #857 are merged we can go ahead and use the new functions from this PR (which require a scratch index) directly, which will finish #737.

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -618,6 +618,51 @@ public:
     /*!
      * \brief Spread a density from the FE mesh to the Cartesian grid using the
      * default spreading spec.
+     *
+     * @note This function may spread forces from points near the physical
+     * boundary into ghost cells outside the physical domain. To account for
+     * these forces one should usually call
+     * IBTK::RobinPhysBdryPatchStrategy::accumulateFromPhysicalBoundaryData()
+     * after this function to move said force values into the physical domain.
+     */
+    void spread(int f_data_idx,
+                libMesh::NumericVector<double>& F,
+                libMesh::NumericVector<double>& X,
+                const std::string& system_name);
+
+    /*!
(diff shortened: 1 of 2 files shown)


ibtk/include/ibtk/FEDataManager.h
```diff
@@ -618,6 +618,53 @@ class FEDataManager : public SAMRAI::tbox::Serializable
...
+                libMesh::NumericVector<double>& F,
+                libMesh::NumericVector<double>& X,
+                const std::string& system_name,
+                const SpreadSpec& spread_spec);
+
+    /*!
+     * \brief Spread a density from the FE mesh to the Cartesian grid using the
+     * default spreading spec.
+     *
+     * @deprecated This version of spread is deprecated - users should close
```
boyceg: I would prefer to keep some "simple" spreading routines that do not require manual use of physical boundary operator objects. What is the rationale for deprecating?

boyceg: Is there some helper code to make it easy to apply those operators at patch boundaries across a level and/or hierarchy?

drwells: I deprecated them since they are the wrong choice when one wants to use multiple parts - they require using a scratch index and using the boundary operator for each part.

boyceg: Are there helper functions to help to do this correctly with the scratch hierarchy?

drwells: There are not.

Lets fix this by just not deprecating the convenience routines.

drwells:
Deprecating the convenience functions is a subjective move - lets just keep them if they might be useful.

PR #861, opened 2020-02-10
Updated bundled boost and added configuring checks.

### LINKED ISSUE #845: Build failure using bundled boost
I'm using IBAMR's bundled boost, and libmesh is configured without boost. I'm getting a build error that looks like

`../../IBAMR/lib/../src/wave_generation/WaveDampingFunctions.cpp:26:10: fatal error: boost/math/tools/roots.hpp: No such file or directory
 #include <boost/math/tools/roots.hpp>`

It appears that our bundled copy of boost isn't complete. We should either remove this dependency from `WaveDampingFunctions` or update our bundled boost.

abarret:
Fixes #845. I've added necessary functions to use `boost/math/tools/roots.hpp` as well as added configure checks for all headers we currently use from boost.


drwells:
I removed boost from my laptop to double-check this: everything works so lets merge!

PR #863, opened 2020-02-10
Fix restarts on the scratch hierarchy.

drwells:
A bug in SAMRAI prevents the usage of GriddingAlgorithm objects that do not correspond to a restart database entry (i.e., one cannot dynamically create GriddingAlgorithm objects in a restarted simulation). Get around this by setting up the GriddingAlgorithm object and friends in the constructor of IBFEMethod so that they are always available.

@marshallrdavey this fixes the problem you pointed out to me earlier. It turns out that we don't have any tests for restarts with the scratch hierarchy - we now do.

### DIFF
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -1234,12 +1234,46 @@ protected:
     /**
      * database for the GriddingAlgorithm used with the scratch hierarchy.
      */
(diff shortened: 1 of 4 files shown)


drwells:
The same problem is present in SAMRAI 3 :crying_cat_face:

PR #866, opened 2020-02-11
Correctly delete test working directories.

drwells:
This fixes a python bug where `bool(conf['keep_work_directories'])` would always evaluate to `True` since the truth-value of `"False"` is `True` (its not an empty string).


drwells:
Since this just pertains to the test runner I will merge myself.

Issue #865, opened 2020-02-11
Bug: Coarsen operator for density in class INSVCStaggeredNonConservativeHierarchyIntegrator


nishantnangia:
Hi folks,

The coarsen operator is incorrectly specified in INSVCStaggeredNonConservativeHierarchyIntegrator. Specifically, this line:
https://github.com/IBAMR/IBAMR/blob/master/src/navier_stokes/INSVCStaggeredNonConservativeHierarchyIntegrator.cpp#L989

should use `d_rho_coarsen_type` instead.

Without this fix, all non-conservative VC cases will crash during the initial time step with error message (e.g. examples/vc_navier_stokes/ex0):

> P=00000:Coarsen operator 'CONSERVATIVE_LINEAR_REFINE' not found for variable 'rho_var

The bug originated from this commit:
https://github.com/IBAMR/IBAMR/commit/0b6e65552aeaa7f7f26cbba30c8c095c6aff3fa8#diff-ed1029e41f9b13ac6a438e15afe1d5fbR958
although at a glance, it does not appear that this mistake was made anywhere else.

==
Aside: @boyceg, could you add me back to the IBAMR organization? I have been looking into some multiphase cases and may want to push some branches in the near future.

drwells:
Nice catch. Could you post the input file you used? It would be good to reproduce this and add a test.

nishantnangia:
The input file was input3d from examples/vc_navier_stokes/ex0, but any of
the input files within should exhibit the same behavior — including the 2D
ones.

nishantnangia:
Are there any instructions on how to create new tests (the last time I wrote a test for IBAMR was back when we were using gtest)? If so, could you point me to them -- I'd be happy to take care of this PR.

Raamkrishnan:
Hi, I recently followed this link to write a test:

https://github.com/IBAMR/IBAMR/wiki/Tests-for-IBAMR

nishantnangia:
Ah wonderful, thanks! I'll take a look.

amneetb:
@nishantnangia 

I built `numdiff` on OSX by using:

```
asbhalla@Amneets-MBP-2:~/Softwares/tars/numdiff-5.9.0$ head config.log 
This file contains any messages produced by compilers while
running configure, to aid debugging if configure makes a mistake.

It was created by Numdiff configure 5.8, which was
generated by GNU Autoconf 2.63.  Invocation command line was

[...]

PR #870, opened 2020-02-14
Fix coarsen specification in non-conservative VC integrator

nishantnangia:
This PR fixes and implements tests for the issue described by https://github.com/IBAMR/IBAMR/issues/865


tests/vc_navier_stokes/SetFluidProperties.cpp
```diff
@@ -0,0 +1,148 @@
...
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+// Filename: SetFluidProperties.cpp
```
boyceg: @drwells, I think we are omitting this material from source code files now --- is that right?

drwells: Yes.

tests/vc_navier_stokes/SetFluidProperties.h
```diff
@@ -0,0 +1,166 @@
...
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE
```
drwells: same here - we just need the new and equivalent text.

drwells:
Thanks a lot for the fix! Would you please also add a changelog entry so you get credit for it at release time?

nishantnangia:
I trimmed the license stuff, added the changelog, and squashed.

drwells:
I can confirm that the output from these tests matches the output from versions of IBAMR before the bug was introduced, so lets merge!

PR #872, opened 2020-02-14
Complex fluid boundary conditions and projection updates

abarret:
The boundary conditions were not consistently applied to the correct form of the conformation tensor. The boundary conditions for the tensor are now consistently applied to the quantity which is being solved for (e.g. square root or logarithm). In doing so, an extra scratch index has been eliminated since we do decompositions in place now.

We are also now applying the projection of the conformation to the index that is owned by the advection diffusion integrator. Previously, we only projected the scratch index used to compute the divergence of the stress.

@fuhuifang I'm not sure if you were using them, but this should make simulating higher Weissenberg numbers easier and more accurate.

### DIFF
--- a/ibtk/include/ibtk/ibtk_utilities.h
+++ b/ibtk/include/ibtk/ibtk_utilities.h
@@ -28,6 +28,7 @@ IBTK_ENABLE_EXTRA_WARNINGS
 
 #include <algorithm>
 #include <array>
+#include <utility>
 
 /////////////////////////////// MACRO DEFINITIONS ////////////////////////////
 
@@ -143,6 +144,36 @@ level_can_be_refined(int level_number, int max_levels)
 {
     const int finest_level_number = max_levels - 1;
     return level_number < finest_level_number;
+}
+
+/*!
+ * Convert a Voigt notation index to the corresponding symmetric tensor index. This function only returns the upper
+ * triangular index.
+ */
+inline std::pair<int, int>
+voigt_to_tensor_idx(const int k)
+{
+    if (k < NDIM)
+        return std::make_pair(k, k);
+    else
+#if (NDIM == 2)
+        return std::make_pair(0, 1);
+#endif
+#if (NDIM == 3)
+    return std::make_pair(k > 3 ? 0 : 1, k > 4 ? 1 : 2);
+#endif
+}
+
+/*!
+ * Convert a symmetric tensor index to the corresponding Voigt notation index.
+ */
+inline int
+tensor_idx_to_voigt(const std::pair<int, int>& idx)
+{
+    if (idx.first == idx.second)
+        return idx.first;
+    else
+        return 3 * NDIM - 3 - idx.first - idx.second;
 }
 
 /*!
--- a/include/ibamr/ibamr_enums.h
+++ b/include/ibamr/ibamr_enums.h
@@ -473,7 +473,7 @@ inline TensorEvolutionType
 string_to_enum<TensorEvolutionType>(const std::string& val)
 {
     if (strcasecmp(val.c_str(), "STANDARD") == 0) return STANDARD;
-    if (strcasecmp(val.c_str(), "SQUARE_ROOT") == 0) return STANDARD;
+    if (strcasecmp(val.c_str(), "SQUARE_ROOT") == 0) return SQUARE_ROOT;
     if (strcasecmp(val.c_str(), "LOGARITHM") == 0) return LOGARITHM;
     return UNKNOWN_TENSOR_EVOLUTION_TYPE;
 }
--- a/src/complex_fluids/CFINSForcing.cpp
+++ b/src/complex_fluids/CFINSForcing.cpp
@@ -150,7 +150,10 @@ CFINSForcing::commonConstructor(const Pointer<Database> input_db,
     // Read parameters from input file
     d_lambda = input_db->getDouble("relaxation_time");
     d_eta = input_db->getDouble("viscosity");
-    d_evolve_type = string_to_enum<TensorEvolutionType>(input_db->getStringWithDefault("evolution_type", "STANDARD"));
+    if (input_db->keyExists("evolution_type"))
+        d_evolve_type = string_to_enum<TensorEvolutionType>(input_db->getString("evolution_type"));
+    if (input_db->keyExists("evolve_type"))
+        d_evolve_type = string_to_enum<TensorEvolutionType>(input_db->getString("evolve_type"));
     if (input_db->keyExists("D")) d_adv_diff_integrator->setDiffusionCoefficient(d_W_cc_var, input_db->getDouble("D"));
     d_log_det = input_db->getBoolWithDefault("log_determinant", d_log_det);
     d_convec_oper_type = input_db->getStringWithDefault("convective_operator_type", "CENTERED");
--- a/src/complex_fluids/CFINSForcing.cpp
+++ b/src/complex_fluids/CFINSForcing.cpp
@@ -146,7 +146,6 @@ CFINSForcing::commonConstructor(const Pointer<Database> input_db,
     d_adv_diff_integrator->registerTransportedQuantity(d_W_cc_var);
     d_adv_diff_integrator->setInitialConditions(d_W_cc_var, d_init_conds);
     d_W_scratch_idx = var_db->registerVariableAndContext(d_W_cc_var, d_context, ghosts_cc);
-    d_W_cc_idx = var_db->registerClonedPatchDataIndex(d_W_cc_var, d_W_scratch_idx);
 
     // Read parameters from input file
     d_lambda = input_db->getDouble("relaxation_time");
@@ -160,6 +159,10 @@ CFINSForcing::commonConstructor(const Pointer<Database> input_db,
     d_error_on_spd = input_db->getBoolWithDefault("error_on_spd", d_error_on_spd);
     d_log_divW = input_db->getBoolWithDefault("log_divergence", d_log_divW);
     d_project_conform = input_db->getBoolWithDefault("project_conformation_tensor", d_project_conform);
+    // Have advection diffusion integrator project conformation tensor if necessary.
+    if (d_project_conform)
+        d_adv_diff_integrator->registerIntegrateHierarchyCallback(&apply_project_tensor_callback,
+                                                                  static_cast<void*>(this));
 
     // Set up drawing variables if necessary
     d_conform_draw = input_db->getBoolWithDefault("output_conformation_tensor", d_conform_draw);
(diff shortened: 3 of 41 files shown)


drwells:
I'm surprised that the test output didn't change - do none of the tests check things relevant to these boundary condition changes?

abarret:
None of the tests check physical boundaries. I'll update the divergence test to check this.

src/complex_fluids/CFINSForcing.cpp
```diff
@@ -279,56 +281,51 @@ CFINSForcing::setDataOnPatchHierarchy(const int data_idx,
...
+            hier_cc_data_ops->copyData(d_W_scratch_idx, W_current_idx);
         }
         else
         {
-            hier_cc_data_ops->linearSum(d_W_cc_idx, 0.5, W_current_idx, 0.5, W_new_idx);
+            hier_cc_data_ops->linearSum(d_W_scratch_idx, 0.5, W_current_idx, 0.5, W_new_idx);
         }
     }
 
+    typedef HierarchyGhostCellInterpolation::InterpolationTransactionComponent InterpolationTransactionComponent;
```
boyceg: `using`? :-)

src/complex_fluids/CFINSForcing.cpp
```diff
@@ -279,56 +281,51 @@ CFINSForcing::setDataOnPatchHierarchy(const int data_idx,
...
+    ghost_fill_op.fillData(data_time);
+
     HierarchyDataOpsManager<NDIM>* hier_data_ops_manager = HierarchyDataOpsManager<NDIM>::getManager();
     Pointer<HierarchyDataOpsReal<NDIM, double> > hier_cc_data_ops =
         hier_data_ops_manager->getOperationsDouble(d_W_cc_var, hierarchy);
     switch (d_evolve_type)
     {
     case SQUARE_ROOT:
-        squareMatrix(d_W_scratch_idx, d_W_cc_var, hierarchy, data_time, initial_time, coarsest_ln, finest_ln);
+        squareMatrix(d_W_scratch_idx, d_W_cc_var, hierarchy, data_time, initial_time, coarsest_ln, finest_ln, true);
```
boyceg: Could you add a comment in front of the hard-coded boolean values to indicate the parameter name?

drwells:
If adding a test is going to be unreasonably difficult or if we need this right now then we should merge. I don't understand this part of the code that well but I am hopeful that we can add another test that extends a current test to the boundary without much work.

abarret:
I'll update a test today.

abarret:
Good thing @drwells makes me write tests. I've found two bugs because of the updated tests.  I'll push a fix and updated tests tomorrow.

drwells:
> Good thing @drwells makes me write tests. I've found two bugs because of the updated tests. I'll push a fix and updated tests tomorrow.

Excellent :)

abarret:
After updating the tests, I noticed small discrepancies that led me to find two different bugs with specifying how the conformation tensor should be evolved. I've fixed those bugs in this PR as well as updating relevant tests/examples. @drwells, I can put them in a new PR if that would be better.

include/ibamr/ibamr_enums.h
```diff
@@ -473,7 +473,7 @@ inline TensorEvolutionType
 string_to_enum<TensorEvolutionType>(const std::string& val)
 {
     if (strcasecmp(val.c_str(), "STANDARD") == 0) return STANDARD;
-    if (strcasecmp(val.c_str(), "SQUARE_ROOT") == 0) return STANDARD;
+    if (strcasecmp(val.c_str(), "SQUARE_ROOT") == 0) return SQUARE_ROOT;
```
drwells: :)

tests/complex_fluids/cf_forcing_op_01_2d.cell.logarithm.output
```diff
@@ -1,6 +1,6 @@
 AdvDiffHierarchyIntegrator::initializePatchHierarchy(): tag_buffer = 0
 Conformation tensor is SPD
 Error norms:
-  L1-norm:  0.077081549
-  L2-norm:  0.049351098
-  max-norm: 0.040101043
+  L1-norm:  0.97983536
+  L2-norm:  1.0673658
+  max-norm: 2.7850786
```
drwells: Should these error norms be so high?

include/ibamr/ibamr_enums.h
abarret: They both start with S so those words are basically the same, right?

drwells:
I don't really understand this part of the code, but is there some reason why we cannot keep the current tests as-is and just add new tests? It seems odd to me that we are replacing tests that compute low error norms with new tests that have much higher errors.

tests/complex_fluids/cf_forcing_op_01_2d.cell.logarithm.output
abarret: Yes. These are absolute norms. The relative norms are ~0.05 plus this is a fairly coarse grid.

boyceg:
@abarret, did the old tests look at the conformation tensor C, but now are looking at errors in log(C)? If so, it is natural for the errors to change.

abarret:
> I don't really understand this part of the code, but is there some reason why we cannot keep the current tests as-is and just add new tests? It seems odd to me that we are replacing tests that compute low error norms with new tests that have much higher errors.

The simple answer is that the original tests were wrong. I've updated the tests so they check the complete function of `setDataOnPatchHierarchy` which includes computing the divergence and filling in correct boundary conditions for all possible evolution types. 

Before there was one test that computed the divergence. That test is now `cf_forcing_2d_01.cell.standard.input`.

abarret:
> @abarret, did the old tests look at the conformation tensor C, but now are looking at errors in log(C)? If so, it is natural for the errors to change.

The tests are now looking at errors of div(exp(C)), div(C*C), and div(C) where C is a matrix with entries that are ~1-3

drwells:
I don't understand this well enough to review the PR. @boyceg are you happy with the second round of changes? If so lets merge :)

src/complex_fluids/CFINSForcing.cpp
```diff
@@ -674,40 +690,49 @@ CFINSForcing::squareMatrix(const int data_idx,
...
+
             for (CellIterator<NDIM> it(box); it; it++)
             {
                 CellIndex<NDIM> i = *it;
 #if (NDIM == 2)
-                (*tens_data)(i, 0) = (*half_data)(i, 0) * (*half_data)(i, 0) + (*half_data)(i, 2) * (*half_data)(i, 2);
-                (*tens_data)(i, 1) = (*half_data)(i, 2) * (*half_data)(i, 2) + (*half_data)(i, 1) * (*half_data)(i, 1);
-                (*tens_data)(i, 2) = (*half_data)(i, 0) * (*half_data)(i, 2) + (*half_data)(i, 1) * (*half_data)(i, 2);
+                Eigen::Matrix2d mat;
+                mat(0, 0) = (*data)(i, 0);
```
boyceg: Is the data ordering amenable to using some loops here instead of hard-coding indexing?

src/complex_fluids/CFINSForcing.cpp
```diff
@@ -757,39 +782,39 @@ CFINSForcing::exponentiateMatrix(const int data_idx,
...
             for (CellIterator<NDIM> it(box); it; it++)
             {
                 CellIndex<NDIM> i = *it;
 #if (NDIM == 2)
                 Eigen::Matrix2d mat;
-                mat(0, 0) = (*log_data)(i, 0);
-                mat(1, 0) = (*log_data)(i, 2);
-                mat(0, 1) = (*log_data)(i, 2);
-                mat(1, 1) = (*log_data)(i, 1);
+                mat(0, 0) = (*data)(i, 0);
```
boyceg: Same question as above.

src/complex_fluids/CFINSForcing.cpp
abarret: It's not pretty, but the best I can think of is something like

```
for (int i = 0; i < NDIM; ++i)
{
    for (int j = 0; j <= i; ++j)
    {
        mat(i,j) = mat(j,i) = i == j ? (*data)(i,0) : (*data)(3*NDIM-3 - i - j);
    }
}
```

Would this be preferable?

boyceg: This is Voigt notation, right? Can you make a helper function for Voigt notation?

abarret: Sure. Something like `voigtToTensor` and `tensorToVoigt`?

boyceg: Sure --- if it is static member function `voigt_to_tensor` or `voigt_to_matrix` sounds good.

boyceg: Or `voigt_to_matrix_index` or `voigt_to_tensor_idx`.

src/complex_fluids/CFINSForcing.cpp
```diff
@@ -655,28 +657,16 @@ CFINSForcing::checkPositiveDefinite(const int data_idx,
...
-                Eigen::Matrix3d tens(Eigen::Matrix3d::Zero());
-                tens(0, 0) = (*s_data)(ci, 0);
-                tens(1, 1) = (*s_data)(ci, 1);
-                tens(2, 2) = (*s_data)(ci, 2);
-                tens(0, 1) = tens(1, 0) = (*s_data)(ci, 5);
-                tens(0, 2) = tens(2, 0) = (*s_data)(ci, 4);
-                tens(1, 2) = tens(2, 1) = (*s_data)(ci, 3);
-                Eigen::LLT<Eigen::Matrix3d> llt;
-#endif
+                MatrixNd tens;
```
boyceg: Isn't that nicer? :-D

abarret:
@boyceg take a look now. I replaced the hard coded indices with loops.

src/complex_fluids/CFINSForcing.cpp
abarret: I hate to admit it, but yes.

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -145,6 +146,29 @@ level_can_be_refined(int level_number, int max_levels)
     return level_number < finest_level_number;
 }
 
+static inline std::pair<int, int>
```
drwells:
```suggestion
inline std::pair<int, int>
```
would you please add a docstring?

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -145,6 +146,29 @@ level_can_be_refined(int level_number, int max_levels)
...
+    else
+#if (NDIM == 2)
+        return std::make_pair(0, 1);
+#endif
+#if (NDIM == 3)
+    return std::make_pair(k > 3 ? 0 : 1, k > 4 ? 1 : 2);
+#endif
+}
+
+static inline int
```
drwells: same

ibtk/include/ibtk/ibtk_utilities.h
abarret: Why should we remove the `static` keyword?

drwells: static here means that the function will only be available in the current translation unit, which doesn't make any sense when it is declared and defined in a header.

abarret:
@drwells, I've added some comments on the functions and removed the `static` keyword.

Issue #509, opened 2019-03-26
Add more tests for interpolation and spreading


drwells:
This is for after #494: we should add more tests that cover the spread and interpolation operators in parallel. #494 has some tests for velocity interpolation.

drwells:
#494 just gained a spread test.

drwells:
Partly addressed by #597: we now have regression tests for the commonly used spreading kernels. We should write equivalents for interpolation.

drwells:
There are a few more things I would like to add before calling this issue resolved:
- [ ] interpolation tests that just print values on patches instead of computing errors. This will verify that we don't accidentally lower an order of accuracy somewhere.
- [ ] spread tests that work in parallel: we should verify that with a wacky partitioning and geometry we get the correct force values on each patch on each processor.

drwells:
I added another thing to do from a comment @boyceg made in #857.

drwells:
I checked the interpolation tests again and they print out max error norms - that means we will know if accuracy is degraded at particular points. Hence we can check off that issue.

PR #873, opened 2020-02-17
More spread tests 2

drwells:
Part of #509 and requested in #857.

This PR adds more spread tests that check:
1. force spreading from a structure that covers the whole domain
2. force spreading from a multi-part structure that covers the whole domain

To be extra-careful I checked these results against those generated by IBAMR 0.3 and they match up to roundoff (relative errors are about 1e-15) so we haven't made any breaking spread changes yet.


drwells:
I have two other PRs that I would like to put up that depend on these changes - it would be very helpful if someone could give this a quick lookover and approve the PR.

drwells:

PR #874, opened 2020-02-18
Fix the discontinuous linear 3D spread kernel.

drwells:
`ic_center` was read before initialized in the 3D version of this kernel. I fixed it and cleaned up the other kernel implementations to match.

The test output for this test is still unreliable but I have a more general fix for that in the pipeline.

Part of #784.

### DIFF
--- a/ibtk/src/lagrangian/fortran/lagrangian_interaction3d.f.m4
+++ b/ibtk/src/lagrangian/fortran/lagrangian_interaction3d.f.m4
@@ -280,7 +280,6 @@ c
      &           NINT((X_shifted(d)-x_lower(d))/dx(d)-0.5d0)
             X_cell(d) = x_lower(d) +
      &           (dble(ic_center(d)-ilower(d))+0.5d0)*dx(d)
-
             if ( d.eq.axis ) then
                if ( X_shifted(d).lt.X_cell(d) ) then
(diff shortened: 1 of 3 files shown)


ibtk/src/lagrangian/fortran/lagrangian_interaction3d.f.m4
```diff
@@ -406,12 +405,11 @@ c
 c     Determine the interpolation stencils and weights.
 c
          do d = 0,NDIM-1
+            ic_center(d) = ilower(d) +
+     &           NINT((X_shifted(d)-x_lower(d))/dx(d)-0.5d0)
```
drwells: Here's the problem - `ic_center` is not correctly calculated when `d.neq.axis`.

drwells:
I fixed a lot of roundoff problems in the test suite this week - if you still got inconsistent output with a test then that should no longer be an issue.

drwells:
I really want to merge this PR because it is the last thing we need (that I know of now) to get the spread tests passing on all tests I have access to at the moment :)

boyceg:
I thought I approved, but I see now that my comment was obscure --- I meant go ahead and fix the thing that this PR fixes now (i.e. merge the PR).

I think that I need to think more about how discontinuous kernels should work in the first place.

PR #875, opened 2020-02-18
Add some missing header inclusions.

drwells:
We cannot compile on some systems without explicitly including the memory header in IBStrategy.h - add it everywhere else we might need it while we are at it.


drwells:
This is both important to @jabrown893 and uncontroversial (I just added more ‘#include’s) so I will merge.

PR #876, opened 2020-02-19
Don't use Triangle in the spread tests.

drwells:
libMesh recently fixed a bug in their Triangle interface which can result in different node numberings. This in turn results in different quadrature formulas, which (finally) results in different spread force values. Since these tests just dump force values this change breaks our tests.

Get around the issue by just using the equivalent MeshTools::Generation function, which builds meshes by hand and is not likely to change.

This is necessary but not sufficient to resolve some problems with test portability.


drwells:

PR #877, opened 2020-02-19
Fix with unstable quadrature point placement.

drwells:
With roundoff, some quadrature points could end up at 0.4999 etc. or 0.5 - in the former case the cell center is calculated differently than in the later (e.g., 3/8 vs 4/8) when rounding.

Additionally, modify two tests to use hex grids to get more reproducible results for harder to understand reasons.

This should be the last set of changes needed to get stable test output :tada: 

This includes the 'correct' output file for the 3D discontinuous linear kernel - even though that test fails on this branch it will work once #874 is in.


drwells:
This works with a variety of machines and debug settings so I will merge myself for the same reason as the last few test suite PRs.

PR #878, opened 2020-02-19
extend IBFEMethod to allow parts to be activated/inactivated

boyceg:
Inactivated parts do not apply forces to the background grid or interpolate velocity from the background grid.

### DIFF
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -287,11 +287,11 @@ namespace IBAMR
  * </ol>
  *
  * <h2>Handling Restart Data</h2>
- * The caching of the IBFE restart data is not managed by SAMRAI's RestartManager. It 
+ * The caching of the IBFE restart data is not managed by SAMRAI's RestartManager. It
  * is instead handled by IBFEMethod::writeFEDataToRestartFile() given a restart_dump_dirname
  * and time_step_number. Each instance of IBFEMethod is registered for restart by default,
  * but the this option can be turned off. During a restart, the data is handled by the
- * RestartManager automatically to reinitiate the IBFEMethod.  
+ * RestartManager automatically to reinitiate the IBFEMethod.
  */
 class IBFEMethod : public IBStrategy
 {
@@ -343,6 +343,21 @@ public:
      * Indicate that a part should use stress normalization.
      */
     void registerStressNormalizationPart(unsigned int part = 0);
+
+    /**
+     * Activate a FE mesh part.
+     */
+    void activatePart(unsigned int part = 0);
+
+    /**
+     * Inactivate a FE mesh part.
+     */
+    void inactivatePart(unsigned int part = 0);
+
+    /**
+     * Check whether a FE part is activated.
+     */
+    bool getPartIsActivated(unsigned int part = 0);
 
     /*!
      * Typedef specifying interface for coordinate mapping function.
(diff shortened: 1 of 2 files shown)


include/ibamr/IBFEMethod.h
```diff
@@ -844,21 +859,21 @@ class IBFEMethod : public IBStrategy
...
+     * read back into the system along with the RestartManager data during restart.
+     *
+     * @code
+     * if (dump_restart_data && (iteration_num % restart_dump_interval == 0 || last_step))
+     * {
+     *     RestartManager::getManager()->writeRestartFile(restart_dump_dirname, iteration_num);
+     *     ib_method_ops->writeFEDataToRestartFile(restart_dump_dirname, iteration_num);
+     * }
+     * @endcode
+     */
```
drwells: oopsies!

boyceg:
@jabrown893 I think this will do what you want for activating/inactivating structures.

Basically all that happens when you "inactivate" a part is that the part stops spreading its force to the background grid, and stops interpolating velocity from the background grid. All other computations are done as usual.

src/IB/IBFEMethod.cpp
```diff
@@ -1115,8 +1145,6 @@ IBFEMethod::spreadFluidSource(const int q_data_idx,
         assertStructureOnFinestLevel();
         getScratchToPrimarySchedule(d_hierarchy->getFinestLevelNumber(), q_data_idx).fillData(data_time);
     }
-
-    return;
```
drwells: ?

boyceg: I've been starting to remove "unnecessary" `return` statements at the end of function bodies. (I used to include explicit `return` statements in all function bodies.) I can revert if you prefer to have these in all functions.

drwells:
I'd like to make a few small amendments to this PR:
- [ ] add a velocity interpolation and force spreading test to ensure that we really do disable things
- [ ] add a changelog entry
- [ ] restore the `return;` statement

@boyceg this should be quick to do (about an hour's work). Can I go ahead and push these changes to your branch and then merge myself once the CI is happy?

boyceg:
> I'd like to make a few small amendments to this PR:
> 
> * [ ]  add a velocity interpolation and force spreading test to ensure that we really do disable things
> * [ ]  add a changelog entry
> * [ ]  restore the `return;` statement
> 
> @boyceg this should be quick to do (about an hour's work). Can I go ahead and push these changes to your branch and then merge myself once the CI is happy?

Fine with me.

src/IB/IBFEMethod.cpp
drwells: I am ambivalent about the style but I would prefer to delete these all at once so that the style is consistent.

boyceg: I am not sure that the current style is very consistent at this point. However, if you want to revert this change, that is fine with me.

drwells:
@boyceg I shouldn't be surprised, but this is a bit trickier than I first thought it would be :)

Do we require that people regrid right after they reenable structures in other parts of the code?

boyceg:
> Do we require that people regrid right after they reenable structures in other parts of the code?

It should not be necessary to regrid right after re-enabling structures. What happens if you don't?

The only thing that should happen when you disable a structure is that it stops moving & it stops applying forces. Then when you turn it on again, it will start moving again and start applying forces again.

Now there _is_ a "collision" between the fluid and the structure that can cause bad things to happen, but that is sort of a separate issue.

drwells:
I want to be careful that we avoid the following circumstance:
1. user disables structure.
2. user regrids and a patch covering the structure is coarsened.
3. user reenables structure which is now on a different level (or split between multiple levels).

drwells:
I think the right answer is to ask the user to regrid immediately after reactivating a structure.

boyceg:
This should not be a problem -- the structure still is treated as something that interacts with the background grid. Literally all that happens is that the structure is assigned a velocity of 0, and its forces no longer show up in the momentum equation. Data structure management is not affected by activation/inactivation. In particular, the AMR grid will still cover the inactivated structure. (This is consistent with how activation/inactivation works with `LDataManager`/`IBMethod`.)

drwells:
@jabrown893 do you have to do anything special w.r.t. regridding when you reactivate structures to keep things working?

drwells:
You posted that at the same time I pinged Jordan. Makes sense.

boyceg:
> @jabrown893 do you have to do anything special w.r.t. regridding when you reactivate structures to keep things working?

I think you may have already figured this out, but the answer to this question is no. It might be better if the inactivated structures really disappeared from the computation, but they still trigger AMR, which means that users do not need to do anything special to re-activate them.

If inactivating structures impacted AMR, then absolutely we would need to ensure that regridding happened right after the status changed.

boyceg:
Superseded by #879.

PR #879, opened 2020-02-20
Inactivate parts api

drwells:
Alternative to #878 - I conferred with @jabrown893 to see how this is done in practice and we came up with some new ideas. In particular, we now have a common API for disabling structures that works in IBFEMethod and IBMethod.

I also added a test based on my favorite (IBFE ex4) that disables a massive structure in the middle of the domain to check that we really can turn things off correctly.

### DIFF
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -846,21 +846,21 @@ public:
     void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db) override;
 
     /*!
-    * For technical reasons this class does not use SAMRAI's RestartManager, so 
-    * restart files must be separately written for the IBFE objects. This function 
-    * saves the solutions to the defined EquationSystems in an xdr file in 
-    * restart_dump_dirname for each FE part. An example snippet is included below to show 
-    * the distinct IBFE restart data saving step. The data will then be automatically 
-    * read back into the system along with the RestartManager data during restart.
-    *
-    * @code
-    * if (dump_restart_data && (iteration_num % restart_dump_interval == 0 || last_step))
-    * {
-    *     RestartManager::getManager()->writeRestartFile(restart_dump_dirname, iteration_num);
-    *     ib_method_ops->writeFEDataToRestartFile(restart_dump_dirname, iteration_num);
-    * }
-    * @endcode   
-    */
+     * For technical reasons this class does not use SAMRAI's RestartManager, so
+     * restart files must be separately written for the IBFE objects. This function
+     * saves the solutions to the defined EquationSystems in an xdr file in
+     * restart_dump_dirname for each FE part. An example snippet is included below to show
+     * the distinct IBFE restart data saving step. The data will then be automatically
+     * read back into the system along with the RestartManager data during restart.
+     *
+     * @code
+     * if (dump_restart_data && (iteration_num % restart_dump_interval == 0 || last_step))
+     * {
+     *     RestartManager::getManager()->writeRestartFile(restart_dump_dirname, iteration_num);
+     *     ib_method_ops->writeFEDataToRestartFile(restart_dump_dirname, iteration_num);
+     * }
+     * @endcode
+     */
     void writeFEDataToRestartFile(const std::string& restart_dump_dirname, unsigned int time_step_number);
 
     /*!
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -618,6 +618,35 @@ public:
                         SAMRAI::tbox::Pointer<SAMRAI::mesh::GriddingAlgorithm<NDIM> > gridding_alg) const override;
 
     /*!
+     * Inactivate a structure/part. See IBAMR::IBStrategy::inactivateLagrangianStructure().
+     *
+     * @note Since this class assumes that structures live on the finest grid
+     * level the second argument is ignored.
+     */
+    virtual void inactivateLagrangianStructure(int structure_number = 0,
+                                               int level_number = std::numeric_limits<int>::max()) override;
+
+    /*!
+     * Activate a previously inactivated structure/part to be used again in
+     * FSI calculations. See IBAMR::IBStrategy::activateLagrangianStructure().
+     *
+     * @note Since this class assumes that structures live on the finest grid
+     * level the second argument is ignored.
+     */
+    virtual void activateLagrangianStructure(int structure_number = 0,
+                                             int level_number = std::numeric_limits<int>::max()) override;
+
+    /*!
+     * Determine whether or not the given structure or part is currently
+     * activated. See IBAMR::IBStrategy::getLagrangianStructureIsActivated().
+     *
+     * @note Since this class assumes that structures live on the finest grid
+     * level the second argument is ignored.
+     */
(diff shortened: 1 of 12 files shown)


drwells:
I'm not as familiar with how to set up structures with IBMethod. It would be nice to add a test for it but someone else should write it.

jabrown893:
I created such a test when Boyce fixed the function for deactivating lagrangian structures. It was PR #699 
Could we just swap the old deactivating function out for your new one?

boyceg:
Seems like a good idea, although note that at present, `IBFEMethod` does not really have a notion of assigning different parts to different levels.

(In fact, my plan is to embed this information in the mesh itself, so that different elements within a given FE structure will interact with different AMR grid levels.)

drwells:
I forgot about #699 - we should modify that test so that an input file determines which code path we use for deactivating so that we know things work the same way if we go through `LDataManager` or `IBStrategy`. I'll take a look at that later today.

drwells:
I added a test variant based on the new test in #699 that shows that we can deactivate a structure correctly from IBMethod as well as LDataManager.

drwells:
@boyceg Since the second argument (the level number) is optional this should be forward-compatible with any new indexing schemes we apply to structures owned by IBFEMethod - in fact, the level number is already unconditionally ignored.

@jabrown893 Does this patch work well for your use case?

boyceg:
Sounds fine.

src/IB/IBFEMethod.cpp
```diff
@@ -884,34 +928,50 @@ IBFEMethod::computeLagrangianForce(const double data_time)
     batch_vec_ghost_update(d_X_half_vecs, INSERT_VALUES, SCATTER_FORWARD);
     for (unsigned part = 0; part < d_num_parts; ++part)
     {
-        if (d_is_stress_normalization_part[part])
+        if (d_part_is_active[part])
```
boyceg: Let's not include this guard --- I think that I would like for all computations to proceed "as usual" for inactivated parts, except that at the end of the day, forces are not spread and velocity is not interpolated.

src/IB/IBFEMethod.cpp
```diff
@@ -820,7 +860,9 @@ IBFEMethod::forwardEulerStep(const double current_time, const double new_time)
     const double dt = new_time - current_time;
     for (unsigned int part = 0; part < d_num_parts; ++part)
     {
-        int ierr = VecWAXPY(d_X_new_vecs[part]->vec(), dt, d_U_current_vecs[part]->vec(), d_X_current_vecs[part]->vec());
+        if (!d_part_is_active[part]) continue;
```
boyceg: Let's not include this guard --- I think that I would like for all computations to proceed "as usual" for inactivated parts, except that at the end of the day, forces are not spread and velocity is not interpolated.

src/IB/IBFEMethod.cpp
```diff
@@ -840,6 +882,7 @@ IBFEMethod::midpointStep(const double current_time, const double new_time)
     const double dt = new_time - current_time;
     for (unsigned int part = 0; part < d_num_parts; ++part)
     {
+        if (!d_part_is_active[part]) continue;
```
boyceg: Let's not include this guard --- I think that I would like for all computations to proceed "as usual" for inactivated parts, except that at the end of the day, forces are not spread and velocity is not interpolated.

src/IB/IBFEMethod.cpp
```diff
@@ -860,6 +903,7 @@ IBFEMethod::trapezoidalStep(const double current_time, const double new_time)
...
```
boyceg: Let's not include this guard --- I think that I would like for all computations to proceed "as usual" for inactivated parts, except that at the end of the day, forces are not spread and velocity is not interpolated.

src/IB/IBFEMethod.cpp
```diff
@@ -884,34 +928,50 @@ IBFEMethod::computeLagrangianForce(const double data_time)
...
-        d_F_systems[part]->solution->zero();
-        d_active_fe_data_managers[part]->computeL2Projection(*d_F_systems[part]->solution,
-                                                             *d_F_rhs_vecs[part],
-                                                             FORCE_SYSTEM_NAME,
-                                                             d_use_consistent_mass_matrix,
-                                                             /*close_U*/ false,
-                                                             /*close_F*/ false);
-        *d_F_half_vecs[part] = *d_F_systems[part]->solution;
-        if (d_direct_forcing_kinematics_data[part])
+        if (d_part_is_active[part])
```
boyceg: Let's not include this guard --- I think that I would like for all computations to proceed "as usual" for inactivated parts, except that at the end of the day, forces are not spread and velocity is not interpolated.

src/IB/IBFEMethod.cpp
```diff
@@ -991,6 +1053,7 @@ IBFEMethod::computeLagrangianFluidSource(double data_time)
     for (unsigned int part = 0; part < d_num_parts; ++part)
     {
         if (!d_lag_body_source_part[part]) continue;
+        if (!d_part_is_active[part]) continue;
```
boyceg: Merge with the previous `if`? I think it is easier to follow the logic when there is just one `continue` in a block.

src/IB/IBFEMethod.cpp
```diff
@@ -1100,6 +1163,8 @@ IBFEMethod::spreadFluidSource(const int q_data_idx,
...
```
boyceg: Merge with the previous `if`? I think it is easier to follow the logic when there is just one `continue` in a block.

src/IB/IBFEMethod.cpp
```diff
@@ -1612,6 +1679,7 @@ IBFEMethod::addWorkloadEstimate(Pointer<PatchHierarchy<NDIM> > hierarchy, const
         std::vector<std::size_t> dofs_per_processor(n_processes);
         for (unsigned int part = 0; part < d_num_parts; ++part)
         {
+            if (!d_part_is_active[part]) continue;
```
boyceg: Let's not include this guard --- I think that I would like for all computations to proceed "as usual" for inactivated parts, except that at the end of the day, forces are not spread and velocity is not interpolated.

include/ibamr/IBFEMethod.h
```diff
@@ -617,6 +615,35 @@ class IBFEMethod : public IBStrategy
...
+    /*!
+     * Determine whether or not the given structure or part is currently
+     * activated. See IBAMR::IBStrategy::getLagrangianStructureIsActivated().
+     *
+     * @note Since this class assumes that structures live on the finest grid
+     * level the second argument is ignored.
+     */
+    virtual bool getLagrangianStructureIsActivated(int structure_number = 0,
+                                                   int level_number = std::numeric_limits<int>::max()) const override;
+
```
boyceg: Could we also go ahead and add some methods to keep track of the "names" of parts and relate them to part numbers? This functionality is provided in `LDataManager` to make it easier to manage activating/inactivating IB parts, and might be useful in general.

drwells: At some point we need to actually implement `class Part {}` which should include such functionality. Could we just open a separate issue to track that instead of adding such features to this PR?

boyceg: Sure thing.

drwells:
Fixed!

PR #880, opened 2020-02-21
move some vector allocations into initialization functions

boyceg:


amneetb:
@nishantnangia : Can you test this branch locally for multiphase solver?

nishantnangia:
I checked some cases earlier this evening and nothing seemed to
break/change — so I approved the PR.

src/navier_stokes/VCStaggeredStokesOperator.cpp
```diff
@@ -85,9 +85,6 @@ VCStaggeredStokesOperator::apply(SAMRAIVectorReal<NDIM, double>& x, SAMRAIVector
 {
     IBAMR_TIMER_START(t_apply);
 
-    // Allocate scratch data.
-    d_x->allocateVectorData();
-
```
drwells: It looks like you forgot one.

boyceg:
> @drwells requested changes on this pull request.
> We need to make a similar fix to VCStaggeredStokesOperator (its missing the allocate now, AFAICT)
> 

Those happen in the base class I think?

drwells:
Whoops - you're absolutely right.

drwells:
@nishantnangia and I are both happy so lets merge!

drwells:
@marshallrdavey This PR might actually help with the increase in memory usage problem. If its easy to do, could you use `master` for the next round of heart simulations and see if there is any change in the output from `getrusage`?

src/navier_stokes/VCStaggeredStokesOperator.cpp
boyceg: Just to close the loop --- allocation/deallocation is being handled in the base class.

PR #882, opened 2020-02-21
Consolidated various scripts into scripts directory.

### LINKED ISSUE #795: Reorganize scripts in src/tools
From #788: we have a bunch of useful perl scripts buried in the source folder. We should put them with the rest of the scripts.

abarret:
Fixes #795. I've consolidated `src/tools/` and scripts directory and provided an explanation of what each script does. I've also organized them into different directories. I deleted one script that seemed to not have much use.


abarret:
Forced pushed to fix the makefiles.

PR #883, opened 2020-02-22
Accumulate forces in some spread tests.

drwells:
This was the original intent of these tests.


drwells:
This works with a variety of machines and debug settings so I will merge myself for the same reason as the last few test suite PRs.

PR #887, opened 2020-02-24
Fix a bug in FEDataManager::spread().

drwells:
Found while working on #857 - it looks like this function hasn't ever worked correctly when one did not provide the boundary ops object since we spread values into ghost regions and then do not add those values back to the provided force index.


boyceg:
> Found while working on #857 - it looks like this function hasn't ever worked correctly when one did not provide the boundary ops object since we spread values into ghost regions and then do not add those values back to the provided force index.

You could imagine modifying the definition of the kernel function near the boundary, so that nothing gets spread into the exterior ghost cells. In fact, this is exactly how we originally handled interactions at the physical boundary. The main advantage of the current approach is that it makes the kernel specification sort of independent from the handling physical BCs.

drwells:
That's a good reason to keep this parameter optional.

PR #889, opened 2020-02-25
Explicitly test interpolated velocity values.

drwells:
This commit adds two sets of tests:
1. test that we can reproduce bilinear functions exactly
2. record interpolated values of a trigonometric function (values are checked by eyeball).

Unlike the current velocity tests these check pointwise values and work without libMesh.

I wrote this as an extra check before we merge #854.


drwells:
I think I can fix a minor issue with linear refinement - don't merge this yet.

drwells:
Its not what I originally planned on doing but this is enough that I'll call #509 done once this is merged.

drwells:
Since this just adds tests I will go ahead and merge myself.

PR #890, opened 2020-02-26
Remove some tests that have inconsistent output.

drwells:
The original objective of these tests was to verify that these kernels work - this is now duplicated by the new interpolate_01 test. Hence, since we have lots of other tests for interpWeighted, we can get rid of these four tests (which have inconsistent point placement due to roundoff at different optimization levels).

Part of #784 - these tests don't have memory problems but they are listed there since the output is unstable.


drwells:
The way to get tests to pass is to delete them :joy_cat:

PR #891, opened 2020-02-27
Don't use libMesh's compilation flags.

drwells:
This is not necessary and its better that we set these flags ourselves.


drwells:
That's a good point - unfortunately libMesh does use glibcxx-debug (an incompatible ABI) in debug mode, but thats not compatible with SAMRAI or other C++ libraries so I always disable it (and I recommend disabling it in the IBAMR build instructions).

boyceg:
Are you sure this breaks SAMRAI? I am not sure, but it has been a long time.

Can/do we verify in configure that we can link libMesh works with the provided compiler flags?

drwells:
> Are you sure this breaks SAMRAI?

I looked at it ages ago but I don't remember if I tried. Since SAMRAI and libMesh both use std::string I think things are likely to break since that container has a different debug implementation.

I'll take a look to see if I can reproduce some bad stuff.

drwells:
In particular:
```cpp
#include <iostream>
#include <string>

#include <debug/string>

int main()
{
    std::string a;
    __gnu_debug::string b;

    std::cout << "sizeof(std::string) = " << sizeof(a) << '\n';
    std::cout << "sizeof(__gnu_debug::string) = " << sizeof(b) << '\n';
}

```
outputs
```
sizeof(std::string) = 32
sizeof(__gnu_debug::string) = 56
```

boyceg:
OK, well, feel free to merge as you see fit.

drwells:
Since the build directions for libMesh explicitly state that one should disable glibcxx debugging (and do not enable it for any of our other C++ dependencies) I think that this change is fine, so I will merge.

PR #893, opened 2020-02-29
Add ghost accumulator

drwells:
Another large chunk of #830. This PR adds a new class that can accumulate ghost data (usually obtained by spreading forces into ghost cells) into the patch/processor that actually owns that cell. This works with cell and side data - I added 20 tests to verify the implementation.

Most of the tests I added actually use `LEInteractor::spread()` - i.e., the tests show that the planned approach actually works.


drwells:
I also verified that this works by turning off accumulation - we get the correct values (except at periodic boundaries) in that case when we have a single patch.

drwells:
I threw this into the heart model and while its not a panacea it lowers the amount of time we spend in post spreading barriers by about 75% (when summed across processors). Very crudely measured: we spend 1000 seconds waiting after we spread without the scratch hierarchy, 361 with the scratch hierarchy, and 81 with the scratch hierarchy + this class (which includes some other relevant changes - e.g., we associate elements with a patch if they are within 1 cell of the patch interior instead of the ghost width). We do have other performance problems up ahead but we should definitely use this class to redo the way we spread forces.

At the same time the performance in this class is awful!
```
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |       616s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| assemble                        |       100 |      7.97s |       1.3% |
| copy_back                       |       100 |      15.8s |       2.6% |
| copy_to_vec                     |       100 |       0.4s |         0% |
+---------------------------------+-----------+------------+------------+
```

we spend 5% of our total runtime just adding things up because we are using `VecStash` and not the ghost arrays correctly. I think this is solvable if I spend some time following the advice fox index management given in the PETSc manual.

drwells:
I rewrote this class to improve performance (its about 5x faster) because we now store local indices and use those to directly populate SAMRAI data structures:

```
+---------------------------------------------+------------+------------+
| Total CPU time elapsed since start          |       838s |            |
|                                             |            |            |
| Section                         | no. calls |  CPU time  | % of total |
+---------------------------------+-----------+------------+------------+
| assemble_backward               |       100 |      1.56s |      0.19% |
| assemble_forward                |       100 |      3.79s |      0.45% |
| copy_back                       |       100 |     0.268s |         0% |
| copy_to_vec                     |       100 |     0.372s |         0% |
+---------------------------------+-----------+------------+------------+
```
the communication steps are still surprisingly expensive - it takes a lot more time than the equivalent code for libMesh vectors. I suppose we just have a lot more data to move.

> Looks good. Do we still implement spreading to node-centered data?

Yes - its supported by LEInteractor (edge data is also supported). To the best of my knowledge we have no tests or code that actually use this so I would prefer to not implement that now. Its not difficult to do (it should be the same code path as cell-centered data) but there is no reason to implement it now if no one is going to use it.

boyceg:
> I rewrote this class to improve performance (its about 5x faster) because we now store local indices and use those to directly populate SAMRAI data structures:
> 
> ```
> +---------------------------------------------+------------+------------+
> | Total CPU time elapsed since start          |       838s |            |
> |                                             |            |            |
> | Section                         | no. calls |  CPU time  | % of total |
> +---------------------------------+-----------+------------+------------+
> | assemble_backward               |       100 |      1.56s |      0.19% |
> | assemble_forward                |       100 |      3.79s |      0.45% |
> | copy_back                       |       100 |     0.268s |         0% |
> | copy_to_vec                     |       100 |     0.372s |         0% |
> +---------------------------------+-----------+------------+------------+
> ```
> 
> the communication steps are still surprisingly expensive - it takes a lot more time than the equivalent code for libMesh vectors. I suppose we just have a lot more data to move.

SAMRAI communication?

drwells:
We don't actually use SAMRAI for any communication in this function (aside from initially setting the DoF numbering) - all communication happens with PETSc in specified ghost regions.

boyceg:
Are you now avoiding `VecStash`?

drwells:
That was my intent since we only use `VecGhostUpdateBegin`/`VecGhostUpdateEnd`. VecStash has a way of creeping in when I don't want it to - I should double-check.

drwells:
I added some calls to `VecStashView` and it looks like the stash is always empty (which is what I expected). I suggest we go with the current version of this PR - we are doing things in the way the PETSc manual recommends. We can always go back and tweak it further later.

drwells:
@boyceg Do you want to take a second look? If not I'll just merge.

boyceg:
Go ahead.

drwells:
GitHub's settings are somehow screwed up - give me a minute (don't merge this yet).

drwells:
Fixed.

PR #896, opened 2020-03-02
all changes for redoing force spreading

### LINKED ISSUE #830: Reimplement force spreading accumulation from ghost cells

drwells:
Fixes #830. This change is mostly backwards compatible (we have some new restrictions on the ghost cell width of what we spread to in IBFEMethod::spread and FEDataManager::spread works in a completely different way) - all the standard IB stuff still works in the same way it did before.

The current implementation works just fine for volumetric meshes but there are still some problems that need to be addressed:
- [x] IBFESurfaceMethod needs to be updated in the same way (since FEDataManager changed, all classes that spread with FEDataManager need to also change)
- [x] all the `ib_partitioning` tests are now broken since we now associate an element with a patch if it is in the patch's bounding box extended by one cell instead of extended by 1 + stencil_width/2 cells. (this is a good thing - we cut down on the total number of elements we loop over by about 25%)
- [x] We need a way to check that the points we use for spreading are globally unique. AFAIK the only issue here is that a quadrature point might end up exactly on the border of two patches and then be counted twice. We can overcome this by looping over the interaction points and zeroing out forces corresponding to points on said borders.
- [x] One of the `ex8` test now fails due to very small (O(10^-6) changes - this might be due to double counting quadrature points but we should try to understand this. (Fixed by #936)

Edit: clarify the ghost region specification for the `ib_partitioning` tests


boyceg:
> all the `ib_partitioning` tests are now broken since we only need 1 cell in the ghost region instead of 1 + stencil_width/2

A question that I should have already asked: Doesn't the accumulation code look at cells beyond the first layer of ghost cells? If so, are we still using patch data with a large ghost cell width to handle physical boundaries?

drwells:
Good question. My original statement was ambiguous: I should have written "*when associating elements with patches* we only need to use a ghost region one cell wide instead of 1 + stencil_width/2". We definitely still use sufficiently wide ghost regions (though we should be able to cut that down to stencil_width/2 in IBFE now - thats not in this patch).

drwells:
OK, there are two remaining unrelated problems now:
1. I added some code to show when we get quadrature points on patch boundaries. This happens basically never for unstructured grids but it makes significant differences for things like IBFE/ex8 (where structures are rectangles aligned with the grid).
2. I wrote two tiny tests for ex8 and I get identical (down to the last bit) results on master with and without the scratch hierarchy and on this branch without the scratch hierarchy. For some reason combining this branch with the scratch hierarchy causes issues. I'm working on this one now.

drwells:
This is now done :tada:

src/IB/IBFEMethod.cpp
```diff
@@ -2682,23 +2693,9 @@ IBFEMethod::spreadTransmissionForceDensity(const int f_data_idx,
...
     if (!integrate_normal_force && !integrate_tangential_force) return;
 
-    const int coarsest_ln = 0;
-    const int finest_ln = d_hierarchy->getFinestLevelNumber();
-    VariableDatabase<NDIM>* var_db = VariableDatabase<NDIM>::getDatabase();
-
-    // Make a copy of the Eulerian data.
-    Pointer<hier::Variable<NDIM> > f_var;
-    var_db->mapIndexToVariable(f_data_idx, f_var);
-    const int f_copy_data_idx = var_db->registerClonedPatchDataIndex(f_var, f_data_idx);
```
drwells: Since we already use a temporary variable in `spreadForce` we can completely remove this extra patch index here.

drwells:
This ASCII art only makes sense on GitHub (sorry). Suppose we have two patches on a physical boundary:
```
+-+-+-+-+-+-B-+-+-+-+-+-+
| | | | | | B | | | | | |
+-+-+-+-+-+-B-+-+-+-+-+-+
| | | | | | B | | | | | |
+-+-+-+-+-+-B-+-+-+-+-+-+
| | | | | | B | | | | | |
+-+-+-+-+-+-B-+-+-+-+-+-+
| | | | | |xB | | | | | |
+-+-+-+-+-+-B-+-+-+-+-+-+
```
here `x` is the location of an IB point and the `B`s form the edge separating the two patches. When we spread values we get the following:
```
+-+-+-+-+-+-B-+-+-+-+-+-+
| | | | | | B | | | | | |
+-+-+-+-+-+-B-+-+-+-+-+-+
| | | | | | B | | | | | |
+-+-+-+-+-+-B-+-+-+-+-+-+
| | | | |X|XBX| | | | | |
+-+-+-+-+-+-B-+-+-+-+-+-+
| | | | |X|XBX| | | | | |
+-+-+-+-+-+-B-+-+-+-+-+-+
         G G G
```
i.e., four values on the current patch interior (as expected), two on the neighboring patch interior (handled by the accumulation step), and three outside the current patch (handled by RobinPhysPatchBdry)

The question: is the rightmost G accumulated into the ghost region of the left patch?

drwells:
I might write some more tests later to verify that this does what I think it does.

boyceg:
> This ASCII art only makes sense on GitHub (sorry). Suppose we have two patches on a physical boundary:
> 
> ```
> +-+-+-+-+-+-B-+-+-+-+-+-+
> | | | | | | B | | | | | |
> +-+-+-+-+-+-B-+-+-+-+-+-+
> | | | | | | B | | | | | |
> +-+-+-+-+-+-B-+-+-+-+-+-+
> | | | | | | B | | | | | |
> +-+-+-+-+-+-B-+-+-+-+-+-+
> | | | | | |xB | | | | | |
> +-+-+-+-+-+-B-+-+-+-+-+-+
> ```
> 
> here `x` is the location of an IB point and the `B`s form the edge separating the two patches. When we spread values we get the following:
> 
> ```
> +-+-+-+-+-+-B-+-+-+-+-+-+
> | | | | | | B | | | | | |
> +-+-+-+-+-+-B-+-+-+-+-+-+
> | | | | | | B | | | | | |
> +-+-+-+-+-+-B-+-+-+-+-+-+
> | | | | |X|XBX| | | | | |
> +-+-+-+-+-+-B-+-+-+-+-+-+
> | | | | |X|XBX| | | | | |
> +-+-+-+-+-+-B-+-+-+-+-+-+
>          G G G
> ```
> 
> i.e., four values on the current patch interior (as expected), two on the neighboring patch interior (handled by the accumulation step), and three outside the current patch (handled by RobinPhysPatchBdry)
> 
> The question: is the rightmost G accumulated into the ghost region of the left patch?

It depends on how the boundary boxes are "trimmed" in the accumulation routine.

drwells:
Do you mean the interpatch (SAMRAIGhostDataAccumulator) or intrapatch (RobinPhysPatchBdry) accumulations? In the former case we do not trim anything (but since ghost cells outside the physical domain do not have DoFs they are ignored)

drwells:
I think I convinced myself that this is fine - we have some spread tests where the structure is the same size as the domain and those still work.

drwells:
I added tests just to triple-check my understanding and they pass both on master and on this branch. Hence this is good to go.

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -1012,14 +1041,15 @@ class FEDataManager : public SAMRAI::tbox::Serializable
 
     /*!
      * Collect all of the active elements which are located within a local
-     * Cartesian grid patch grown by the specified ghost cell width.
+     * Cartesian grid patch grown by a ghost width of 1 (like
+     * IBTK::LEInteractor::getMinimumGhostWidth(), we assume that IB points
+     * are allowed to move no more than one cell width between regridding
```
boyceg, line 1046: Can we _easily_ encode this assumption about the # of cells points can move between regridding operations somewhere, perhaps as a constant member variable? (If it is not easy to do, let's not worry about it for now.)

drwells:
That's a good question - the assumption "IB points move less than 1 cell length between regrids` is baked into `LEInteractor::getMinimumGhostCellWidth()`. Would you mind if I addressed this in a followup PR where we add a constant to `LEInteractor` that hard-codes `1`, e.g., `LEInteractor::max_cells_moved_between_regrids = 1`?

ibtk/include/ibtk/FEDataManager.h
drwells: Not really; see below.

boyceg:
Sounds fine.

drwells:
All right :) Let's merge!

PR #897, opened 2020-03-03
store part activity state in IBFEMethod restart database

boyceg:

PR #898, opened 2020-03-03
Add a test for inactivation + restart

drwells:
Quick follow-up to #897 - this test verifies that we do inactivate the structure correctly after restarting.


drwells:
You aren't bad - we just need a process improvement :)

drwells:
I added a changelog entry and also incremented `IBFE_METHOD_VERSION`.

boyceg:
Thank you!

PR #902, opened 2020-03-04
Force ghosted assembly

drwells:
From the heart model meeting - this makes a marginal performance difference (maybe 0.25% improvement?) but it should make things more consistent in parallel. It also shows how we can modify other systems if necessary to use the same tricks.

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -937,7 +937,8 @@ IBFEMethod::computeLagrangianForce(const double data_time)
         assembleInteriorForceDensityRHS(
             *d_F_rhs_vecs[part], *d_X_half_vecs[part], d_Phi_half_vecs[part], data_time, part);
     }
-    batch_vec_assembly(d_F_rhs_vecs);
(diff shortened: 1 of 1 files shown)


boyceg:
Can we also / do we need to do this in `FEDataManager::computeL2Projection()`?

drwells:
Do you mean `interpWeighted`? We already do ghosted assembly there - see #557.

boyceg:
Yeah.

Is this the only place left in the code at present where we do FE-type assembly without using ghosts?

drwells:
Surprisingly it is - except for the non-default behavior we left in for velocity interpolation.

drwells:
On second thought - I'm not sure. We still call `VecAssemblyBegin` in `CIBMethod`, `CIBStrategy`, and a few other places. I'm not sure if we are relying on `VecStash or not for those.

boyceg:
Ah, I really mean to be asking if this is the only place where we do FE-type assembly in `IBFEMethod` and friends.

drwells:
I believe so - we don’t have any mandatory communication in FEDataManager. We definitely do use ‘PetscVector::close()’ in IBFESurfaceMethod but we can’t really change things there without conflicting with #828.

PR #903, opened 2020-03-05
IBTK Initialization class

abarret:
This is a follow up from #287. We need to add an initialization class to ensure that the default communicator in `IBTK_MPI` is set. This also initializes standard libraries to make a transition to SAMRAI 3 easier (SAMRAI 3 adds required initialization calls). I also went ahead and moved templated `IBTK_MPI` functions to a separate private file.

This PR just adds the class, but does change any of the examples.

### DIFF
--- a/ibtk/include/ibtk/IBTKInit.h
+++ b/ibtk/include/ibtk/IBTKInit.h
@@ -1,0 +1,108 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2014 - 2019 by the IBAMR developers
+// All rights reserved.
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+#ifndef included_IBTKInit
+#define included_IBTKInit
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include <IBTK_config.h>
+
+#include "SAMRAI_config.h"
+#include "tbox/SAMRAIManager.h"
+#include "tbox/SAMRAI_MPI.h"
+
+#include <petscsys.h>
+#ifdef IBTK_HAVE_LIBMESH
+#include "libmesh/libmesh.h"
+#include "libmesh/reference_counter.h"
+#endif
+
+namespace IBTK
+{
+/**
+ * @brief Initialization for IBAMR programs.
+ *
+ * The singleton IBTKInit class handles the initializations for PETSc, LibMesh, and SAMRAI. This object should be
+ * created using the initialize() function at the start of the main() function. The destruction of the object correctly
+ * closes the libraries.
+ *
+ */
+
+class IBTKInit
+{
+public:
+    /**
+     * Constructor for IBTKInit. Initializes libraries and sets the SAMRAI world communicator. Attempts to create a
+     * second IBTKInit object will result in a run time error.
+     */
+    IBTKInit(int argc,
+             char** argv,
+             MPI_Comm communicator = MPI_COMM_WORLD,
+             char* petsc_file = nullptr,
+             char* petsc_help = nullptr);
+    /*!
+     * \brief Default constructor. This function is not implemented and should not be used.
+     */
+    IBTKInit() = delete;
+
+    /*!
+     * \brief Copy constructor. This function is not implemented and should not be used.
+     */
+    IBTKInit(const IBTKInit& from) = delete;
+
+    /*!
+     * \brief Assignment operator. This function is not implemented and should not be used.
+     */
+    IBTKInit& operator=(const IBTKInit& that) = delete;
+
+    /**
+     * Destructor. Closes libraries appropriately.
+     */
+    ~IBTKInit();
+
+#ifdef IBTK_HAVE_LIBMESH
+    /**
+     * Get libMesh initialization object.
+     */
+    libMesh::LibMeshInit& getLibMeshInit()
+    {
+        return d_libmesh_init;
+    }
+#endif
+
+    /*!
+     * \brief Check if the library has been initialized. Throw an error if it has not been initialized.
+     */
+    inline static bool check_initialized()
+    {
+        if (!s_initialized)
+        {
+            TBOX_ERROR(
+                "IBAMR is not initialized! IBAMR must be initialized by an appropriate call to IBTKInit::initialize() "
+                "prior to using any of the library.");
+        }
+        return s_initialized;
+    }
+
+private:
+
+#ifdef IBTK_HAVE_LIBMESH
+    libMesh::LibMeshInit d_libmesh_init;
+#endif
+    static bool s_initialized;
+};
+
+} // namespace IBTK
+
+#endif
--- a/ibtk/lib/Makefile.am
+++ b/ibtk/lib/Makefile.am
@@ -36,7 +36,8 @@ clean-local:
 libIBTK_a_SOURCES = \
 ../src/utilities/ParallelEdgeMap.cpp \
 ../src/utilities/FixedSizedStream.cpp \
-../src/utilities/IBTK_MPI.cpp
+../src/utilities/IBTK_MPI.cpp \
+../src/utilities/IBTKInit.cpp
 
 if USING_BUNDLED_MUPARSER
 libIBTK_a_SOURCES += \
@@ -239,6 +240,7 @@ pkg_include_HEADERS += \
 ../include/ibtk/HierarchyIntegrator.h \
 ../include/ibtk/HierarchyMathOps.h \
 ../include/ibtk/IBTK_MPI.h \
+../include/ibtk/IBTKInit.h \
 ../include/ibtk/IndexUtilities.h \
 ../include/ibtk/JacobianOperator.h \
 ../include/ibtk/KrylovLinearSolver.h \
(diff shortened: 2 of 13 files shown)


ibtk/src/utilities/IBTKInit.cpp
```diff
@@ -0,0 +1,79 @@
...
+#ifdef IBTK_HAVE_LIBMESH
+    s_libmesh_init = new LibMeshInit(argc, argv, communicator);
+    libMesh::ReferenceCounter::disable_print_counter_info();
+    NULL_USE(petsc_file);
+    NULL_USE(petsc_help);
+#else
+    // We need to initialize PETSc.
+    PetscInitialize(&argc, &argv, petsc_file, petsc_help);
+#endif
+#if SAMRAI_VERSION_MAJOR > 2
```
drwells: :tada:

ibtk/src/utilities/IBTKInit.cpp
```diff
@@ -0,0 +1,79 @@
...
+    }
+    static IBTKInit instance(argc, argv, communicator, petsc_file, petsc_help);
+}
+
+/////////////////////////////// PRIVATE //////////////////////////////////////
+
+IBTKInit::IBTKInit(int argc, char** argv, IBTK_MPI::comm communicator, char* petsc_file, char* petsc_help)
+{
+#ifdef IBTK_HAVE_LIBMESH
+    s_libmesh_init = new LibMeshInit(argc, argv, communicator);
```
drwells: We (perhaps only me) sometimes use this in programs - we should make it returnable by reference.

ibtk/src/utilities/IBTKInit.cpp
```diff
@@ -0,0 +1,79 @@
...
+    libMesh::ReferenceCounter::disable_print_counter_info();
```
drwells: I'd prefer not to disable this unconditionally - it is occasionally useful to see. Users should disable it on their own if they don't want to see what libMesh has to say.

boyceg: I am not sure, but I think that this may be to prevent emitting some libMesh messages in models that do not explicitly use libMesh.

drwells: That's a good point - for that reason we should disable it.

boyceg: Can we re-enable it after it is disabled? I like this counter also.

Alternatively, would it be possible to somehow conditionally disable?

drwells: I suspect that it will be pretty rare for users to actually care so we should recommend they call `libMesh::ReferenceCounter::enable_print_counter_info()` on their own if they want this information.

ibtk/src/utilities/IBTKInit.cpp
abarret: It's returnable by pointer. Would reference be better?

abarret:
> Looks good. Could I convince you to write a tiny test that verifies that we do set up everything as promised?

I'll work on writing a test.

ibtk/src/utilities/IBTKInit.cpp
drwells: I missed that function in the header - oops.

I would prefer to have it returned by reference.

boyceg: Would it be possible to use pointers / smart pointers in a way that makes access to `LibMeshInit` safer after shutting things down?

drwells: Yes  - we did this a while back in deal.II with the `SmartPointer` class. essentially, one must also store a boolean indicating whether or not the pointed-to object is alive.

drwells: Alternatively, we could make `LibMeshInit` a member of this class instead of statically allocating it - that is probably the better solution.

boyceg: That seems best to me -- seems like it avoids most of the de-initialization issues that I am imagining.

drwells: Yes. In particular, the only static member of this class should be a boolean storing whether or not we have already created it.

abarret:
@drwells take a look now. I added a test and made the libMesh initialization object returnable by reference.

ibtk/include/ibtk/IBTKInit.h
```diff
@@ -0,0 +1,119 @@
...
+     */
+    IBTKInit& operator=(const IBTKInit& that) = delete;
+
+    using comm = MPI_Comm;
+
+    /**
+     * The initialize function creates and returns a static IBTKInit object which initializes all pertinent libraries.
+     * Repeated calls to initialize() return the IBTKInit object.
+     */
+    static IBTKInit& initialize(int argc,
```
boyceg: @drwells, do you have an opinion about whether this should use an initializer function instead of a constructor?

If there is an initializer should there also be a deinitializer?

ibtk/src/utilities/IBTKInit.cpp
boyceg: Is there a safe way to lazy-initialize libMesh? Not all IBAMR models will use libMesh, and it seems better not to initialize it if it is not used.

ibtk/include/ibtk/IBTKInit.h
drwells: We should definitely just use a constructor and destructor.

ibtk/lib/Makefile.am
```diff
@@ -36,7 +36,8 @@ clean-local:
 libIBTK_a_SOURCES = \
 $(top_builddir)/src/utilities/ParallelEdgeMap.cpp \
 $(top_builddir)/src/utilities/FixedSizedStream.cpp \
-$(top_builddir)/src/utilities/IBTK_MPI.cpp
+$(top_builddir)/src/utilities/IBTK_MPI.cpp \
+$(top_builddir)/src/utilities/IBTKInit.cpp
```
drwells: This conflicts with #904 - just replace `$(top_builddir)` with `../src/`.

ibtk/src/utilities/IBTKInit.cpp
drwells: I don't think so - I think we are stuck always initializing it.

ibtk/src/utilities/IBTK_MPI.cpp
```diff
@@ -35,12 +36,14 @@ IBTK_MPI::setCommunicator(IBTK_MPI::comm communicator)
 IBTK_MPI::comm
 IBTK_MPI::getCommunicator()
 {
+    IBTKInit::check_initialized();
```
drwells: This set of changes seems overly invasive - lets just set the default communicator to `MPI_COMM_NULL` and let MPI crash if we make any calls with that.

ibtk/src/utilities/IBTKInit.cpp
boyceg: What if we leave application codes to setup `LibMeshInit` or not, and to have them pass in to the constructor if they are using libMesh?

boyceg: This is sort of the status quo now -- we do not automatically setup libMesh (or PETSc for that matter) in application codes.

Is this too complicated: `IBTK_MPI` does not setup PETSc or libMesh, but we provide some convenience classes that can manage standard setup tasks (e.g. `IBTK_PETSc_Initialize` for models that use PETSc but not libMesh, or `IBTK_libMesh_Initialize` for models that use libMesh)?

abarret: > Is this too complicated: `IBTK_MPI` does not setup PETSc or libMesh, but we provide some convenience classes that can manage standard setup tasks (e.g. `IBTK_PETSc_Initialize` for models that use PETSc but not libMesh, or `IBTK_libMesh_Initialize` for models that use libMesh)?

These would need to be matched with corresponding `finalize` and `shutdown` calls, as well as be called in the correct order. If this is the preferred method, I can switch things around. I don't have strong feelings on it.

boyceg: @drwells what do you think?

I guess that as long as the various libraries "stay out of the way" when they aren't being used, it doesn't really matter if they are activated or not.

drwells: This sounds too complicated. We should just have one class set up all our dependencies but make it clear in the documentation of that class what exactly it does (i.e., IBAMR does not depend on initialization functions itself, but its dependencies do, so these functions always need to be called).

boyceg: That's what I thought. Thanks! :-)

abarret:
I've addressed the comments. @drwells, this is ready for another look.

ibtk/include/ibtk/IBTKInit.h
```diff
@@ -0,0 +1,110 @@
...
+class IBTKInit
+{
+public:
+    /**
+     * Constructor for IBTKInit. Initializes libraries and sets the SAMRAI world communicator. Attempts to create a
+     * second IBTKInit object will result in a run time error.
+     */
+    IBTKInit(int argc,
+             char** argv,
+             IBTK_MPI::comm communicator = MPI_COMM_WORLD,
```
drwells: Lets just use `MPI_comm` here instead of the alias. That makes it a bit clearer that `IBTK_MPI::comm` isn't some C++ wrapper class.

abarret: Fixed.

PR #909, opened 2020-03-10
Add some notes on using MPI.

### LINKED ISSUE #884: Add notes on getting good MPI performance
We should summarize some of the discussion we had over email on how to run jobs efficiently on a cluster. The PETSc manual has a really good section on this - we should start by summarizing that and adding some more novice-friendly explanations of terminology.

Topics to include:
1. how to set up the streams benchmark and what it means
2. how to pin jobs to cores and other `mpiexec` tips and tricks 
3. how to disable openmp (this has occassionally bitten me when I compile hypre on my own) and other threading stuff

If anyone has notes on this they should post them here as a start :)

drwells:
Fixes #884. This problem is quite difficult but we should at least provide enough information for users to get started (and to figure out how to saturate memory bandwidth).

There is one missing section on good slurm parameters - I will fill that in tomorrow. We should also cite some references (like adding a link to the PETSc manual).

### DIFF
--- a/doc/mpi.md
+++ b/doc/mpi.md
@@ -1,0 +1,190 @@
+# Tips and Tricks for using MPI with IBAMR
+
+## Overview
+IBAMR is parallelized with MPI. MPI, the message-passing interface, is the most
+popular network communication standard for large-scale (dozens, hundreds, or
+more processors) parallelization of scientific programs at the current time.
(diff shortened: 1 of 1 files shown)


drwells:
I added some notes on SLURM.

These notes are not exhaustive but I hope others will add any additional tips to the document so that we can have everything in one place.

PR #910, opened 2020-03-11
Remove call to IBFEMethod::reinitializeFEData().

drwells:
This doesn't change anything after the first time we call it. This function accounts for about 15% of the total time spent setting things up after regridding so its worth removing for performance reasons.

We regrid about every other timestep when we start the TAVR code so this is actually noticeable in some regimes of some applications.

### DIFF
--- a/doc/news/changes/minor/20200311DavidWells
+++ b/doc/news/changes/minor/20200311DavidWells
@@ -1,0 +1,4 @@
+Fixed: IBFEMethod::reinitializeFEData() is no longer called after every time
+regrid since it does the same thing every time.
+<br>
+(David Wells, 2020/03/11)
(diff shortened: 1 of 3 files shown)


boyceg:
I think that my original intent was to facilitate libMesh AMR. If might be worthwhile to be able to control if this happens (but to default to "off").

boyceg:
Also, @jabrown893, do you know why the TAVR model is regridding so frequently? I thought time step sizes were v. small.

drwells:
That makes sense. Since we don't support AMR yet could we just unconditionally disable it for now? I'm not sure how we could set up the logic correctly (i.e., in a somewhat automatic way) when that code path doesn't yet exist.

boyceg:
Yes, let's add a flag but default it to `false`.

jabrown893:
@boyceg, my guess is because the beginning of the TAVR run (which is what David is profiling) involves a lot of forces and very fast movement because of the crimping. The forces have to be super strong to crimp in order to overcome the stiffness of the stent, but that results in very fast velocities???
But yes, the timestep is still very small.

drwells:
I added a boolean to IBFEMethod.h that is always `false` which indicates whether or not we are using AMR in the finite element mesh.

PR #912, opened 2020-03-12
Slightly improve test suite performance.

drwells:
A scheduler may assign multiple MPI jobs to the same processors on a cluster unless we explicitly disable binding. This is necessary for the test suite since we have multiple concurrent MPI programs running.


drwells:
Good question - looks like `-bind-to` is generic but `--oversubscribe` is not. I'll update the PR.

drwells:
Fixed!

PR #914, opened 2020-03-13
Print IBAMR version and git commit hash in config.log.

### LINKED ISSUE #881: Tracking git hash/IBAMR verison in config.log
I mentioned this to @drwells. It would be nice to output the git hash/IBAMR version in the config.log when we are doing comparisons with various builds as a way to check the build.

drwells:
This is sometimes useful when comparing builds. Fixes #881. 

It would be nice to also make this information available in something like
```cpp
print_ibamr_hash(std::ostream &out);
```
but this would require rerunning `configure` after every commit to make sure we always have the correct hash in the source code. Its not clear to me how to do this with autotools and I don't think its worthwhile at this time since running configure takes a long time.

### DIFF
--- a/configure
+++ b/configure
@@ -41148,6 +41148,29 @@ ac_config_files="$ac_config_files Makefile config/make.inc examples/Makefile exa
 
 subdirs="$subdirs ibtk"
 
+
+ibamr_version_string=$(cat "$srcdir/VERSION")
+{ $as_echo "$as_me:${as_lineno-$LINENO}: Configured version of IBAMR is $ibamr_version_string" >&5
+$as_echo "$as_me: Configured version of IBAMR is $ibamr_version_string" >&6;}
+# check if git is available:
+which git>/dev/null
+FOUND_GIT=$?
+if test "$FOUND_GIT" -eq 0 # 0 indicates 'which git' succeeded
+then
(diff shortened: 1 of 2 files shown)


boyceg:
https://stackoverflow.com/a/12368262 ?

drwells:
Another approach is to add a make rule that is dependent on `.git/HEAD` but this approach (rather the approach in the linked issue to get updated information) looks more robust. Dealing with these edge cases is going to make this PR a lot harder but I'll give it a shot.

drwells:
I thought about this some more and I don't think adding `print_ibamr_hash` is worth doing:
1. it is has limited value: it can only capture information stored in git. If any files have been locally modified but don't correspond to a commit then the hash is no longer meaningful (two hashes can correspond two different library behavior).
2. we need to make sure that both git and the `.git/` directory are available to extract the hash - from that we need to automatically regenerate version information by detecting when git hashes change. This logic is pretty complicated and has to be implemented in a makefile. This is possible but its going to take a lot of effort to get right, and due to 1. this effort just isn't worth it.

Hence I think we should proceed with the current state of the PR: it solves the original problem (print version info and git hash in `config.log`). If anyone can find some code that solves 2. then we can go ahead with that too.

PR #916, opened 2020-03-16
Improve fe coupling

drwells:
I took a look at our mass matrix solves and I noticed (via `-ksp_log`) that our matrices allocated three times as many entries as they actually used in 3D. This is because, by default, libMesh assumes all variables couple to each-other.

We can get around this by providing an extra argument to libMesh. This saves us some memory (about 10% for the heart model). Fortunately PETSc is already smart enough to ignore the zero entries so it does not really change the performance.

### DIFF
--- a/doc/news/changes/minor/20200316DavidWells
+++ b/doc/news/changes/minor/20200316DavidWells
@@ -1,0 +1,5 @@
+Improved: Systems created inside IBFEMethod no longer couple all variables to
+each-other since these systems are only used to solve projection problems. This
+lowers memory usage in IBFE applications by about 10-15%.
+<br>
+(David Wells, 2020/03/16)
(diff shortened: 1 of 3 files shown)


drwells:
Yes. I could only find correct usage for it from a mailing list. The documentation from `dof_map.h` is telling:
```cpp

  /**
   * Degree of freedom coupling.  If left empty each DOF
   * couples to all others.  Can be used to reduce memory
   * requirements for sparse matrices.  DOF 0 might only
   * couple to itself, in which case \p dof_coupling(0,0)
   * should be 1 and \p dof_coupling(0,j) = 0 for j not equal
   * to 0.
   *
   * This variable is named as though it were class private,
   * but it is in the public interface.  Also there are no
   * public methods for accessing it...  This typically means
   * you should only use it if you know what you are doing.
   */
  CouplingMatrix * _dof_coupling;
```

PR #917, opened 2020-03-16
Run IWYU and fix the resulting problems.

### LINKED ISSUE #807: Post 0.6: run IWYU
We should run this early in the release cycle so that we can shake out any overly zealous header deletions.

drwells:
Fixes #807.

IWYU made a lot of mistakes, particularly with Eigen and Boost headers. It is probably worthwhile to run this tool every couple of years but it took hours to get everything working again - its just not worthwhile to run this for every release.


boyceg:
Uhhh, you want me to review this? :-D

drwells:
I don't know. I read through the changes and deleted all the `private/` headers and other weird stuff IWYU did - it should be fine.

src/wave_generation/IrregularWaveGenerator.cpp
```diff
@@ -15,11 +15,14 @@
 
 #include "ibamr/IrregularWaveGenerator.h"
 #include "ibamr/RNG.h"
-#include "ibamr/app_namespaces.h"
```
drwells: A major problem with IWYU pragmas is that its really hard to remember to put them everywhere we need them.

PR #919, opened 2020-03-17
Fixing regrid projection for multiphase_flow/ex3

nishantnangia:
This PR fixes multiphase_flow/ex3 by adding level set initial conditions to the input file. Previously this case was carrying out an incorrect velocity projection, which caused the initial velocity field inside the bubble to be ~0.5, when it should have been ~1.

I also added a test for this example.


drwells:
@amneetb is there anything else we should do here?

amneetb:
This looks good to me. 

Another way to achieve the same objective is the way I do it in `IBLevelSet` where the level set cannot be prescribed analytically. See the following lines of code:

https://github.com/IBAMR/IBAMR/blob/master/examples/IBLevelSet/ex0/example.cpp#L895-L902

PR #920, opened 2020-03-17
Fix some IWYU issues.

drwells:
1. We don't need to include unistd.h in these headers - we aren't doing any
system calls.
2. We don't need stddef.h.
3. Fix a compilation problem where IWYU added an internal GCC header.
4. Remove IWYU from the list of release tasks - its just not worth the pain.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -76,7 +76,6 @@
 #include "tbox/RestartManager.h"
 #include "tbox/SAMRAI_MPI.h"
 #include "tbox/Utilities.h"
-#include <ext/new_allocator.h>
 
 #include "libmesh/boundary_info.h"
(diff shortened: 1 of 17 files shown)


src/IB/IBFEMethod.cpp
```diff
@@ -76,7 +76,6 @@
 #include "tbox/RestartManager.h"
 #include "tbox/SAMRAI_MPI.h"
 #include "tbox/Utilities.h"
-#include <ext/new_allocator.h>
```
drwells: This is an internal header that IWYU added.

drwells:
This fixes compilation on macs so merging it is somewhat urgent.

Issue #778, opened 2019-12-05
Enable floating point exceptions in tests


drwells:
It would be useful to track these via `feenableexcept`: this helped debug #725.

drwells:
I tried this this morning and found at least one issue with our interface to hypre.

I found another interesting problem in the stokes solve: we get floating point exceptions when we try to do`1e-15*inf + -inf` (an axpy operation) deep inside GMRES. @boyceg do you have any idea why we would get vectors with `-inf` or `inf` inside the solver?

boyceg:
it is hard to see how that would happen unless the RHS had junk in it, so that the residual was crazy. @knepley do you have thoughts about

knepley:
The easiest way to get inf is dividing by zero. We have PetscIsInfOrNanReal() which can be used anywhere in the code to track down the first instance. We call it from the convergence check in GMRES, so you must be getting the exception before it makes it there. I just usually keep putting in this check until I find the source.

drwells:
I ran the test suite with `feenableexcept(FE_DIVBYZERO|FE_INVALID)` - in this case it looks like we get infs in some other way because the programs abort when we do `inf + -inf` and not before.

knepley:
Did you valgrind this? Another easy way to get inf is from uninitialized data.

drwells:
I did not - that's a good idea.

drwells:
The standalone navier-stokes tests pass so I suspect its not an initialization problem but there is only one way to find out!

drwells:
valgrind doesn't complain so it is not an initialization problem.

drwells:
However - we immediately get an overflow error at time zero in both debug and optimized modes since some values appear to be initialized to FLOAT_MAX - we multiply those values by 52, yielding an overflow error.

amneetb:
Could this explain parallel CIB bug?

drwells:
The only way to know would be to try running some CIB tests with and without floating point exceptions to see if it triggers the problem. Since the Stokes solve produces correct output when we disable floating point exceptions I doubt that it is related.

PR #921, opened 2020-03-18
Zero some stored hypre residuals.

drwells:
Additionally, exit successfully if there were no iterations and the residual norm is nan - this can happen when the right-hand side is the zero vector (which occurs durring initialization, like in regridProjection).

Part of #778 - we need this change to get most of the tests passing with floating point exceptions enabled.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/minor/20200318DavidWells
+++ b/doc/news/changes/minor/20200318DavidWells
@@ -1,0 +1,6 @@
+Improved: IBTK::CCPoissonHypreLevelSolver::solveSystem() and
+IBTK::SCPoissonHypreLevelSolver::solveSystem() now both return <code>true</code> in
+the case where the right-hand side vector is zero (which causes the relative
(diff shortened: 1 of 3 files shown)


doc/news/changes/minor/20200318DavidWells
```diff
@@ -0,0 +1,6 @@
+Improved: IBTK::CCPoissonHypreLevelSolver::solveSystem() and
+IBTK::SCPoissonSolverManager::solveSystem() now both return <code>true</code> in
```
boyceg:
```suggestion
IBTK::SCPoissonSolver::solveSystem() now both return <code>true</code> in
```

drwells: Both should be `PoissonHypreLevelSolver`.

PR #924, opened 2020-03-20
Can register a callback function with HierarchyIntegrator to allow specialization for HierarchyIntegrator::regridHierarchy()

abarret:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/HierarchyIntegrator.h
+++ b/ibtk/include/ibtk/HierarchyIntegrator.h
@@ -494,6 +494,24 @@ public:
     void registerApplyGradientDetectorCallback(ApplyGradientDetectorCallbackFcnPtr callback, void* ctx = nullptr);
 
     /*!
+     * Callback function specification to enable further specialization of regridHierarchy().
+     */
+    using RegridHierarchyCallbackFcnPtr =
+        void (*)(SAMRAI::tbox::Pointer<SAMRAI::hier::BasePatchHierarchy<NDIM> > hierarchy,
+                 double data_time,
+                 bool initial_time,
+                 void* ctx);
+
+    /*!
+     * Register a callback function to enable further specialization of regridHierarchy().
+     *
+     * \note This function will be called after all data owned by the hierarchy is initialized and set, but before the
+     * data is synchronized. If data owned by the integrator is used in this function, it should be synchronized prior
+     * to use by an appropriate coarsening operation.
+     */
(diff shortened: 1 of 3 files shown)


drwells:
Should these functions be called after synchronization?

abarret:
> Should these functions be called after synchronization?

This is designed for data that isn't controlled by the `HierarchyIntegrator`. I put it here so that if you modify any data owned by the integrator, they won't have to synchronize it. But at the same time this means if you want to _use_ the data, you need to synchronize it first. I think the best thing is to make it clear when this function is being called.

ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -493,6 +493,24 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
...
+                 double data_time,
+                 bool initial_time,
+                 void* ctx);
+
+    /*!
+     * Register a callback function to enable further specialization of regridHierarchy().
+     *
+     * \note This function will be called after all data owned by the hierarchy is initialized and set, but before the
+     * data is synchronized. If data owned by the integrator is used in this function, it should be synchronized prior
+     * to use.
```
drwells: From the other conversation: if this is called prior to synchronization but a user needs synchronized data, then how should they do it?

abarret: Using an appropriate coarsening operation. I've updated the comment.

drwells:
@abarret Do you think its necessary to add a new test? I'm not sure if its worth waiting until we use this or not. Additionally, can you run the test suite locally?

abarret:
> Do you think its necessary to add a new test? I'm not sure if its worth waiting until we use this or not.

I've been using this patch for the past several weeks. Now that CIB is fixed, we need this patch for related work. If you approve of the updated documentation, I propose we merge now, and I'll commit to writing tests for all the `HierarchyIntegrator` callback functions soon.

> Additionally, can you run the test suite locally?

Yes. I'm running it now. Again, I've been using this patch for several weeks now so it _should_ be fine.

drwells:
Sounds good but lets run the test suite just to be sure. Would you please merge this correctly once the test suite is done and there are no unexpected failures?

abarret:
The test suite passes so I will merge.

PR #925, opened 2020-03-20
add missing include in FEDataManager.cpp

boyceg:
- [x] Was clang-format run?
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -18,6 +18,8 @@
 #include "ibtk/FECache.h"
 #include "ibtk/FEDataManager.h"
 #include "ibtk/IBTK_CHKERRQ.h"
+#include "ibtk/IndexUtilities.h"
+#include "ibtk/JacobianCalculator.h"
 #include "ibtk/JacobianCalculatorCache.h"
 #include "ibtk/LEInteractor.h"
 #include "ibtk/QuadratureCache.h"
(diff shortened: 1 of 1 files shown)


boyceg:
@drwells looks like another IWYU casualty.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -18,6 +18,8 @@
 #include "ibtk/FECache.h"
 #include "ibtk/FEDataManager.h"
 #include "ibtk/IBTK_CHKERRQ.h"
+#include "ibtk/IndexUtilities.h"
+#include "ibtk/JacobianCalculator.h"
```
drwells: I need to change these over to `<` and `>` at some point...

drwells:
Its a header - lets merge without me running the test suite.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: I always thought that the "rule" was to use angle brackets for system headers, quotes for non-system headers.

drwells: The only difference is that using `"`s also includes the current directory in the preprocessor path. Ultimately its not that big of a deal but I prefer to use the slightly stricter pathing rule.

drwells:
@boyceg its not a big deal but could we keep all the bullet points, even the irrelevant ones? It is easier to tell at a glance what has been done if we use the same list.

boyceg:
Sure thing. I was thinking that we should only include the relevant ones, to make sure to capture whether all relevant items have been done.

PR #931, opened 2020-03-22
update location of clang++ in setup_compile_commands.sh

boyceg:


drwells:
I think this only effects people who use this script directly (like @boyceg) so lets merge.

PR #932, opened 2020-03-22
add full IIM support to IBFESurfaceMethod (part 1)

boyceg:
- [x] Was clang-format run?
- [X] Were relevant issues cited? Please remember to add `Fixes #12345` to close
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/include/ibamr/IBFESurfaceMethod.h
+++ b/include/ibamr/IBFESurfaceMethod.h
@@ -88,17 +88,33 @@ namespace IBAMR
  * \brief Class IBFESurfaceMethod is an implementation of the abstract base
  * class IBStrategy that provides functionality required by the IB method with
  * a finite element representation of a surface mesh.
+ *
+ * Coupling schemes include both IB formulations (integral operations with
+ * regularized delta function kernels) and an immersed interface method (IIM)
+ * scheme (E. M. Kolahdouz, A. P. S. Bhalla, B. A. Craven, and B. E. Griffith.
+ * An immersed interface method for discrete surfaces. J Comput Phys,
+ * 400:108854 (37 pages), 2020).
+ *
+ * \note When using the IIM implementation, it is recommended that users set
+ * all linear solvers to use tight relative tolerances (1e-10).
  */
 class IBFESurfaceMethod : public IBStrategy
 {
 public:
-    static const std::string COORDS_SYSTEM_NAME;
     static const std::string COORD_MAPPING_SYSTEM_NAME;
+    static const std::string COORDS_SYSTEM_NAME;
     static const std::string FORCE_SYSTEM_NAME;
     static const std::string NORMAL_VELOCITY_SYSTEM_NAME;
+    static const std::string PRESSURE_IN_SYSTEM_NAME;
     static const std::string PRESSURE_JUMP_SYSTEM_NAME;
+    static const std::string PRESSURE_OUT_SYSTEM_NAME;
     static const std::string TANGENTIAL_VELOCITY_SYSTEM_NAME;
+    static const std::string TAU_IN_SYSTEM_NAME;
+    static const std::string TAU_OUT_SYSTEM_NAME;
     static const std::string VELOCITY_SYSTEM_NAME;
+    static const std::string WSS_IN_SYSTEM_NAME;
+    static const std::string WSS_OUT_SYSTEM_NAME;
+    static const std::array<std::string, NDIM> VELOCITY_JUMP_SYSTEM_NAME;
 
     /*!
      * \brief Constructor.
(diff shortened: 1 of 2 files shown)


boyceg:
This is just support for pressure + velocity gradient jumps, along with some changes in the way that libMesh vectors are tracked and setting up some new systems for tracking wall shear stress and traction vectors along the interface.

boyceg:
Once this is finalized, I need to make sure that @AminKolah is listed as authoring the code before merging.

include/ibamr/IBFESurfaceMethod.h
```diff
@@ -521,26 +566,139 @@ class IBFESurfaceMethod : public IBStrategy
...
+        }
+
+        std::vector<libMesh::System*> systems;
+        std::vector<libMesh::PetscVector<double>*> current_vecs, half_vecs, new_vecs, IB_ghost_vecs;
+
+    private:
+        const std::string d_system_name;
+        const bool d_has_current_vecs, d_has_half_vecs, d_has_new_vecs, d_has_IB_ghost_vecs;
+        bool d_initialized = false;
+    };
```
AminKolah: This is really great, and makes it much more impact!

src/IB/IBFESurfaceMethod.cpp
```diff
@@ -2131,7 +2763,11 @@ IBFESurfaceMethod::getFromRestart()
         TBOX_ERROR(d_object_name << ":  Restart file version different than class version." << std::endl);
     }
     db->getIntegerArray("d_ghosts", d_ghosts, NDIM);
-    d_use_jump_conditions = db->getBool("d_use_jump_conditions");
+    d_use_pressure_jump_conditions = db->getBool("d_use_pressure_jump_conditions");
+    d_use_velocity_jump_conditions = db->getBool("d_use_velocity_jump_conditions");
+    d_compute_fluid_traction = db->getBool("d_compute_fluid_traction");
+    d_traction_interior_side = db->getBool("d_d_traction_interior_side");
```
AminKolah: I think that we can remove `d_traction_interior_side`? It's not being used anywhere

boyceg:
@AminKolah, what is the intention of `traction_activation_time`?

AminKolah:
> @AminKolah, what is the intention of `traction_activation_time`?

For RBD-FSI this is the time when we want the fluid forces to drive the motion. It's set to 0 by default but can be larger if we want to activate it at later time.

boyceg:
> > @AminKolah, what is the intention of `traction_activation_time`?
> 
> For RBD-FSI this is the time when we want the fluid forces to drive the motion. It's set to 0 by default but can be larger if we want to activate it at later time.

It seems better to do this in the RBD code and not in the IIM code. How about just have the dynamics code ignore the tractions as appropriate?

AminKolah:
> > > @AminKolah, what is the intention of `traction_activation_time`?
> > 
> > 
> > For RBD-FSI this is the time when we want the fluid forces to drive the motion. It's set to 0 by default but can be larger if we want to activate it at later time.
> 
> It seems better to do this in the RBD code and not in the IIM code. How about just have the dynamics code ignore the tractions as appropriate?

OK I see what you mean. I guess when I added to the source code I was thinking that could also reduce run time? Although I imagine it might not be that significant to worry about.

boyceg:
> > > > @AminKolah, what is the intention of `traction_activation_time`?
> > > 
> > > 
> > > For RBD-FSI this is the time when we want the fluid forces to drive the motion. It's set to 0 by default but can be larger if we want to activate it at later time.
> > 
> > 
> > It seems better to do this in the RBD code and not in the IIM code. How about just have the dynamics code ignore the tractions as appropriate?
> 
> OK I see what you mean. I guess when I added to the source code I was thinking that could also reduce run time? Although I imagine it might not be that significant to worry about.

I'm going to drop it.

boyceg:
> This PR is still huge. I'll take an initial stab at it by pulling out `LibMeshSystemData` and putting it in IBTK.

Feel free to change the way that class is set up / works. I think that the libMesh `System` class may provide ways to manage these kinds of extra vectors that is easier than rolling our own --- perhaps that is preferable.

drwells:
Are we going to ever want to use this class with cell-centered data?

AminKolah:
> Are we going to ever want to use this class with cell-centered data?

I think not, at least no in the current form? That would require implementation of new functions for imposing velocity jumps

boyceg:
I would say that we are not going to use IIM + cell-centered velocities for the foreseeable future. A sort of research question is whether we can generate the "correction" terms automatically.

drwells:
Do we really need all these libMesh systems? Could we get away with, e.g., one system for TAU and then having `"half_in"` and `"half_out"` vectors owned by the same system?

In general, we only need multiple systems in `IBFEMethod` because the systems have different constraints - AFAICT no system added here has new constraints (and they all use the same finite element space) so we should implement it as more vectors in the same system.

boyceg:
One problem is visualization. I suppose that these stress variables could be in a “stress reconstruction” system.

drwells:
I'm not too familiar with exodus but is it a big deal if the vectors just have a different name? What is the primary change in doing things in one system?

boyceg:
libMesh outputs the data in the “solution” vector, splitting out whatever variables are registered with the system.

So we cannot easily output alternative vectors. We could store these fields as separate variables in one system, however.

drwells:
It's unfortunate that you're right and there is no easy workaround for outputting arbitrary vectors in a system :(

boyceg:
However, we could switch to using a single `System` to hold all of the IIM-specific variables: pressure and viscous stress jump conditions, inner and outer pressures, wall shear stresses, and total fluid traction vectors.

boyceg:
One thing to note is that I've added in some flexibility in the choice of basis functions for different IIM-related fields. I think we may ultimately want to use discontinuous representations of the jump conditions for geometries with sharp corners.

include/ibamr/IBFESurfaceMethod.h
AminKolah: The issue had to do with the class `LibMeshSystemData`. I have everything working by keeping all new changes except removing this class. Please let me know if you want me to push them.... If we still want to make the system of equations more compact perhaps  you and @drwells can come up with a proper way to handle them differently. Currently everything including those shared with `fe_data_manager` like `COORDINATE_SYSTEM_NAME, `VELOCITY_SYSTEM_NAME`, and `FORCE_SYSTEM_NAME` are handled through `LibMeshSystemData`, while I think that might be a good place to only keep the jump systems and not everything?

PR #936, opened 2020-03-23
Fix SAMRAIGhostDataAccumulator + SideData.

drwells:
Fixes one of the two remaining issues with #896.

The box returned by `getGhostBox` is for the patch: it is not the offset box required to correctly use side data. Using the wrong box caused the function to skip the last column which results in wrong accumulation values.

In the modified test output, note that 6 really should be the maximum value since that entry shows up four times on the patch interior (twice on each patch due to periodicity) and twice more on one patch in ghost entries.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
Are we using the SAMRAI geometry classes to generate appropriate boxes? E.g. [`SideGeometry<DIM>::toSideBox()`](https://ibamr.github.io/IBAMR-docs/samrai/2.4.4/html/classSAMRAI_1_1pdat_1_1SideGeometry.html#b84ba95e9888e7ffe365795f43ca2908)?

drwells:
We should have, but we did not. This patch uses a much dumber solution - we just read everything in the arrays regardless of what the box actually is.

drwells:
I'm not sure I understand your question. Are you asking about whether we should be using boxes somewhere else?

boyceg:
> I'm not sure I understand your question. Are you asking about whether we should be using boxes somewhere else?

Sorry for not being clear --- probably the easiest thing to do is to rely on the boxes provided by the `ArrayData`. It looks like that is what you are doing?

However, if you want to know the range of indices that corresponds to the "interior" for a given data centering, one way to do that is to use the corresponding `Geometry` class.

drwells:
The point of this patch is that we don't need boxes at all - we need all the data on the patch (interior and ghost are treated the same way for the purpose of accumulation) so we just read the underlying array. The actual box is irrelevant.

ibtk/src/math/SAMRAIGhostDataAccumulator.cpp
```diff
@@ -124,6 +111,8 @@ add_or_get(Pointer<PatchLevel<NDIM> >& level,
         IBTK_CHKERRQ(ierr);
         for (unsigned int d = 0; d < dofs_ptrs.size(); ++d)
         {
+            const PetscInt size = dofs_ptrs[d]->getDepth() * dofs_ptrs[d]->getOffset();
```
drwells: This is the part that replaces the box usage - `getDepth() * getOffset()` is exactly the size of the array.

PR #938, opened 2020-03-24
Add LibMeshSystemVectors

drwells:
Part of #932, but applied to IBFEMethod first. @boyceg what do you think?

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/LibMeshSystemVectors.h
+++ b/ibtk/include/ibtk/LibMeshSystemVectors.h
@@ -1,0 +1,186 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2020 - 2020 by the IBAMR developers
+// All rights reserved.
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+#ifndef included_IBTK_LibMeshSystemVectors
+#define included_IBTK_LibMeshSystemVectors
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include <ibtk/FEDataManager.h>
+
+#include <libmesh/equation_systems.h>
+#include <libmesh/petsc_vector.h>
+#include <libmesh/system.h>
+
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBTK
+{
+/*!
+ * \brief Class LibMeshSystemVectors is a convenience class that manages
+ * access to libMesh vectors for the same system defined on multiple
+ * parts. This class supports access to vectors ghosted with the Lagrangian
+ * partitioning (i.e., libMesh's computed partitioning) as well as the IB
+ * partitioning (i.e., the partitioning based on the distribution of SAMRAI
+ * data).
+ *
+ * A libMesh::System stores vectors in two different ways: the
+ * <tt>solution</tt> and <tt>current_local_solution</tt> vectors are stored
+ * explicitly by the System object (see libMesh::System::solution and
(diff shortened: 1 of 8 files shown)


boyceg:
Looks good to me. What do you think about possibly making the accessor function names a bit more succinct?

drwells:
I like `IBGhostedVectors` since that's the same name we use in `FEDataManager` - it makes it clear that they do the same thing.

I also find `getVector` a bit wordy. I thought about also adding `getVec` to return a PETSc vector immediately but it looks a little confusing to me since it could also just be shorthand for `getVector` (and we only save three characters). Ultimately I'd rather it be a little longer and clear because I can't think of good shorthand that's not overly obscure (we could also abuse `operator()` or `operator[]` but thats going to make things harder to read).

boyceg:
Sounds good.

Are you able to go ahead and implement this change more completely in `IBFEMethod`? I actually have some code that I have been dragging my heels to write because of the painful way that these vectors are currently set up. :-D

drwells:
> Are you able to go ahead and implement this change more completely in IBFEMethod?

Yes. I wanted to post the in-progress version in case someone sees something else we could do to make things easier. I'll convert the rest of the systems (the nice thing about this class is we can do it one system at a time).

ibtk/include/ibtk/LibMeshSystemVectors.h
```diff
@@ -0,0 +1,185 @@
...
+     * Get a specific vector for a specific part.
+     */
+    libMesh::PetscVector<double>& getVector(const std::string& vec_name, const unsigned int part)
+    {
+        TBOX_ASSERT(part < d_fe_data_managers.size());
+        if (!vec_stored_in_map(vec_name))
+        {
+            if (vec_name == "solution") return dynamic_cast<libMesh::PetscVector<double>&>(*d_systems[part]->solution);
+            if (vec_name == "current local solution")
+                return dynamic_cast<libMesh::PetscVector<double>&>(*d_systems[part]->current_local_solution);
```
drwells: Note that we hard-code some special vector names here since libMesh doesn't put these in its map of vectors - `system->solution` is not the same as `system->request_vector("solution")` but for convenience we are slightly inconsistent with libMesh here and identify every vector with a string.

drwells:
On second thought - `vecs->getVectors` is a little redundant. I suppose we could just write  `vecs->get` and `vecs->getIBGhosted` instead.

boyceg:
> On second thought - `vecs->getVectors` is a little redundant. I suppose we could just write `vecs->get` and `vecs->getIBGhosted` instead.

That is exactly what I was thinking when I asked. :-) But this relies on this particular variable naming convention, which does not seem ideal.

drwells:
well, the point of this class is to be used internally, so that's not so bad.

boyceg:
OK, in that case, I would just use `get`.

Also, can we use `"current_local_solution"` (or make it a synonym)?

drwells:
Good idea. I also made `"current"` an alias for `"current_local_solution"`.

drwells:
OK, that should do it.

We shouldn't convert the rest of the systems right now because we have zero tests (and also zero examples) for fluid sources and stress normalization - changing these might break something I don't understand so we'll skip it.

Notably, I did not add any tests for this class because it is really just an implementation detail for how IBFEMethod manages libMesh vectors - since all the IBFEMethod tests still pass this is fine.

ibtk/src/utilities/LibMeshSystemVectors.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+/////////////////////////////// STATIC ///////////////////////////////////////
+
+namespace
+{
+bool
```
boyceg: Should this be `inline`?
```suggestion
inline bool
```

drwells: If its in the same translation unit the compiler will consider it for inlining, so I don't think this matters.

PR #944, opened 2020-03-26
update generated files using automake 1.16.2

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
Homebrew updated my automake to 1.16.2 ...

drwells:
This was just realized on Sunday - the Homebrew team really has their act together!

I will update to this soon on my personal machines and its not too hard to get it on other devices I frequently use but this is going to make contributing harder for most people with older tool sets. Can we also update the 'download autotools' script to install this new version to ease the pain?

boyceg:
I could try to figure out how to downgrade.

The bootstrap script actually only checks that the version and subversion number match --- so it is only checking to see if you are using autoconf 1.16. Maybe it needs to be updated to pin down the full version number?

drwells:
We will need to given that there are nontrivial output differences.

boyceg:
Right now the `bootstrap` script checks to see if minimum version requirements are met. How much work would it be to make it check that a precise version is being used?

drwells:
It's not hard. In fact, the current version check only looks at the minor version, not the major or subminor versions. I'm writing a patch.

boyceg:
OK, great. We should just standardize specific tools.

abarret:
The bundled automake needs to be updated as well. Or we need to update the script to download the correct configure tools

boyceg:
Let's stick with `automake` 1.16.1 for now and handle updates in concert with specifying versions of tools in `bootstrap`, per #945.

PR #948, opened 2020-03-30
Fix spreadTransmissionForceDensity() + AMR.

drwells:
I'm skipping fluid sources for the last step of #896 but I want to be sure that we handle jump conditions correctly since they are used in a lot of examples. It turns out that this didn't work correctly with AMR - fortunately the fix is pretty simple. I added an interesting test (a thin elastic ring) to make sure we are accumulating forces correctly.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
It's a good thing I added this test. Even though we have an ex0 test with jump terms that works with the new force spreading this does not.

PR #951, opened 2020-03-31
lazy initialize IB ghosted DOF indexing

### LINKED ISSUE #947: Call FEDataManager::reinitElementMappings() by request for each system
I ran callgrind on TAVR again over the weekend and took a look at the results. I don't have time to work on this now but I figured we should summarize the problem so that we don't forget about it.

Some applications add a lot of their own libMesh systems to the EquationSystems object managed by FEDataManager - most of these never need to access the IB-ghosted partitioning, but we still unconditionally compute this partitioning for them in `reinitElementMappings()`. This problem is made worse by use of the scratch hierarchy where we do everything for both the primary and scratch partitioning. Since the TAVR code regrids a lot this ends up occupying 15% of our total run time.

We should modify the logic to only compute `d_active_patch_ghost_dofs` and `d_system_ib_ghost_vec` on demand instead - this should largely resolve the problem since we rarely access this information for more than two systems (instead of assembling it, for the TAVR case, for about 10 systems).

boyceg:
Fixes #947.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -592,7 +592,7 @@ public:
      * \return A pointer to a vector, with ghost entries corresponding to
      * relevant IB data, associated with the specified system.
      */
-    std::unique_ptr<libMesh::PetscVector<double> > buildIBGhostedVector(const std::string& system_name) const;
+    std::unique_ptr<libMesh::PetscVector<double> > buildIBGhostedVector(const std::string& system_name);
 
     /*!
      * \return A pointer to the unghosted coordinates (nodal position) vector.
@@ -1036,6 +1036,11 @@ private:
                                 const std::string& system_name);
 
     /*!
+     * Reinitialize IB ghosted DOF data structures for the specified system.
+     */
+    void reinitializeIBGhostedDOFs(const std::string& system_name);
+
+    /*!
      * Read object state from the restart file and initialize class data
      * members.  The database from which the restart data is read is determined
      * by the object_name specified in the constructor.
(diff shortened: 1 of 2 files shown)


drwells:
> 198/314 (63.06%) tests passed

At a glance the implementation looks fine but it doesn't work. Can you try running some IBFE examples locally to figure this out?

boyceg:
fixing now

boyceg:
I forgot to initialize the data structures in `buildIBGhostedVector`.

boyceg:
I had to remove the `const` qualifier from `buildIBGhostedVector`. I'm running the test suite now on my Mac.

boyceg:
BTW, the testsuite takes forever to run on my Mac. Any tricks for making it run faster? :-)

drwells:
For something like this I recommend only running IBFE tests: try `./attest -j -R IBFE` (here `IBFE` is a regular expression).

How slow is slow? I can complete `./attest -j6` on my desktop (which has four physical processors and circa 2013 hardware) in about ten minutes.

boyceg:
Debug or opt mode? Much slower for me...

drwells:
optimized mode.

boyceg:
Ah, I see -- I've been running a debugging build.

drwells:
I'm not exactly sure what is going on but libMesh does a ton of extra checking in debug mode for parallel computations that slows it down by more than an order of magnitude. I almost always run the test suite in optimized mode to avoid this problem.

jabrown893:
Would it be helpful for me to test this on some TAVR code, or is there anything else I can ran to help test?

drwells:
This looks reasonable and passes the test suite so lets merge.

@jabrown893 if you already have some timing data for the TAVR code it would be useful to compare these changes against that. I'm also going to rerun things in callgrind to get some more precise readings.

PR #952, opened 2020-03-31
fix nodal interaction

### LINKED ISSUE #722: FEDataManager spread error when called with use_nodal_quadrature
When I modify the example `IBFE/explicit/ex0` to set `spread_use_nodal_quadrature` to `TRUE`, I get a PETSc error about a null pointer. So far, I've only tracked it down to line 885 in FEDataManager.cpp on master:
`F_dX_vec.pointwise_mult(F_vec, dX_vec);`
It appears to be an issue with the `dX_vec` object returned from `buildDiagonalL2MassMatrix()`. This is also a problem with the flag `interp_use_nodal_quadrature`.

I don't know enough libMesh to effectively investigate. Am I using `use_nodal_quadrature` wrong, or is this a bug?

boyceg:
Fixes #722.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -101,12 +101,7 @@
 #include "libmesh/type_vector.h"
 #include "libmesh/variant_filter_iterator.h"
 
-#include "petscksp.h"
-#include "petscsys.h"
 #include "petscvec.h"
-#include <petsclog.h>
-
-#include <mpi.h>
 
 IBTK_DISABLE_EXTRA_WARNINGS
 #include "boost/multi_array.hpp"
@@ -304,7 +299,6 @@ FEData::clearCached()
     d_quadrature_cache.clear();
     d_L2_proj_solver.clear();
     d_L2_proj_matrix.clear();
-    d_L2_proj_matrix_diag.clear();
 }
 
 const boundary_id_type FEDataManager::ZERO_DISPLACEMENT_X_BDRY_ID = 0x100;
@@ -524,6 +518,8 @@ FEDataManager::reinitElementMappings()
     // We reinitialize mappings after repartitioning, so clear the cache since
     // its content is no longer relevant:
     d_fe_data->clearCached();
+    d_L2_proj_matrix_diag.clear();
+    d_L2_proj_matrix_diag_ghost.clear();
 
     // Delete cached hierarchy-dependent data.
     d_active_patch_elem_map.clear();
@@ -823,40 +819,29 @@ FEDataManager::spread(const int f_data_idx,
     JacobianCalculatorCache jacobian_calculator_cache(mesh.spatial_dimension());
 
     // Check to see if we are using nodal quadrature.
-    const bool use_nodal_quadrature =
-        spread_spec.use_nodal_quadrature && (F_fe_type == X_fe_type && F_order == X_order);
+    const bool use_nodal_quadrature = spread_spec.use_nodal_quadrature;
+    if (use_nodal_quadrature) TBOX_ASSERT(F_fe_type == X_fe_type && F_order == X_order);
 
-    // We only use the FECache objects if we do *not* use nodal quadrature so
-    // only perform sanity checks in that case:
-    if (!use_nodal_quadrature)
-    {
-        // This will break, at some point in the future, if we ever use
-        // nonnodal-interpolating finite elements. TODO: more finite elements
-        // will probably work.
-        std::vector<FEFamily> fe_family_whitelist{ LAGRANGE, L2_LAGRANGE };
-        TBOX_ASSERT(std::find(fe_family_whitelist.begin(), fe_family_whitelist.end(), F_fe_type.family) !=
-                    fe_family_whitelist.end());
-        TBOX_ASSERT(std::find(fe_family_whitelist.begin(), fe_family_whitelist.end(), X_fe_type.family) !=
-                    fe_family_whitelist.end());
-    }
+    // This will break, at some point in the future, if we ever use
+    // nonnodal-interpolating finite elements. TODO: more finite elements
+    // will probably work.
+    std::vector<FEFamily> fe_family_whitelist{ LAGRANGE, L2_LAGRANGE };
+    TBOX_ASSERT(std::find(fe_family_whitelist.begin(), fe_family_whitelist.end(), F_fe_type.family) !=
+                fe_family_whitelist.end());
+    TBOX_ASSERT(std::find(fe_family_whitelist.begin(), fe_family_whitelist.end(), X_fe_type.family) !=
+                fe_family_whitelist.end());
 
     if (use_nodal_quadrature)
     {
-        // Store a copy of the F vector.
-        std::unique_ptr<NumericVector<double> > F_vec_bak = F_vec.clone();
-        *F_vec_bak = F_vec;
-
         // Multiply by the nodal volume fractions (to convert densities into
         // values).
-        NumericVector<double>& F_dX_vec = F_vec;
-        const NumericVector<double>& dX_vec = *buildDiagonalL2MassMatrix(system_name);
-        F_dX_vec.pointwise_mult(F_vec, dX_vec);
-        F_dX_vec.close();
+        PetscVector<double>* dX_vec = buildIBGhostedDiagonalL2MassMatrix(system_name);
+        std::unique_ptr<NumericVector<double> > F_x_dX_vec = F_vec.clone();
+        F_x_dX_vec->pointwise_mult(F_vec, *dX_vec);
 
         // Extract local form vectors.
-        auto F_dX_petsc_vec = static_cast<PetscVector<double>*>(&F_dX_vec);
-        const double* const F_dX_local_soln = F_dX_petsc_vec->get_array_read();
-
+        auto F_x_dX_petsc_vec = static_cast<PetscVector<double>*>(F_x_dX_vec.get());
+        const double* const F_x_dX_local_soln = F_x_dX_petsc_vec->get_array_read();
         auto X_petsc_vec = static_cast<PetscVector<double>*>(&X_vec);
         const double* const X_local_soln = X_petsc_vec->get_array_read();
 
@@ -869,33 +854,48 @@ FEDataManager::spread(const int f_data_idx,
             const size_t num_active_patch_nodes = patch_nodes.size();
             if (!num_active_patch_nodes) continue;
 
-            // Store the values of F_JxW and X at the nodes.
-            std::vector<double> F_dX_node, X_node;
-            F_dX_node.reserve(n_vars * num_active_patch_nodes);
+            const Pointer<Patch<NDIM> > patch = level->getPatch(p());
+            const Pointer<CartesianPatchGeometry<NDIM> > patch_geom = patch->getPatchGeometry();
+            const double* const patch_x_lower = patch_geom->getXLower();
+            const double* const patch_x_upper = patch_geom->getXUpper();
+            std::array<bool, NDIM> touches_upper_regular_bdry;
+            for (unsigned int d = 0; d < NDIM; ++d)
+                touches_upper_regular_bdry[d] = patch_geom->getTouchesRegularBoundary(d, 1);
+
+            // Store the values of F_JxW and X at the nodes inside the patch.
+            std::vector<double> F_x_dX_node, X_node;
+            F_x_dX_node.reserve(n_vars * num_active_patch_nodes);
             X_node.reserve(NDIM * num_active_patch_nodes);
             std::vector<dof_id_type> F_idxs, X_idxs;
+            IBTK::Point X;
             for (unsigned int k = 0; k < num_active_patch_nodes; ++k)
             {
                 const Node* const n = patch_nodes[k];
-                for (unsigned int i = 0; i < n_vars; ++i)
-                {
-                    IBTK::get_nodal_dof_indices(F_dof_map, n, i, F_idxs);
-                    for (const auto& F_idx : F_idxs)
-                    {
-                        F_dX_node.push_back(F_dX_local_soln[F_dX_petsc_vec->map_global_to_local_index(F_idx)]);
-                    }
-                }
+                bool inside_patch = true;
                 for (unsigned int d = 0; d < NDIM; ++d)
                 {
                     IBTK::get_nodal_dof_indices(X_dof_map, n, d, X_idxs);
-                    for (const auto& X_idx : X_idxs)
+                    X[d] = X_local_soln[X_petsc_vec->map_global_to_local_index(X_idxs[0])];
+                    inside_patch =
+                        inside_patch && (X[d] >= patch_x_lower[d]) &&
+                        ((X[d] < patch_x_upper[d]) || (touches_upper_regular_bdry[d] && X[d] <= patch_x_upper[d]));
+                }
+                if (inside_patch)
+                {
+                    for (unsigned int i = 0; i < n_vars; ++i)
                     {
-                        X_node.push_back(X_local_soln[X_petsc_vec->map_global_to_local_index(X_idx)]);
+                        IBTK::get_nodal_dof_indices(F_dof_map, n, i, F_idxs);
+                        for (const auto& F_idx : F_idxs)
+                        {
+                            F_x_dX_node.push_back(
+                                F_x_dX_local_soln[F_x_dX_petsc_vec->map_global_to_local_index(F_idx)]);
+                        }
                     }
+                    X_node.insert(X_node.end(), &X[0], &X[0] + NDIM);
                 }
             }
-            TBOX_ASSERT(F_dX_node.size() == n_vars * num_active_patch_nodes);
-            TBOX_ASSERT(X_node.size() == NDIM * num_active_patch_nodes);
+            TBOX_ASSERT(F_x_dX_node.size() <= n_vars * num_active_patch_nodes);
+            TBOX_ASSERT(X_node.size() <= NDIM * num_active_patch_nodes);
 
             // Spread values from the nodes to the Cartesian grid patch.
             //
@@ -903,29 +903,25 @@ FEDataManager::spread(const int f_data_idx,
             // on periodic boundaries will be "double counted".
             //
             // \todo Add warnings for FE structures with periodic boundaries.
-            const Pointer<Patch<NDIM> > patch = level->getPatch(p());
             const Box<NDIM> spread_box = patch->getBox();
             Pointer<PatchData<NDIM> > f_data = patch->getPatchData(f_data_idx);
             if (cc_data)
             {
                 Pointer<CellData<NDIM, double> > f_cc_data = f_data;
                 LEInteractor::spread(
-                    f_cc_data, F_dX_node, n_vars, X_node, NDIM, patch, spread_box, spread_spec.kernel_fcn);
+                    f_cc_data, F_x_dX_node, n_vars, X_node, NDIM, patch, spread_box, spread_spec.kernel_fcn);
             }
             if (sc_data)
             {
                 Pointer<SideData<NDIM, double> > f_sc_data = f_data;
                 LEInteractor::spread(
-                    f_sc_data, F_dX_node, n_vars, X_node, NDIM, patch, spread_box, spread_spec.kernel_fcn);
+                    f_sc_data, F_x_dX_node, n_vars, X_node, NDIM, patch, spread_box, spread_spec.kernel_fcn);
             }
         }
 
         // Restore local form vectors.
-        F_dX_petsc_vec->restore_array();
+        F_x_dX_petsc_vec->restore_array();
         X_petsc_vec->restore_array();
-
-        // Restore the value of the F vector.
-        F_vec = *F_vec_bak;
     }
     else
     {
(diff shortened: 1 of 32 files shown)


ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -849,8 +849,9 @@ FEDataManager::spread(const int f_data_idx,
         // Multiply by the nodal volume fractions (to convert densities into
         // values).
         NumericVector<double>& F_dX_vec = F_vec;
-        const NumericVector<double>& dX_vec = *buildDiagonalL2MassMatrix(system_name);
-        F_dX_vec.pointwise_mult(F_vec, dX_vec);
+        std::unique_ptr<NumericVector<double> > dX_vec = F_dX_vec.clone();
+        *dX_vec = *buildDiagonalL2MassMatrix(system_name);
+        F_dX_vec.pointwise_mult(F_vec, *dX_vec);
```
boyceg: The easiest way to get this to work is to ensure that the vector containing the diagonal entries of the mass matrix has exactly the same parallel layout as the other vectors involved in the operation.

src/IB/IBFEMethod.cpp
```diff
@@ -3662,6 +3662,12 @@ IBFEMethod::getFromInput(Pointer<Database> db, bool /*is_from_restart*/)
         d_default_interp_spec.use_nodal_quadrature = db->getBool("interp_use_nodal_quadrature");
     else if (db->isBool("IB_use_nodal_quadrature"))
         d_default_interp_spec.use_nodal_quadrature = db->getBool("IB_use_nodal_quadrature");
+    if (d_default_interp_spec.use_nodal_quadrature)
```
boyceg: Probably we should use `QSIMPSON` for second-order meshes, but currently there is only a single `default_interp_spec` and a single `default_spread_spec`.

boyceg: But if we don't specify _anything_ here, then bad things happen in determining the bounding boxes.

src/IB/IBFEMethod.cpp
```diff
@@ -3662,6 +3662,12 @@ IBFEMethod::getFromInput(Pointer<Database> db, bool /*is_from_restart*/)
...
+    {
+        d_default_interp_spec.quad_type = QTRAP;
+        d_default_interp_spec.quad_order = FIRST;
+    }
```
drwells: I'm not sure we have code in place to handle it, but we should disable adaptive quadrature here too.

boyceg: Yes. Also we should reorder the way things are setup to avoid overriding user settings. Will push an update in 1 minute.

boyceg:
Some of the PETSc `#include`s (that IWYU added???) are causing problems with Jenkins. Not sure what is going on...

drwells:
IWYU might have done it but I don't see how `#include "petscksp.h"` could cause problems - that header has to be available. Do you remember what the error was?

boyceg:
It was complaining about `#include "petscsystypes.h"`.

boyceg:
that fixed Jenkins......

boyceg:
@drwells I think I just added a test.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -842,19 +835,19 @@ FEDataManager::spread(const int f_data_idx,
...
+        *dX_vec = *buildDiagonalL2MassMatrix(system_name);
+        F_dX_vec->pointwise_mult(F_vec, *dX_vec);
 
         // Extract local form vectors.
-        auto F_dX_petsc_vec = static_cast<PetscVector<double>*>(&F_dX_vec);
+        auto F_dX_petsc_vec = static_cast<PetscVector<double>*>(F_dX_vec.get());
+        int ierr = VecGhostUpdateBegin(F_dX_petsc_vec->vec(), INSERT_VALUES, SCATTER_FORWARD);
+        IBTK_CHKERRQ(ierr);
+        ierr = VecGhostUpdateEnd(F_dX_petsc_vec->vec(), INSERT_VALUES, SCATTER_FORWARD);
+        IBTK_CHKERRQ(ierr);
```
drwells, line 850: We typically don't do communication inside FEDataManager. Could this be moved somewhere else?

boyceg: Perhaps, but we would somehow need to know the layout of the `F` vector in order to setup the ghosted entries of the vector of the diagonal mass matrix entries.

An alternative could be to apply the diagonal mass matrix in a matrix-free manner.

boyceg: Anyway, if you have suggestions on where to move it / how to avoid it, I am happy to do it; but I couldn't think of a simple way to do it, which is why it is where it is now.

drwells: I don't understand why this is necessary. Applying the diagonal mass matrix should be a purely local operation. Let me think about this a little more.

boyceg: Because we need the forces on the IB ghost nodes. Upon entry to the function, `F` has units of force density. We need to scale the forces by the mass matrix to get the nodal forces.

boyceg: (Notice that this is essentially what happens for the other code path, but it is done in a matrix-free way.)

drwells:
Could you also add a changelog entry?

boyceg:
Sure — at nap time. :-)

boyceg:
BTW, the reason I didn’t do this in a matrix free way is so that we only define the lumping procedure in one place (because there are different ways to do lumping).

boyceg:
I guess it would be easy to add a helper function.

boyceg:
@drwells how about this? I am not sure if we can get away with just using `NumericVector<double>::pointwise_mult()` here, which would be much more compact...

boyceg:
In fact we can ... let me do that instead.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: @drwells how is this?

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2389,6 +2365,19 @@ FEDataManager::buildDiagonalL2MassMatrix(const std::string& system_name)
     return d_fe_data->d_L2_proj_matrix_diag[system_name].get();
 } // buildDiagonalL2MassMatrix
 
+NumericVector<double>*
+FEDataManager::buildGhostedDiagonalL2MassMatrix(const std::string& system_name)
+{
+    if (!d_fe_data->d_L2_proj_matrix_diag_ghost.count(system_name))
+    {
+        std::unique_ptr<NumericVector<double> > M_vec = buildGhostedSolutionVector(system_name)->clone();
```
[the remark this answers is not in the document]
drwells: I think we should avoid using `NumericVector<double>` whenever possible - we always assume later on that it's a `PetscVector<double>` so we might as well cast it here.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2389,6 +2365,19 @@ FEDataManager::buildDiagonalL2MassMatrix(const std::string& system_name)
...
+        *M_vec = *buildDiagonalL2MassMatrix(system_name);
+        M_vec->close();
+        d_fe_data->d_L2_proj_matrix_diag[system_name] = std::move(M_vec);
+    }
+    return d_fe_data->d_L2_proj_matrix_diag_ghost[system_name].get();
```
drwells, line 2378: Putting this in `d_fe_data` is wrong - this mass matrix depends on the IB partitioning, but `FEData` doesn't know about the IB partitioning. This is going to cause problems with the scratch hierarchy.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1657,17 +1633,17 @@ FEDataManager::interpWeighted(const int f_data_idx,
                     F_node, n_vars, X_node, NDIM, f_sc_data, patch, interp_box, interp_spec.kernel_fcn);
             }
 
-            // Insesrt the values of F at the nodes.
+            // Insert the values of F at the nodes.
             F_vec.insert(F_node, F_node_idxs);
         }
+        F_vec.close();
```
drwells, line 1639: We can completely remove all parallel communication in this code by utilizing the fact that we already have all the ghost data we need - the values of the lumped mass matrix don't change. If we work with the local PETSc versions (i.e., `VecGhostGetLocalForm()` and `VecGhostRestoreLocalForm()`) then we can scale things and get correct ghost values.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -826,37 +821,26 @@ FEDataManager::spread(const int f_data_idx,
...
-
         // Multiply by the nodal volume fractions (to convert densities into
         // values).
-        NumericVector<double>& F_dX_vec = F_vec;
-        const NumericVector<double>& dX_vec = *buildDiagonalL2MassMatrix(system_name);
-        F_dX_vec.pointwise_mult(F_vec, dX_vec);
-        F_dX_vec.close();
+        NumericVector<double>* dX_vec = buildGhostedDiagonalL2MassMatrix(system_name);
+        std::unique_ptr<NumericVector<double> > F_dX_vec = F_vec.clone();
+        F_dX_vec->pointwise_mult(F_vec, *dX_vec);
```
drwells, line 839: Maybe I'm being dense but I still don't know what `F_dx` is supposed to stand for here.

ibtk/src/lagrangian/FEDataManager.cpp
drwells: In retrospect it doesn't make much sense for the other lumped mass matrix to live here either.

ibtk/src/lagrangian/FEDataManager.cpp
drwells: Better, but we can improve it more. Given our other problems with multiple parts I would like to completely avoid communication in this class.

boyceg: Yes I got rid of this --- this is out of date.

ibtk/src/lagrangian/FEDataManager.cpp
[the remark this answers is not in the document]
boyceg: Should we avoid using it in `FEDataManager::buildGhostedCoordsVector()`?

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: I will move both out into the `FEDataManager` class.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: F * dX

boyceg: I will change it to `F_x_dX`.

boyceg: `F` has units of density; for nodal interaction we need to get to total amount of stuff associated with the node. `dX` is the volume associated with the node.

ibtk/src/lagrangian/FEDataManager.cpp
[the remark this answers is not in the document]
drwells: Yes, but we should address that in another issue.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: So what we are doing here is to interpolate the the IB points that correspond to the given Cartesian grid data. Those IB points may be owned by another processor, and so I don't see how we can avoid communication.

That said, what this particular routine is doing is a bit perverse. With nodal interpolation, we do not have to do an L2 projection; we are interpolating the background field directly to the nodes of the FE mesh.

Here I am doing interpolating to the nodes and then _weighting_ the result using the diagonal mass matrix, so that the units are the same as in the other code path. Then, we go back and "solve" the L2 projection in `IBFEMethod` --- except that for nodal interpolation, it is the trivial projection using the diagonal mass matrix.

boyceg: (In other words, if we did not weight the result in `interWeighted()` using the diagonal mass matrix, the we also would not have to go back to do a subsequent solve.)

boyceg:
@drwells, I have changed the way that `interpWeighted()` works for nodal interpolation, and now skip the (trivial) L2 projection when using nodal interpolation.

boyceg:
Grrrr, something is still wrong.

boyceg:
OK, this really does seem to work correctly now.

@drwells could you take a look? I will squash down to a couple commits if this looks OK.

drwells:
I'm looking at this now. Both tests fail for me :(

boyceg:
Let me check.

boyceg:
I screwed up the test output. Should be fixed now.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -882,7 +871,7 @@ FEDataManager::spread(const int f_data_idx,
                     IBTK::get_nodal_dof_indices(F_dof_map, n, i, F_idxs);
                     for (const auto& F_idx : F_idxs)
                     {
-                        F_dX_node.push_back(F_dX_local_soln[F_dX_petsc_vec->map_global_to_local_index(F_idx)]);
+                        F_x_dX_node.push_back(F_x_dX_local_soln[F_x_dX_petsc_vec->map_global_to_local_index(F_idx)]);
                     }
```
drwells, line 875: These loops (over n_vars then over nodes per element) look wrong to me - we expect the input to the spread function to have the index over vars running fastest (i.e., we have `REAL X(0:NDIM-1,0:*)` in Fortran). We definitely use 'vars fastest' in the nonnodal case (see `sum_weighted_elem_solution`).

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -848,15 +839,13 @@ FEDataManager::spread(const int f_data_idx,
 
         // Multiply by the nodal volume fractions (to convert densities into
         // values).
-        NumericVector<double>& F_dX_vec = F_vec;
-        const NumericVector<double>& dX_vec = *buildDiagonalL2MassMatrix(system_name);
-        F_dX_vec.pointwise_mult(F_vec, dX_vec);
-        F_dX_vec.close();
+        PetscVector<double>* dX_vec = buildIBGhostedDiagonalL2MassMatrix(system_name);
+        std::unique_ptr<NumericVector<double> > F_x_dX_vec = F_vec.clone();
+        F_x_dX_vec->pointwise_mult(F_vec, *dX_vec);
```
drwells, line 844: I verified that this does update ghost values.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2382,13 +2371,27 @@ FEDataManager::buildDiagonalL2MassMatrix(const std::string& system_name)
...
+    return d_L2_proj_matrix_diag[system_name].get();
 } // buildDiagonalL2MassMatrix
 
+PetscVector<double>*
+FEDataManager::buildIBGhostedDiagonalL2MassMatrix(const std::string& system_name)
+{
+    if (!d_L2_proj_matrix_diag_ghost.count(system_name))
+    {
+        std::unique_ptr<PetscVector<double> > M_vec(
+            static_cast<PetscVector<double>*>(buildIBGhostedVector(system_name)->clone().release()));
```
drwells, line 2387:
```suggestion
        std::unique_ptr<PetscVector<double> > M_vec = buildIBGhostedVector(system_name);
```

drwells: That function already returns a new vector every time it is called.

src/IB/IBFEMethod.cpp
```diff
@@ -774,6 +776,15 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
                                                             data_time,
                                                             /*close_F*/ false,
                                                             /*close_X*/ false);
```
drwells, line 778: This logic is really complicated since `interpWeighted` no longer does weighted interpolation in the nodal quadrature case. Could we change `interpWeighted` to always create a weighted RHS and then just solve for things normally later on so we avoid this kind of problem? I would rather be consistent than maximally efficient.

src/IB/IBFEMethod.cpp
```diff
@@ -783,18 +794,22 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
 
     if (d_use_ghosted_velocity_rhs)
     {
-        batch_vec_ghost_update(U_rhs_vecs, ADD_VALUES, SCATTER_REVERSE);
-        batch_vec_ghost_update(U_rhs_vecs, INSERT_VALUES, SCATTER_FORWARD);
+        batch_vec_ghost_update(U_active_elemental_quadrature_rhs_vecs, ADD_VALUES, SCATTER_REVERSE);
+        batch_vec_ghost_update(U_active_elemental_quadrature_rhs_vecs, INSERT_VALUES, SCATTER_FORWARD);
+        batch_vec_assembly(U_active_nodal_quadrature_rhs_vecs);
```
drwells, line 799: That would also get rid of the need to change things here as well.

src/IB/IBFEMethod.cpp
```diff
@@ -806,11 +821,14 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
                                                                  /*close_F*/ false);
```
drwells, line 821: Similarly, it's dumb, but I would prefer to keep calling the linear solver (even if it takes exactly one iteration to solve) in every case to simplify control flow.

drwells:
I'm also not sure we are using the lumped mass matrix correctly in the case where one element is spread across multiple patches - I need to ponder that some more.

drwells:
One way to use the same code path in all cases would be to add an `IBSolver` to `FEDataManager`, which is either aliased to the normal mass matrix or returns the lumped one (should that option be relevant).

src/IB/IBFEMethod.cpp
boyceg: Sure — I just thought of a way to do this without calling close().

However, because nodal interpolation uses insert instead of add, I think we cannot avoid either calling assemble or using INSERT_VALUES at some point, which is different from the other code path when using ghosted assembly.

drwells: I need to reread our code for setting up a lumped mass matrix but I believe we still need to call 'add' to get the element contributions that are not on the current processor.

drwells:
Another thing I would like to add is explicit interpolation tests - we should really verify that we get nodally second order accurate results in parallel. That would verify that a few things above (such as not setting up the lumped mass matrix, data orderings, etc) are fixed.

boyceg:
Could you set up the interpolation tests while I try to sort out the rest?

boyceg:
I would like to try to get this working as a way to try to speed up some big models that are already using fine structure meshes.

drwells:
Sure.

drwells:
Hmm.. I get different errors with different processor counts. I'm investigating.

drwells:
Well, that's embarrassing - I forgot to do a min reduction in `interpolate_01.cpp` over the error. Good thing we are taking a second look!

boyceg:
> Furthermore - what is the relationship between `use_consistent_mass_matrix` and `IB_use_nodal_quadrature`? The former just refers to computing the Lagrangian forces, right?

`use_consistent_mass_matrix` controls how the structural forces are computed.

`IB_use_nodal_quadrature` controls whether nodal integration is used (or not) in interaction.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: Note that we are in fact looping over in this order, because at each node we then loop over the variables (which for vector-valued quantities correspond to the components).

src/IB/IBFEMethod.cpp
boyceg: OK I have reverted this.

ibtk/src/lagrangian/FEDataManager.cpp
drwells: Oh, I see - that makes sense.

drwells:
I tried to create a PR twice and failed twice (GitHub's unicorn of death). I put the tests on slack.

boyceg:
I think I have finally sorted this out (unless I am not understanding how spreading now works --- which, sadly, is quite possible!)

drwells:
> (unless I am not understanding how spreading now works --- which, sadly, is quite possible!)

Welcome to the club :)

boyceg:
> > (unless I am not understanding how spreading now works --- which, sadly, is quite possible!)
> 
> Welcome to the club :)

Is there some documentation about how ghosted data is/should be handled in spreading? It looks like the implementation only spreads from points that are in the patch interiors, which is good, but I did not see where ghost values are handled.

drwells:
I'll put up some additional documentation in IBFEMethod summarizing this comment in a minute - the flow is pretty complicated (which is the main reason I don't want to add even more branches to this function).

You probably know most of this, but this is the complete IBFE spreading pipeline:

First, looping over each structure (and looping over each patch that intersects said structure):
1. Elements are associated with a patch when they intersect the bounding box of the patch enlarged by one ghost layer. Hence, between regrids, we are guaranteed that every element that could intersect the patch is considered.
2. We compute coordinates of quadrature points and values of the finite element force times `JxW` (which has units of force * volume) on all elements associated with a patch.
3. To prevent the (highly unlikely) case of a quadrature point on a patch boundary we call `zeroExteriorValues` on the data so that all force values outside the patch are zero.
4. We then provide a box to LEInteractor which specifies that we should only spread values at coordinates inside each patch. This somewhat duplicates the previous step (but the previous step also includes points that could be on shared patch faces). A point on a patch interior near the boundary can spread outside the patch - we account for this by having enough ghost cells that we can capture all the spread values.

Then for all the patches (no more structure loop):
1. We accumulate values spread outside the physical domain back into the physical domain. This process should be thought of as accumulating values from non-DoF cells to cells with DoFs - whether or not the cells are ghosts is irrelevant (i.e., we can accumulate values into the ghost cells of a patch).
2. We do a global sum via `SAMRAIGhostDataAccumulator` to combine the contributions from both the previous step and step 4 above so that forces spread into ghost cells are now sent to the correct owning patch. I think of this process as essentially the same as doing finite element vector assembly - we can sum into regions the local processor doesn't own.


Is this what you are looking for? Now seems like a good time to exhaustively document the topic.

boyceg:
Yes, that's it; I was looking for the ghost data accumulation code in `FEDataManager` and not finding it.

Which leads to a question: I wonder if it would make sense to put the `SAMRAIGhostDataAccumulator` stuff into `FEDataManager`, possibly turning spreading into a two-phase operation -- maybe `spreadBegin()` and `spreadEnd()`?

boyceg:
Anyway, those changes would be for a different PR, of course.

For this PR, if you take a look at the nodal spreading and nodal interpolation code, you should see that I try to exclude points that are outside of the patch boxes, so I think that even though the nodal partitioning does have some overlap, all of the right points are available and are hit only once.

drwells:
> Which leads to a question: I wonder if it would make sense to put the SAMRAIGhostDataAccumulator stuff into FEDataManager, possibly turning spreading into a two-phase operation -- maybe spreadBegin() and spreadEnd()?

I don't see how that could work with multiple parts - since multiple structures can sum into the same patch we really need to finish the loop over all structures before we proceed to the last two steps.

boyceg:
Ah yes, I had forgotten about multiple parts. Would it make sense for `FEDataManager` to provide a `static` helper function to finalize spreading? I am imagining my future self forgetting again that there is this extra step beyond calling `spread()`. :-)

drwells:
Unfortunately a static function isn't going to work with other things we have - we set up two sets of FEDataManagers so that we can use the scratch hierarchy and the primary hierarchy.

drwells:
Things are in order so lets merge.

drwells:
I am very happy that we ultimately did not need to change IBFEMethod!

PR #956, opened 2020-04-01
Required changes for IIM, part 0

drwells:
First part of #932 - these are the minor changes that don't really effect anything but add some fields we will need later.

1. switch to d_use_jump_conditions
2. add system names

#932 currently has a merge conflict - this also fixes that. Once this is merged I will open a successor to #932.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/include/ibamr/IBFESurfaceMethod.h
+++ b/include/ibamr/IBFESurfaceMethod.h
@@ -89,17 +89,33 @@ namespace IBAMR
  * \brief Class IBFESurfaceMethod is an implementation of the abstract base
  * class IBStrategy that provides functionality required by the IB method with
  * a finite element representation of a surface mesh.
+ *
+ * Coupling schemes include both IB formulations (integral operations with
+ * regularized delta function kernels) and an immersed interface method (IIM)
+ * scheme (E. M. Kolahdouz, A. P. S. Bhalla, B. A. Craven, and B. E. Griffith.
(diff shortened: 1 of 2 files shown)


include/ibamr/IBFESurfaceMethod.h
```diff
@@ -89,17 +89,33 @@ namespace IBAMR
...
  */
 class IBFESurfaceMethod : public IBStrategy
 {
 public:
-    static const std::string COORDS_SYSTEM_NAME;
     static const std::string COORD_MAPPING_SYSTEM_NAME;
+    static const std::string COORDS_SYSTEM_NAME;
     static const std::string FORCE_SYSTEM_NAME;
     static const std::string NORMAL_VELOCITY_SYSTEM_NAME;
+    static const std::string PRESSURE_IN_SYSTEM_NAME;
```
boyceg, line 109: One way to reduce the number of systems would be to have separate variables for the "inner" and "outer" reconstructions in the same `System` (e.g. variable `0` is inner and `1` is outer).

drwells: That's a good idea. I think at the end of this set of patches we should try to minimize the number of systems but it's not a big deal - I think it would be better to keep extra systems for now to avoid extra breakage.

PR #959, opened 2020-04-05
add support for different time stepping schemes in IBFEMethod

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/ibtk_utilities.h
+++ b/ibtk/include/ibtk/ibtk_utilities.h
@@ -18,6 +18,7 @@
 
 #include "ibtk/ibtk_macros.h"
 
+#include "tbox/MathUtilities.h"
 #include "tbox/PIO.h"
 #include "tbox/Utilities.h"
 
@@ -112,6 +113,28 @@ static const bool ENABLE_TIMERS = true;
 
 namespace IBTK
 {
+inline std::string
+get_data_time_str(const double data_time, const double current_time, const double new_time)
+{
+    const double half_time = 0.5 * (current_time + new_time);
+    if (SAMRAI::tbox::MathUtilities<double>::equalEps(data_time, current_time))
+    {
+        return "current";
+    }
+    else if (SAMRAI::tbox::MathUtilities<double>::equalEps(data_time, half_time))
+    {
+        return "half";
+    }
+    else if (SAMRAI::tbox::MathUtilities<double>::equalEps(data_time, new_time))
+    {
+        return "new";
+    }
+    else
+    {
+        return "unknown";
+    }
+}
+
 template <class T, unsigned N>
 inline std::array<T, N>
 array_constant(const T& v)
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -687,6 +687,12 @@ public:
 
     /*!
      * Advance the positions of the Lagrangian structure using the (explicit)
+     * backward Euler method.
+     */
+    void backwardEulerStep(double current_time, double new_time) override;
+
+    /*!
+     * Advance the positions of the Lagrangian structure using the (explicit)
      * midpoint rule.
      */
     void midpointStep(double current_time, double new_time) override;
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -18,6 +18,7 @@
 #include "ibamr/IBFEDirectForcingKinematics.h"
 #include "ibamr/IBFEMethod.h"
 #include "ibamr/IBHierarchyIntegrator.h"
+#include "ibamr/ibamr_enums.h"
 #include "ibamr/ibamr_utilities.h"
 #include "ibamr/namespaces.h" // IWYU pragma: keep
 
@@ -651,7 +652,14 @@ IBFEMethod::preprocessIntegrateData(double current_time, double new_time, int nu
     // U^{n+1/2} and U^{n+1} to equal U^{n}.
     d_X_vecs->copy("solution", { "current", "new", "half" });
     d_U_vecs->copy("solution", { "current", "new", "half" });
-    d_F_vecs->copy("solution", { "half" });
+    switch (d_ib_solver->getTimeSteppingType())
+    {
+    case MIDPOINT_RULE:
+        d_F_vecs->copy("solution", { "current", "half" });
+        break;
+    default:
+        d_F_vecs->copy("solution", { "current", "new", "half" });
+    }
 
     // Update the mask data.
     getVelocityHierarchyDataOps()->copyData(mask_new_idx, mask_current_idx);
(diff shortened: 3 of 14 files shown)


boyceg:
The intention here is to get time stepping schemes other than just explicit `MIDPOINT_RULE` working with `IBFEMethod`. I still need to test it out.

boyceg:
@drwells, I am running the IBFE part of the test suite now. Let me know if there are any changes you'd like made.

src/IB/IBFEMethod.cpp
```diff
@@ -18,6 +18,7 @@
 #include "ibamr/IBFEDirectForcingKinematics.h"
 #include "ibamr/IBFEMethod.h"
 #include "ibamr/IBHierarchyIntegrator.h"
+#include "ibamr/ibamr_enums.h"
```
boyceg, line 21: For better or worse, `clangd` knows how to use `IWYU`. :-D

src/IB/IBFEMethod.cpp
```diff
@@ -318,6 +319,27 @@ assemble_poisson(EquationSystems& es, const std::string& /*system_name*/)
     }
 }
 
+std::string
+get_data_time_str(const double data_time, const double current_time, const double half_time, const double new_time)
```
boyceg, line 323: Is it worth making this more general?

boyceg:
I get the same failures for the IBFE tests with master and the branch.

src/IB/IBFEMethod.cpp
drwells: It seems fine to me - those are the time points we expect everywhere in the library.

src/IB/IBFEMethod.cpp
```diff
@@ -651,7 +673,14 @@ IBFEMethod::preprocessIntegrateData(double current_time, double new_time, int nu
...
     d_U_vecs->copy("solution", { "current", "new", "half" });
-    d_F_vecs->copy("solution", { "half" });
+    switch (d_ib_solver->getTimeSteppingType())
+    {
+    case MIDPOINT_RULE:
+        d_F_vecs->copy("solution", { "current", "half" });
+        break;
+    default:
+        d_F_vecs->copy("solution", { "current", "new", "half" });
+    }
```
drwells, line 683: This is a nice application of the copy syntax. I think I see why you wanted to use it immediately.

src/IB/IBFEMethod.cpp
```diff
@@ -733,35 +770,16 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
...
-        U_rhs_vecs = d_U_vecs->getIBGhosted("tmp");
-    else
-        U_rhs_vecs = d_U_vecs->get("RHS Vector");
-
-    std::vector<Pointer<RefineSchedule<NDIM> > > no_fill(u_ghost_fill_scheds.size(), nullptr);
-
+    std::vector<PetscVector<double>*> U_vecs = d_U_vecs->get(data_time_str);
+    std::vector<PetscVector<double>*> X_vecs = d_X_vecs->get(data_time_str);
+    std::vector<PetscVector<double>*> U_rhs_vecs =
+        d_use_ghosted_velocity_rhs ? d_U_vecs->getIBGhosted("tmp") : d_U_vecs->get("RHS Vector");
```
drwells, line 776: :heavy_check_mark:

src/IB/IBFEMethod.cpp
drwells: It would be nice to, at some point, move this switch statement into a more general place but we can handle that in another PR.

boyceg: I could move this to `ibtk_utilities.h`.

src/IB/IBFEMethod.cpp
drwells: That's good because it sure doesn't work with GCC!

src/IB/IBFEMethod.cpp
drwells: Let's add that to this PR.

drwells: or just merge - up to you.

boyceg:
@drwells, before I merge, can you check that the new tests pass for you (and update outputs as needed if not)? The new tests should be captured by `-R FORWARD_EULER` and `-R TRAPEZOIDAL_RULE`.

drwells:
Sure.

drwells:
I'd like to get rid of some duplicated test input and output files. Would you mind if I pushed a commit to this branch that replaces some of the duplicated input and output files with symbolic links?

include/ibamr/IBFEMethod.h
```diff
@@ -685,6 +685,12 @@ class IBFEMethod : public IBStrategy
      */
     void forwardEulerStep(double current_time, double new_time) override;
 
+    /*!
+     * Advance the positions of the Lagrangian structure using the (explicit)
+     * backward Euler method.
+     */
+    void backwardEulerStep(double current_time, double new_time) override;
```
drwells, line 692: Quick question - why is this implemented but not mentioned in the changelog?

boyceg: Because it is not really backward Euler.

boyceg: Because it is not really backward Euler.

src/IB/IBFEMethod.cpp
boyceg: I will move it and squash after you update tests.

drwells:
It looks like tests fail because of inconsistent DoF enumeration between processors, which is a bit surprising - the whole point of libMesh's `LinearPartitioner` is that it is very consistent. Otherwise the test output matches. I'll disable IBFE logging to fix the problem.

drwells:
I cleaned up the tests. Since the problem is due to partitioning (which we no longer print) things should now work for both of us.

boyceg:
OK I will finish up in a bit and merge.

src/IB/IBFEMethod.cpp
boyceg: Since `half_time = 0.5*(current_time + new_time)`, I thought it might make sense just to have folks pass in `current_time` and `new_time`:
```
template <class T>
inline std::string
get_data_time_str(const T data_time, const T current_time, const T new_time)
{
    const T half_time = static_cast<T>(0.5) * (current_time + new_time);
    if (SAMRAI::tbox::MathUtilities<T>::equalEps(data_time, current_time))
    {
        return "current";
    }
    else if (SAMRAI::tbox::MathUtilities<T>::equalEps(data_time, half_time))
    {
        return "half";
    }
    else if (SAMRAI::tbox::MathUtilities<T>::equalEps(data_time, new_time))
    {
        return "new";
    }
    else
    {
        return "unknown";
    }
}
```
OK?

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -112,6 +113,29 @@ static const bool ENABLE_TIMERS = true;
 
 namespace IBTK
 {
+template <class T>
+inline std::string
+get_data_time_str(const T data_time, const T current_time, const T new_time)
+{
+    const T half_time = static_cast<T>(0.5) * (current_time + new_time);
```
boyceg, line 120: @drwells, should I make this simply
```suggestion
    const T half_time = 0.5 * (current_time + new_time);
```
?

drwells: Templating on the time type seems like a premature optimization to me - what is the advantage of doing it?

boyceg: Fair enough --- let's just leave it as `double`.

include/ibamr/IBFEMethod.h
boyceg: It is not a real implementation of backward Euler --- it is basically still forward Euler, but a little more stable.

boyceg:
jenkins retest this please

Issue #482, opened 2019-02-17
Use FECache in FEDataInterpolation


drwells:
We seriously need to do this: fixing this should be easy and we still spend a ton of time recomputing shape functions.

drwells:
Between this, the dof caching, and #444 we should be able to cut off another 15-20% of total run time, which would be nearly as good as the first round of serial performance patches.

drwells:
Another function to tackle in this class is interpolateCommon. We call this so many times that it may be worthwhile to use some more expensive optimizations:
1. Use denser objects than `std::vector<std::vector<std::vector<std::vector<VectorValue>>>>`. We can probably use `boost::multi_array`.
2. Template the function based on the number of basis functions per cell so that we can unroll some loops.
3. Template the function on the number of variables so we can unroll more loops.
4. Make sure that computing gradients is optional.

drwells:
I want to get to this soon but, given the C++14 issues, I think this should wait until after we push out another release.

The heart model spends about 17% of its time calling this function, most of which is spent recomputing shape function values. However, since the current version of this class handles multiple systems and finite elements its not trivial to put in FECache and initialize things correctly,

drwells:
`FEDataInterpolation` might also be a source of excessive memory usage, as discussed in #684. A few notes:
1. The primary user of this class is `IBFE(Surface)Method::assembleInteriorForce()`: `IBFECentroidPostProcessor` also uses it. I don't think it is ever directly used in applications.
2. This class does two things:
   a. Save some computational effort for vector-valued FE fields (e.g., only compute derivatives of shape functions for one component of `X` and copy them to the others).
   b. Calculate values of the FE field at quadrature points (like [dealii::FEValues::get_function_gradients()](https://dealii.org/current/doxygen/deal.II/classFEValuesBase.html#ad1f4e0deb5d982e8172d82141c634a67)).

It can also compute values on faces. I suspect it would be clearer to split face functionality into a separate class (i.e., like [dealii::FEValues](https://dealii.org/current/doxygen/deal.II/classFEValues.html) and [dealii::FEFaceValues](https://dealii.org/current/doxygen/deal.II/classFEFaceValues.html), which have somewhat different reinitialization logic)

We can definitely improve performance by using our own cache objects inside this class (and, once #444 is in, that too).

drwells:
Now that most of the pieces have been written (e.g., the JxW stuff) replacing this class is going to be a project for January.

drwells:
[...]

PR #961, opened 2020-04-07
Implement contravariants

drwells:
First part of #482: we need to efficiently sidestep libMesh's computations of the contravariant so that we can compute both JxW values and gradients.

This first PR doesn't do much - it just explicitly stores the computed contravariant in a renamed class. The new few PRs will add other useful mapping features (array of jacobians, mapped quadrature points). Once we have all the features of the mapping we can move on to replace `libMesh::FEBase` for a subset of elements.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/JacobianCalculator.h
+++ b/ibtk/include/ibtk/JacobianCalculator.h
@@ -19,6 +19,7 @@
 #include <IBTK_config.h>
 
 #include <ibtk/ibtk_macros.h>
+#include <ibtk/ibtk_utilities.h>
 
 #include "tbox/Utilities.h"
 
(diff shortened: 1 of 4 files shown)


boyceg:
Should we try to send some of these things upstream to libMesh?

drwells:
Good question - that would be a substantial project.

There are two primarily difficulties with generalizing this:
1. We currently only work with Lagrange type elements (continuous and discontinuous) which makes doing isoparametric elements much easier. libMesh also supports a few other things, like Nedelec elements, that don't fit this paradigm. In principle we should be able to generalize our own `Mapping` class (which is a total ripoff of deal.II's) to cover Hdiv and Hcurl but thats going to be a huge undertaking.
2. libMesh relies on the whole `get_phi()` idiom of storing access to shape values. That's not going to work in quite the same way with this code since we are not recomputing things on every element - in particular, for a composite mesh, we would store things (by quadrature key) in a map and return different precomputed arrays and their calculators as needed.

It might still be conceivable to handle this for lagrange elements. Lets investigate the question again once we are done with our own changes.

Issue #657, opened 2019-08-16
Excessive memory usage in applications


drwells:
@boyceg @marshallrdavey Lets collect what we know in one place. I spent some more time poking around yesterday and I have a few ideas on how to proceed.

For those who are new to the discussion: under some circumstances on some machines the heart model exhibits a (slow) linear growth in the amount of memory (as measured by `getrusage`) over time.

The bad news is that I ran a few tests with a memory sanitizer and I have not yet been able to actually detect any *leaks*: the memory is allocated and (at least for IBFE/explicit/ex4) still deallocated at the end of the program. MPI does leak a little bit of memory, but it only does it in two places and it is not substantial enough to cause our current problems.

Hence, I don't think valgrind or the compiler's leak sanitizer are going to help us find this bug.

My current plan is to run the heart model for a long time on my workstation with `heaptrack`, which intercepts calls to `malloc` and then records the results. I attached a graphical depiction of our memory usage in `ex4`. *warning*: I haven't used this tool before: it generates nice plots but tools built around libunwind sometimes print obviously nonsense answers, so its possible this is misleading us. I'll assume for the rest of the discussion that its reasonably accurate, but we should do more testing.

Here is a visualization of our memory usage generated by heaptrack:
![heaptrack-screenshot](https://user-images.githubusercontent.com/3917035/63180780-f85d6b80-c01c-11e9-8a63-05f43dd2098f.png)
This is a flamegraph describing where we allocate memory: most of our allocations come from doing stuff with SAMRAI. We allocate and deallocate an incredible amount of memory: heaptrack measured that, for this 32^3 grid, we allocated/deallocated nearly a terabyte over a thousand timesteps, which seems wrong (but nevertheless here we are!)

My current best guess is that we are breaking SAMRAI's memory pool implementation with our incredible number of allocations and deallocations, but we need to investigate further before I will be sure.

drwells:
The previous picture is for one MPI process. I'll post more data when I get it.

knepley:
Note that you can use Massif (part of valgrind) to do the same job. I have liked its interface.

abarret:
You should consider tracking the number of calls to `PatchLevel::allocatePatchData` and `PatchLevel::deallocatePatchData` where SAMRAI does most of its data allocation. Currently, I think we allocate and deallocate pretty freely. We could probably do a more efficient job of maintaining data storage.

drwells:
Those show up very prominently in the profile.

[...]

PR #963, opened 2020-04-08
Remove most calls to localize.

drwells:
We still use it in IBFESurfaceMethod which is in the process of being rewritten anyway.

@jabrown893 Please make similar fixes to the TAVR code, should that use `localize` in a similar way anywhere.

Related to #657.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/examples/IBFE/explicit/ex0/example.cpp
+++ b/examples/IBFE/explicit/ex0/example.cpp
@@ -531,7 +531,7 @@ main(int argc, char* argv[])
             System& X_system = equation_systems->get_system<System>(IBFEMethod::COORDS_SYSTEM_NAME);
             NumericVector<double>* X_vec = X_system.solution.get();
             NumericVector<double>* X_ghost_vec = X_system.current_local_solution.get();
-            X_vec->localize(*X_ghost_vec);
+            copy_and_synch(*X_vec, *X_ghost_vec);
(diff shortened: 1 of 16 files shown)


examples/IBFE/explicit/ex0/example.cpp
```diff
@@ -531,7 +531,8 @@ main(int argc, char* argv[])
             System& X_system = equation_systems->get_system<System>(IBFEMethod::COORDS_SYSTEM_NAME);
             NumericVector<double>* X_vec = X_system.solution.get();
             NumericVector<double>* X_ghost_vec = X_system.current_local_solution.get();
-            X_vec->localize(*X_ghost_vec);
+            *X_ghost_vec = *X_vec;
```
boyceg, line 534: cannot check right now, but is this equivalent to what `copy_and_synch` does?

drwells: It does! Should we just use that instead?

boyceg: Yes?

drwells:
Good point. Fixed!

jabrown893:
Yes, the TAVR code called this function four times.

PR #964, opened 2020-04-09
Fix a memory consumption bug in SAMRAIGhostDataAccumulator.

drwells:
Caught by massif while looking around for other problems. I modified a test to create `SAMRAIDataAccumulator`s in a loop - it leaks on master but doesn't with this patch.

We shouldn't create multiple variables - this actually uses a noticeable amount of memory in long time integrations.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
I'm not sure if we can do it in a general way. Here we want to create either a cell-centered or side-centered integer variable - since those conditions change every time we want setup variables that may already exist I don't think there's a simple way to do it.

PR #966, opened 2020-04-13
add standalone explicit dynamics solver

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
@drwells this is still very much WIP, but I wanted to be able to use `LibMeshSystemVectors` in cases where there are not corresponding `FEDataManager`s. Let me know what you think about the changes made here to `LibMeshSystemVectors`.

drwells:
I think it would make more sense to tackle this with inheritance instead of having two different ways in which this class can operate - e.g., we could make the current (on master) implementation `LibmeshIBSystemVectors` with a different constructor (and also have it offer the IB functions) and then make the 'new' class here the base class (which takes the array of `EquationSystems` in its constructor instead).

boyceg:
@drwells, how about this? If you are happy with this, I will squash and cherry pick into a separate PR.

boyceg:
I split off the changes to `LibMeshSystemVector` in #967.

PR #967, opened 2020-04-13
split LibMeshSystemVectors into base and derived classes

boyceg:
The base class, `LibMeshSystemVectors`, now only needs to use `libMesh::EquationSystem` objects to track the vectors. The new derived class, `LibMeshSystemIBVectors`, uses `FEDataManager` objects to determine IB ghosting.

Part of #966.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/LibMeshSystemIBVectors.h
+++ b/ibtk/include/ibtk/LibMeshSystemIBVectors.h
@@ -1,0 +1,112 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2020 - 2020 by the IBAMR developers
+// All rights reserved.
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+#ifndef included_IBTK_LibMeshSystemIBVectors
+#define included_IBTK_LibMeshSystemIBVectors
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include <ibtk/FEDataManager.h>
+#include <ibtk/LibMeshSystemVectors.h>
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBTK
+{
+/*!
+ * \brief Class LibMeshSystemIBVectors is a convenience class that manages
+ * access to libMesh vectors for the same system defined on multiple
+ * parts. It extends the base class LibMeshSystemVectors to provide access
+ * to vectors ghosted with both the Lagrangian partitioning (i.e., libMesh's
+ * computed partitioning, as in the base class) as well as the IB
+ * partitioning (i.e., the partitioning based on the distribution of SAMRAI
+ * data).
+ *
+ * This class stores information that depends on SAMRAI's parallel
+ * partitioning: LibMeshSystemIBVectors::reinit() should be called after
+ * regridding to clear this data.
+ *
+ * @note This class is intended for internal use in IBAMR. It is used in
+ * IBFEMethod and IBFESurfaceMethod.
+ */
+class LibMeshSystemIBVectors : public LibMeshSystemVectors
+{
+public:
+    /*!
+     * Constructor.
+     *
+     * @param[in] fe_data_managers IBTK::FEDataManager objects for each
+     * part. These are used to get libMesh vectors with ghost regions
+     * corresponding to the Eulerian partitioning.
+     *
+     * @param[in] system_name Name of the libMesh::System whose vectors we are accessing.
+     */
+    LibMeshSystemIBVectors(const std::vector<IBTK::FEDataManager*>& fe_data_managers, std::string system_name);
+
+    /*!
+     * Constructor, taking a mask argument indicating which systems actually exist.
+     *
+     * @param[in] fe_data_managers IBTK::FEDataManager objects for each
+     * part. These are used to get libMesh vectors with ghost regions
+     * corresponding to the Eulerian partitioning.
+     *
+     * @param[in] part_mask A vector indicating on which parts the given
+     * system actually exists (for example, only some parts have fluid
+     * sources)
+     *
+     * @param[in] system_name Name of the libMesh::System whose vectors we are accessing.
+     */
+    LibMeshSystemIBVectors(const std::vector<IBTK::FEDataManager*>& fe_data_managers,
+                           const std::vector<bool>& part_mask,
+                           std::string system_name);
+
+    /*!
+     * Get an IB-ghosted vector for a specific part.
+     *
+     * @param[in] vec_name Name of the vector.
+     *
+     * @param[in] part Part number.
+     */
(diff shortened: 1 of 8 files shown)


boyceg:
@drwells, if this looks OK, can you check the test suite?

ibtk/include/ibtk/LibMeshSystemIBVectors.h
```diff
@@ -0,0 +1,129 @@
...
+ * All vectors added to a libMesh::System by this class are ghosted (with
+ * ghost data computed by the System) and registered for projection (i.e., if
+ * the mesh changes, these vectors will be automatically updated).
+ *
+ * This class stores information that depends on SAMRAI's parallel
+ * partitioning: LibMeshSystemIBVectors::reinit() should be called after
+ * regridding to clear this data.
+ *
+ * @note This class is intended for internal use in IBAMR. It is used in
+ * IBFEMethod and IBFESurfaceMethod.
```
drwells, line 57: The class documentation needs to be updated so that it explains what is different in this class vs. the base class.

ibtk/include/ibtk/LibMeshSystemVectors.h
```diff
@@ -57,7 +55,7 @@ namespace IBTK
  * ghost data computed by the System) and registered for projection (i.e., if
  * the mesh changes, these vectors will be automatically updated).
  *
- * This class stores information that depends on SAMRAI's parallel
+ * This class may store information that depends on SAMRAI's parallel
  * partitioning: LibMeshSystemVectors::reinit() should be called after
  * regridding to clear this data.
```
drwells, line 60: This  comment is no longer true and should be removed; i.e., this class doesn't know anything about SAMRAI any more.

ibtk/include/ibtk/LibMeshSystemVectors.h
```diff
@@ -158,14 +134,17 @@ class LibMeshSystemVectors
     void zero(const std::string& vec_name);
 
 protected:
-    /// Check if a vector exists and, if not, add it.
-    void maybeAdd(const std::string& vec_name);
+    LibMeshSystemVectors(const std::vector<bool>& part_mask, std::string system_name);
```
drwells, line 137: As mentioned below, we should not have this constructor since it doesn't fully construct the object.

ibtk/src/utilities/LibMeshSystemIBVectors.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+/////////////////////////////// NAMESPACE ////////////////////////////////////
+
+namespace IBTK
+{
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+LibMeshSystemIBVectors::LibMeshSystemIBVectors(const std::vector<IBTK::FEDataManager*>& fe_data_managers,
+                                               std::string system_name)
+    : LibMeshSystemVectors(std::vector<bool>(fe_data_managers.size(), true), system_name),
```
drwells, line 39: This is a bit risky because the parent class is going to be set up in some half-completed state. Lets move the rest of the code in the constructor into a convenience function that converts `std::vector<FEDataManager *>` to `std::vector<EquationSystems *>`.

ibtk/src/utilities/LibMeshSystemIBVectors.cpp
```diff
@@ -0,0 +1,108 @@
...
+    {
+        libMesh::EquationSystems* es = fe_data_manager->getEquationSystems();
+        d_systems.push_back(&es->get_system(d_system_name));
+    }
+}
+
+LibMeshSystemIBVectors::LibMeshSystemIBVectors(const std::vector<IBTK::FEDataManager*>& fe_data_managers,
+                                               const std::vector<bool>& part_mask,
+                                               std::string system_name)
+    : LibMeshSystemVectors(part_mask, system_name), d_fe_data_managers(fe_data_managers)
```
drwells, line 52: same

ibtk/include/ibtk/LibMeshSystemVectors.h
```diff
@@ -158,14 +134,17 @@ class LibMeshSystemVectors
...
 
-    /// Add an IB ghosted vector.
-    std::vector<std::unique_ptr<libMesh::PetscVector<double> > >& maybeAddIBGhosted(const std::string& vec_name);
+    /// Returns false if a standard libMesh vector; otherwise returns true.
+    inline bool vec_stored_in_map(const std::string& vec_name)
```
drwells, line 140: We should make this static so that its clear that it doesn't depend on the object.

ibtk/src/utilities/LibMeshSystemVectors.cpp
```diff
@@ -29,48 +28,35 @@
...
 
 namespace IBTK
 {
 /////////////////////////////// CLASS DEFINITION /////////////////////////////
 
-LibMeshSystemVectors::LibMeshSystemVectors(const std::vector<IBTK::FEDataManager*>& fe_data_managers,
+LibMeshSystemVectors::LibMeshSystemVectors(const std::vector<libMesh::EquationSystems*>& equation_systems,
                                            std::string system_name)
-    : d_fe_data_managers(fe_data_managers), d_system_name(std::move(system_name))
+    : d_system_name(std::move(system_name))
```
drwells, line 39: I can't remember what I originally did but we should just switch this constructor out for a call to the other one with `std::vector<bool>(equation_systems.size(), true)` (a delegated constructor)

drwells:
@boyceg and I communicated over slack and agreed that I can make the minor changes necessary to finish this.

boyceg:
@drwells I think I have addressed all comments.

boyceg:
Hmmm, except that now there are some extra commits that show up --- did I screw up rebase?

boyceg:
OK I think I fixed the rebase.

ibtk/include/ibtk/LibMeshSystemIBVectors.h
drwells: What was changed here?

boyceg: Not much.

boyceg: I think it is still correct though. :-)

drwells: It is but I would prefer to delete all the repeated text so that this doc string only explains the new stuff.

boyceg: OK I got rid of duplicated text.

drwells: :heavy_check_mark:

drwells:
The last round of changes broke a bunch of tests so they no longer run. Do you have time to take a look?

boyceg:
I think I was misusing `std::move()`. Can you give it another try?

boyceg:
Nope, that's not it. Unfortunately I am out of time to work on this right now.

drwells:
Bah. Would you like me to wrap it up? I think it's close.

drwells:
I went ahead and force-pushed a fix for the bug - we were populating the array of `System *`s twice which caused a check on its length to fail.

boyceg:
Thank you!!

PR #968, opened 2020-04-14
factor out L2 projection code into new class FEProjector

boyceg:
Part of #966.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEProjector.h
+++ b/ibtk/include/ibtk/FEProjector.h
@@ -1,0 +1,123 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2011 - 2019 by the IBAMR developers
+// All rights reserved.
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+#ifndef included_IBTK_FEProjector
+#define included_IBTK_FEProjector
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include <libmesh/equation_systems.h>
+#include <libmesh/petsc_linear_solver.h>
+#include <libmesh/petsc_matrix.h>
+#include <libmesh/petsc_vector.h>
+
+#include <map>
+#include <string>
+
+namespace IBTK
+{
+class FEData;
+} // namespace IBTK
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBTK
+{
+/*!
+ * \brief Class FEProjector coordinates data structures for projecting
+ * fields in FE models.
+ */
+class FEProjector
+{
+public:
+    /// Constructor.
+    FEProjector(libMesh::EquationSystems* equation_systems, bool enable_logging = true);
+
+    /// Alternative constructor that takes in a shared pointer to an FEData object.
+    FEProjector(std::shared_ptr<FEData> fe_data, bool enable_logging = true);
+
+    /// Deleted default constructor.
+    FEProjector() = delete;
+
+    /// Deleted copy constructor.
+    FEProjector(const FEProjector& from) = delete;
+
+    /// Deleted assignment operator.
+    FEProjector& operator=(const FEProjector& that) = delete;
+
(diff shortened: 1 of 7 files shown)


boyceg:
@drwells, if this looks OK, can you check the testsuite?

boyceg:
I noticed that I was rebuilding the solvers differently in the branch than in `master`; should be fixed now.

ibtk/lib/Makefile.am
```diff
@@ -354,12 +355,15 @@ pkg_include_HEADERS += \
 ../include/ibtk/private/PETScSAMRAIVectorReal-inl.h \
 ../include/ibtk/private/StreamableManager-inl.h
 
+## TODO: Should JacobianCalculator.h be in include or src?
+
```
drwells, line 359: This isn't relevant to the PR - lets open an issue.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -518,7 +514,8 @@ FEDataManager::reinitElementMappings()
     // We reinitialize mappings after repartitioning, so clear the cache since
     // its content is no longer relevant:
     d_fe_data->clearCached();
-    d_L2_proj_matrix_diag.clear();
+
+    // The ghosted diagonal mass matrix needs to be rebuilt after repartitioning.
     d_L2_proj_matrix_diag_ghost.clear();
```
drwells, line 519: We shouldn't have this and the other calls any more since it is no longer this class' responsibility to manage the solvers.

ibtk/include/ibtk/FEProjector.h
```diff
@@ -0,0 +1,123 @@
...
+
+namespace IBTK
+{
+/*!
+ * \brief Class FEProjector coordinates data structures for projecting
+ * fields in FE models.
+ */
+class FEProjector
+{
+public:
```
drwells, line 42: We also need a `clearCached` function here that resets (at least) any data dependent on the Eulerian partitioning.

We don't have proper separation of concerns right now regarding regridding libMesh data - I think this is done in IBFEMethod. We really need some clear distinction here between 'update the IB data' and 'update libMesh's data that depends on its own parallel partitioning'.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -0,0 +1,426 @@
...
+FEProjector::setLoggingEnabled(const bool enable_logging)
+{
+    d_enable_logging = enable_logging;
+}
+
+bool
+FEProjector::getLoggingEnabled() const
+{
+    return d_enable_logging;
+}
```
drwells, line 416: I think 100% of the actual content in this file is copied and pasted so I'm not going to read it :)

drwells:
@boyceg and I communicated over slack and agreed that I can make the minor changes necessary to finish this.

drwells:
Note to self - I misread the intent of this class during the initial review. `FEProjector` doesn't know anything about the Eulerian partitioning so the clear function is *probably* not needed (and the IB stuff stays in FEDataManager).

ibtk/lib/Makefile.am
boyceg: #969

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: @drwells notes this below, but the IB ghosted representation should be owned by `FEDataManager` or possibly `FEData`.

ibtk/include/ibtk/FEProjector.h
boyceg: Perhaps we should change the method name to something other than `clearCached` --- maybe -- this is kind of lengthy, but something like `clearCartGridDependentData()` or `clearCartPartitionDependentData()`?

ibtk/lib/Makefile.am
boyceg: Removed comment from that `Makefile.am`.

ibtk/include/ibtk/FEProjector.h
boyceg: I propose `clearPatchHierarchyDependentData`.

boyceg:
@drwells I think this is done. Can you run the testsuite? :-D

PR #971, opened 2020-04-15
avoid using FEDataManager in FEDataInterpolation

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEDataInterpolation.h
+++ b/ibtk/include/ibtk/FEDataInterpolation.h
@@ -60,13 +60,13 @@ class NumericVector;
 namespace IBTK
 {
 /*!
- * \brief Class FEDataInterpolation manages data requred to evaluate one or more FE field variables at a collection of
+ * \brief Class FEDataInterpolation manages data required to evaluate one or more FE field variables at a collection of
  * points, possibly (not not necessarily) corresponding to the points of a quadrature rule.
  */
 class FEDataInterpolation
 {
 public:
-    FEDataInterpolation(unsigned int dim, FEDataManager* const fe_data_manager);
+    FEDataInterpolation(unsigned int dim, std::shared_ptr<FEData> fe_data);
 
     ~FEDataInterpolation() = default;
 
@@ -235,7 +235,7 @@ public:
     size_t registerInterpolatedSystem(const libMesh::System& system,
                                       const std::vector<int>& vars = std::vector<int>(1, 0),
                                       const std::vector<int>& grad_vars = std::vector<int>(),
-                                      libMesh::NumericVector<double>* system_data = nullptr);
+                                      libMesh::NumericVector<double>* system_vec = nullptr);
 
     /*!
      * \brief Get the variable data for all of the systems.
@@ -274,11 +274,11 @@ public:
                                      const unsigned int qp);
 
     /*!
-     * \brief Initialize all of the data structures requred to evaluate the FE shape functions, quadrature rules, etc.
+     * \brief Initialize all of the data structures required to evaluate the FE shape functions, quadrature rules, etc.
      *
      * NOTE: This method must be called before reinitializing data on individual elements.
      */
-    void init(bool use_IB_ghosted_vecs);
+    void init();
 
     /*!
      * \brief Reinitialize the FE shape functions, quadrature rules, etc. for the specified element.
@@ -290,7 +290,7 @@ public:
                 const std::vector<double>* weights = nullptr);
 
     /*!
-     * \brief Reinitialize the FE shape functions, quadrature rules, etc. for the specified side of the specificed
+     * \brief Reinitialize the FE shape functions, quadrature rules, etc. for the specified side of the specified
      * element.
      *
      * NOTE: Nodal values are set by calling collectDataForInterpolation().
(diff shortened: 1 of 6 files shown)


ibtk/include/ibtk/FEDataInterpolation.h
```diff
@@ -235,7 +243,7 @@ class FEDataInterpolation
     size_t registerInterpolatedSystem(const libMesh::System& system,
                                       const std::vector<int>& vars = std::vector<int>(1, 0),
                                       const std::vector<int>& grad_vars = std::vector<int>(),
-                                      libMesh::NumericVector<double>* system_data = nullptr);
+                                      libMesh::NumericVector<double>* system_vec = nullptr);
```
boyceg, line 246: There is a class named `SystemData` so I thought I would try to make it clearer that this is the vector of DOFs.

drwells:
Makes sense - we only use `FEDataManager` in two places in this class.

Could we get rid of the whole 'use IB ghosted' logic and FEDataManager completely and just rely on the user to provide a ghosted vector?

boyceg:
That would require more invasive changes. For instance, we evaluate the stress tensors in the IB ghosted representation in some places, and the stress tensors themselves may have dependencies on `Systems`, so we would need to go in and manually communicate to IB ghosted form.

The current implementation is not so pretty but I think it is simpler than requiring callers to figure out which vectors to pass in.

boyceg:
Actually I think it wouldn't be too hard to make a little converter. Let me give it a shot.

boyceg:
OK I think I did it.

boyceg:
Jenkins retest this please

boyceg:
@drwells test suite seems to work for me. Can you check and let me know if there are other changes?

src/IB/IBFEMethod.cpp
```diff
@@ -330,6 +330,27 @@ libmesh_restart_file_name(const std::string& restart_dump_dirname,
...
+    std::vector<SystemData> ghosted_system_data;
+    EquationSystems* equation_systems = fe_data_manager->getEquationSystems();
+    for (const SystemData& system_data : unghosted_system_data)
+    {
+        const std::string& system_name = system_data.system_name;
+        const System& system = equation_systems->get_system(system_name);
+        NumericVector<double>* original_system_vec = system_data.system_vec;
+        if (!original_system_vec) original_system_vec = system.current_local_solution.get();
+        NumericVector<double>* ghost_system_vec =
+            fe_data_manager->buildGhostedSolutionVector(system_name, /*synch_data*/ false);
```
drwells, line 346: The logic in this function partially duplicates the logic in `LibMeshSystemVectors. Could we set up a private utility function in IBFEMethod that does the same thing here but uses that class to access vectors? That would also allow us to not use `buildGhostedSolutionVector` here, which is a deprecated function.

boyceg:
Which tests?

boyceg:
My original rewrite here split the class into base/derived, but if we are just going to put the onus of setting up ghosting correctly on the caller, less invasive changes are needed. I just reverted most of the changes. @drwells can you see if this at least passes the test suite? I can work on switching to use `LibMeshSystemVectors`.

drwells:
The test failure is related to something else - I'm working on a fix now.

drwells:
> but if we are just going to put the onus of setting up ghosting correctly on the caller, less invasive changes are needed

Now seems like a good time to reach a consensus on this issue - I think that we should *always* require the caller to provide sufficient ghost information. This creates more work for the user but makes logic a lot clearer and prevents unnecessary blocks in the multipart case.

I would like to apply this principle everywhere - what do others think?

boyceg:
> > but if we are just going to put the onus of setting up ghosting correctly on the caller, less invasive changes are needed
> 
> Now seems like a good time to reach a consensus on this issue - I think that we should _always_ require the caller to provide sufficient ghost information. This creates more work for the user but makes logic a lot clearer and prevents unnecessary blocks in the multipart case.
> 
> I would like to apply this principle everywhere - what do others think?

Sounds good to me.

boyceg:
I just rebased off master to grab the updated test output.

drwells:
All the `IBFE/explicit_ex4` tests pass so this is presently working.

src/IB/IBFEMethod.cpp
boyceg: The way that I could see to use `LibMeshSystemVectors` seemed too complicated, and so I just changed this to use `buildIBGhostedVector` and to return a vector of `unique_ptr`s to the built vectors.

boyceg: (One issue is that functions associated with different parts could have different systems associated with them, and so it is not clear that `LibMeshSystemVectors` is the best way to organize things. The `LibMeshSystemVectors` interface is seems best suited for cases where all of the systems do more-or-less the same thing.)

boyceg:
if this looks OK I can squash

src/IB/IBFEMethod.cpp
drwells: Rats. Well, it was worth trying.

drwells:
This looks good. I would prefer to ultimately get all of the code that uses the IB partitioning out of IBFEMethod but that is far beyond the scope of this PR.

I the test suite passes on 3dafc8c0cd9cb200e058cfdc14fc9145116f69d2. Let me know if you change anything when you squash.

boyceg:
I merely squashed down to a single commit.

PR #972, opened 2020-04-16
Fix test output related to FEProjector.

drwells:
It looks like I managed to merge #968 without running the test suite - this causes a bunch of test failures for trivial reasons (we log the creation of the solver in a different class now). This is good motivation for us to get the CI running the test suite.

Since this PR only affects the test suite I suggest we merge asap.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
Since Jenkins won't ever read the files changed in this PR lets merge.

PR #976, opened 2020-04-26
Replace SAMRAI_MPI with IBTK_MPI

abarret:
This replaces all instances of `SAMRAI_MPI` with the equivalent `IBTK_MPI` function. This also modifies all the examples and tests to use `IBTKInit` to initialize all the libraries. There is a perl script in `scripts/code_updates` that should automatically do most of the changes except for initializing the library. I couldn't find an effective way to automatically set up appropriate `IBTKInit` calls since that could change based on what libraries are being used, but it is simple to do by hand.

I also added an `IBTK_MPI` test and fixed a couple of bugs in `IBTK_MPI`.

 This will be the first set of changes that require users to change their code, so I suggest we don't merge until closer to a release (perhaps 1.0?). I'm posting this now so that we can start a review and so we don't forget about it. Any merge conflicts (of which there will probably be many) can be handled easily. This is a required patch before we can proceed with SAMRAI 3 updates.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/private/IBTK_MPI-inl.h
+++ b/ibtk/include/ibtk/private/IBTK_MPI-inl.h
@@ -87,7 +87,7 @@ IBTK_MPI::bcast(const T x, const int root, IBTK_MPI::comm communicator)
     int size = 1;
     T temp_copy = x;
     bcast(&temp_copy, size, root, communicator);
-    return x;
+    return temp_copy;
 }
 
 template <typename T>
@@ -170,7 +170,7 @@ IBTK_MPI::minMaxReduction(T* x, const int n, int* rank, MPI_Op op, IBTK_MPI::com
     for (int i = 0; i < n; ++i)
     {
         x[i] = recv[i].first;
-        rank[i] = send[i].second;
+        rank[i] = recv[i].second;
     }
 } // minMaxReduction
 } // namespace IBTK
--- a/ibtk/src/utilities/IBTKInit.cpp
+++ b/ibtk/src/utilities/IBTKInit.cpp
@@ -53,7 +53,6 @@ IBTKInit::IBTKInit(int argc, char** argv, MPI_Comm communicator, char* petsc_fil
 
 IBTKInit::~IBTKInit()
 {
-    pout << "IBTKInit destructor called. Shutting down libraries.\n";
     SAMRAIManager::shutdown();
 #if SAMRAI_VERSION_MAJOR > 2
     SAMRAIManager::finalize();
--- a/doc/news/changes/incompatibilities/20200425AaronBarrett
+++ b/doc/news/changes/incompatibilities/20200425AaronBarrett
@@ -1,0 +1,8 @@
+New: IBAMR now requires the initialization object IBTKInit to be created upon
+program start. IBTKInit is responsible for initializing and deinitializing all
+libraries as well as registering the MPI communicator with IBTK_MPI. All
+SAMRAI_MPI calls in the library have been replaced with respective IBTK_MPI
+calls. The perl script scripts/code_updates/update_mpi_calls.pl will
+automatically complete most of the changes.
+<br>
+(Aaron Barrett, 2020/04/25)
--- a/examples/CIB/ex0/example.cpp
+++ b/examples/CIB/ex0/example.cpp
@@ -41,6 +41,7 @@
 #include <ibamr/app_namespaces.h>
 
 #include <ibtk/AppInitializer.h>
+#include <ibtk/IBTKInit.h>
 #include <ibtk/LData.h>
 #include <ibtk/LDataManager.h>
 #include <ibtk/muParserCartGridFunction.h>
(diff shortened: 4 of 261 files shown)


drwells:
It looks like we are missing a few places:

```

LSLocateStructureInterface.cpp: In member function 'double LSLocateStructureInterface::getMinimumWedgeCoord(SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps>)':
LSLocateStructureInterface.cpp:280:17: error: 'IBTK_MPI' has not been declared
     wedge_min = IBTK_MPI::minReduction(wedge_min);
                 ^
Makefile:675: recipe for target 'main2d-LSLocateStructureInterface.o' failed
make[3]: *** [main2d-LSLocateStructureInterface.o] Error 1
  CXX      main3d-LSLocateStructureInterface.o
  CXX      main3d-RigidBodyKinematics.o
  CXXLD    main3d
LSLocateStructureInterface.cpp: In member function 'double LSLocateStructureInterface::getMinimumWedgeCoord(SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps>)':
LSLocateStructureInterface.cpp:280:17: error: 'IBTK_MPI' has not been declared
     wedge_min = IBTK_MPI::minReduction(wedge_min);
                 ^
```

abarret:
@boyceg do I need to be on a UNC VPN to access the Jenkins server? I'm having trouble accessing the server.

drwells:
I'm not Boyce but the answer is yes :)

abarret:
> I'm not Boyce but the answer is yes :)

Well my onyen got disabled, so I can't access the server anymore. Is there someway we can post the build results that I can access? Otherwise you'll have to tell me what the errors are every time :)

drwells:
That's annoying. I think it might be easier for now to add you as an external collaborator (like Charles) so that you can still access the VPN.

drwells:
@abarret Since this patch is mostly search and replace you should be able to find all the problems by compiling all the examples locally. To avoid using incredible amounts of disk space I recommend doing this in release mode. Can you give that a try?

abarret:
Hmmm I though I had done that since implementing `IBTKInit` wasn't a search and replace, but I found a few compilation errors. I forced pushed a change.

drwells:
This looks fine and all the tests (except for the ones we just added) pass.

We don't have an explicit policy on it, but since the goal is to release the next version of IBAMR at the end of the month I would prefer to not include any large breaking changes (like this) at this point in the release cycle. Instead, I would much rather merge this after we create the release branch.

We might as well break the world all at once. Will get to SAMRAI 3 this year?

abarret:
>  I would much rather merge this after we create the release branch.
>
> We might as well break the world all at once. Will get to SAMRAI 3 this year?

Agreed. This largely depends on when we decide that it's a good time and when we will have time to review the massive number of changes.

I have time to do it this summer (what else am I gonna do?), but if we have a lot of outstanding PR's (e.g. IIM), then it might make sense to wait longer.

drwells:
That's a good point - we need to ping the community and see if anyone has outstanding changes long before switching to SAMRAI 3. That really means we should wait until early 2021 to throw the switch on SAMRAI 3 so that people have June - December to finish up whatever they want to get into IBAMR without needing to do the transition themselves.

boyceg:
I think that we should not hold off until 2021 to migrate to SAMRAI 3 unless we get a lot of push back. I would like to aim to do it this summer. If Aaron's scripts are really good, then any lagging branches or forks should be able to make the transition without too much pain. :-D

abarret:
> I think that we should not hold off until 2021 to migrate to SAMRAI 3 unless we get a lot of push back. I would like to aim to do it this summer. If Aaron's scripts are really good, then any lagging branches or forks should be able to make the transition without too much pain. :-D

I'm anticipating that it'll take me a week to get the entire library to compile (correctly??). A good number of changes are very easy to automate, assuming the code is written relatively cleanly. Although there are changes that I could not automate. In those cases, it should be easy to see what's wrong and how to fix it, unless people are writing their own `RefineOperator` or using `BoxList` or `BoxTree` where it gets more complicated.

drwells:
@abarret could you take a second look now that we are done with the release?

abarret:
@drwells I've fixed the conflicts.

ibtk/include/ibtk/private/IBTK_MPI-inl.h
```diff
@@ -87,7 +87,7 @@ IBTK_MPI::bcast(const T x, const int root, IBTK_MPI::comm communicator)
     int size = 1;
     T temp_copy = x;
     bcast(&temp_copy, size, root, communicator);
-    return x;
+    return temp_copy;
```
drwells, line 90: Interesting. I guess this was a bug.

ibtk/src/utilities/IBTKInit.cpp
```diff
@@ -53,7 +53,6 @@ IBTKInit::IBTKInit(int argc, char** argv, MPI_Comm communicator, char* petsc_fil
 
 IBTKInit::~IBTKInit()
 {
-    pout << "IBTKInit destructor called. Shutting down libraries.\n";
```
drwells, line 56: 👍

abarret:
It's worth noting that since this PR has been merged, user code will require a new initialization. I think the first failure they will get will be MPI complaining about a null communicator.

drwells:
Will it? I was able to revert an example and run it just fine.

abarret:
I think the MPI communicator defaults to a null communicator unless `IBTKInit` is constructed properly.

drwells:
Oh, I see - I made a mistake in my check.

Let me quickly put up another patch which changes that.

PR #977, opened 2020-04-27
Factor mechanics code from IBFEMethod into FEMechanicsBase

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/include/ibamr/FEMechanicsBase.h
+++ b/include/ibamr/FEMechanicsBase.h
@@ -1,0 +1,616 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2014 - 2019 by the IBAMR developers
+// All rights reserved.
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+#ifndef included_IBAMR_FEMechanicsBase
+#define included_IBAMR_FEMechanicsBase
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include "ibamr/ibamr_enums.h"
+
+#include "ibtk/FEDataManager.h"
+#include "ibtk/LibMeshSystemVectors.h"
+#include "ibtk/ibtk_utilities.h"
+#include "ibtk/libmesh_utilities.h"
+
+#include "tbox/Serializable.h"
+
+#include "libmesh/coupling_matrix.h"
+#include "libmesh/enum_fe_family.h"
+#include "libmesh/enum_order.h"
+#include "libmesh/enum_quadrature_type.h"
+#include "libmesh/explicit_system.h"
+
+#include <string>
+#include <utility>
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * @brief      Class FEMechanicsBase provides core finite element mechanics
+ *             functionality and data management.
+ */
+class FEMechanicsBase : public SAMRAI::tbox::Serializable
+{
+public:
+    // TODO: Update the names of these systems:
+    // COORDS_SYSTEM_NAME --> CURRENT_COORDINATES_SYSTEM_NAME
+    // COORD_MAPPING_SYSTEM_NAME --> DISPLACEMENT_SYSTEM_NAME
+    static const std::string COORDS_SYSTEM_NAME;
+    static const std::string COORD_MAPPING_SYSTEM_NAME;
+    static const std::string FORCE_SYSTEM_NAME;
+    static const std::string PRESSURE_SYSTEM_NAME;
+    static const std::string VELOCITY_SYSTEM_NAME;
+
+    /*!
+     * Constructor for a single-part model.
+     */
+    FEMechanicsBase(const std::string& object_name,
+                    const SAMRAI::tbox::Pointer<SAMRAI::tbox::Database>& input_db,
+                    libMesh::MeshBase* mesh,
+                    bool register_for_restart = true,
+                    const std::string& restart_read_dirname = "",
+                    unsigned int restart_restore_number = 0);
+
+    /*!
+     * Constructor for a multi-part model.
+     */
+    FEMechanicsBase(const std::string& object_name,
+                    const SAMRAI::tbox::Pointer<SAMRAI::tbox::Database>& input_db,
+                    const std::vector<libMesh::MeshBase*>& meshes,
+                    bool register_for_restart = true,
+                    const std::string& restart_read_dirname = "",
+                    unsigned int restart_restore_number = 0);
+
+    /*!
+     * Deleted default constructor.
+     */
+    FEMechanicsBase() = delete;
+
+    /*!
+     * Deleted copy constructor.
+     */
+    FEMechanicsBase(const FEMechanicsBase& from) = delete;
+
+    /*!
+     * Deleted assignment operator.
+     */
+    FEMechanicsBase& operator=(const FEMechanicsBase& that) = delete;
+
+    /*!
+     * Destructor.
+     */
+    virtual ~FEMechanicsBase() override;
+
+    /*!
+     * Return a pointer to the FEData object for the specified part.
+     */
+    std::shared_ptr<IBTK::FEData> getFEData(unsigned int part = 0) const;
+
+    /*!
+     * Typedef specifying interface for coordinate mapping function.
+     */
+    using CoordinateMappingFcnPtr = void (*)(libMesh::Point& x, const libMesh::Point& X, void* ctx);
+
+    /*!
+     * Struct encapsulating coordinate mapping function data.
+     */
+    struct CoordinateMappingFcnData
+    {
+        CoordinateMappingFcnData(CoordinateMappingFcnPtr fcn = nullptr, void* ctx = nullptr) : fcn(fcn), ctx(ctx)
+        {
+        }
+
+        CoordinateMappingFcnPtr fcn;
+        void* ctx;
+    };
+
+    /*!
+     * Register the (optional) function used to initialize the physical
+     * coordinates from the Lagrangian coordinates.
+     *
+     * @note       If no function is provided, the initial physical coordinates
+     *             are taken to be the same as the Lagrangian coordinate system,
+     *             i.e., the initial coordinate mapping is assumed to be the
+     *             identity mapping.
+     */
+    void registerInitialCoordinateMappingFunction(const CoordinateMappingFcnData& data, unsigned int part = 0);
+
+    /*!
+     * Get the initial coordinate mapping function data.
+     */
+    CoordinateMappingFcnData getInitialCoordinateMappingFunction(unsigned int part = 0) const;
+
+    /*!
+     * Typedef specifying interface for initial velocity specification function.
+     */
+    using InitialVelocityFcnPtr = void (*)(libMesh::VectorValue<double>& U0, const libMesh::Point& X0, void* ctx);
+
+    /*!
+     * Struct encapsulating initial velocity specification function data.
+     */
+    struct InitialVelocityFcnData
+    {
+        InitialVelocityFcnData(InitialVelocityFcnPtr fcn = nullptr, void* ctx = nullptr) : fcn(fcn), ctx(ctx)
+        {
+        }
+
+        InitialVelocityFcnPtr fcn;
+        void* ctx;
+    };
+
+    /*!
+     * Register the (optional) function used to initialize the velocity of the
+     * solid mesh.
+     *
+     * @note       If no function is provided, the initial velocity is taken to
+     *             be zero.
+     */
+    void registerInitialVelocityFunction(const InitialVelocityFcnData& data, unsigned int part = 0);
+
+    /*!
+     * Get the initial velocity function data.
+     */
+    InitialVelocityFcnData getInitialVelocityFunction(unsigned int part = 0) const;
+
+    /*!
+     * Typedef specifying interface for PK1 stress tensor function.
+     */
+    using PK1StressFcnPtr = IBTK::TensorMeshFcnPtr;
+
+    /*!
+     * Struct encapsulating PK1 stress tensor function data.
+     */
+    struct PK1StressFcnData
+    {
+        PK1StressFcnData(PK1StressFcnPtr fcn = nullptr,
+                         std::vector<IBTK::SystemData> system_data = {},
+                         void* const ctx = nullptr,
+                         const libMesh::QuadratureType& quad_type = libMesh::INVALID_Q_RULE,
+                         const libMesh::Order& quad_order = libMesh::INVALID_ORDER)
+            : fcn(fcn), system_data(std::move(system_data)), ctx(ctx), quad_type(quad_type), quad_order(quad_order)
+        {
+        }
+
+        PK1StressFcnPtr fcn;
+        std::vector<IBTK::SystemData> system_data;
+        void* ctx;
+        libMesh::QuadratureType quad_type;
+        libMesh::Order quad_order;
+    };
+
+    /*!
+     * Register the (optional) function to compute the first Piola-Kirchhoff
+     * stress tensor, used to compute the forces on the Lagrangian finite
+     * element mesh.
+     *
+     * @note       It is possible to register multiple PK1 stress functions with
+     *             this class.  This is intended to be used to implement
+     *             selective reduced integration.
+     */
+    void registerPK1StressFunction(const PK1StressFcnData& data, unsigned int part = 0);
+
+    /*!
+     * Get the PK1 stress function data.
+     */
+    std::vector<PK1StressFcnData> getPK1StressFunction(unsigned int part = 0) const;
+
+    /*!
+     * Typedef specifying interface for Lagrangian body force distribution
+     * function.
+     */
+    using LagBodyForceFcnPtr = IBTK::VectorMeshFcnPtr;
+
+    /*!
+     * Struct encapsulating Lagrangian body force distribution data.
+     */
+    struct LagBodyForceFcnData
+    {
+        LagBodyForceFcnData(LagBodyForceFcnPtr fcn = nullptr,
+                            std::vector<IBTK::SystemData> system_data = {},
+                            void* const ctx = nullptr)
+            : fcn(fcn), system_data(std::move(system_data)), ctx(ctx)
+        {
+        }
+
+        LagBodyForceFcnPtr fcn;
+        std::vector<IBTK::SystemData> system_data;
+        void* ctx;
+    };
+
+    /*!
+     * Register the (optional) function to compute body force distributions on
+     * the Lagrangian finite element mesh.
+     *
+     * @note       It is @em NOT possible to register multiple body force
+     *             functions with this class.
+     */
+    void registerLagBodyForceFunction(const LagBodyForceFcnData& data, unsigned int part = 0);
+
+    /*!
+     * Get the Lagrangian body force function data.
+     */
+    LagBodyForceFcnData getLagBodyForceFunction(unsigned int part = 0) const;
+
+    /*!
+     * Typedef specifying interface for Lagrangian pressure force distribution
+     * function.
+     */
+    using LagSurfacePressureFcnPtr = IBTK::ScalarSurfaceFcnPtr;
+
+    /*!
+     * Struct encapsulating Lagrangian surface pressure distribution data.
+     */
+    struct LagSurfacePressureFcnData
+    {
+        LagSurfacePressureFcnData(LagSurfacePressureFcnPtr fcn = nullptr,
+                                  std::vector<IBTK::SystemData> system_data = {},
+                                  void* const ctx = nullptr)
+            : fcn(fcn), system_data(std::move(system_data)), ctx(ctx)
+        {
+        }
+
+        LagSurfacePressureFcnPtr fcn;
+        std::vector<IBTK::SystemData> system_data;
+        void* ctx;
+    };
+
+    /*!
+     * Register the (optional) function to compute surface pressure
+     * distributions on the Lagrangian finite element mesh.
+     *
+     * @note       It is @em NOT possible to register multiple pressure
+     *             functions with this class.
+     */
+    void registerLagSurfacePressureFunction(const LagSurfacePressureFcnData& data, unsigned int part = 0);
+
+    /*!
+     * Get the Lagrangian surface pressure function data.
+     */
+    LagSurfacePressureFcnData getLagSurfacePressureFunction(unsigned int part = 0) const;
+
+    /*!
+     * Typedef specifying interface for Lagrangian surface force distribution
+     * function.
+     */
+    using LagSurfaceForceFcnPtr = IBTK::VectorSurfaceFcnPtr;
+
+    /*!
+     * Struct encapsulating Lagrangian surface force distribution data.
+     */
+    struct LagSurfaceForceFcnData
+    {
+        LagSurfaceForceFcnData(LagSurfaceForceFcnPtr fcn = nullptr,
+                               std::vector<IBTK::SystemData> system_data = {},
+                               void* const ctx = nullptr)
+            : fcn(fcn), system_data(std::move(system_data)), ctx(ctx)
+        {
+        }
+
+        LagSurfaceForceFcnPtr fcn;
+        std::vector<IBTK::SystemData> system_data;
+        void* ctx;
+    };
+
+    /*!
+     * Register the (optional) function to compute surface force distributions
+     * on the Lagrangian finite element mesh.
+     *
+     * @note       It is @em NOT possible to register multiple surface force
+     *             functions with this class.
+     */
+    void registerLagSurfaceForceFunction(const LagSurfaceForceFcnData& data, unsigned int part = 0);
+
+    /*!
+     * Get the Lagrangian surface force function data.
+     */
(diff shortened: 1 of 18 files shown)


boyceg:
@drwells, I think this is passing most of the tests that usually work for me on MacOS. The restart files have changed because of the refactoring, and so I think it will be necessary to update those tests.

boyceg:
Ah, the problem with restart is that I screwed up `FEMechanicsBase::putToDatabase()`. Should be fixed now...

include/ibamr/IBFEMethod.h
```diff
@@ -302,280 +302,72 @@ class IBFEDirectForcingKinematics;
...
     SAMRAI::tbox::Pointer<SAMRAI::pdat::SideVariable<NDIM, double> > mask_var;
     int mask_current_idx, mask_new_idx, mask_scratch_idx;
 
     /*!
-     * \brief Constructor.
+     * \brief Constructor for a single-part model.
      */
     IBFEMethod(const std::string& object_name,
-               SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db,
+               const SAMRAI::tbox::Pointer<SAMRAI::tbox::Database>& input_db,
```
drwells, line 323: :tada:

include/ibamr/FEMechanicsBase.h
```diff
@@ -0,0 +1,606 @@
...
+    std::vector<std::unique_ptr<libMesh::EquationSystems> > d_equation_systems;
+
+    /// FEData objects provide key FE data management.
+    std::vector<std::shared_ptr<IBTK::FEData> > d_fe_data;
+
+    /// FEProjector objects provide L2 projection functionality.
+    std::vector<std::shared_ptr<IBTK::FEProjector> > d_fe_projectors;
+
+    /// Number of parts owned by the present object.
+    const unsigned int d_num_parts = 1;
```
drwells, line 462: This didn't make any sense in IBFEMethod either so we should remove it.

include/ibamr/IBFEMethod.h
```diff
@@ -636,8 +428,8 @@ class IBFEMethod : public IBStrategy
      * @note Since this class assumes that structures live on the finest grid
      * level the second argument is ignored.
      */
-    virtual void inactivateLagrangianStructure(int structure_number = 0,
-                                               int level_number = std::numeric_limits<int>::max()) override;
+    void inactivateLagrangianStructure(int structure_number = 0,
+                                       int level_number = std::numeric_limits<int>::max()) override;
```
drwells, line 432: Can we move this and related functions into `FEMechanicsBase`? Since that class already stores data structures per part, it makes more sense that these should be there.

src/IB/IBFEMethod.cpp
```diff
@@ -885,8 +693,7 @@ IBFEMethod::backwardEulerStep(const double current_time, const double new_time)
             TBOX_ERROR("IBFEDirectForcingKinematics does not implement backwardEulerStep().\n");
         }
     }
-    return;
-} // backwardEulerStep
+}
```
drwells, line 696: Could we please avoid these types of code updates in this PR? I like this change but its making this patch bigger than it needs to be.

include/ibamr/IBFEMethod.h
```diff
@@ -1162,47 +866,45 @@ class IBFEMethod : public IBStrategy
...
 
     /*!
-     * Vectors of pointers to the fluid source or sink density vectors. All of
-     * these vectors are owned by the libMesh::System objects except for the
-     * ones in d_Q_IB_ghost_vecs, which are owned by the FEDataManager
-     * objects.
+     * Convenience pointer for object managing access to libMesh system vectors
+     * for the force.
      */
-    std::vector<libMesh::PetscVector<double>*> d_Q_half_vecs, d_Q_rhs_vecs, d_Q_IB_ghost_vecs;
```
drwells, line 1190: I deliberately left all the fluid source/sink vectors alone since we don't have any tests - are you sure that we should proceed with this change at this time? I would prefer to handle this later.

boyceg:
Hmmm, I thought that tests were working. I will take another look later — but could be a while — I need to bake a birthday cake and get started on a grant tonight. :-D

src/IB/IBFEMethod.cpp
boyceg: I knew you would say this, but let me tell you why I went ahead and did it: I am now running `clang-tidy` interactively, and getting rid of these `return` statements makes `clang-tidy` happy. (And I figured if I am wiping out the unneeded `return` I also might as well get rid of the unneeded comment, which dates back to days when I did not have good block matching tools in my editor.)

include/ibamr/IBFEMethod.h
boyceg: Yes let's do it now.

include/ibamr/IBFEMethod.h
boyceg: I can move the data structure, but `inactivateLagrangianStructure` and related functions are in `IBStrategy`, so we cannot completely get rid of them in `IBFEMethod`.

boyceg:
Well, I got the cake baked. I think I also fixed most of the problems that were causing tests to fail, and reverted some of the "improvements" that were getting in the way of review.

@drwells can you take another look?

drwells:
I'm on it!

drwells:
The tests now pass, so that's a good start. I'm reading more carefully now.

include/ibamr/IBFEMethod.h
drwells: Oh duh. Lets definitely not move this into `FEMechanicsBase` - its too complicated to think about which parent class has the right virtual function definition.

src/IB/IBFEMethod.cpp
```diff
@@ -354,24 +343,50 @@ build_ib_ghosted_system_data(std::vector<SystemData>& ghosted_system_data,
...
+    system.remove_vector(vector_name);
+    system.add_vector(vector_name, /*projections*/ true, /*type*/ GHOSTED);
+
+    if (clone_vector != nullptr)
+    {
+        const auto& parallel_vector = dynamic_cast<const PetscVector<Number>&>(*clone_vector);
+        auto& ghosted_vector = dynamic_cast<PetscVector<Number>&>(system.get_vector(vector_name));
+        insert_parallel_into_ghosted(parallel_vector, ghosted_vector);
+    }
+};
```
drwells, line 374: AFAICT we shouldn't need to use these in IBFEMethod any more. If we do then we should promote them to protected static functions inside `FEMechanicsBase` so that we don't need to duplicate code.

src/IB/IBFEMethod.cpp
```diff
@@ -402,163 +417,71 @@ IBFEMethod::IBFEMethod(const std::string& object_name,
...
-{
-    TBOX_ASSERT(part < d_num_parts);
-    d_PK1_stress_fcn_data[part].push_back(data);
-    if (data.quad_type == INVALID_Q_RULE)
+    // Setup cached system vectors.
+    const std::vector<std::string> vector_names{ "half", "tmp", "RHS Vector" };
+    for (const std::string& vector_name : vector_names)
     {
-        d_PK1_stress_fcn_data[part].back().quad_type = d_default_quad_type[part];
+        convert_parallel_to_ghosted(vector_name, Phi_system);
```
drwells, line 445: Could we move this into the restart data load section somehow instead? IIRC this problem only occurs when we load vectors, not when we create them.

src/IB/IBFEMethod.cpp
```diff
@@ -402,163 +417,71 @@ IBFEMethod::IBFEMethod(const std::string& object_name,
...
-    return d_lag_surface_pressure_fcn_data[part];
-} // getLagSurfacePressureFunction
-
-void
-IBFEMethod::registerLagSurfaceForceFunction(const LagSurfaceForceFcnData& data, const unsigned int part)
-{
-    TBOX_ASSERT(part < d_num_parts);
-    d_lag_surface_force_fcn_data[part] = data;
+    // libMesh also stores a convenience pointer to the RHS that we need to reset:
+    Phi_system.rhs = &Phi_system.get_vector("RHS Vector");
```
drwells, line 448: same

src/IB/IBFEMethod.cpp
```diff
@@ -402,163 +417,71 @@ IBFEMethod::IBFEMethod(const std::string& object_name,
...
     d_lag_body_source_part[part] = true;
     d_lag_body_source_fcn_data[part] = data;
     auto& Q_system = d_equation_systems[part]->add_system<ExplicitSystem>(SOURCE_SYSTEM_NAME);
     // This system has a single variable so we don't need to also specify diagonal coupling
     Q_system.add_variable("Q", d_fe_order[part], d_fe_family[part]);
+    // Setup cached system vectors.
+    const std::vector<std::string> vector_names{ "half", "tmp", "RHS Vector" };
+    for (const std::string& vector_name : vector_names)
+    {
+        convert_parallel_to_ghosted(vector_name, Q_system);
```
drwells, line 467: same

src/IB/IBFEMethod.cpp
```diff
@@ -693,50 +568,36 @@ IBFEMethod::preprocessIntegrateData(double current_time, double new_time, int nu
...
         if (d_direct_forcing_kinematics_data[part])
         {
             d_direct_forcing_kinematics_data[part]->postprocessIntegrateData(current_time, new_time, num_cycles);
         }
+    }
 
-        // Update the coordinate mapping dX = X - s.
+    // Update the coordinate mapping dX = X - s.
+    for (unsigned part = 0; part < d_meshes.size(); ++part)
+    {
```
drwells, line 597: This doesn't seem right - we also call updateCoordinateMapping in `FEMechanicsBase::postprocessIntegrateData()`.

src/IB/IBFEMethod.cpp
```diff
@@ -1393,37 +1258,8 @@ IBFEMethod::initializeFEEquationSystems()
...
-                const auto& parallel_vector = dynamic_cast<const PetscVector<Number>&>(*clone_vector);
-                auto& ghosted_vector = dynamic_cast<PetscVector<Number>&>(system.get_vector(vector_name));
-                insert_parallel_into_ghosted(parallel_vector, ghosted_vector);
-            }
-        };
-
-        const std::array<std::string, 2> system_names{ { COORDS_SYSTEM_NAME, VELOCITY_SYSTEM_NAME } };
-        const std::array<std::string, 2> vector_names{ { "new", "half" } };
+        const std::vector<std::string> system_names{ COORDS_SYSTEM_NAME, VELOCITY_SYSTEM_NAME };
+        const std::vector<std::string> vector_names{ "new", "half" };
```
drwells, line 1262: Isn't all this stuff done in `FEMechanicsBase` now?

src/IB/IBFEMethod.cpp
```diff
@@ -3638,24 +2872,32 @@ IBFEMethod::commonConstructor(const std::string& object_name,
...
         }
         if (mesh_has_second_order_elems)
         {
             d_fe_order[part] = SECOND;
+            d_fe_order_position[part] = SECOND;
+            d_fe_order_force[part] = SECOND;
+            d_fe_order_pressure[part] = SECOND;
             d_default_quad_order[part] = FIFTH;
+            d_default_quad_order_stress[part] = FIFTH;
+            d_default_quad_order_force[part] = FIFTH;
```
drwells, line 2899: Shouldn't all of this also be in the base class now?

src/IB/IBFEMethod.cpp
boyceg: In the initial implementation, `IBFEMethod::pre/postprocessIntegrateData()` did not call the base class implementation.

If you prefer to leverage the base class implementation, that's OK with me, but then I think the base class should not do any data allocation --- for instance, because subclasses could be using different time stepping schemes and need different system vector instances. (This already happens even within `IBFEMethod` depending on whether one uses midpoint or trapezoidal rule, for instance.)

include/ibamr/IBFEMethod.h
```diff
@@ -1162,47 +862,45 @@ class IBFEMethod : public IBStrategy
...
-    std::unique_ptr<IBTK::LibMeshSystemIBVectors> d_X_vecs;
+    std::unique_ptr<IBTK::LibMeshSystemVectors> d_Q_vecs;
 
     /*!
-     * Object managing access to libMesh system vectors for the structure velocity.
+     * Convenience pointer for object managing access to libMesh system vectors
+     * for the position.
      */
-    std::unique_ptr<IBTK::LibMeshSystemIBVectors> d_U_vecs;
+    IBTK::LibMeshSystemIBVectors* d_X_IB_vecs;
```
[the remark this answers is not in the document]
boyceg: I just added extra `LibMeshSystemIBVectors` objects to the derived class. Seems easier than creating an interface --- maybe we can revisit after there are additional implementations.

src/IB/IBFEMethod.cpp
boyceg: These calls are needed to setup all of the vectors used in the different systems. I've renamed the key function `setup_system_vectors` and moved it into the base class as a static member function.

src/IB/IBFEMethod.cpp
boyceg: See above.

src/IB/IBFEMethod.cpp
boyceg: See above.

src/IB/IBFEMethod.cpp
boyceg: See above.

src/IB/IBFEMethod.cpp
boyceg: OK, so the base class now does essentially no data initialization or allocation. Once we have multiple implementations of the base class, perhaps we can think about whether it makes sense to factor some of that into the base class.

src/IB/IBFEMethod.cpp
boyceg: Nope; see above.

src/IB/IBFEMethod.cpp
boyceg: Dropped.

boyceg:
@drwells, I think this now addresses most of your comments. I need to re-run the test suite and add a couple tests.

boyceg:
Looks like the usual MacOS suspects are failing:
```
The following tests FAILED:

    IBFE/explicit_ex4_2d.scratch_hier.mpirun=4.input
    IBFE/explicit_ex4_2d.postprocessor.mpirun=4.input
    IBFE/explicit_ex4_2d.scratch_hier.restart=25.mpirun=4.input
    IBFE/explicit_ex4_2d.inactive_1.mpirun=4.input
    IBFE/explicit_ex4_2d.inactive_1.mpirun=4.restart=50.input
    IBFE/explicit_ex4_2d.mpirun=4.input
    IBFE/explicit_ex1_2d.mpirun=4.input
    IBFE/explicit_ex4_2d.postprocessor.scratch_hier.mpirun=4.input
    IBFE/explicit_ex5_2d.regrid.mpirun=2.input
    IBFE/explicit_ex0_2d.input
```

boyceg:
Ah, `IBFE/explicit_ex1_2d.mpirun=4.input` and `IBFE/explicit_ex0_2d.input` should not have been failing for me --- found a bug; now fixed. Others it is hard to know, as they also fail for me with master.

drwells:
IBFE/explicit_ex4_2d.source_test.input fails for me but the rest of the tests currently pass.

boyceg:
> IBFE/explicit_ex4_2d.source_test.input fails for me but the rest of the tests currently pass.

Fails catastrophically or just gives different results?

drwells:
it fails at the diff stage.

boyceg:
Oh, it is because I named the input/output files incorrectly. Can you take another look?

include/ibamr/FEMechanicsBase.h
```diff
@@ -0,0 +1,616 @@
...
+     * file in restart_dump_dirname for each FE part. An example snippet is
+     * included below to show the distinct FE restart data saving step. The data
+     * will then be automatically read back into the system along with the
+     * RestartManager data during restart.
+     *
+     * @code
+     * if (dump_restart_data && (iteration_num % restart_dump_interval == 0 || last_step))
+     * {
+     *     RestartManager::getManager()->writeRestartFile(restart_dump_dirname, iteration_num);
+     *     ib_method_ops->writeFEDataToRestartFile(restart_dump_dirname, iteration_num);
```
drwells, line 367:
```suggestion
     *     fe_mechanics->writeFEDataToRestartFile(restart_dump_dirname, iteration_num);
```

src/IB/IBFEMethod.cpp
```diff
@@ -1278,1241 +1107,739 @@ void
...
-                EquationSystems& equation_systems = *d_active_fe_data_managers[part]->getEquationSystems();
-                MeshBase& mesh = equation_systems.get_mesh();
-                BoxPartitioner partitioner(*d_hierarchy, equation_systems.get_system(COORDS_SYSTEM_NAME));
-                partitioner.repartition(mesh);
-            }
-        }
+    // Set up the coupling matrix which will be used by each system.
+    d_diagonal_system_coupling.resize(NDIM);
+    for (unsigned int i = 0; i < NDIM; ++i)
+        for (unsigned int j = 0; j < NDIM; ++j) d_diagonal_system_coupling(i, j) = i == j ? 1 : 0;
```
drwells, line 1558: `d_diagonal_system_coupling` is a member of the base class, so we should set it up inside a base class member function.

src/IB/IBFEMethod.cpp
```diff
@@ -1278,1241 +1107,739 @@ void
...
 
+    // Create the FE data managers that manage mappings between the FE mesh
+    // parts and the Cartesian grid.
+    d_equation_systems.resize(d_meshes.size());
+    d_primary_fe_data_managers.resize(d_meshes.size(), nullptr);
+    d_scratch_fe_data_managers.resize(d_meshes.size(), nullptr);
+    d_active_fe_data_managers.resize(d_meshes.size(), nullptr);
+    d_fe_data.resize(d_meshes.size(), nullptr);
```
drwells, line 1566: we don't need the `nullptr`s

src/IB/IBFEMethod.cpp
```diff
@@ -1278,1241 +1107,739 @@ void
...
-
-    // First handle the stress contributions.  These are handled separately
-    // because each stress function may use a different quadrature rule.
-    const size_t num_PK1_fcns = d_PK1_stress_fcn_data[part].size();
-    for (unsigned int k = 0; k < num_PK1_fcns; ++k)
+    // The choice of FEDataManager set is important here since it determines the
+    // IB ghost regions.
+    d_X_vecs.reset(new LibMeshSystemIBVectors(d_active_fe_data_managers, COORDS_SYSTEM_NAME));
+    d_U_vecs.reset(new LibMeshSystemIBVectors(d_active_fe_data_managers, VELOCITY_SYSTEM_NAME));
+    d_F_vecs.reset(new LibMeshSystemIBVectors(d_active_fe_data_managers, FORCE_SYSTEM_NAME));
```
drwells, line 1679: We don't need to do this reset now.

src/IB/IBFEMethod.cpp
```diff
@@ -2521,152 +1848,100 @@ IBFEMethod::assembleInteriorForceDensityRHS(PetscVector<double>& G_rhs_vec,
...
             {
+                // X:     reference coordinate
+                // x:     current coordinate
+                // FF:    deformation gradient associated with x = chi(X,t)
+                //        (FF = dchi/dX)
+                // J:     Jacobian determinant (J = det(FF))
+                // N:     unit normal in the reference configuration
+                // n:     unit normal in the current configuration
+                // dA_da: reference surface area per current surface area (from Nanson's
+                //        relation)
```
drwells, line 1911: :heavy_check_mark:

include/ibamr/FEMechanicsBase.h
```diff
@@ -0,0 +1,616 @@
...
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * @brief      Class FEMechanicsBase provides core finite element mechanics
+ *             functionality and data management.
+ */
+class FEMechanicsBase : public SAMRAI::tbox::Serializable
```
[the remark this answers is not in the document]
boyceg: You tell me. :-)

tests/IBFE/explicit_ex4_source_test.cpp
```diff
@@ -0,0 +1,612 @@
+// ---------------------------------------------------------------------
```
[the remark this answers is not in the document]
boyceg: Will do.

tests/IBFE/explicit_ex4_2d_source_test.output
```diff
@@ -0,0 +1,300 @@
+
+IBFEMethod: mesh part 0 is using SECOND order LAGRANGE finite elements.
+
+IBHierarchyIntegrator::initializePatchHierarchy(): tag_buffer = 0
+INSStaggeredHierarchyIntegrator::initializeCompositeHierarchyData():
+  projecting the interpolated velocity field
+INSStaggeredHierarchyIntegrator::regridProjection(): regrid projection solve number of iterations = 0
+INSStaggeredHierarchyIntegrator::regridProjection(): regrid projection solve residual norm        = 0
+Total number of elems: 309
```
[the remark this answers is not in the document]
boyceg: Interesting/horrifying.

include/ibamr/FEMechanicsBase.h
boyceg: 👍

src/IB/IBFEMethod.cpp
boyceg: 👍

src/IB/IBFEMethod.cpp
boyceg: 👍 

I removed some others while I was at it...

tests/IBFE/explicit_ex4_source_test.cpp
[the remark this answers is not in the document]
boyceg: OK I added an option, but it also required some other changes to the example code --- not sure if that impacts outputs for other tests. Can you take another look?

boyceg:
I think I have addressed requested changes.

tests/IBFE/explicit_ex4_2d_source_test.output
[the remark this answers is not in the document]
boyceg: This is setup to use `QUAD` elements; here's the logic to setup the mesh:
```
// NOTE: number of segments along boundary is 4*2^r.
            const double num_circum_segments = 2.0 * M_PI * R / ds;
            const int r = log2(0.25 * num_circum_segments);
            MeshTools::Generation::build_sphere(mesh, R, r, Utility::string_to_enum<ElemType>(elem_type));
```
FWIW, here is what is done for `TRI` elements:
```
            const int num_circum_nodes = ceil(2.0 * M_PI * R / ds);
            for (int k = 0; k < num_circum_nodes; ++k)
            {
                const double theta = 2.0 * M_PI * static_cast<double>(k) / static_cast<double>(num_circum_nodes);
                mesh.add_point(libMesh::Point(R * cos(theta), R * sin(theta)));
            }
            TriangleInterface triangle(mesh);
            triangle.triangulation_type() = TriangleInterface::GENERATE_CONVEX_HULL;
            triangle.elem_type() = Utility::string_to_enum<ElemType>(elem_type);
            triangle.desired_area() = 1.5 * sqrt(3.0) / 4.0 * ds * ds;
            triangle.insert_extra_points() = true;
            triangle.smooth_after_generating() = true;
            triangle.triangulate();
```
I think that most of the discrepancies are for `ex4` and `ex5`. I wonder if the root of the problem is that we are not getting consistent meshes?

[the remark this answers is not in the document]
drwells: I think it might be due to the use of triangle - e.g., #876 had related problems that I could trace back to libMesh's patches to triangle.

boyceg:
Should we remove Triangle from all tests?

drwells:
I think we already have at this point. However, I'm not completely convinced the failure is due to triangle - the change libMesh made only affected node numberings, not the actual mesh geometry.

boyceg:
It looks like `explicit_ex4.cpp` and `explicit_ex5.cpp` both use `TriangleInterface` to generate triangular meshes, and they both have input files that specify `TRI3` or `TRI6` elements.

drwells:
Hm. It must not be triangle, then.

drwells:
We should still get rid of it to get more consistent output for the same reasons specified in #876. Let me open an issue for that.

boyceg:
Aren't you getting different numbers of elements than me? Couldn't that be because ... Triangle is generating different meshes for us?

drwells:
Maybe. I am not completely convinced yet that Triangle is producing substantially different output since, IIRC, most ex4 tests do currently pass on macs. There must be a different explanation.

boyceg:
@drwells, if the only issue is the failing source test, how about you run it in Linux and update the output, and then we can worry about it later when trying to get the tests working on MacOS? I would like to get this merged so I can setup the FE mechanics solver.

drwells:
Sure.

include/ibamr/FEMechanicsBase.h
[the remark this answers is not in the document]
drwells: I think the current version is fine. In particular, since restarts still work, what we have suffices.

drwells:
I force pushed a change to the test output file.

boyceg:
Command line git diff should be better.

drwells:
You're right - `diff -y` makes this a lot easier to read.

I think this is fine so lets merge.

PR #979, opened 2020-04-30
Fix a busy wait problem in attest.

drwells:
This prevents us from running python at 100% when we have more available processors than work. This can happen when, for example, we run

    ./attest -j6 -R 'mpirun=4'

here we will never fully utilize the number of allocated cores, so the loop will endlessly attempt to schedule more jobs without waiting on the lone worker to finish first.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/attest
+++ b/attest
@@ -452,6 +452,7 @@ def main():
     # 2. Run the tests!
     processors_in_use = 0
     n_tests = len(unstarted_tests)
+    n_executing_tests = 0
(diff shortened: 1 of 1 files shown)


drwells:
Lets see if the CI is happy again.

boyceg:
It is!

PR #981, opened 2020-04-30
Fix the merging algorithm used by MergingLoadBalancer.

drwells:
We get different partitions with the libc++ version of std::sort() than the
libstdc++ version of std::sort(), so just use std::stable_sort() to get the same
thing in both cases.

Note that this is a problem with the original algorithm: since we sort by magnitudes many entries have (as far as the sorting algorithm can see) the same value. Hence two different unstable sorting algorithms yield two different results.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
Probably `std:sort()` in `LDataManager` could also be converted to `std::stable_sort()` for the same reason.

drwells:
The only two sort calls I see in that class both sort things based on indices - why would that require a stable sort?

boyceg:
I think that in some places we are actually sorting by position under the hood.

PR #982, opened 2020-05-01
Fixed namespace issue for Vector in CIBMethod

benlvr:
Added IBTK namespace to a call to Vector in CIBMethod that was leading to compilation errors on the UNC longleaf system when using gcc 4.8.5 compilers.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/CIBMethod.cpp
+++ b/src/IB/CIBMethod.cpp
@@ -593,7 +593,7 @@ CIBMethod::initializePatchHierarchy(Pointer<PatchHierarchy<NDIM> > hierarchy,
         for (const auto& node_idx : local_nodes)
         {
             const int local_idx = node_idx->getLocalPETScIndex();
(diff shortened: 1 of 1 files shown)


abarret:
@drwells At some point we talked about namespacing all `Index` calls (did we do that?). I'm confused about why this namespace issue exists, but perhaps we should include `Vector` as well?

drwells:
This namespace issue exists because we have `ibamr/app_namespaces.h`, which imports all namespaces. We fixed it for `Index` in #762.

What other class is conflicting with `Vector`?

abarret:
I don't know. I couldn't find one. My only thought was that somehow this interfered with one of `Eigen`'s `typedef` since `CIBMethod` has other `Eigen` includes. I'm confused because it doesn't show up in the rest of the library.

PR #983, opened 2020-05-02
do not use constrain_element_vector() with too many DOFs

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
OK, here's the problem --- `DofMap::constrain_element_XXX()` is designed to be used for numbers of DOFs that occur on individual elements, but I wrote `FEDataManager::interpWeighted()` to call it on, potentially, all of the DOFs in the mesh when using nodal interaction. Whoops.

boyceg:
In particular, `DofMap::constrain_element_vector()` will create an n-by-n `DenseMatrix` in which n is set to the total number of DOF indices to check for constraints.

I think it should be fine to use `DofMap::enforce_constraints_exactly()` instead, and so that is what this PR does instead of calling `DofMap::constrain_element_vector()`.

The only thing I am not sure about is whether we can avoid calling `close()` on `F_vec`...

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1847,6 +1844,14 @@ FEDataManager::interpWeighted(const int f_data_idx,
     // Accumulate data.
     if (close_F) F_vec.close();
 
+    // Enforce constraints here if using nodal quadrature.  (If using
+    // element-based quadrature, constraints have already been imposed above.)
+    if (use_nodal_quadrature && F_dof_map.n_constrained_dofs() > 0)
+    {
+        if (!close_F) F_vec.close();
+        F_dof_map.enforce_constraints_exactly(F_system, &F_vec);
```
drwells, line 1852: This doesn't seem right - here F_vec is a RHS vector, not a solution vector. We should be performing assembly in the same way in both the nodal and nonnodal case.

drwells:
The tests fail in the diff stage for me - there are substantial (6th decimal place) differences in the computed volumes.

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: No --- nodal interpolation is just direct interpolation at the nodes --- it is nothing like the elemental case. We sort of fake it to be like the elemental case by multiplying by the diagonal mass matrix.

boyceg: Maybe we need to impose the constraints on the assembled interpolated vector and then multiply by the diagonal mass matrix.

boyceg:
> The tests fail in the diff stage for me - there are substantial (6th decimal place) differences in the computed volumes.

Because I cannot reliably run the test suite, I will need your help in fixing this. :-D

drwells:
As discussed elsewhere: the root cause of the crash is the application of constraints when we use nodal quadrature. Since the degrees of freedom are uncoupled in this case we can completely skip any imposition of constraints until after the solution process, which is what this patch now does.

PR #984, opened 2020-05-04
Fix attest's busy wait again.

drwells:
The old check was too imprecise: if we are executing a single test we may still
be able to run more tests at the current time. Instead of waiting if we are only
running one test, instead wait if we couldn't launch any new tests.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
This saga is a good example of why threads are really hard to get right. We do need them, however, for `attest`.

PR #985, opened 2020-05-04
Make the ex5 test more robust.

drwells:
As previously noted, these values depend on the version of libMesh. The
test where we regrid a lot fails with some configurations - get around
that failure by making the whole test more robust.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


tests/IBFE/explicit_ex5.cpp
```diff
@@ -732,7 +732,7 @@ postprocess_data(Pointer<Database> input_db,
         // This computed value is not stable between debug and optimized mode
         // (and also different versions of libMesh): make it slightly more
         // stable by increasing the order of magnitude
-        pout << loop_time << " " << 0.01 + -F_integral[1] / (0.5 * rho * U_max * U_max * D) << endl;
+        pout << loop_time << " " << 0.01 + std::abs(F_integral[1] / (0.5 * rho * U_max * U_max * D)) << endl;
```
boyceg, line 735: Why is this the right thing to do?

drwells: In general? I don't remember the details but changing the version of libMesh gave us results different enough to upset numdiff.

In this case? The value goes negative in the newish ex5 regridding test. Since the value passes near zero this old trick of adding a value to it isn't enough to get consistent output.

boyceg: You could just look at `F_integral[0]` --- that should be nonnegative.

drwells: That works.

drwells:
This should be good to go now - the ex5 tests all pass with several libMesh configurations.

tests/IBFE/explicit_ex5.cpp
```diff
@@ -732,7 +732,7 @@ postprocess_data(Pointer<Database> input_db,
         // This computed value is not stable between debug and optimized mode
         // (and also different versions of libMesh): make it slightly more
         // stable by increasing the order of magnitude
-        pout << loop_time << " " << 0.01 + -F_integral[1] / (0.5 * rho * U_max * U_max * D) << endl;
+        pout << loop_time << " " << F_integral[1] << endl;
```
boyceg, line 735: Is the comment still correct?

boyceg: Maybe also use the dimensional expression for `F_integral[0]` too?

drwells: No (I will fix it) and good idea - I'll change that as well.

drwells: The extra factors for the first term were equivalent to multiplying by 0.5, which in turn multiplied the error by 0.5 - as a result things now pass that once failed. I'm working on it.

drwells: Fixed!

PR #994, opened 2020-05-19
Fe values codim

drwells:
This PR actually uses the new `FEValues` class - it should make IBFE programs run about 5-10% faster (based on how large their structures are).

I also added a horrible test based on a klein bottle to really check that we get the same values as libMesh. Here's the mesh:

klein

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
Part of #482.

boyceg:
There are a number of places where we still us `FEBase` directly --- are any of those candidate for replacement by `FEValues`?

drwells:
Yes, but this replacement class is still not full-featured (e.g., we do not yet know how to compute normal vectors).

I started here since `FEDataInterpolation` takes up a huge amount of our total runtime.

drwells:
Part of why I went to all this effort to add a tricky test was to make it easier in the future to implement normal vectors and be able to show that we get the same thing as libMesh.

PR #997, opened 2020-05-22
Revert "Use FEValues in FEDataInterpolation."

drwells:
This reverts commit 5dba10c26cfe55027d47cbebee88d63d5d0ff0f1.

This feature passes the test suite but causes problems in user codes
that we don't currently have time to address.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/major/20200519DavidWells
+++ b/doc/news/changes/major/20200519DavidWells
@@ -1,7 +1,0 @@
-New: Added a new class IBTK::FEValues, which is a replacement for libMesh::FE
-for Lagrange elements that is 4-5 times faster. This is now used inside various
-parts of the IBFE stack and makes, in particular, assembling the interior force
-density much less expensive. For some models this leads to a 10% overall run
-time reduction.
-<br>
-(David Wells, 2020/05/19)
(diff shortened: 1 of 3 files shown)


drwells:
@jabrown893 @marshallrdavey could you take a quick look, before we merge, and make sure that this fixes your problems?

Also, @jabrown893: I still want to meet to try and fix the bug that occurred with this patch so that I at least know what went wrong.

jabrown893:
Which bug is this supposed to fix? The issues with using nodal quadrature or the random seg faults?

drwells:
The issue with nodal interpolation (where things fail in a weird way inside FEDataInterpolation).

drwells:
This fix fixes @jabrown893's case with nodal quadrature, so lets merge.

drwells:
We need to double-check that previous statement - hold on.

jabrown893:
Yes, this does fix one of the errors I was experiencing with nodal quadrature (the boost error)

drwells:
@marshallrdavey that should fix the problem you reported - let me know if it does not.

PR #1001, opened 2020-05-28
Fix edge case in HR limiters

nishantnangia:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

This PR fixes a small bug in an edge case in the HR limiters used for the CUI, MGAMMA, and FBICS convective operator. I verified that this fix does not significantly change the order of accuracy of these operators, but it does change the output of two of the tests (which I went ahead and modified so that they pass).


amneetb:
This is an edge case in which far-upwind and downwind values are the same, but there in an extrema in between. One needs to do upwinding in this case; earlier the code was doing far-upwinding, now it is doing near-node upwinding, which tackles very coarse mesh cases.

abarret:
@nishantnangia Can you squash these commits so that the tests will pass on the commit with the bug fix?

nishantnangia:
@abarret Good point -- should be squashed now!

PR #1002, opened 2020-05-29
Fix the scratch hierarchy with more than two levels.

drwells:
The fix here is easy - we only want to regrid one level (the finest) so we
should not hard-code zero as the coarser level in the call to
regridAllFinerLevels.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1362,8 +1362,15 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
             // TODO: d_current_time is actually nan here. Since we don't do
             // any sort of tagging based on the current time I think we can
             // just put in a bogus (nonnegative) value.
+            int max_level = d_scratch_hierarchy->getFinestLevelNumber();
+            if (max_level == 0)
+            {
+                TBOX_ERROR("IBFEMethod::endDataRedistribution()\n"
+                           << "At the current time the scratch hierarchy is only available\n"
+                           << "with AMR (i.e., there must be at least two levels.)\n");
+            }
             d_scratch_gridding_algorithm->regridAllFinerLevels(
-                d_scratch_hierarchy, 0, 0.0 /*d_current_time*/, tag_buffer);
+                d_scratch_hierarchy, max_level - 1, 0.0 /*d_current_time*/, tag_buffer);
             if (d_do_log) plog << "IBFEMethod: finished scratch hierarchy regrid" << std::endl;
         }
 
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1364,13 +1364,10 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
             // just put in a bogus (nonnegative) value.
             int max_level = d_scratch_hierarchy->getFinestLevelNumber();
             if (max_level == 0)
-            {
-                TBOX_ERROR("IBFEMethod::endDataRedistribution()\n"
-                           << "At the current time the scratch hierarchy is only available\n"
-                           << "with AMR (i.e., there must be at least two levels.)\n");
-            }
-            d_scratch_gridding_algorithm->regridAllFinerLevels(
-                d_scratch_hierarchy, max_level - 1, 0.0 /*d_current_time*/, tag_buffer);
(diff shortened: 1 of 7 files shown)


src/IB/IBFEMethod.cpp
```diff
@@ -1362,8 +1362,15 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
...
+            int max_level = d_scratch_hierarchy->getFinestLevelNumber();
+            if (max_level == 0)
+              {
+                TBOX_ERROR("IBFEMethod::endDataRedistribution()\n"
+                           << "At the current time the scratch hierarchy is only available\n"
+                           << "with AMR (i.e., there must be at least two levels.)\n");
+              }
             d_scratch_gridding_algorithm->regridAllFinerLevels(
-                d_scratch_hierarchy, 0, 0.0 /*d_current_time*/, tag_buffer);
+                d_scratch_hierarchy, max_level - 1, 0.0 /*d_current_time*/, tag_buffer);
```
boyceg, line 1373: To get this to work with a single level, wouldn't this do the trick:
```suggestion
                d_scratch_hierarchy, std::max(0, max_level - 1), 0.0 /*d_current_time*/, tag_buffer);
```

drwells: I don't think so because if there is only one level it doesn't make sense to regrid all finer levels.

boyceg: Can you give it a try? I think this is how patches are distributed for single-level runs.

drwells: I'm on it!

boyceg: Ah, [here](https://ibamr.github.io/IBAMR-docs/samrai/2.4.4/html/classSAMRAI_1_1mesh_1_1GriddingAlgorithm.html#4dd69a9bf65bfe13b4cd6712e868c607) is the relevant documentation.

boyceg: Should use `GriddingAlgorithm::makeCoarsestLevel()` for a single-level run.

drwells: you're right - surprisingly this works.

drwells: As in, I am surprised that nothing else was broken in this case.

drwells:
Could someone take a look and merge soon? @jabrown893 would like to use this.

abarret:
I'm not qualified to review this as I don't know about the scratch hierarchy, but I think Boyce is probably occupied with grant stuff. Does the fact that `regridAllFinerLevels` can only add one level to the hierarchy per call affect how things happen as levels appear and disappear?

src/IB/IBFEMethod.cpp
```diff
@@ -1362,8 +1362,12 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
...
             // any sort of tagging based on the current time I think we can
             // just put in a bogus (nonnegative) value.
-            d_scratch_gridding_algorithm->regridAllFinerLevels(
-                d_scratch_hierarchy, 0, 0.0 /*d_current_time*/, tag_buffer);
+            int max_level = d_scratch_hierarchy->getFinestLevelNumber();
+            if (max_level == 0)
+                d_scratch_gridding_algorithm->makeCoarsestLevel(d_scratch_hierarchy, 0.0 /*d_current_time*/);
+            else
+                d_scratch_gridding_algorithm->regridAllFinerLevels(
+                    d_scratch_hierarchy, max_level - 1, 0.0 /*d_current_time*/, tag_buffer);
```
drwells, line 1370: @abarret:
> Does the fact that regridAllFinerLevels can only add one level to the hierarchy per call affect how things happen as levels appear and disappear?

You're not wrong but that's irrelevant - this patch is a lot dumber than perhaps you think it is. At the point this is called all relevant levels exist (we are regridding for the first time). The original bug was that we called `regridAllFinerLevels` from level 0, which regrids level 1 - this level doesn't exist in the scratch hierarchy if we have, e.g., three total levels.

`makeCoarsestLevel` is oddly named in this context, but according to SAMRAI's documentation if the level already exists it applies load balancing to level 0 - which is exactly what we want to do.

drwells: @abarret even if you don't believe me all the tests pass which shows that I am right :)

PR #1005, opened 2020-06-01
Fix tests on Mac, part 1

drwells:
It looks like a subset of tests fail on Macs for two related but distinct reasons:
1. libMesh generates slightly different partitionings in parallel with `LinearPartitioner` on Macs.
2. libMesh sometimes (not always) generates slightly different meshes (different element counts) on Macs.

This PR fixes the first issue. Don't merge this yet - it depends on #1002.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
Since #1002 is in this can now be merged.

tests/IBFE/explicit_ex4_2d.inactive_1.mpirun=4.output
```diff
@@ -2177,10 +2177,10 @@ workload estimate on processor 0 = 1024
...
 workload estimate on processor 2 = 1024
 workload estimate on processor 3 = 1024
-local active DoFs on processor 0 = 3222
-local active DoFs on processor 1 = 2187
-local active DoFs on processor 2 = 2187
-local active DoFs on processor 3 = 1962
+local active DoFs on processor 0 = 2655
+local active DoFs on processor 1 = 2299
+local active DoFs on processor 2 = 2361
+local active DoFs on processor 3 = 2243
```
drwells, line 2183: @abarret The large jump is a result of `LinearPartitioner` being a bad choice for getting equal numbers of degrees of freedom on each cell since it completely ignores both the geometry and topology of the mesh. The new partitioner does an OK job for quasi-uniform meshes. Thats why the differences are so large.

drwells:
This fixed more tests than it broke on macs so lets merge :)

PR #1006, opened 2020-06-03
Fix tests on mac 2

drwells:
This PR resolves a couple of outstanding issues with the test suite and mac:
1. I deleted the long-suffering `adv_diff_01_2d` test that has never worked.
2. Triangle generates different grids (including different numbers of elements) on different machines for some inputs (e.g., `explicit_ex4_2d.source_test.input`). To get around this I just saved the two 2D triangulations that we use in practice.
3. Many structured grids have centroids with the same x coordinate - this is problematic with `libMesh::CentroidPartitioner` since that partitioner does not use a stable sorting algorithm. I worked around this in a few different ways that should be pretty robust and wrote another partitioner for us to use in the test suite.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
I would prefer to keep it this way so that we don't have to drastically change the tests, though that is also an option since we have a simple geometry (I am pretty sure that libMesh can split quads into triangles).

drwells:
We still use Triangle in a few other places - I don't completely understand why but only a minority of Triangle-using tests didn't work on mac. We should ultimately get rid of using it but I am content with only fixing the problematic tests.

Speaking of which - some of these tests no longer pass in debug mode. I'm taking a look.

drwells:
This doesn't work in debug mode due to overzealous libMesh assertions. I'm working on a fix.

drwells:
That should do it!

PR #1008, opened 2020-06-05
Cleanup our libMesh/boost detection a bit.

### LINKED ISSUE #1007: Try to detect some common problems with libMesh installations
A common mistake with libMesh is to reinstall a new copy of libMesh with a different configuration in the same directory as an older installation of libMesh. This is problematic since some headers from the old installation might remain - e.g., boost.

We should try to detect some inconsistencies when we care about our dependencies' dependencies.

drwells:
Fixes #1007, I think - @rossisimone does this address all the weird problems you had?

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
Older versions of libMesh don't define `LIBMESH_HAVE_EXTERNAL_BOOST` - let me try to write a workaround.

Issue #824, opened 2020-01-06
IBAMR 0.7


drwells:
I closed a number of issues today that are no longer relevant for one reason or another. I also added everything that, to me, appears significant enough that we should try hard to get it in the next release into the 0.7 milestone.

I set a tentative due date of May 31 for the next release: I would prefer to get this done in May since people have more limited availability over the summer.

I assigned roughly half the issues to myself, which is going to form the majority of my 'time spent generically improving IBAMR' this semester (particularly #482). I don't have time to do everything - it would help a lot if I could get some people to volunteer (or to volunteer their students) to help tackle the remaining issues.

Feel free to add more issues to the 0.7 milestone as necessary. With a little luck this may actually be the last pre-SAMRAI 3 release :)

drwells:
@amneetb Do you think that June 15 is sufficient time to get in the patches mentioned elsewhere?

amneetb:
Yes, that should be a sufficient time to get all of the patches in.

drwells:
@IBAMR/ibamr-write-team  The deadline is not strict, but we have planned on releasing IBAMR 0.7 on June 15. The test suite is currently in good shape, with the exception of mesh serialization in libMesh 1.2.1 (which is only an issue inside the testsuite itself and not the library). I plan on throwing a day or two at release preparation at the end of the week.

I would like to keep up our current release cadence of 'twice a year, about six months apart' - hence it doesn't make sense to push back the release by more than a few days. With that in mind, I would like to ask that all non-documentation fixes get put up in PR form in the next few days so that I will have time to deal with various platform-dependent problems.

drwells:
We might not get the release done today but lets take what we have and run with it.

# IBAMR release tasks

## Summary

This document describes the process for creating a new release of IBAMR on
GitHub.

## static analysis

- [ ] verify that neither `namespaces.h` nor `app_namespaces.h` are in IBTK or
  IBAMR headers
- [ ] clear basic compilation warnings with GCC and clang on linux
- [ ] compile with the latest libMesh release (which should be configured with
  `--disable-deprecated`) to verify that we do not use any deprecated
  functionality (this will cause compilation errors if we do use them). Add
  version checks: for example, here is one (lightly edited) in
  `IBFEInstrumentPannel.cpp` that uses a macro defined in `libmesh_version.h`:
```cpp
    // new API in 1.4.0
#if LIBMESH_VERSION_LESS_THAN(1, 4, 0)
[...]

PR #1010, opened 2020-06-10
Fixing restart for IBLevelSet/ex0

amneetb:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/minor/20200612AmneetBhalla
+++ b/doc/news/changes/minor/20200612AmneetBhalla
@@ -1,0 +1,3 @@
+Bug Fix: Fixed restart small bug in examples/IBLevelSet/ex0. 
+<br>
+(Amneet Bhalla, 2020/06/12)
--- a/examples/IBLevelSet/ex0/example.cpp
+++ b/examples/IBLevelSet/ex0/example.cpp
@@ -518,14 +518,17 @@ main(int argc, char* argv[])
         }
         navier_stokes_integrator->registerAdvDiffHierarchyIntegrator(adv_diff_integrator);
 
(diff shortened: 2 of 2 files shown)


amneetb:
#824

boyceg:
If `IBFEMethod` is only used internally in `IBLevelSetMethod`, could it be responsible for creating it as needed?

amneetb:
I think so. The design is such that the application can control when to abort IBFEMethod amidst a simulation, or even not use it at all (for example when FEDistanceEvaluator is not used for calculating level set). The possible number of combinations are many, so I left it to the application.

boyceg:
It seems like the application code has to do things differently for de novo vs. restarted runs. Typically, restartable objects allow the application code to set up everything in the same way whether things are from restart or not, and the implementations sorts out the details internally.

Can `IBLevelSetMethod` can be set up to "do the right thing" in either case, to simplify the application code?

drwells:
@amneetb I'd like to get this into 0.7 but its missing a lot of checkmarks. What do you plan on doing?

amneetb:
I think this only changing an example, and no other file in the library, so it should be fine to merge. The CI says that all checks have passed. Most of the checkboxes don't apply to this commit, I think. 

Regarding "Does the test suite pass?". Isn't this be checked by the CI system?  Github says "All checks have passed", so I presume I can checkmark the first box?

drwells:
The CI is not yet set up to run the test suite, unfortunately.

Would you please add a changelog entry? Once that's done lets merge.

amneetb:
@drwells:  I've added it.

PR #1015, opened 2020-06-16
Fix 64-bit hypre.

drwells:
I get this error when I install hypre through homebrew. Since no one has mentioned it before I don't think its a big deal - unless someone complains I don't plan on putting it in 0.7.0.

This is still WIP since it causes a bunch of weird crashes.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/solvers/impls/SCPoissonHypreLevelSolver.cpp
+++ b/ibtk/src/solvers/impls/SCPoissonHypreLevelSolver.cpp
@@ -704,6 +704,8 @@ SCPoissonHypreLevelSolver::solveSystem(const int x_idx, const int b_idx)
     IBTK_TIMER_START(t_solve_system_hypre);
 
     d_current_iterations = 0;
+    // HYPRE_INT may be either long or int
+    HYPRE_Int current_iterations = d_current_iterations;
     d_current_residual_norm = 0.0;
 
     if (d_solver_type == "SysPFMG")
@@ -719,7 +721,7 @@ SCPoissonHypreLevelSolver::solveSystem(const int x_idx, const int b_idx)
             HYPRE_SStructSysPFMGSetZeroGuess(d_solver);
         }
         HYPRE_SStructSysPFMGSolve(d_solver, d_matrix, d_rhs_vec, d_sol_vec);
-        HYPRE_SStructSysPFMGGetNumIterations(d_solver, &d_current_iterations);
+        HYPRE_SStructSysPFMGGetNumIterations(d_solver, &current_iterations);
         HYPRE_SStructSysPFMGGetFinalRelativeResidualNorm(d_solver, &d_current_residual_norm);
     }
     else if (d_solver_type == "Split")
@@ -735,7 +737,7 @@ SCPoissonHypreLevelSolver::solveSystem(const int x_idx, const int b_idx)
             HYPRE_SStructSplitSetZeroGuess(d_solver);
         }
         HYPRE_SStructSplitSolve(d_solver, d_matrix, d_rhs_vec, d_sol_vec);
-        HYPRE_SStructSplitGetNumIterations(d_solver, &d_current_iterations);
+        HYPRE_SStructSplitGetNumIterations(d_solver, &current_iterations);
         HYPRE_SStructSplitGetFinalRelativeResidualNorm(d_solver, &d_current_residual_norm);
     }
     else if (d_solver_type == "PCG")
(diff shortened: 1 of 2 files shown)


ibtk/include/ibtk/GeneralSolver.h
```diff
@@ -322,7 +324,7 @@ class GeneralSolver : public virtual SAMRAI::tbox::DescribedClass
     double d_rel_residual_tol = 0.0;
     double d_abs_residual_tol = 0.0;
     int d_max_iterations = 100;
-    int d_current_iterations = 0;
+    HYPRE_Int d_current_iterations = 0;
```
boyceg, line 327: Let's not do this here.

ibtk/src/solvers/impls/PETScKrylovLinearSolver.cpp
```diff
@@ -209,7 +209,9 @@ PETScKrylovLinearSolver::solveSystem(SAMRAIVectorReal<NDIM, double>& x, SAMRAIVe
     d_A->imposeSolBcs(x);
 
     // Get iterations count and residual norm.
-    ierr = KSPGetIterationNumber(d_petsc_ksp, &d_current_iterations);
+    PetscInt current_iter = d_current_iterations;
```
boyceg, line 212: Should we go ahead and systematically use PETSc `typedef`s?

boyceg:
Should we try to fix this for 0.8.0?

drwells:
Since we are extending the deadline I should be able to finish this in time.

drwells:
Well, that didn't happen - this is a low priority item so I'll remove it from the release.

ibtk/include/ibtk/GeneralSolver.h
```diff
@@ -21,6 +21,8 @@
 #include "tbox/DescribedClass.h"
 #include "tbox/Pointer.h"
 
+#include <HYPRE_utilities.h>
```
boyceg, line 24: Let's not chain ourselves to the hypre interface for setting the number of iterations.

Hopefully we never need to avail ourselves of 64-bit integers to store the number of iterations used in a hypre solver. 😀

boyceg:
@drwells this looks mostly done. Can we just finish it for 0.8?

ibtk/src/solvers/impls/PETScKrylovLinearSolver.cpp
drwells: That would be nice to do but it doesn't work well with SAMRAI - there are places where we need to store index information but SAMRAI doesn't support `long int`.

drwells:
My initial version compiled but introduced some bugs where I didn't copy data correctly - those are all now fixed.

drwells:
I had to fix one more small thing but this now works (all tests pass) with homebrew's hypre.

ibtk/src/solvers/impls/PETScKrylovLinearSolver.cpp
boyceg: Should we drop this change since we are not changing the type of `d_current_iterations`? If `PetscInt` is different from `int`, we are already going to run into a lot of problem. 😬

ibtk/src/solvers/impls/PETScNewtonKrylovSolver.cpp
```diff
@@ -209,7 +209,9 @@ PETScNewtonKrylovSolver::solveSystem(SAMRAIVectorReal<NDIM, double>& x, SAMRAIVe
 
     // Get iterations counts and residual norm.
     IBTK_CHKERRQ(ierr);
-    ierr = SNESGetIterationNumber(d_petsc_snes, &d_current_iterations);
+    PetscInt current_iterations = d_current_iterations;
```
boyceg, line 212: Same comment as above.

ibtk/src/solvers/impls/SCPoissonHypreLevelSolver.cpp
```diff
@@ -65,6 +66,16 @@ static Timer* t_solve_system_hypre;
 static Timer* t_initialize_solver_state;
 static Timer* t_deallocate_solver_state;
 
+// HYPRE can use 64bit indices, but SAMRAI IntVectors are always 32 - add
```
boyceg, line 69: Should we pull this up to a helper header since it now shows up in a few places?

boyceg (changes requested):
Thanks for not putting `HYPRE_Int` in the interface. 😁  A few suggested changes below...

ibtk/src/solvers/impls/SCPoissonHypreLevelSolver.cpp
drwells: I'd rather not since these two files are the only places it shows up (and we don't interface with hypre anywhere else) - having an extra header `hypre_compatibility.h` or whatever is more work than copying and pasting this seven-line function once.

boyceg: That's fine --- I was thinking about using an existing header like `IndexUtilities.h` or `ibtk_utilities.h`.

ibtk/src/solvers/impls/PETScKrylovLinearSolver.cpp
drwells: Yeah, I guess we can just drop this commit - its only a partial fix for a much more substantial problem.

drwells:
I deleted the `PetscInt` commit.

boyceg (approved):

PR #1018, opened 2020-06-17
Fix some tests with libMesh 1.2.1.

drwells:
Same as #1017, but cherry-picked onto the 0.7 branch.


drwells:
The CI doesn't yet run the test suite so lets just merge.

PR #1023, opened 2020-06-26
Bug fixes to compute time steps in parallel.

### LINKED ISSUE #1022: Different processors use different time steps
https://github.com/IBAMR/IBAMR/blob/59afafa326986b0f27e5e957b8c09e87a8ef3a69/ibtk/src/utilities/HierarchyIntegrator.cpp#L330-L350

@boyceg Is there a reason why we don't do a `minReduction` when we compute time step sizes? I'm getting patches that run at different timesteps depending on how the patches are laid out on processors.

abarret:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

Fixes #1022. `AdvDiffHierarchyIntegrator` did not compute timestep sizes consistent across all processors. I've updated the documentation in `HierarchyIntegrator` to note that time step sizes should be consistent in implementations of `getMinimumTimeStepSizeSpecialized` and `getMaximumTimeStepSizeSpecialized`. ~~This is also fixes a minor bug in `HierarchyIntegrator` to select the smallest of all timesteps in computing the minimum timestep.~~

### DIFF
--- a/ibtk/include/ibtk/HierarchyIntegrator.h
+++ b/ibtk/include/ibtk/HierarchyIntegrator.h
@@ -699,7 +699,8 @@ protected:
 
     /*!
      * Virtual method to compute an implementation-specific minimum stable time
-     * step size.
+     * step size. Implementations should ensure that the returned time step is
+     * consistent across all processors.
      *
      * A default implementation is provided that returns
      * min(dt_max,dt_growth_factor*dt_current).  The growth condition prevents
@@ -709,7 +710,8 @@ protected:
 
     /*!
      * Virtual method to compute an implementation-specific maximum stable time
-     * step size.
+     * step size. Implementations should ensure that the returned time step is
+     * consistent across all processors.
      *
      * A default implementation is provided that returns
      * min(dt_max,dt_growth_factor*dt_current).  The growth condition prevents
--- a/src/adv_diff/AdvDiffHierarchyIntegrator.cpp
+++ b/src/adv_diff/AdvDiffHierarchyIntegrator.cpp
@@ -978,7 +978,7 @@ AdvDiffHierarchyIntegrator::getMaximumTimeStepSizeSpecialized()
             }
         }
     }
-    return dt;
+    return SAMRAI_MPI::minReduction(dt);
 } // getMaximumTimeStepSizeSpecialized
 
 void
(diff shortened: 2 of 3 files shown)


ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -699,7 +699,8 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
 
     /*!
      * Virtual method to compute an implementation-specific minimum stable time
-     * step size.
+     * step size. Implementations should ensure that the returned time step is
+     * consistent across all processors.
```
boyceg, line 703: 👍

ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -709,7 +710,8 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
 
     /*!
      * Virtual method to compute an implementation-specific maximum stable time
-     * step size.
+     * step size. Implementations should ensure that the returned time step is
+     * consistent across all processors.
```
boyceg, line 714: 👍

ibtk/src/utilities/HierarchyIntegrator.cpp
```diff
@@ -333,7 +333,7 @@ HierarchyIntegrator::getMinimumTimeStepSize()
     double dt = getMinimumTimeStepSizeSpecialized();
     for (const auto& child_integrator : d_child_integrators)
     {
-        dt = std::max(dt, child_integrator->getMinimumTimeStepSize());
+        dt = std::min(dt, child_integrator->getMinimumTimeStepSize());
```
boyceg, line 336: this should be `std::max()` not `std::min()`

src/adv_diff/AdvDiffHierarchyIntegrator.cpp
```diff
@@ -978,7 +978,7 @@ AdvDiffHierarchyIntegrator::getMaximumTimeStepSizeSpecialized()
             }
         }
     }
-    return dt;
+    return SAMRAI_MPI::minReduction(dt);
```
boyceg, line 981: 👍

ibtk/src/utilities/HierarchyIntegrator.cpp
boyceg: The minimum time step size will be the max over all of the minimums.

The maximum time step size will be the min over all of the maximums.

abarret: Yes, I see. This isn't choosing smallest stable time steps, it's choosing minimum timesteps.

abarret: I'm pretty sure this is the second time I've gotten confused on this :laughing:

PR #1025, opened 2020-06-29
Fix some uninitialized variable warnings.

drwells:
- [x] Does the test suite pass? (I didn't run IBFE tests since this is completely isolated)
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/wave_generation/IrregularWaveBcCoef.cpp
+++ b/src/wave_generation/IrregularWaveBcCoef.cpp
@@ -196,13 +196,14 @@ IrregularWaveBcCoef::setBcCoefs(Pointer<ArrayData<NDIM, double> >& acoef_data,
         TBOX_ASSERT(!gcoef_data || bc_coef_box == gcoef_data->getBox());
 #endif
 
-        double dof_posn[NDIM];
         for (Box<NDIM>::Iterator b(bc_coef_box); b; b++)
         {
             const SAMRAI::hier::Index<NDIM>& i = b();
             if (acoef_data) (*acoef_data)(i, 0) = 1.0;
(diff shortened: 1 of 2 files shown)


drwells:
Fixed!

src/wave_generation/IrregularWaveBcCoef.cpp
```diff
@@ -196,13 +196,14 @@ IrregularWaveBcCoef::setBcCoefs(Pointer<ArrayData<NDIM, double> >& acoef_data,
...
 #endif
 
-        double dof_posn[NDIM];
         for (Box<NDIM>::Iterator b(bc_coef_box); b; b++)
         {
             const SAMRAI::hier::Index<NDIM>& i = b();
             if (acoef_data) (*acoef_data)(i, 0) = 1.0;
             if (bcoef_data) (*bcoef_data)(i, 0) = 0.0;
 
+            IBTK::Point dof_posn;
```
boyceg, line 205: Does `Point` have a default constructor?

drwells: Surely you mean a not-default constructor?

The online documentation for `Matrix(const Scalar &s)` states (probably incorrectly) that it creates a 1x1 matrix, so I went for the unambiguous option.

boyceg: I think what I meant was --- a "reasonable default constructor".

drwells: The default constructor doesn't initialize data, so its not going to help in this case. Besides - we should throw an error if we read a value which we didn't set, so `signaling_NaN()` is the right choice here.

PR #1026, opened 2020-06-29
Prevent some code from being overoptimized.

drwells:
GCC 10.1.0 got a lot better at vectorizing code, but in this case I am pretty sure it is doing something wrong: it is generating wrong code w.r.t. where the array index is placed which causes bus errors. I will ultimately report this to the GCC guys but I want to finish up a few other things first - for now this workaround seems to defeat the loop vectorization function.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
I think the comment is enough. Besides, we may want to continue supporting 10.1.0 for a long time.

PR #1029, opened 2020-07-01
Allow fe amr

### LINKED ISSUE #249: nodal FE coupling does not respect periodic boundaries of FE meshes
The specialized FE coupling implementation loops over all of the nodes in the FE mesh to spread forces / interpolate velocities. This is what we want for FE meshes without periodic boundaries, but can result in "double counting" for FE meshes with periodic boundaries.

### LINKED ISSUE #476: apparent parallel libMesh periodic BC issue
It seems like there is a longstanding bug in handling periodic boundary conditions with `examples/IBFE/explicit/ex1` in parallel. I can reproduce this on a number of different points in the history by setting
```
   largest_patch_size {
      level_0 = N/2,  N  // all finer levels will use same values as level_0
   }
   smallest_patch_size {
      level_0 = N/2,  N  // all finer levels will use same values as level_0
   }
```
and running on four processors. There will be a clear aberration in the x-component of the velocity field at the location where the periodic boundaries of the structure meet.

### LINKED ISSUE #656: IB post processor in parallel (with periodic boundary conditions)
maybe related to issue #476.  

I have some issues with the post processor not working correctly in parallel.  I am not sure yet if this is because of periodic boundary conditions, but maybe.

drwells:
Fixes #249. Fixes #476. Fixes #656 (@cpuelz could you verify this?)

The much-discussed patch to fix our usage of hanging nodes. It turns out that this patch also fixes our other problems with constraints.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


src/IB/IBFEMethod.cpp
```diff
@@ -636,6 +636,33 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
...
     }
 
+    // If the velocity system has constraints then those need to be dealt with
+    // here and not during assembly because we do not distribute the
+    // constraints corresponding to the IB partitioning.
+    std::vector<PetscVector<double>*> vecs_for_second_summation;
+    for (unsigned int part = 0; part < d_meshes.size(); ++part)
+    {
+        // We do not need to constrain anything when we have nodal quadrature
+        // since all RHS values are independent
```
boyceg, line 646: I don't understand this comment. There are still hanging node values that need to be constrained, regardless of the quadrature scheme. Should this say something like "constraints are dealt with elsewhere"?

drwells: I need to reword this. My original point was that if we use nodal quadrature then we do not need to use constraints when setting up the linear system since no dof depends on any other dof.

drwells:
I thought we had a policy of not substantially modifying examples. How do you want to do this?

src/IB/IBFEMethod.cpp
drwells: I added some more text explaining that FEProjector will always enforce constraints after we solve the linear system.

boyceg:
> I thought we had a policy of not substantially modifying examples. How do you want to do this?

I think it would be OK to add an optional flag to turn on local mesh refinement for a couple of the existing examples and to add new input files that turn on the functionality, making it easy for folks to compare the behavior with/without local refinement.

drwells:
It turns out that the fix for #249 is related to this patch. I'll update it in a few minutes.

drwells:
OK, that should do it.

I added a test to verify that we get the same output for periodic structures when we use nodal quadrature as when we use the trapezoid rule. This shows that we have resolved the nodal problems.

boyceg:
> OK, that should do it.
> 
> I added a test to verify that we get the same output for periodic structures when we use nodal quadrature as when we use the trapezoid rule. This shows that we have resolved the nodal problems.

Oh yeah, this is a good idea. Can we do it for each of the nodal tests?

drwells:
We can test it on our own, but its going to be hard to integrate into the test suite since we will need to use a different quadrature rule to assemble the mass matrix stored by `FEProjector`.

drwells:
@boyceg I'd like to get this merged since I'll be unavailable next week (so that y'all can get started with upping the mesh resolution). Is there anything else you would like to see in this PR? A few things, like modifying an example, are not necessary for the heart model to use this feature.

cpuelz:
@drwells I can check #656, sorry I haven't gotten to this yet.  feel free to merge whenever tho.

PR #1034, opened 2020-07-03
Fix an ABI problem in some Fortran code.

drwells:
Enabling link-time optimization with GCC has the interesting side effect of enabling type-checking of our Fortran interfaces. In particular, it looks like LOGICAL is not the same as unsigned int: some FORTRAN compilers make logical true equal to -1 so this is not the right choice.

After reading through some gfortran documentation I'm not sure at all what integral type LOGICAL is supposed to be, so lets completely avoid the issue and use an integer set to 1 if we want the corner code on.

Here is one of the four GCC warnings:
```
/home/drwells/Documents/Code/CPP/ibamr/src/advect/AdvectorExplicitPredictorPatchOps.cpp:411:10:
warning: type of 'advect_predict_ppm_with_source3d_' does not match original declaration
  411 |     void ADVECT_PREDICT_PPM_WITH_SOURCE_FC(const double*,
      |          ^
/home/drwells/Documents/Code/CPP/ibamr/build/src/advect/fortran/advect_predictors3d.f:984:
note: 'advect_predict_ppm_with_source3d' was previously declared here
  984 |       subroutine advect_predict_PPM_with_source3d(
      |
```

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/include/ibamr/AdvectorExplicitPredictorPatchOps.h
+++ b/include/ibamr/AdvectorExplicitPredictorPatchOps.h
@@ -374,11 +374,12 @@ private:
      *    d_limiter_type ........ specifies the type of slope limiting type used in
      *                            computing numerical fluxes
      *    d_using_full_ctu ...... specifies whether full corner transport
-     *                            upwinding is used for 3D computations
+     *                            upwinding is used for 3D computations. Set to 1 to
+     *                            enable it.
      */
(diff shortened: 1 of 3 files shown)


boyceg:
Is that the only place I used that type?

drwells:
It's the only place where it is a function argument.

PR #1035, opened 2020-07-03
Move a Fortran function used in IBTK into IBTK.

drwells:
We need to do this in the CMake build so that there are no undefined symbols in libIBTK2d and libIBTK3d. With static libraries this is not a problem but it is with shared libraries on mac (the linker fails if any symbols are undefined).

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
But, can you move this into a `NDIM`-independent source code file?

drwells:
Fixed!

PR #1039, opened 2020-07-13
Fix the ghost cell size given to SAMRAIGhostDataAccumulator.

drwells:
Thanks to @jabrown893 for pointing this out.

IBFEMethod can determine the ghost cell width required by itself and by FEDataManager, but it doesn't know about ghost cell with requirements of other classes. Hence we should not use its own ghost width for setting up force spreading accumulation data structures - instead we can just read it from the input data.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -928,10 +928,19 @@ IBFEMethod::spreadForce(const int f_data_idx,
     }
 
     {
-        auto hierarchy = d_use_scratch_hierarchy ? d_scratch_hierarchy : d_hierarchy;
         if (!d_ghost_data_accumulator)
+        {
+            // If we have multiple IBMethod objects we may end up with a wider
+            // ghost region than the one required by this class. Hence, set the
+            // ghost width by just picking whatever the data actually has at the
+            // moment.
+            const Pointer<PatchLevel<NDIM> > level = hierarchy->getPatchLevel(ln);
+            const IntVector<NDIM> gcw =
+                level->getPatchDescriptor()->getPatchDataFactory(f_scratch_data_idx)->getGhostCellWidth();
+
             d_ghost_data_accumulator.reset(new SAMRAIGhostDataAccumulator(
-                hierarchy, f_var, d_ghosts, d_hierarchy->getFinestLevelNumber(), d_hierarchy->getFinestLevelNumber()));
+                hierarchy, f_var, gcw, d_hierarchy->getFinestLevelNumber(), d_hierarchy->getFinestLevelNumber()));
+        }
         d_ghost_data_accumulator->accumulateGhostData(f_scratch_data_idx);
     }
 
--- a/doc/news/changes/minor/2020713DavidWells
+++ b/doc/news/changes/minor/2020713DavidWells
@@ -1,0 +1,4 @@
+Fix: Fixed a problem with IBFEMethod when the ghost data size required by
+IBFEMethod is not equal to that required by another class.
+<br>
+(David Wells, 2020/07/13)


src/IB/IBFEMethod.cpp
```diff
@@ -928,10 +928,28 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+        IntVector<NDIM> gcw;
+        for (PatchLevel<NDIM>::Iterator p(level); p; p++)
+        {
+            const Pointer<Patch<NDIM> > patch = level->getPatch(p());
+            Pointer<PatchData<NDIM> > f_data = patch->getPatchData(f_scratch_data_idx);
+            gcw = f_data->getGhostCellWidth();
+            break;
+        }
+        // Now cover the case that this processor has no patches:
+        SAMRAI_MPI::maxReduction(gcw, NDIM);
```
boyceg, line 948: If this corresponds to a `Variable` registered with the `VariableDatabase`, I think you can ask the `VariableDatabase` for the ghost cell width directly and avoid this collective.

drwells: Do you know how to do this? The `VariableDatabase` lets you register things with a ghost width but you cannot get them out. If you ask the `PatchDataFactory` then you get the wrong answer - I am pretty sure that this is the only way to get this value out.

boyceg: It is convoluted, but you need to access the `PatchDescriptor`, which knows about ghost cell widths.

boyceg: The `PatchDescriptor` has the `PatchDataFactory` that is actually used to construct `PatchData`. This factory should tell you the ghost cell width.

drwells: I tried that and it returns a vector of all zeros - I think its a SAMRAI bug.

boyceg: I am sure that we use this elsewhere in the library to get ghost cell widths...

boyceg: (Also, some patch data will indeed have zero GCW.)

boyceg: Oh, yeah, if you have a patch ID and a `Patch` you should be able to get the `PatchDataFactory` directly from the patch without going through `VariableDatabase` at all. You might need to properly cast to the concrete `PatchDataFactory` class.

drwells: The return type of `Variable<NDIM>::getPatchDataFactory()` is `Pointer<PatchDataFactory<NDIM> >` so no extra cast is required. For the following code

```cpp
        const Pointer<PatchLevel<NDIM> > level = hierarchy->getPatchLevel(ln);
        IntVector<NDIM> gcw;
        pout << "GCW from patch data factory: "
             << f_var->getPatchDataFactory()->getGhostCellWidth()
             << '\n';
        {
            const Pointer<Patch<NDIM> > patch = level->getPatch(p());
            Pointer<PatchData<NDIM> > f_data = patch->getPatchData(f_scratch_data_idx);
            gcw = f_data->getGhostCellWidth();
            break;
        }
        // Now cover the case that this processor has no patches:
        SAMRAI_MPI::maxReduction(gcw, NDIM);
        pout << "GCW from max reduction: "
             << gcw
             << '\n';
```
I get the output

```
GCW from patch data factory: (0,0)
GCW from max reduction: (3,3)
```

so using the patch data factory isn't going to work, even in serial.

boyceg: You cannot use the patch data factory from the variable --- it does not know what the ghost cell width is. (In SAMRAI, ghost cell widths are not properties of `Variables`.) Instead, you have to use the one associated with the `PatchDescriptor`. It _must_ know what the ghost cell width is --- otherwise it would not allocate patch data with the correct ghost cell width.

drwells: Oh, I see. Let me try that instead.

drwells: This is strange inconsistency in SAMRAI - `VariableDatabase::registerVariableAndContext()` takes a ghost width. Therefore it should be able to tell me the ghost width of a variable.

Nevertheless getting the patch data factory out of the patch descriptor works so lets go with that.

boyceg: The way that SAMRAI works, the `Variable` + `VariableContext` together are assigned a ghost cell width. The `Variable` by itself does not correspond to _any_ particular data storage.

drwells: That explains it!

drwells:
I ran the test suite again and everything seems to be in order.

drwells:
@boyceg are you happy with the fix? If so lets merge.

boyceg:
Looks good.

PR #1041, opened 2020-07-14
Remove CACHED assembly from IBFEMethod.

drwells:
It isn't possible for this to work correctly with constraints since we need ghost values to do the two-stage constraint resolution process implemented in e2e62c4d1bef6bfd817998ba87d0c37c913be1fd.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
Is this code path faster for problems without constraints?

drwells:
Yes. Using `VecCache` for assembly is slower because it has to recompute the communication pattern and a lot of duplicated information ends up in the cache (e.g., if we sum to the same entry 10 times, we get 10 cache entries to communicate).

boyceg:
Then should we keep it for cases where we can avoid using constraints?

drwells:
Oh, I think I misunderstood your question. The faster code path is to always use ghosted assembly since it is much faster than relying on `VecCache`. Therefore there is no reason to use `CACHED` (it is slower and doesn't work in all cases).

boyceg:
We should totally get rid of slower code that does not work. :-)

PR #1049, opened 2020-07-21
Fix hanging nodes plus nodal quad

drwells:
Supercedes #1047.

We previously de facto interpolated the hanging node constraints in a manner similar to that described in Kormann's 2016 paper on mass lumping for the Schrodinger equation. This algorithm is problematic for us, in particular, since it results in ignoring the values interpolated at hanging nodes while spreading forces from hanging nodes. Get around this by properly forming the lumped mass matrix (which, if there are hanging nodes, is no longer diagonal).

We should write a clearer discussion than this for the upcoming mass lumping paper, but the reason this newer approach is preferable is that we are now using the same mass matrix C^T M C for both spreading and interpolation. For interpolation we use the new lumped (and not necessarily diagonal) mass matrix to solve the linear system. For spreading we use the unconstrained diagonal mass matrix, but we only use it with finite element solution vectors (i.e., vectors x where C x = x): hence the inner product induced by C^T M C is equal to that of just M in the nodal quadrature case. Similarly, we use the same JxW values and point locations for both spreading and interpolation (they come from the diagonal mass matrix), while previously we ignored the values interpolated at hanging nodes.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -240,12 +266,14 @@ FEProjector::buildDiagonalL2MassMatrix(const std::string& system_name)
 
         // Extract the FE system and DOF map, and setup the FE object.
         System& system = d_fe_data->getEquationSystems()->get_system(system_name);
-        const int sys_num = system.number();
         DofMap& dof_map = system.get_dof_map();
         FEData::SystemDofMapCache& dof_map_cache = *d_fe_data->getDofMapCache(system_name);
         dof_map.compute_sparsity(mesh);
         FEType fe_type = dof_map.variable_type(0);
-        std::unique_ptr<QBase> qrule = fe_type.default_quadrature_rule(dim);
+        const Order fe_order = fe_type.order;
```
boyceg, line 273: Note that this will only work for certain element types.

Perhaps we could use `QNODAL` if available and otherwise throw an error if an element type is not supported.

Also, although I think that `QTRAP` is equivalent to the originally implemented lumping technique for `FIRST` order elements, I don't know whether `QSIMPSON` is equivalent for `SECOND` order elements.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -268,109 +295,136 @@ FEProjector::buildDiagonalL2MassMatrix(const std::string& system_name)
...
+
+        // Extract the FE system and DOF map, and setup the FE object.
+        System& system = d_fe_data->getEquationSystems()->get_system(system_name);
+        DofMap& dof_map = system.get_dof_map();
+        FEData::SystemDofMapCache& dof_map_cache = *d_fe_data->getDofMapCache(system_name);
+        dof_map.compute_sparsity(mesh);
+        FEType fe_type = dof_map.variable_type(0);
+        const Order fe_order = fe_type.order;
+        TBOX_ASSERT(fe_order == libMesh::FIRST || fe_order == libMesh::SECOND);
+        std::unique_ptr<QBase> qrule =
```
boyceg, line 359: Same comment as above.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -389,40 +443,29 @@ FEProjector::computeL2Projection(PetscVector<double>& U_vec,
...
-        converged = reason > 0;
-    }
-    else
-    {
-        PetscVector<double>* M_diag_vec = buildDiagonalL2MassMatrix(system_name);
-        ierr = VecPointwiseDivide(U_vec.vec(), F_vec.vec(), M_diag_vec->vec());
-        IBTK_CHKERRQ(ierr);
-        converged = true;
-    }
+    std::pair<PetscLinearSolver<double>*, PetscMatrix<double>*> proj_solver_components =
```
boyceg, line 446: Can we just do a pointwise divide in cases where the lumped mass matrix remains diagonal?

ibtk/src/lagrangian/FEProjector.cpp
drwells: `QNODAL` actually turns into `QSIMPSON`: see

https://github.com/libMesh/libmesh/blob/7f11f2f7247d487176cde060fa8c60be35acf402/src/quadrature/quadrature_nodal_2D.C#L138

```cpp
    case QUAD9:
    case TRI6:
      {
        QSimpson rule(/*dim=*/2, /*ignored*/_order);
        rule.init(_type, /*ignored*/_p_level);
        _points.swap (rule.get_points());
        _weights.swap(rule.get_weights());
        return;
      }
```

The correct answer for tensor-product elements is Simpson's rule since, in that case, it is equal to the Gauss-Lobatto rule. It doesn't look like libMesh's triangular Simpson's rule achieves a higher order than composite trapezoid. I'll see if I can find an alternative quadrature rule that can integrate quadratics correctly.

ibtk/src/lagrangian/FEProjector.cpp
drwells: That should be possible (or a pointwise mult). I'll take a look.

ibtk/src/lagrangian/FEProjector.cpp
drwells: We do get different lumped mass matrices when we use `QNODAL` or equivalent but this doesn't matter for velocity interpolation, since in that case the value on the main diagonal of the lumped mass matrix is the same thing we multiply the interpolated velocity by (i.e., if we are doing velocity interpolation, we really do end up interpolating the velocity).

For force spreading this choice does matter: we get a different `JxW` value. Its not clear to me if one should be better than the other. I am a little suspicious of the current implementation because I do not know what quadrature rule it corresponds to, and for a given set of points there should only be one set of weights.

drwells:
I suspect that we might get better results with the present implementation of quadrature at higher order. Nevertheless the point of this PR is to fix something else so I will revert that - we should try switching later.

ibtk/src/lagrangian/FEProjector.cpp
drwells: Detecting whether or not we have constraints or something else that causes a non-diagonal lumped mass matrix is tricky with the way this is currently set up. I've already spent half an hour on this and I suspect its going to take me at least another hour to finish this for virtually no benefit (converging in one iteration is approximately the same amount of work). Can we skip this?

drwells:
This is done aside from switching from doing 1 linear solver iteration to dividing by matrix entries.

ibtk/src/lagrangian/FEProjector.cpp
boyceg: Can you quickly confirm that there is negligible runtime performance impact for problems without constraints?

drwells: Sure!

Another thought I had: if we ever do time-dependent FE AMR + nodal quadrature then this check will get even more complicated.

drwells: It looks like the linear solve takes up about 5% of the total run time of `interpolateVelocity`: it's low but since its measurable I suppose we should try to skip the solve.

drwells: Fixed!

drwells:
Done!

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -390,11 +460,23 @@ FEProjector::computeL2Projection(PetscVector<double>& U_vec,
...
     const System& system = d_fe_data->getEquationSystems()->get_system(system_name);
-    const DofMap& dof_map = system.get_dof_map();
-    if (consistent_mass_matrix)
+
+    // We can use the diagonal mass matrix directly if we do not need a
+    // consistent mass matrix *and* there are no constraints.
+    //
+    // TODO: this would also work with Dirichlet boundary constraints but its not
+    // as easy to detect those constraints.
+    if (!consistent_mass_matrix && system.get_dof_map().n_constrained_dofs() == 0)
```
boyceg, line 469: 👍

ibtk/include/ibtk/FEProjector.h
```diff
@@ -65,6 +65,15 @@ class FEProjector
     std::pair<libMesh::PetscLinearSolver<double>*, libMesh::PetscMatrix<double>*>
     buildL2ProjectionSolver(const std::string& system_name);
 
+    /*!
+     * \return Pointers to a linear solver and sparse matrix corresponding to a
+     * L2 projection operator generated with the relevant nodal quadrature rule
+     * (i.e., QSIMPSON for second-order elements and QTRAP for first-order
```
boyceg, line 71: Does this comment need to be updated (since we are not currently using `QTRAP` or `QSIMPSON`)?

ibtk/src/lagrangian/FEProjector.cpp
drwells: I wrote something much more complicated at first which is why this took so long!

ibtk/include/ibtk/FEProjector.h
drwells: You're right: let me add a citation of Hinton instead.

boyceg: I'm also OK to do something else

boyceg: ...but maybe that should be another PR.

drwells: Nah, I added it to this one. It only took a minute to fix the documentation :)

PR #1052, opened 2020-07-23
Use a better algorithm for L2 projection.

drwells:
It occurred to me that, now that we have a lumped mass matrix, we have a much better preconditioner available to us for mass matrix solves than just using ILU or a Jacobi variant.

Using MINRES and a Jacobi preconditioner based on the lumped mass matrix is about twice as fast as the previous approach (GMRES and ILU) largely because ILU is overkill for this problem.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/major/20200723DavidWells
+++ b/doc/news/changes/major/20200723DavidWells
@@ -1,0 +1,5 @@
+Improved: IBTK::FEProjector now uses a much better algorithm for solving L2
+projections (Jacobi preconditioner based on the lumped mass matrix and MINRES)
+which is about twice as fast for large tetrahedral meshes.
+<br>
+(David Wells, 2020/07/23)
--- a/ibtk/src/lagrangian/FEProjector.cpp
+++ b/ibtk/src/lagrangian/FEProjector.cpp
@@ -25,6 +25,8 @@
 #include <libmesh/boundary_info.h>
 #include <libmesh/dof_map.h>
 #include <libmesh/elem.h>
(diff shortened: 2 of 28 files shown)


drwells:
I need to copy and paste a few more `FEProjector::buildLumpedL2ProjectionSolver(): building lumped L2 projection solver for system: p_f interpolation system`s. Let me do that real quick to fix some stray tests.

drwells:
Fixed.

knepley:
@drwells Why would you use MINRES instead of CG? The mass matrix should be SPD. MINRES is for symmetric indefinite.

drwells:
The honest answer is that, in practice, I find GMRES slightly faster than CG so I went with MINRES here without thinking that hard about the choice since the system is symmetric. I don't remember if I actually tried out CG or not - let me do that quickly to see if something changes.

drwells:
I used callgrind to compare the performance of MINRES vs CG: for this problem MINRES takes about 10% fewer instructions on my little benchmark (a 3D tetrahedral mesh) so I think we should stick with it for now.

knepley:
Are you measuring convergence in L2 or in the A-norm? I think that is the difference. CG should not be beatable in the A-norm.

drwells:
I am doing whatever libMesh does by default.

drwells:
It's probably worthwhile to examine this in the future but I am content with the current state of things: we don't spend a ton of time on the L2 projection compared to other stuff. Let me open an issue for this so that I remember what to look for.

PR #1056, opened 2020-07-27
Remove the buggy IB6 delta function.

### LINKED ISSUE #853: IB-6 kernel disagrees with inlined version
I noticed this while working on #615 - when I tried to inline the delta function I got different results than the ones I get with `master`. We should investigate and figure out which version is right.

drwells:
We do not use this anywhere in the library and it disagrees with the version of this kernel used in the interpolation and spread functions (which is known to be correct).

Fixes #853. I don't see any reason to actually fix this function since it isn't used anywhere in the library nor do users have any reason to call it themselves. It's just not a good use of our time to bother adding something which no one will use.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


doc/news/changes/incompatibilities/20200727DavidWells
```diff
@@ -0,0 +1,5 @@
+Removed: The IB6 delta function (i.e., lagrangian_ib_6_delta) has been removed
```
boyceg, line 1: This comment suggests that we don't support the IB_6 kernel function, but we do -- just not though this particular interface.

drwells: Good point. I clarified the comment.

PR #1057, opened 2020-07-28
Refactor jacobian calculator

drwells:
We added a new class `IBTK::Mapping` in #961, which provided a superset of the functionality of `JacobianCalculator`. This PR finally cleans up the implementation to just use `Mapping` everywhere.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
That's a good idea but its also a preexisting problem. Let me open an issue so we remember to do that but I would prefer to not fix it in this PR (i.e., this PR's purpose should be to just get rid of `JacobianCalculator`).

boyceg:
> That's a good idea but its also a preexisting problem. Let me open an issue so we remember to do that but I would prefer to not fix it in this PR (i.e., this PR's purpose should be to just get rid of `JacobianCalculator`).

But isn't this also the PR where the class name "`Mapping`" is introduced in the first place? Why not fix it now before merging?

drwells:
> But isn't this also the PR where the class name "Mapping" is introduced in the first place? Why not fix it now before merging?

Yes and no. We previously had the following compatibility shim:

```cpp
 class NodalMapping : public Mapping<dim, spacedim>, public JacobianCalculator
```
i.e., all the `Mapping` classes we actually used also were `JacobianCalculator`s so that we could start implementing all these new `Mapping` classes without changing how we use `JacobianCalculator` in older code.

drwells:
If you think it would be easier to add those changes to this PR I will do so - I think it would be easier to keep it separate so that the inheritance changes are not mixed in with the mass renaming operations.

boyceg:
> > But isn't this also the PR where the class name "Mapping" is introduced in the first place? Why not fix it now before merging?
> 
> Yes and no. We previously had the following compatibility shim:
> 
> ```c++
>  class NodalMapping : public Mapping<dim, spacedim>, public JacobianCalculator
> ```
> 
> i.e., all the `Mapping` classes we actually used also were `JacobianCalculator`s so that we could start implementing all these new `Mapping` classes without changing how we use `JacobianCalculator` in older code.

I had missed that `Mapping` already showed up. I agree; let's change the name in another PR.

PR #1059, opened 2020-07-29
Fe values fe interpolation

### LINKED ISSUE #482: Use FECache in FEDataInterpolation

drwells:
Closes #482 :tada:

Followup to #997: the problem was that we need to support more than isoparametric finite elements in this class (which isn't too hard to do - we just need to restrict ourselves to elements with a straightforward reference element implementation.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/major/20200519DavidWells
+++ b/doc/news/changes/major/20200519DavidWells
@@ -1,0 +1,7 @@
+New: Added a new class IBTK::FEValues, which is a replacement for libMesh::FE
+for Lagrange elements that is 4-5 times faster. This is now used inside various
+parts of the IBFE stack and makes, in particular, assembling the interior force
+density much less expensive. For some models this leads to a 10% overall run
+time reduction.
+<br>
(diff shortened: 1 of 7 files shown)


doc/news/changes/major/20200519DavidWells
```diff
@@ -0,0 +1,7 @@
+New: Added a new class IBTK::FEValues, which is a replacement for libMesh::FE
```
boyceg, line 1: Is this a replacement for both codimension 0 and 1 use cases, or just codimension 0?

drwells: It works in codimension 1 but it cannot yet compute normal vectors so I think the comment is close enough.

drwells:
We will eventually compute normal vectors here so I think its fine to accept the simplified log message.

PR #1062, opened 2020-07-31
Implement a faster bounding box intersection function.

drwells:
We can get much better performance by implementing a few extra things:
1. Inline the function
2. Explicit short-circuit logic in the loop over spatial dimensions
3. Skip a fourth redundant line segment check

This seems trivial, but it makes a huge difference in models with large numbers
of elements (such as the TAVR code) that regrid frequently since we presently
call this function, for N local patches and K global elements, K*N times: i.e.,
if one processor has twice as many patches it has to do twice as much work, so
making this much cheaper has a large benefit to both serial preformance and time
spent waiting in parallel.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -211,6 +211,46 @@ get_JxW(const std::tuple<libMesh::ElemType, libMesh::QuadratureType, libMesh::Or
         return mapping.getJxW();
     }
 }
+
+// libMesh's box intersection code is slow and not in a header (i.e., cannot be
+// inlined). This is problematic for us since we presently call this function
+// for every element on every patch: i.e., for N patches on the current
+// processor and K *total* (i.e., on all processors) elements we call this
+// function K*N times, which takes up a lot of time in regrids. For example:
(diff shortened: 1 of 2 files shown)


ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2866,10 +2906,12 @@ FEDataManager::collectActivePatchElements(std::vector<std::vector<Elem*> >& acti
         {
             if ((*el_it)->active())
             {
-#if LIBMESH_VERSION_LESS_THAN(1, 2, 0)
-                if (bbox.intersect(patch_bbox)) elems.insert(*el_it);
+#if LIBMESH_VERSION_LESS_THAN(1, 6, 0)
+                if (bbox_intersects(bbox, patch_bbox)) elems.insert(*el_it);
```
drwells, line 2910: See also https://github.com/libMesh/libmesh/pull/2666.

drwells:
Good point. Fixed!

drwells:
I rebased to fix the merge conflict.

PR #1066, opened 2020-08-04
Add compatibility headers for CMake.

drwells:
Part of #1020: We need to ultimately get rid of `IBTK_config.h` and `IBAMR_config.h` since they are top-level files. This PR adds shims that are compatible with both the new CMake build system (which doesn't have top level files) to ease the transition. Upcoming PRs will remove direct `IBTK_config.h` and `IBAMR_config.h` inclusions from the library.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
Could we merge this soon-ish? Its holding back other progress on the CMake branch.

PR #1067, opened 2020-08-05
Improve copyright headers

drwells:
Related to #1020: I noticed a bunch of odd headers when working on an automatic conversion.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/examples/wave_tank/ex1/RigidBodyKinematics.cpp
+++ b/examples/wave_tank/ex1/RigidBodyKinematics.cpp
@@ -11,8 +11,6 @@
 //
 // ---------------------------------------------------------------------
 
-
-
 /////////////////////////////////// INCLUDES /////////////////////////////////////
 
 // SAMRAI INCLUDES
@@ -29,6 +27,7 @@
 // IBTK THIRD-PARTY INCLUDES
 #include "muParser.h"
 
+// APPLICATION INCLUDES
 #include "RigidBodyKinematics.h"
 
 // C++ INCLUDES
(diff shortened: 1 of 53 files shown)


examples/wave_tank/ex1/RigidBodyKinematics.cpp
```diff
@@ -58,6 +25,8 @@
 // IBTK INCLUDES
 
 // IBTK THIRD-PARTY INCLUDES
+#include "RigidBodyKinematics.h"
```
boyceg, line 28: is the class header a "third party" include?

drwells: Oops :)

drwells: Fixed.

examples/wave_tank/ex1/RigidBodyKinematics.cpp
```diff
@@ -60,6 +27,9 @@
 // IBTK THIRD-PARTY INCLUDES
 #include "muParser.h"
 
+// APPLICATION INCLUDES
+#include "RigidBodyKinematics.h"
```
boyceg, line 31: 👍 

These separated sections are leftover from the dark days before clang-format could readily keep headers grouped together automatically.

drwells: It would be nice to clean all this up, but ultimately since SAMRAI 2 doesn't put all its `#include`s in a folder we don't have a good way to distinguish between `PatchLevel.h` and `RigidBodyKinematics.h`. Hopefully this is fixed in newer versions.

PR #1069, opened 2020-08-06
Nishant/brinkman adv diff

Raamkrishnan:
Implemented a flux-based Brinkman Penalization approach which can apply Dirichlet as well as flux (homogeneous and Inhomogeneous Neumann) boundary conditions on the surface. This is implemented only in the advection-diffusion solver. Added a flag in AdvDiffSemiImplicitHierarchyIntegrator to zero out the temporal contribution to solve the Penalized Poisson equation. Also, in BrinkmanPenalizationRigidBodyDynamics class, ghost cells are filled for the entire hierarchy now. 
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


include/ibamr/BrinkmanPenalizationAdvDiff.h
```diff
@@ -0,0 +1,295 @@
...
+ * \brief BrinkmanPenalizationAdvDiff is an abstract class that provides an interface
+ * to implement Brinkman penalization body force in the advection-diffusion equation
+ * in order to enforce Dirichlet and Neumann boundary conditions on surfaces of rigid
+ * immersed bodies. A single instance of this class is meant to handle all of the Brinkman
+ * penalization zones for multiple transported quantities with various boundary conditions.
+
+ * TODO: A description of the penalization terms for both Dirichlet and Neumann BCs
+ * For Dirichlet BCs, inhomogeneous and homogeneous BCs are treated in the same way.
+ * For Neumann BCs, homogeneous BCs are fairly straightforward, while inhomogeneous BCs require
+ * some additional input by the user. Add a reference to Sakurai's paper
```
drwells, line 63: Would you please finish this TODO?

src/IB/BrinkmanPenalizationAdvDiff.cpp
```diff
@@ -0,0 +1,781 @@
...
+                        }
+                        else if (std::abs(phi) <= alpha)
+                        {
+                            Hphi = 0.5 + 0.5 * phi / alpha + 1.0 / (2.0 * M_PI) * std::sin(M_PI * phi / alpha);
+                        }
+                        else
+                        {
+                            Hphi = 1.0;
+                        }
+                        // Note: assumes that chi is positive when phi is negative
```
drwells, line 467: Could you change this to an assertion?

src/IB/BrinkmanPenalizationAdvDiff.cpp
```diff
@@ -0,0 +1,781 @@
...
```
drwells, line 648: same

include/ibamr/BrinkmanPenalizationAdvDiff.h
```diff
@@ -0,0 +1,295 @@
...
+     * The user must set the patch data B_idx such that \f$ n \dot B = \gamma \f$.
+     * The applied forcing term is then computed internally as \f$ \nabla \dot (\chi B) - \chi \nabla \dot B \f$.
+     * Note that B_idx contains side-centered patch data.
+     */
+    using BrinkmanInhomogeneousNeumannBCsFcnPtr = void (*)(int B_idx,
+                                                           SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps> hier_math_ops,
+                                                           double time,
+                                                           void* ctx);
+
+    /*
```
drwells, line 145:
```suggestion
    /*!
```

include/ibamr/BrinkmanPenalizationAdvDiff.h
```diff
@@ -0,0 +1,295 @@
...
+
+    void
+    registerDirichletHomogeneousNeumannBC(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > Q_var,
+                                          SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > ls_solid_var,
+                                          std::string bc_type,
+                                          double bc_val,
+                                          double num_interface_cells = 2.0,
+                                          double eta_penalty_coef = 1.0e-4);
+
+    /*
```
drwells, line 161:
```suggestion
    /*!
```

include/ibamr/BrinkmanPenalizationAdvDiff.h
```diff
@@ -0,0 +1,295 @@
...
+     * inputs.
+     */
+    void registerInhomogeneousNeumannBC(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > Q_var,
+                                        SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > ls_solid_var,
+                                        BrinkmanInhomogeneousNeumannBCsFcnPtr callback,
+                                        void* ctx,
+                                        double num_interface_cells = 2.0,
+                                        double eta_penalty_coef = 1.0e-4);
+
+    /*
```
drwells, line 175: etc. Please add the extra `!` so that these show up in doxygen.

drwells:
In particular: this changes the output of the wave tank test in the second significant digit.

include/ibamr/BrinkmanPenalizationAdvDiff.h
Raamkrishnan: I have added the test for BPAdvDiff which you can find in tests/adv_diff. I have also investigated the test ```wave_tank/nwt_cylinder```. We now use a more appropriate formula for interface width while computing Heaviside/smooth indicator function and this is implemented in ```BrinkmanPenalizationRigidBodyDynamics``` class which was the reason the test was failing. Now I have updated the wave_tank test output.

drwells:
Ah - if the new algorithm is better then lets use it and update the test.

Would someone who is more familiar with Brinkman penalization please give this a second look?

drwells:
@abarret I think most concerns in your review have been addressed - would you please take a second look?

Raamkrishnan:
> @abarret I think most concerns in your review have been addressed - would you please take a second look?

I am working on @abarret some of his suggestions. I will push it today evening. @abarret can take a look after that.

src/IB/BrinkmanPenalizationAdvDiff.cpp
```diff
@@ -0,0 +1,781 @@
...
+            d_div_B_chi_scratch_idx, d_div_var, 1.0, d_B_chi_scratch_idx, d_B_var, NULL, d_current_time, false);
+        hier_math_ops->div(d_div_B_scratch_idx, d_div_var, 1.0, d_B_scratch_idx, d_B_var, NULL, d_current_time, false);
+        hier_cc_data_ops.multiply(d_div_B_scratch_idx, d_chi_scratch_idx, d_div_B_scratch_idx);
+        // Ensure that we don't overwrite contributions from other BCs
+        hier_cc_data_ops.add(F_idx, F_idx, d_div_B_chi_scratch_idx);
+        hier_cc_data_ops.subtract(F_idx, F_idx, d_div_B_scratch_idx);
+
+        // Nullify patch data just to be safe
+        hier_sc_data_ops.setToScalar(d_B_scratch_idx, std::numeric_limits<double>::quiet_NaN());
+        hier_sc_data_ops.setToScalar(d_B_chi_scratch_idx, std::numeric_limits<double>::quiet_NaN());
```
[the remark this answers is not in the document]
Raamkrishnan: Now I set patch data to NAN only in debug mode. Now, I have incorporated all your suggestions @abarret. Please take a look.

boyceg:
@Raamkrishnan should @abarret take another look at this point?

abarret:
Yes, I need to revisit this. It was on my to-do list for  yesterday, but we lost power and we still don't have it back. I'll review again ASAP.

examples/adv_diff/ex3/example.cpp
```diff
@@ -0,0 +1,400 @@
...
+
+        // Brinkman penalization for Transport equation.
+        const double eta = input_db->getDouble("ETA");
+        const double num_of_interface_cells = input_db->getDouble("NUMBER_OF_INTERFACE_CELLS");
+        Pointer<BrinkmanPenalizationAdvDiff> brinkman_adv_diff =
+            new BrinkmanPenalizationAdvDiff("BrinkmanPenalizationAdvDiff", time_integrator);
+
+        // setting inhomogeneous Neumann at the cylinder surface.
+        brinkman_adv_diff->registerInhomogeneousNeumannBC(
+            q_var, phi_solid_var, &evaluate_brinkman_bc_callback_fcn, nullptr, num_of_interface_cells, eta);
```
abarret, line 218: :+1:

abarret (changes requested):
This is close. I left a few more comments, mostly about the public interface.

Two big things:

1. Do you have a reason to put `BrinkmanPenalizationAdvDiff.cpp` in the `IB` folder? Can you move this class to the `adv_diff` folder? While you're doing that, can you rename it `AdvDiffBrinkmanPenalization`? Class names should generally go from least specific (`AdvDiff`) to most specific (`BrinkmanPenalization`)
2. The `wave_tank/nwt_cylinder.mpirun=8.input` test is failing again. Can you take another look at that?

include/ibamr/BrinkmanPenalizationAdvDiff.h
```diff
@@ -0,0 +1,306 @@
...
+
+    /*!
+     * \brief Register a transported quantity with this object, along with the solid level set
+     * variable for which to apply a specified boundary condition.
+     *
+     * \note This function can only be used to register homogeneous/inhomogeneous Dirichlet BCs, and homogeneous Neumann
+     * BCs.
+     */
+    void
+    registerDirichletHomogeneousNeumannBC(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > Q_var,
```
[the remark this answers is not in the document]
boyceg: Could we use the existing BC infrastructure? I would like to avoid propagating different ways of setting BCs if possible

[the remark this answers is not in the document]
boyceg: (Or is this doing a different kind of BC setting?)

examples/adv_diff/ex3/LevelSetInitialCondition.h
```diff
@@ -0,0 +1,95 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2018 - 2018 by the IBAMR developers
```
boyceg, line 3: Can we update copyright dates?

examples/adv_diff/ex3/LevelSetInitialCondition.h
```diff
@@ -0,0 +1,95 @@
...
+                        Pointer<SAMRAI::hier::Variable<NDIM> > var,
+                        Pointer<Patch<NDIM> > patch,
+                        const double data_time,
+                        const bool initial_time = false,
+                        Pointer<PatchLevel<NDIM> > patch_level = Pointer<PatchLevel<NDIM> >(NULL)) override;
+
+    //\}
+
+private:
+    LevelSetInitialCondition();
```
boyceg, line 65:
```suggestion
    LevelSetInitialCondition() = delete;
```

examples/adv_diff/ex3/LevelSetInitialCondition.h
```diff
@@ -0,0 +1,95 @@
...
+     */
+    LevelSetInitialCondition(const std::string& object_name,
+                             const double radius,
+                             const IBTK::VectorNd& origin,
+                             const bool fluid_is_interior_to_cylinder);
+
+    /*!
+     * \brief Empty destructor.
+     */
+    ~LevelSetInitialCondition();
```
boyceg, line 44:
```suggestion
    ~LevelSetInitialCondition() = default;
```

include/ibamr/AdvDiffHierarchyIntegrator.h
```diff
@@ -474,6 +487,21 @@ class AdvDiffHierarchyIntegrator : public IBTK::HierarchyIntegrator
      */
     int getResetPriority(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > Q_var) const;
 
+    /*!
+     * \brief Register BrinkmanPenalizationAdvDiff object to add Brinkman penalization terms to the advection-diffusion
```
boyceg, line 491: The way that we handle this for IB methods for FSI is to have the IB solver tell the field solver about forces through a `CartGridFunction`. That way, the Navier-Stokes solver is oblivious to the details of the IB formulation.

Is it possible to use the same design pattern here?

include/ibamr/BrinkmanPenalizationAdvDiff.h
```diff
@@ -0,0 +1,306 @@
...
+     * \brief Function to compute the Brinkman forcing contribution to the RHS of the advection-diffusion solver
+     * for a specified transported quantity Q_var.
+     *
+     * For Dirichlet BCs, \f$ F_i = \chi_i/\eta Q_{bc}\f$ where \f$\chi_i = 1-H_i\f$.
+     * For homogeneous Neumann BCs, \f$ F_i = 0\f$.
+     * For inhomogeneous Neumann BCs, \f$ F_i = \nabla \dot (\chi_i B) - \chi_i \nabla \dot B_i \f$, with a user defined
+     * \f$B_i\f$.
+     * The overall functional form of the Brinkman body force is \f$F = \sum_{i} F_i\f$.
+     */
+    void computeForcing(int F_idx, SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > Q_var);
```
boyceg, line 222: For instance, in the analogy to IB for FSI, this forcing could be passed to the advection-diffusion solver though the `CartGridFunction` interface without having to tell the `AdvDiffHierarchyIntegrator` class about the Brinkmann class.

include/ibamr/AdvDiffHierarchyIntegrator.h
abarret: Possibly? I'm not entirely sure. This method requires masking the diffusion coefficient and the convective operator, so the solvers need to be adjusted prior to each cycle.

boyceg: In that case, would it make sense for this to be a subclass of `AdvDiffSemiImplicitHierarchyIntegrator`?

abarret: I think that would be the best option.

boyceg: It looks like the methods that would be needed in the subclass are mostly already provided by the Brinkmann class, and so it shouldn't be much work to move the code from `AdvDiffSemiImplicitHierarchyIntegrator` to the Brinkmann class.

@Raamkrishnan, is it clear what is being suggested and does it make sense? Before doing anything --- @drwells, do you have an opinion?

include/ibamr/BrinkmanPenalizationAdvDiff.h
boyceg: When I made this comment I did not realize that this involved changing the linear operator, and so this suggestion is not relevant.

include/ibamr/BrinkmanPenalizationAdvDiff.h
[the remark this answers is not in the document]
amneetb: This is imposing BCs on embedded interfaces and *not* on the domain boundaries; so it is an entirely different thing. We do provide a description in the class header.

amneetb:
> A couple of comments:
> 
> 1. It seems like Brinkmann penalization is only implemented for `AdvDiffSemiImplicitHierarchyIntegrator`. If that is right, then I think that either the other implementations need to generate an error message if Brinkmann penalization is set. Alternatively, only `AdvDiffSemiImplicitHierarchyIntegrator` should know about Brinkmann at this time. (@drwells do you have an opinion?)

This class needs to solve diffusion equations implicitly, hence  `AdvDiffSemiImplicitHierarchyIntegrator`.

> 2. In this branch, there is now fairly tight coupling between the Brinkmann class and `AdvDiffSemiImplicitHierarchyIntegrator`. It seems like the Brinkmann class is actually changing the way that `AdvDiffSemiImplicitHierarchyIntegrator` works in a fairly substantial way. I wonder if it would be easier to maintain if it is a subclass of `AdvDiffSemiImplicitHierarchyIntegrator` that overrides the functionality that should change. (Again, I would ask @drwells for an opinion.)

This is not true. This class *does not* touch `AdvDiffSemiImplicitHierarchyIntegrator` class. It just places pieces of code related to BrinkmanPenalization at appropriate places in `AdvDiffSemiImplicitHierarchyIntegrator` class.

include/ibamr/BrinkmanPenalizationAdvDiff.h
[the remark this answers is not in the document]
boyceg: Perhaps the comment here (and in related functions) could state that it is specifying the BC along the interface determined by the level set variable.

Also, instead of having separate functions for Dirichlet, Neumann, and homogeneous/inhomogenous, how about just one function that takes in a `bc_type` `enum` and the callback, but if the callback is `NULL` treat as homogeneous?

include/ibamr/AdvDiffHierarchyIntegrator.h
amneetb: `CartGridFunction` are not useful, unless everything could be described analytically, which never happens in practice and stuff related to level sets. One needs to do some (complex) computations to arrive at a forcing field. This computation is usually integrated with time-stepping scheme as well.

include/ibamr/BrinkmanPenalizationAdvDiff.h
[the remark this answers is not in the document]
boyceg: > > A couple of comments:
> > 
> > 1. It seems like Brinkmann penalization is only implemented for `AdvDiffSemiImplicitHierarchyIntegrator`. If that is right, then I think that either the other implementations need to generate an error message if Brinkmann penalization is set. Alternatively, only `AdvDiffSemiImplicitHierarchyIntegrator` should know about Brinkmann at this time. (@drwells do you have an opinion?)
> 
> This class needs to solve diffusion equations implicitly, hence `AdvDiffSemiImplicitHierarchyIntegrator`.

Sure, that is fine --- but in that case, I am asking why it makes sense for `AdvDiffHierarchyIntegrator` to know about the Brinkmann object. It seems more natural just to have the `AdvDiffSemiImplicitHierarchyIntegrator` know about the Brinkmann object.

If you want `AdvDiffHierarchyIntegrator` to know about the Brinkmann object, then it seems like the default implementation should be that it emits an error message if you try to use it with something other than `AdvDiffSemiImplicitHierarchyIntegrator`.

boyceg:
> > A couple of comments:
> > 
> > 1. It seems like Brinkmann penalization is only implemented for `AdvDiffSemiImplicitHierarchyIntegrator`. If that is right, then I think that either the other implementations need to generate an error message if Brinkmann penalization is set. Alternatively, only `AdvDiffSemiImplicitHierarchyIntegrator` should know about Brinkmann at this time. (@drwells do you have an opinion?)
> 
> This class needs to solve diffusion equations implicitly, hence `AdvDiffSemiImplicitHierarchyIntegrator`.

Sure, that is fine --- but in that case, I am asking why it makes sense for AdvDiffHierarchyIntegrator to know about the Brinkmann object. It seems more natural just to have the AdvDiffSemiImplicitHierarchyIntegrator know about the Brinkmann object.

If you want AdvDiffHierarchyIntegrator to know about the Brinkmann object, then it seems like the default implementation should be that it emits an error message if you try to use it with something other than AdvDiffSemiImplicitHierarchyIntegrator.

amneetb:
> This is close. I left a few more comments, mostly about the public interface.
> 
> Two big things:
> 
> 1. Do you have a reason to put `BrinkmanPenalizationAdvDiff.cpp` in the `IB` folder?

This is imposing BCs on embedded/immersed interfaces. It more appropriate to put it in IB folder. 
 
 While you're doing that, can you rename it `AdvDiffBrinkmanPenalization`? Class names should generally go from least specific (`AdvDiff`) to most specific (`BrinkmanPenalization`)

This is not related to AdvDiff system, but a helper object. It is a separate system of classes related to Brinkman penalization techniques. We have BrinkmanPenalizationRigidBodyDynamics, BrinkmanPenalizationAdvDiff, more to come in the future. 

> 2. The `wave_tank/nwt_cylinder.mpirun=8.input` test is failing again. Can you take another look at that?

amneetb:
> > > A couple of comments:
> > > 
> > > 1. It seems like Brinkmann penalization is only implemented for `AdvDiffSemiImplicitHierarchyIntegrator`. If that is right, then I think that either the other implementations need to generate an error message if Brinkmann penalization is set. Alternatively, only `AdvDiffSemiImplicitHierarchyIntegrator` should know about Brinkmann at this time. (@drwells do you have an opinion?)
> > 
> > 
> > This class needs to solve diffusion equations implicitly, hence `AdvDiffSemiImplicitHierarchyIntegrator`.
> 
> Sure, that is fine --- but in that case, I am asking why it makes sense for `AdvDiffHierarchyIntegrator` to know about the Brinkmann object. It seems more natural just to have the `AdvDiffSemiImplicitHierarchyIntegrator` know about the Brinkmann object.

Yes, it should be moved to `AdvDiffSemiImplicitHierarchyIntegrator`. 
> 
> If you want `AdvDiffHierarchyIntegrator` to know about the Brinkmann object, then it seems like the default implementation should be that it emits an error message if you try to use it with something other than `AdvDiffSemiImplicitHierarchyIntegrator`.

include/ibamr/AdvDiffHierarchyIntegrator.h
boyceg: I disagree with the statement that `CartGridFunction` is not useful. :-)

When I was looking at the interface changes, I did not realize that the solver was being changed under the hood, and so the explicit coupling through forcing functions is not what you want to do.

boyceg:
> > 2. In this branch, there is now fairly tight coupling between the Brinkmann class and `AdvDiffSemiImplicitHierarchyIntegrator`. It seems like the Brinkmann class is actually changing the way that `AdvDiffSemiImplicitHierarchyIntegrator` works in a fairly substantial way. I wonder if it would be easier to maintain if it is a subclass of `AdvDiffSemiImplicitHierarchyIntegrator` that overrides the functionality that should change. (Again, I would ask @drwells for an opinion.)
> 
> This is not true. This class _does not_ touch `AdvDiffSemiImplicitHierarchyIntegrator` class. It just places pieces of code related to BrinkmanPenalization at appropriate places in `AdvDiffSemiImplicitHierarchyIntegrator` class.

It looks like the linear solver and timestepping both change when Brinkmann penalization is enabled. Is that right?

include/ibamr/AdvDiffHierarchyIntegrator.h
amneetb: BrinkmanPenalizationAdvDiff is a helper object that helps modify solver coefficients, when penalization is desired. It is not an integrator, hence we did not subclass from AdvDiffSemiImplicitHI.

abarret: > I disagree with the statement that `CartGridFunction` is not useful. :-)

(The complex fluid solver is completely built on top of `CartGridFunction`)

amneetb:
> > > 1. In this branch, there is now fairly tight coupling between the Brinkmann class and `AdvDiffSemiImplicitHierarchyIntegrator`. It seems like the Brinkmann class is actually changing the way that `AdvDiffSemiImplicitHierarchyIntegrator` works in a fairly substantial way. I wonder if it would be easier to maintain if it is a subclass of `AdvDiffSemiImplicitHierarchyIntegrator` that overrides the functionality that should change. (Again, I would ask @drwells for an opinion.)
> > 
> > 
> > This is not true. This class _does not_ touch `AdvDiffSemiImplicitHierarchyIntegrator` class. It just places pieces of code related to BrinkmanPenalization at appropriate places in `AdvDiffSemiImplicitHierarchyIntegrator` class.
> 
> It looks like the linear solver and timestepping both change when Brinkmann penalization is enabled. Is that right?

Linear solver yes, time stepping no. Some terms are organized at different places because solver coefficients depend upon the updated values of other scalars.

include/ibamr/AdvDiffHierarchyIntegrator.h
amneetb: So is our turbulence code. But we can not use them at *all*  places.

boyceg:
> > > > 1. In this branch, there is now fairly tight coupling between the Brinkmann class and `AdvDiffSemiImplicitHierarchyIntegrator`. It seems like the Brinkmann class is actually changing the way that `AdvDiffSemiImplicitHierarchyIntegrator` works in a fairly substantial way. I wonder if it would be easier to maintain if it is a subclass of `AdvDiffSemiImplicitHierarchyIntegrator` that overrides the functionality that should change. (Again, I would ask @drwells for an opinion.)
> > > 
> > > 
> > > This is not true. This class _does not_ touch `AdvDiffSemiImplicitHierarchyIntegrator` class. It just places pieces of code related to BrinkmanPenalization at appropriate places in `AdvDiffSemiImplicitHierarchyIntegrator` class.
> > 
> > 
> > It looks like the linear solver and timestepping both change when Brinkmann penalization is enabled. Is that right?
> 
> Linear solver yes, time stepping no. Some terms are organized at different places because solver coefficients may depend upon the updated values of other scalars.

There is an existing interface that is intended to facilitate changes to the linear solvers and linear operators: `AdvDiffHierarchyIntegrator::setHelmholtzSolver()` and `AdvDiffHierarchyIntegrator::setHelmholtzRHSOperator()`. Could this do what you need if it was extended to know, for instance, that in some cases the solvers / operators need to be reset before each time step or time step cycle?

What I would like to try to minimize is logic in the class that depends on whether or not Brinkmann penalization is enabled or not.

include/ibamr/AdvDiffHierarchyIntegrator.h
amneetb: > I disagree with the statement that `CartGridFunction` is not useful. :-)
> 
> When I was looking at the interface changes, I did not realize that the solver was being changed under the hood, and so the explicit coupling through forcing functions is not what you want to do.

Yes, we are modifying solver coefficients, masking convection and other forces in the fictitious region, and in case of flux BCs, we are computing a forcing term that in general is a complex function of level set and other scalars. Moreover, there could be some interfaces that have Dirichlet BC and some may have Neumann BCs; all of this taken care by this helper class.

drwells:
> 1. It seems like Brinkmann penalization is only implemented for AdvDiffSemiImplicitHierarchyIntegrator. If that is right, then I think that either the other implementations need to generate an error message if Brinkmann penalization is set. Alternatively, only AdvDiffSemiImplicitHierarchyIntegrator should know about Brinkmann at this time. (@drwells do you have an opinion?)
> 2. In this branch, there is now fairly tight coupling between the Brinkmann class and AdvDiffSemiImplicitHierarchyIntegrator. It seems like the Brinkmann class is actually changing the way that AdvDiffSemiImplicitHierarchyIntegrator works in a fairly substantial way. I wonder if it would be easier to maintain if it is a subclass of AdvDiffSemiImplicitHierarchyIntegrator that overrides the functionality that should change. (Again, I would ask @drwells for an opinion.)

I am not that familiar with this part of the library - I'll have to take a close look to see what is going on. In general I don't think its a good idea to substantially alter existing classes in incompatible ways - we might be able to get away with it if its just us using them but its going to make our users really unhappy if they upgrade and things stop working in strange ways.

One good way around this would be to start splitting up this PR - 4,000 lines is quite a lot.

boyceg:
At any rate, when I put together these interfaces, the idea was to make the solvers extensible so that users do not have to modify `AdvDiffSemiImplicitHierarchyIntegrator` to get it to do what they want it to do.

My goal here is to see if we can make some _*small*_ (or at least small-ish) changes to the interface to `AdvDiffSemiImplicitHierarchyIntegrator` and/or `AdvDiffHierarchyIntegrator` so that this can be implemented either by using the (possibly extended) interface, or by subclassing, but not by injecting implementation-specific logic into `AdvDiffSemiImplicitHierarchyIntegrator`. One reason to try to do this is to try to ensure that users also are able to extend functionality with the existing interface, not by forking and hacking the "base" classes. If it looks like a major roadblock, then we can proceed for now with the current approach of having `AdvDiffSemiImplicitHierarchyIntegrator` know about the Brinkmann class.

drwells (changes requested):
The `nwt_cylinder` test is still failing so we need to look at that.

This PR is much easier to read in whitespace-independent mode (i.e., append `?w=1` to the URL). Most of the substantial changes are to the `AdvDiffSemiImplicitHierarchyIntegrator` class but these are backwards-compatible: essentially all the changes are guarded by `if (apply_brinkman)`.

Like @boyceg I think that this approach is probably not ideal for the future. There is just too much Brinkman-specific information inside a more abstract class (`AdvDiffSemiImplicitHierarchyIntegrator`). I don't use level set or Brinkman penalization but after reading through the code I don't see why we have both `BrinkmanPenalizationAdvDiff` and those changes to the existing class - neither object seems to be completely responsible for actually doing Brinkman calculations which is not a good design. In principle we should have a clear API for adding penalization terms fo `AdvDiffSemiImplicitHierarchyIntegrator` instead of hard-coding in one of them.

I agree with everything else @boyceg wrote in https://github.com/IBAMR/IBAMR/pull/1069#issuecomment-690798948.

drwells:
In particular the current implementation changes the output to have a nonzero rotational velocity, but an original goal of the test was to show that we achieved zero rotation.

Raamkrishnan:
> In particular the current implementation changes the output to have a nonzero rotational velocity, but an original goal of the test was to show that we achieved zero rotation.

The implementation of ```Brinkman penalizationAdvDiff``` was not affecting the wave tank test. There was a bug in ```BrinkmanPenalizationRigidBodyDynamics``` class while storing the rotational velocity in vector. I fixed the bug in my last commit but that was correct only for 3D. That is why the wave tank test was failing. Now I corrected that bug.

Raamkrishnan:
@drwells: We could not think of any other design now. would you like to design it better?

amneetb:
We can mark this PR as non-ibfe/non-essential/WIP and revisit later. We have many such PRs already like peridynamics, mls, ...

amneetb:
When reviewers add tag like “non-novel”, “not good design” it is a death sentence, and it leads to rejection of papers/PRs. Reviewers gotta review.

boyceg:
I think you misunderstand my comments. The base class is intended to be extensible — so that its functionality can be modified without directly modifying the implementation. If its design doesn’t allow that then we should try to fix the base class.

I pointed out a few mechanisms above that were intended to allow for the code to be extended and asked whether they were sufficient or not. I find your response of “reviewers gotta review” to be extremely insulting.

boyceg:
> @drwells: We could not think of any other design now. would you like to design it better?

@Raamkrishnan, there are a couple of mechanisms that are included in the design of these classes that are intended to make them extensible without directly modifying the base class.

One is just through inheritance --- e.g. make a subclass of `AdvDiffSemiImplicitHierarchyIntegrator` that changes the functionality that needs to be changed, and leaves everything else alone. Many of the key methods are marked `virtual` so that you can reimplement them in a subclass. You can always call the base class implementation from the subclass. E.g.,
```
BrinkmanAdvDiffSemiImplicitHierarchyIntegrator::preprocessIntegrateHierarchy(...)
{
    // do custom stuff before calling the base class method
    ...

    // call the base class method
    AdvDiffSemiImplicitHierarchyIntegrator::preprocessIntegrateHierarchy(...);

    // do custom stuff after calling the base class method
    ...
}
```
Looking at the changes to `AdvDiffSemiImplicitHierarchyIntegrator` that are in this PR, I think that it should be easy to accomplish what you want by making `BrinkmanPenalizationAdvDiff` a subclass of `AdvDiffSemiImplicitHierarchyIntegrator`.

The other way --- which I think I originally added many years ago for @amneetb :-) --- is by providing callback functions to customize the functionality. The `HierarchyIntegrator` base class lets you specify C-style functions that are called at the end of `preprocessIntegrateHierarchy()`, `integrateHierarchy()`, and `postprocessIntegrateHierarchy()`. This is less flexible, since you cannot easily control when within the routine the custom code is called, but avoids the need to create a derived class. These callbacks can be used with any class that derives from `HierarchyIntegrator`.

In addition, for the `AdvDiffHierarchyIntegrator` classes, it is possible to swap out the linear solvers and RHS operators via `setHelmholtzSolver()` and `setHelmholtzRHSOperators`. There are methods that are specifically intended to force those solvers to be reinitialized, although it may not be exactly what you need in this case.

The question that I asked above is whether this extension of `AdvDiffSemiImplicitHierarchyIntegrator` could be reasonably implemented using that existing functionality that is intended to enable extensibility. If so, I don't think it should require a lot of new code to be written, but instead should mostly consist of reorganizing the code that you already have written.

If this extension cannot be reasonably implemented using the current interfaces for the base class, then that is also important to know, because it means that we need to think about how to improve the base classes to facilitate extension --- which would be something to address at some point in the future, and not in this PR.

Does this make sense?

boyceg:
BTW, to be clear, I am not asking for changes — I’m asking if it is possible to accomplish what you are doing through one of these mechanisms.

amneetb:
inheritance is most obvious candidate here. But whatever we are going to do through inheritance will be exactly what we did in AdvDiffSHI. The new integrator will have some scalars that are not penalized and some scalars that are penalized. So the implementation would be exact same as we did here, albeit the name of the name of the class would be different. Moreover, BP has the advantage that it requires least changes in the existing code to do IB stuff. Furthemore, AdvDiffSHI has a lambda penalty parameter,  which is nothing but a BrinkmanPenalization term. So the existing class already has Brinkman penalization term a decade ago. What we did was formally extended the mechanism through a helper class that encapsulates all penalization terms systematically.

amneetb:
FYI: it was discussed separately that my comment on Reviewers reviving without reading paper holds truth as a separate working paper discussing the theory of this PR was sent before asking the PR to be reviewed well in advance. Because everyone’s busy schedule it got missed and the PR was reviewed without reading the working theory. This has lead to frustration on both ends, as things like modifying linear solvers etc were missed and a time consuming debate on what was being done in the PR started. Moreover, the prior review comments were addressed that did not mention anything about design. The authors of this PR addressed those comments and moved onto different project, only to find new comments on the PR.

PR #1072, opened 2020-08-07
Remove the dX system.

### LINKED ISSUE #1071: Remove dX system?
A drawback to using `ReplicatedMesh` is that every single dof is allocated on every single processor - i.e., every time we add a system to the mesh we have to allocate an amount of memory proportional to the number of processors.

We only ever use this system to calculate the vector indicating the displacement of the mesh - i.e., all the information we need in this system is already available in `X`. Similarly, if for some reason someone wants to plot this, we can use this same information to recover this vector.

I think we are at the point where this is necessary to consider since the heart mesh is big enough that we are having trouble allocating everything we need - the more we can cut out of the mesh, the better.

drwells:
Fixes #1071 - I'm testing it now to see how much of a difference this makes.

Still WIP since this causes tests that print out dof counts to (as expected) print about 10% fewer dofs. I should also take a closer look at the meter code to figure out if I broke anything since that area is under-tested.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/FEMechanicsBase.cpp
+++ b/src/IB/FEMechanicsBase.cpp
@@ -892,30 +892,36 @@ FEMechanicsBase::updateCoordinateMapping(const unsigned int part)
     EquationSystems& equation_systems = *d_fe_data[part]->getEquationSystems();
     MeshBase& mesh = equation_systems.get_mesh();
     auto& X_system = equation_systems.get_system<ExplicitSystem>(COORDS_SYSTEM_NAME);
-    const unsigned int X_sys_num = X_system.number();
-    NumericVector<double>& X_coords = *X_system.solution;
-    auto& dX_system = equation_systems.get_system<ExplicitSystem>(COORD_MAPPING_SYSTEM_NAME);
-    const unsigned int dX_sys_num = dX_system.number();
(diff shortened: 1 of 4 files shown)


drwells:
This lowers memory usage, with 12 processors on my workstation, from about 36.3 GB to 35.1 GB: not a ton but every little bit helps.

boyceg:
Seems like a good idea. As long as initial and current coordinates are available, we can omit "dX". We need to make sure that both are stored in the viz files.

drwells:
Would initial coordinates ever be different from the coordinates at timestep 0? Those are definitely stored.

boyceg:
I am not sure if this is answering your question, but initial coordinates can be different from mesh coordinates.

boyceg:
Also, AFAIK, it is not easy to access solution values at different time steps in VisIt --- so using the initial value of the current position to compute displacements could be tricky.

drwells:
I don't know of any way to save a libMesh vector with a visualization format outside of what they provide for system solutions so I don't think that we have a way to resolve this problem - i.e., there is no good way to visualize this data without actually allocating it. Hence lets close this for now.

PR #1075, opened 2020-08-13
Vector vector to multiarray

drwells:
A decent chunk of our total memory usage (something like ~70~ 150 MB out of 2000 MB) per processor is taken up by allocating tons of these vectors. We can get this somewhat under control by using a denser data structure.

Unfortunately we do need to keep `SystemDofMapCache` for now since it is much more efficient than the native libMesh methods.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

Edit: I misread massif's memory profiling. The number has been updated.

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -1881,7 +1881,6 @@ FEDataManager::interpWeighted(const int f_data_idx,
                               F_rhs_concatenated.begin() + (var_n + 1) * n_basis,
                               F_rhs.get_values().begin());
 
-                    dof_id_scratch = F_dof_indices[var_n];
                     // We do *not* apply constraints here. See the note in the
                     // documentation of this function for an explanation.
                     if (is_ghosted)
@@ -1898,7 +1897,8 @@ FEDataManager::interpWeighted(const int f_data_idx,
                     }
                     else
                     {
-                        F_vec.add_vector(F_rhs, F_dof_indices[var_n]);
+                        copy_dof_ids_to_vector(var_n, F_dof_indices, dof_id_scratch);
+                        F_vec.add_vector(F_rhs, dof_id_scratch);
                     }
                 }
                 qp_offset += n_qp;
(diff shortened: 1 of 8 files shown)


boyceg:
I cannot remember if we try to re-use these for `System`s with identical DOF layouts. If we are not doing that now, is there a reasonable way to do so?

drwells:
We don't do this at the moment. I think its technically possible to implement - we could have `FEData` keep track of what `SystemDofMapCache` objects have been created and if we request one for a system identical to another system then we can avoid making a duplicate.

I'm not sure if libMesh offers any guarantees on whether or not two identical systems will have the same dofs or not but its worth a try.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -329,7 +332,10 @@ FEProjector::buildDiagonalL2MassMatrix(const std::string& system_name)
                 // for spread forces and other places where we already have a
                 // finite element solution vector and are not solving a linear
                 // system.
-                M_vec->add_vector(M_e_vec, dof_indices[var_num]);
+                dof_id_scratch.clear();
```
boyceg, line 335: It is too bad that we have to have these verbose set of statements to reset the `dof_id_scratch` vectors. If `dof_id_scratch` already is the right size, could we do a `copy` instead?

boyceg:
BTW, are we traversing the data structures in the correct order now?

ibtk/src/lagrangian/FEProjector.cpp
drwells: Unfortunately the constraint code will resize the vector, so we don't have any guarantees that the scratch vector will have the right size at this point.

boyceg: It seems common enough that it might be worth having a helper function to turn this in to a self-documenting one-liner?

drwells: Let me see if that ends up being shorter - I'll give it a try.

drwells: That is better.

drwells:
> BTW, are we traversing the data structures in the correct order now?

What do you mean?

boyceg:
Column or row major order as appropriate?

drwells:
I think so - this is the same order as before and since the interpolate functions are now templated they work with the older format too.

A few functions (like `interpolate`) expect the data in a different order, but it looks like we just transpose things appropriately.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2174,11 +2174,11 @@ FEDataManager::restrictData(const int f_data_idx,
                     F_rhs_e[axis](k) += F_qp * phi_F[k][qp];
                 }
             }
-            for (unsigned int i = 0; i < n_vars; ++i)
+            for (unsigned int var_n = 0; var_n < n_vars; ++var_n)
             {
-                dof_id_scratch = F_dof_indices[i];
-                F_dof_map.constrain_element_vector(F_rhs_e[i], dof_id_scratch);
-                F_rhs_vec->add_vector(F_rhs_e[i], dof_id_scratch);
+                copy_dofs_to_vector(var_n, F_dof_indices, dof_id_scratch);
```
boyceg, line 2179: 👍

boyceg: except ---  sorry to be pedantic ---`copy_dof_ids_to_vector`? these are the DOF indices, not the DOFs themselves . . .

drwells: Changed!

PR #1076, opened 2020-08-14
Reduce the total number of SystemDofMapCaches.

drwells:
The dofs are always the same if we have the same FEType.

Followup to #1075.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -256,7 +256,7 @@ protected:
     /*!
      * Mapping between system numbers and SystemDofMapCache objects.
      */
-    std::map<unsigned int, std::unique_ptr<SystemDofMapCache> > d_system_dof_map_cache;
+    std::map<std::pair<unsigned int, libMesh::FEType>, std::unique_ptr<SystemDofMapCache> > d_system_dof_map_cache;
 
     /**
      * Permit FEDataManager to directly examine the internals of this class.
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -331,10 +331,25 @@ FEData::getDofMapCache(const std::string& system_name)
 FEData::SystemDofMapCache*
 FEData::getDofMapCache(unsigned int system_num)
 {
-    std::unique_ptr<FEData::SystemDofMapCache>& dof_map_cache = d_system_dof_map_cache[system_num];
+    System& system = d_es->get_system(system_num);
+    const unsigned int n_vars = system.n_vars();
+    TBOX_ASSERT(n_vars > 0);
+    const FEType& fe_type = system.variable_type(0);
+
+    // we assume that all variables have the same FEType inside IBAMR
+    for (unsigned int var_n = 1; var_n < n_vars; ++var_n)
+    {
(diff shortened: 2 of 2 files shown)


boyceg:
A `System` can have different `FEType`s for different variables --- should this be templated over something like `vector<FEType>` instead of `FEType`?

drwells:
It can, but we never actually do it inside IBAMR - thats part of the reason why we can switch to using `boost::multi_array` instead of `std::vector<std::vector<dof_id_type>>` internally.

boyceg:
OK, great, but the actual DOF layout depends on the number of components --- is that captured in the interface?

drwells:
I don't think that we ever use multiple components but I'll add that too.

boyceg:
Sure we do --- force, velocity, displacement all have `NDIM` components.

drwells:
I think we just have `NDIM` variables, which is not the same thing - I don't totally understand what components are but I don't think we ever use them.

boyceg:
No, in the code, some fields are scalars (one variable), some are vectors (`NDIM` variables), some are tensor (`NDIM*NDIM`).

boyceg:
(At least in the way that we use libMesh --- they introduced vector-valued FEs after I wrote most of `IBFEMethod`.)

drwells:
I think we are saying the same thing here, namely that we have 1, NDIM, or NDIM*NDIM variables in a system. Since we only have scalar variables the number of components is always the same as the number of variables (I think).

Let me double check to make sure I understand what components are.

boyceg:
OK, great --- when I say "components" I really mean "variables".

drwells:
Ah.

I think that if we do, e.g., Stokes, then we have two components but NDIM + 1 variables, but I want to check.

drwells:
No, that's not it - now I am not sure at all what components are (but we don't use them so its not critical that we work this out). For example,

```cpp
unsigned int System::n_components() const
{
  if (_variables.empty())
    return 0;

  const Variable & last = _variables.back();
  return last.first_scalar_number() + last.n_components();
}
```
where I think `first_scalar_number()` will be 0 and `n_components()` is `1` unless we are using the SCALAR family (which we do not). Hence I think we always have one component?

boyceg:
Sorry I messed up my libMesh nomenclature. Internally we always allocate some number of "variables", each with the same FE type. I am not sure what libMesh means by "components". At any rate, I am 99% sure that the DOF indexing will be the same for any system as long as the `FEType` is the same for corresponding variables.

Because the use case that we are trying to make fast is the case where all of the `FEType`s are the same, it should work to key off the number of *variables* + the common `FEType` for each variable.

Alternatively, it seems possible to key off of something like `vector<FEType>`, with `fe_type[n]` being the `FEType` for the `n`th variable.

drwells:
> Because the use case that we are trying to make fast is the case where all of the FETypes are the same, it should work to key off the number of variables + the common FEType for each variable.

Yup - this is what we do (the key is `std::pair<unsigned int, FEType>` where the first entry is the number of variables).

> Alternatively, it seems possible to key off of something like vector<FEType>, with fe_type[n] being the FEType for the nth variable.

This would also be possible but I don't think we should bother at the moment - we have no reason at the moment to have a single system with multiple `FEType`s and since this is a class that is only used with `System` objects created internally its not a big deal to ignore this case.

Issue #802, opened 2019-12-18
Put all configuration info in a single header


drwells:
There are a few places where we conditionally include different libMesh headers based on which version of libMesh we are using. This means that we need to include `libmesh_config.h` before the version check. The best way to do this would be to write a header that contains the configuration headers of all dependencies and put it at the top of each header and source file so that version information is always available.

boyceg:
This could also make it easier to switch build systems. Can we do this ASAP after tagging 0.6?

In the past, I made a concerted effort to avoid including these generated config files in IBAMR header files. What are "best practices" for this kind of thing?

knepley:
We create these files:

  makefile configure header arch-master-debug/lib/petsc/conf/petscvariables
  makefile configure header arch-master-debug/lib/petsc/conf/petscrules
  configure header arch-master-debug/include/petscconf.h
  C specific configure header arch-master-debug/include/petscfix.h
  configure pkg header arch-master-debug/include/petscpkg_version.h

petscconf.h is included by both C and Fortran headers, but petscfix.h only by C headers. The version header includes the versions of packages we installed and macros for checking them because some packages (Hypre) need to check this at compile time.

drwells:
I prefer the approach that PETSc uses in `petscpkg_version.h`; we do approximately the same thing in deal.II but put everything in one header (see https://github.com/dealii/dealii/blob/master/include/deal.II/base/config.h.in).

I would prefer to extract version information in the build system (we have it at that point anyway) and then put it with our own `#define`s in `IBTK_config.h`.

drwells:
We also need to include `libmesh/libmesh_version.h` to get the `LIBMESH_VERSION_LESS_THAN` macro.

drwells:
Lets get to this sooner, rather than later, so that we can start transitioning to a new build system. I will put all configuration information in `ibtk/config.h` and `#include` that in older headers for compatibility. The top level `IBTK_config.h` and `IBAMR_config.h`, `ibtk_macros.h`, `ibamr_macros.h`, `IBTK_CHKERRQ.h`, etc. all need to be absorbed by that header.

drwells:
This is now done for CMake - we just need to backport it (or a shim) to autotools.

drwells:
We are at the point now where `ibtk/config.h` and `ibamr/config.h` satisfy the original goals of this issue.

PR #1079, opened 2020-08-18
Add new config files 2

drwells:
Part of #1020 - CMake makes it a lot simpler to put all the macros and other things determined by the build system into a single header. I would prefer to put everything in one configuration file and have that always be available (i.e., we include it in all headers) - now would be a good time to decide if we want to do something different.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
Also part of #802.

.clang-format
```diff
@@ -54,7 +54,7 @@ UseTab: Never
 IncludeBlocks: Regroup
 IncludeCategories:
 # first our config files:
-  - Regex: "(IBTK|IBAMR)_config.h"
```
boyceg, line 57: Do we have compatibility headers named `IBTK_config.h` and `IBAMR_config.h`? Could this cause problems on OSes that do not use case sensitive file systems by default (e.g. MacOS)?

config/config.h.tmp
```diff
@@ -20,10 +20,11 @@
 
 #include <IBAMR_config.h>
 
+// Ultimately, all IBAMR macros will be aliases of the IBTK macros
```
boyceg, line 23: Are you planning to keep IBTK as a separate package? Should we just roll everything into IBAMR and provide backward compatibility by making IBTK macros aliases to IBAMR macros?

ibtk/include/ibtk/BoxPartitioner.h
```diff
@@ -11,11 +11,15 @@
 //
 // ---------------------------------------------------------------------
 
+/////////////////////// INCLUDE GUARD ////////////////////////////////////
+
 #ifndef included_IBTK_ibtk_boxpartitioner
```
boyceg, line 16:
```suggestion
#ifndef included_IBTK_boxpartitioner
```
?

.clang-format
boyceg: Oops, I misread this --- the new header is `ibtk/config.h` --- so it should be fine.

drwells: They still exist (and will exist as long as the autotools build system is around) but they are not in the CMake build system (these top-level config files don't work with `make install`). I removed them here since nothing in the library will include them once I am done with this round of patches.

ibtk/include/ibtk/AppInitializer.h
```diff
@@ -11,11 +11,15 @@
 //
 // ---------------------------------------------------------------------
 
+/////////////////////// INCLUDE GUARD ////////////////////////////////////
```
boyceg, line 14: I hate making this comment, but ...

I think that all of the comments in this style have the same width, and have the text start from the same column. Is it easy to stick with that?

config/config.h.tmp
drwells: They are no longer separate projects (as far as the build system is concerned) in the CMake build system, so no (though the directory structure is unchanged). Since IBAMR depends on IBTK I set up the macros for IBTK first.

We can do it either way but it doesn't really impact this patch - that's a change that will be made in the cmake code.

ibtk/include/ibtk/BoxPartitioner.h
drwells: How weird - I'll add a fix for that.

ibtk/include/ibtk/AppInitializer.h
drwells: This would be pretty easy to fix with sed: I think we have two different lengths in the headers. Do you want the whole thing to be 80 characters or something else?

config/config.h.tmp
boyceg: The original idea was that IBTK provided lower level discretization tools that could be used to put together different kinds of "immersed" models, with IBAMR being the implementation of a "standard" IB method. For instance, at one point, I actively used an electrophysiology simulation code that used IBTK but not IBAMR.

I've tried to maintain this separation, in which more generic things wind up in IBTK, but I am not sure whether it is really a useful distinction.

drwells: We might want to continue this discussion in #1020 - this isn't the way its set up right now but it would not be hard to split them off since the CMake build system defines `IBTK2d` and `IBTK3d` as distinct targets with their own properties, so in principle they could be split off and installed separately pretty easily (though that is currently not supported).

ibtk/include/ibtk/AppInitializer.h
drwells: My point is I am handy enough with sed that we can do whatever we want here - just tell me how long you want the ascii art guards to be and where the text will start and I can fix all of them in a few minutes.

boyceg: If there is a simple script that can easily standardize, I think that I don't care so much what the width is.

I _think_ that the most common is 78 characters, with the first block of slashes being 31 characters wide. (Why? I don't know!) Maybe stick with that to avoid additional changes to comments?

drwells: Sure. Fixed!

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -11,10 +11,14 @@
 //
 // ---------------------------------------------------------------------
 
+/////////////////////////////// INCLUDE GUARD ////////////////////////////////
+
 #ifndef included_FESurfaceDistanceEvaluator
 #define included_FESurfaceDistanceEvaluator
 
-///////////////////////////// INCLUDES ///////////////////////////////////
+/////////////////////////////// INCLUDES /////////////////////////////////////
```
drwells, line 19: I checked all the other ascii art while I was at it.

drwells:
I'll rerun the tester and then this should be good to go.

drwells:
I think the CI got stuck. Lets see if we can fix it:

Jenkins retest this please

PR #1086, opened 2020-08-25
Ibfemethod patch elem assoc

### LINKED ISSUE #1085: fix interaction between REGRID_CFL_INTERVAL and IBFEMethod
One can disable regridding by setting `REGRID_CFL_INTERVAL` to a huge value. However, this screws up IBFEMethod since that assumes that no particle travels more than one cell width between regrids (and that assumption is how we associate elements with patches).

There are a few things we could do to fix this:
1. Use an independent CFL estimation routine inside IBFEMethod that redoes the association between elements and patches when necessary.
2. Error out if `REGRID_CFL_INTERVAL` is high enough that IBFEMethod would have problems when it encounters problems.
3. Ignore the use-provided value and regrid when we need to.
4. Use `REGRID_CFL_INTERVAL` to associate elements with patches. This would be pretty inefficient since here we would have all elements on all processors.

drwells:
Fixes #1085.

The fix I have here still causes a few tests to fail because I added a check to error out if there are nodes that are not associated with patches. Hence, before proceeding, I think its worth deciding whether or not we want to continue supporting nodes outside the physical domain - if any IB points are outside the physical domain we implicitly assume that they have zero velocity and don't spread forces. The former is a questionable assumption since, e.g., for the lid driven cavity, the velocity very well may not be zero.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -171,7 +171,7 @@ static Timer* t_begin_data_redistribution;
 static Timer* t_end_data_redistribution;
 static Timer* t_apply_gradient_detector;
 // Version of IBFEMethod restart file data.
-const int IBFE_METHOD_VERSION = 4;
+const int IBFE_METHOD_VERSION = 5;
 
 inline boundary_id_type
 get_dirichlet_bdry_ids(const std::vector<boundary_id_type>& bdry_ids)
@@ -352,6 +352,32 @@ build_ib_ghosted_system_data(std::vector<SystemData>& ghosted_system_data,
     return;
 }
 
+/*
+ * Get the smallest cell width on the specified level. This operation is
+ * collective.
+ */
+double
+get_min_patch_dx(const Pointer<PatchHierarchy<NDIM> > hierarchy, const int level_num)
+{
+    double result = std::numeric_limits<double>::max();
+
+    // Some processors might not have any patches so its easier to just quit
+    // after one loop operation than to check
+    Pointer<PatchLevel<NDIM> > level = hierarchy->getPatchLevel(level_num);
+    for (PatchLevel<NDIM>::Iterator p(level); p; p++)
+    {
+        Pointer<Patch<NDIM> > patch = level->getPatch(p());
+        const Pointer<CartesianPatchGeometry<NDIM> > patch_geom = patch->getPatchGeometry();
+        const double* const patch_dx = patch_geom->getDx();
+        const double patch_dx_min = *std::min_element(patch_dx, patch_dx + NDIM);
+        result = std::min(result, patch_dx_min);
+        break; // all patches on the same level have the same dx values
+    }
+
+    result = IBTK_MPI::minReduction(result);
+
+    return result;
+}
 } // namespace
 
 const std::string IBFEMethod::SOURCE_SYSTEM_NAME = "IB source system";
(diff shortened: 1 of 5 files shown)


drwells:
I fixed some, but not all, of the failing tests due to nodes outside the domain.

boyceg:
I prefer the current behavior --- allow for (parts of) structures to be outside of the computational domain. In any case, I'd suggest addressing that issue in a separate PR.

src/IB/IBFEMethod.cpp
```diff
@@ -521,6 +547,30 @@ IBFEMethod::preprocessIntegrateData(double current_time, double new_time, int nu
...
+    assertStructureOnFinestLevel();
+    const double patch_dx = get_min_patch_dx(d_hierarchy, d_hierarchy->getFinestLevelNumber());
+    for (unsigned int part = 0; part < d_meshes.size(); ++part)
+    {
+        // To avoid allocating an extra vector we abuse "new" for a moment since
+        // it gets overwritten below anyway
+        PetscVector<double>& diff = d_X_vecs->get("new", part);
+        PetscVector<double>& last = d_X_vecs->get("last_patch_elem_assoc", part);
+        VecWAXPY(diff.vec(), -1.0, last.vec(), d_X_vecs->get("solution", part).vec());
+        const double max_distance = diff.linfty_norm();
```
boyceg, line 561: Should we look at the pointwise Euclidean norm of the displacement?

drwells: That's a good question. This l-oo (that's a lower-case L) type norm makes more sense to me because a particle could move 0.75 cell widths in each coordinate direction and still be inside the same cell (and here that is reflected by having a maximum distance of 0.75) but if we use the L2 norm then we would get `sqrt(3*0.75**2) = 1.299...`, implying that we have moved outside the cell.

drwells: I'm not sure what the right choice is - this is an odd CFL condition because the CFL condition is inherently Eulerian but we have a Lagrangian description of the structure.

boyceg: I think that we also should consider the motion of the edges/faces. I think we want to think about the furthest that any material point in the element could move. Specifically, we need to detect if is it possible that a material point in the element that was not previously able to interact with the Cartesian grid patch is now able to do so. It may be best to use the max norm in making this determination. I need to draw a picture, but I don’t expect that I will get a chance until this afternoon.

drwells: Independent of the norm choice, we do have some wiggle room here - we associate elements with patches if they are within 1 cell width of the patch but we reassociate (in this case, with the default values) once a point has move 0.75 cell widths in a coordinate direction.

drwells: One more useful point: none of the tests, aside from the new one, actually execute this reassociation procedure because the CFL based regridding is much more conservative.

boyceg: Sure, I understand --- I just want to try to make sure that we don't think we have more wiggle room than we actually do and get surprised later on. :-)

drwells:
The 'a node is outside the patch' error checking is now in #1087.

include/ibamr/FEMechanicsBase.h
```diff
@@ -424,7 +424,7 @@ class FEMechanicsBase : public SAMRAI::tbox::Serializable
      * PARALLEL vectors into GHOSTED vectors for a collection of Systems.
      */
     static void setup_system_vectors(libMesh::EquationSystems* equation_systems,
-                                     const std::vector<std::string>& system_names,
+                                     const std::string& system_name,
```
boyceg, line 427: Why is this being changed?

drwells:
I changed it because we no longer ever actually call it with multiple system names. I can change it back.

drwells:
I reverted that change and rebased.

PR #1087, opened 2020-08-26
Check for nodes that are inside the domain but outside a patch.

drwells:
Part of #1085 - this is just the last patch from #1086.

We could have saved a lot of time trying to diagnose #1085 if we had a way to detect that IBFEMethod was not working correctly. This PR adds a check to ensure that all mesh nodes are actually inside a patch or outside the domain.

To get around a problem with SAMRAI we use bounding boxes for patches that are slightly larger than the patch itself. Similarly, to determine if a point is outside the domain, we use a bounding box slightly smaller than the domain itself. In either case the node is OK.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -22,6 +22,7 @@
 
 #include "ibtk/QuadratureCache.h"
 #include "ibtk/SAMRAIDataCache.h"
+#include "ibtk/ibtk_enums.h"
 #include "ibtk/ibtk_utilities.h"
 
 #include "BasePatchLevel.h"
@@ -274,7 +275,24 @@ protected:
  * element data while this class stores additional data dependent on the
  * Eulerian grid.
  *
- * <h3>Parameters effecting workload estimate calculations</h3>
+ * <h2>Parameters read from the input database</h2>
+ *
+ * At the present time the only parameter read from the input database is
+ * <code>node_outside_patch_check</code>, which controls how this class responds
+ * to mesh nodes outside the finest patch level. In all cases, for backwards
+ * compatibility, nodes that are outside the computational domain are permitted
+ * and are ignored by this check. Possible values are:
+ * <ol>
+ *   <li><code>node_outside_permit</code>: Permit nodes to be outside the finest
+ *   patch level.</li>
+ *   <li><code>node_outside_warn</code>: Permit nodes to be outside the finest
+ *   patch level, but log a warning whenever this is detected.
+ *   <li><code>node_outside_error</code>: Abort the program when nodes are detected
+ *   outside the finest patch level.
+ * </ol>
+ * The default value is <code>node_outside_error</code>.
+ *
+ * <h2>Parameters effecting workload estimate calculations</h2>
  * FEDataManager can estimate the amount of work done in IBFE calculations
  * (such as FEDataManager::spread). Since most calculations use a variable
  * number of quadrature points on each libMesh element this estimate can vary
@@ -1254,6 +1272,15 @@ private:
      */
     const InterpSpec d_default_interp_spec;
     const SpreadSpec d_default_spread_spec;
+
+    /*!
+     * after reassociating patches with elements a node may still lie
+     * outside all patches on the finest level in unusual circumstances
+     * (like when the parent integrator class does not regrid sufficiently
+     * frequently and has more than one patch level). This enum controls
+     * what we do when this problem is detected.
+     */
+    NodeOutsidePatchCheckType d_node_patch_check = NODE_OUTSIDE_ERROR;
 
     /*!
      * SAMRAI::hier::IntVector object which determines the required ghost cell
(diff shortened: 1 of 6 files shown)


boyceg:
I think that I would like to be able to control whether this results in a warning, error, or is just ignored. We could default to error, but include in the error message instructions on how to change to warning or ignore.

drwells:
That's a good idea - in fact we could let users decide if they want to permit points outside the domain or not.

boyceg:
I am not sure that the code necessarily _breaks_ if a node is not associated with any patch interior. Specifically, I think that we can tolerate having the nodes a little outside of the domain. Resulting in an error in this case seems overly restrictive, especially for complex geometries where you may not have as much control as you might like.

drwells:
That should definitely be permitted by default (it is with both master and the current state of this patch).

drwells:
I made the suggested changes. However, with our current setup (where we tag cells for refinement if they contain IB points) we do not ever encounter a situation where this check would be executed so we need to also merge #1086 for this to work.

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -1255,6 +1273,15 @@ class FEDataManager : public SAMRAI::tbox::Serializable
...
     const SpreadSpec d_default_spread_spec;
 
+    /*!
+     * after reassociating patches with elements a node may still lie
+     * outside all patches on the finest level in unusual circumstances
+     * (like when the parent integrator class does not regrid sufficiently
+     * frequently and has more than one patch level). This enum controls
+     * what we do when this problem is detected.
+     */
+    NodeOutsidePatchCheckType d_node_patch_check = NODE_OUTSIDE_ERROR;
```
boyceg, line 1283: Is the default behavior in master currently `NODE_OUTSIDE_PERMIT`? If so, should we make this `NODE_OUTSIDE_WARN` instead?

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -620,6 +620,119 @@ FEDataManager::reinitElementMappings()
...
+                }
+                else
+                {
+                    // Points are allowed to be outside the domain - they simply
+                    // are no longer used for IB calculations.
+                    bool inside_hier = true;
+                    for (unsigned int d = 0; d < NDIM; ++d)
+                    {
+                        // Like above - permit points very close to the boundary
+                        // to pass the check and treat them as being outside
```
boyceg, line 685: Do we want to treat them as "outside" or "inside" if they are just at the boundary? I would vote for "inside".

ibtk/include/ibtk/FEDataManager.h
drwells: The current behavior is `NODE_OUTSIDE_PERMIT`. To be clear: here 'outside' refers to 'outside the finest patch level', not 'outside the domain'.

I'm about 60/40 in favor of raising an error in this case since I cannot imagine a circumstance where one would want the structure to be inside the domain and not interacting with the background grid, even though its not strictly backwards compatible. Is there some use case I am not aware of where we might *not* want to crash in that case?

boyceg: OK, that sounds fine. I guess I was thinking that outside the domain implies outside of the finest patch level. :-)

ibtk/src/lagrangian/FEDataManager.cpp
boyceg: Here I mean physical boundary --- apologies if I am misinterpreting.

drwells: Here points very close to the boundary of the domain (the bounding box on the patch hierarchy itself) are treated as being outside the domain. The logic is complicated - if a Node is inside, or nearly inside, a patch then it passes the check. If a Node is *outside* or *nearly outside* the domain then it also passes the check.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -620,6 +620,119 @@ FEDataManager::reinitElementMappings()
...
+        IBTK_MPI::allToOneSumReduction(node_ranks.data(), node_ranks.size());
+        if (rank == 0)
+        {
+            const std::string message =
+                "At least one node in the current mesh is inside the fluid domain and not associated with any "
+                "patch. This class currently assumes that all elements are on the finest level and will not "
+                "work correctly if this assumption does not hold. This usually happens when you use multiple "
+                "patch levels and set the regrid CFL interval to a value larger than one. To change this check set "
+                "node_outside_patch_check to a different value in the input database: see the documentation of "
+                "FEDataManager for more information.";
```
drwells, line 709: I tried to make the error message useful. This is what I implemented.

ibtk/include/ibtk/FEDataManager.h
drwells: Unfortunately this is not always the case and its where @rossisimone got stuck for way too long.

PR #1102, opened 2020-09-10
Adding wave generation zone code for first order water waves and updating wave tank ex0

Kaustubh-Khedkar:
This code is used to create a generation zone for first-order water waves at the channel inlet.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


include/ibamr/FirstOrderStokesWaveGenerator.h
```diff
@@ -0,0 +1,59 @@
...
+class FirstOrderStokesWaveGenerator : public StokesWaveGeneratorStrategy
+{
+public:
+    FirstOrderStokesWaveGenerator(const std::string& object_name,
+                                  SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db);
+
+    /*!
+     * Get surface elevation at a specified horizontal position and time.
+     */
+    double getSurfaceElevation(double x, double time) const;
```
boyceg, line 49: Should these be marked with `override`?

boyceg (dismissed):
Looks good. Can you check whether `override` should be included and add a test? It should be possible (I think) to re-use the existing wave tank test and just add new input/output files to specify the first-order Stokes wave.

boyceg:
ok to test

include/ibamr/FirstOrderStokesWaveGenerator.h
```diff
@@ -0,0 +1,59 @@
...
+//
+// ---------------------------------------------------------------------
+
+/////////////////////// INCLUDE GUARD ////////////////////////////////////
+
+#ifndef included_FirstOrderStokesWaveGenerator
+#define included_FirstOrderStokesWaveGenerator
+
+///////////////////////////// INCLUDES ///////////////////////////////////
+
```
drwells, line 20:
```suggestion
#include <ibamr/config.h>

```
All of the headers now include this so that our standard macros are always available.

drwells (changes requested):
We need to add an additional header.

Would you please take care of the other items in the checklist?

include/ibamr/FirstOrderStokesWaveGenerator.h
Kaustubh-Khedkar: Yes, it should be marked with override. I will make the change.

Kaustubh-Khedkar: > Looks good. Can you check whether `override` should be included and add a test? It should be possible (I think) to re-use the existing wave tank test and just add new input/output files to specify the first-order Stokes wave.

The test/wave_tank does not have wave generation functionality added to it.

drwells (approved):

drwells:
I ran the testsuite and it passed.

Would you like to add a changelog entry or a test? This is small enough that I am okay with leaving them out. Up to you.

Kaustubh-Khedkar:
> I ran the testsuite and it passed.
> 
> Would you like to add a changelog entry or a test? This is small enough that I am okay with leaving them out. Up to you.

I will write a test.

Kaustubh-Khedkar:
> > I ran the testsuite and it passed.
> > Would you like to add a changelog entry or a test? This is small enough that I am okay with leaving them out. Up to you.
> 
> I will add a test.

Kaustubh-Khedkar:
Hi @drwells, 
I have written a test and tried running it by following the instructions given on:
https://github.com/IBAMR/IBAMR/wiki/Tests-for-IBAMR  
I get RUN FAILED as the output. I also tried running one of the existing tests and the output is as follows:
./attest -R nwt_cylinder
1/1 starting test wave_tank/nwt_cylinder.mpirun=8.input
    wave_tank/nwt_cylinder.mpirun=8.input ----------------------- RUN FAILED

0/1 (0.0%) tests passed

The following tests FAILED:

    wave_tank/nwt_cylinder.mpirun=8.input

I have attached the config.log files for IBAMR and NUMDIFF. Can you please take a look?

[IBAMR_config.log](https://github.com/IBAMR/IBAMR/files/5264794/IBAMR_config.log
[NUMDIFF_config.log](https://github.com/IBAMR/IBAMR/files/5264801/NUMDIFF_config.log)
)

drwells:
Those logs look fine - I think you set up the test suite correctly.

Tests can fail in two ways - they can either abort while running or produce the wrong output. In this case the test aborted (or crashed). Could you try going into that test directory (i.e., `tests/wave_tank/`) and running the test manually (the same way as examples - `mpirun -np 8 ./nwt_cylinder nwt_cylinder.mpirun=8.input`) and see what happens?

Kaustubh-Khedkar:
Yes, the test case ran correctly. Following is the output file (in .txt) and IB2d_cylinder2d.log files
[IB2d_cylinder2d.log](https://github.com/IBAMR/IBAMR/files/5265417/IB2d_cylinder2d.log)
.
[output.txt](https://github.com/IBAMR/IBAMR/files/5265412/output.txt)

drwells:
I don't understand why it would work in case and not the other. If things continue not to work I can just check the patch when you are done by running the test suite myself.

Kaustubh-Khedkar:
> I don't understand why it would work in case and not the other. If things continue not to work I can just check the patch when you are done by running the test suite myself.

I have added the test case. Please let me know if any more changes are required.

ibtk/configure
```diff
@@ -12118,16 +12118,11 @@ $as_echo "$lt_cv_ld_force_load" >&6; }
...
-      case ${MACOSX_DEPLOYMENT_TARGET-10.0},$host in
-	10.0,*86*-darwin8*|10.0,*-darwin[91]*)
-	  _lt_dar_allow_undefined='$wl-undefined ${wl}dynamic_lookup' ;;
-	10.[012][,.]*)
+    darwin*)
+      case ${MACOSX_DEPLOYMENT_TARGET},$host in
+	10.[012],*|,*powerpc*)
 	  _lt_dar_allow_undefined='$wl-flat_namespace $wl-undefined ${wl}suppress' ;;
-	10.*)
+	*)
```
drwells, line 12125: What version of autotools did you use? This file shouldn't be changed.

tests/wave_tank/nwt.mpirun=8.output
```diff
@@ -0,0 +1,11 @@
...
+0.01	-0.03998369123
+0.02	-0.0399505819
+0.03	-0.03991531588
+0.04	-0.03988280904
+0.05	-0.03985496371
+0.06	-0.03983229365
+0.07	-0.03981465848
+0.08	-0.03980157022
+0.09	-0.03979253903
+0.1	-0.03978689494
```
drwells, line 11: I get substantially different output for this test:
```
0.01    -0.03998478538
0.02    -0.03994721241
0.03    -0.0398957478
0.04    -0.03983655776
0.05    -0.0397740696
0.06    -0.03971252944
0.07    -0.03965461961
0.08    -0.03960119585
0.09    -0.03955297905
0.1     -0.03951051114
```
Do you know why this would be the case?

drwells (changes requested):
Looks good but we need to use autotools version 2.69 to avoid creating unnecessary conflicts.

I'm not sure why we get different output for the test. Can you think of a reason why this might be the case? We need to resolve that before we merge since the differences ae pretty large.

ibtk/configure
boyceg: I just verified that running `bootstrap` on the branch reverses most (all?) of these change.

I could commit and squash...

boyceg: @Kaustubh-Khedkar, if you run the `bootstrap` script in the top level of the IBAMR source directory, do you see the same behavior?

IIRC, we tried to set up `bootstrap` so that it automatically installs consistent versions of `autoconf`, `automake`, and `libtool` so that users do not have to manually deal with it --- perhaps it is not working, or is not doing what we intended it to do.

drwells: > I could commit and squash...

Boyce has a good point - there is no reason for us to get tripped up here if we can easily force push a fix. Lets focus on the test.

tests/wave_tank/nwt.mpirun=8.output
drwells: I read through the code and I don't see anything like random number generation that could be giving different results on different platforms.

ibtk/configure
boyceg: > > I could commit and squash...
> 
> Boyce has a good point - there is no reason for us to get tripped up here if we can easily force push a fix. Lets focus on the test.

Sounds good, although I will hold off on doing this until the differences in the NWT test output are resolved.

tests/wave_tank/nwt.mpirun=8.output
Kaustubh-Khedkar: The output file in tests/wave_tank/nwt.mpirun=8.output is run using: 
mpirun -np 4 ./nwt nwt.mpirun=8.input -stokes_ksp_rtol 1.0e-8

I have run it by using the following:
mpirun -np 8 ./nwt nwt.mpirun=8.input -stokes_ksp_rtol 1.0e-10

The output is (unable to upload the output file so pasting the output):

Printing level set at cell center closest to point (4.2396, 0.52)
0.01	        -0.03998368681
0.02	-0.03995058865
0.03	-0.03991532587
0.04	-0.0398828241
0.05	-0.03985500045
0.06	-0.03983235123
0.07	-0.0398147255
0.08	-0.03980168935
0.09	-0.03979270585
0.1	        -0.03978722421

This has a tighter tolerance than the earlier case.

boyceg: > The output file in tests/wave_tank/nwt.mpirun=8.output is run using:
> mpirun -np 4 ./nwt nwt.mpirun=8.input -stokes_ksp_rtol 1.0e-8
> 
> I have run it by using the following:
> mpirun -np 8 ./nwt nwt.mpirun=8.input -stokes_ksp_rtol 1.0e-10
> 
> The output is (unable to upload the output file so pasting the output):
> 
> Printing level set at cell center closest to point (4.2396, 0.52)
> 0.01 -0.03998368681
> 0.02 -0.03995058865
> 0.03 -0.03991532587
> 0.04 -0.0398828241
> 0.05 -0.03985500045
> 0.06 -0.03983235123
> 0.07 -0.0398147255
> 0.08 -0.03980168935
> 0.09 -0.03979270585
> 0.1 -0.03978722421
> 
> This has a tighter tolerance than the earlier case.

If you run the test (e.g. using `attest -R nwt`), does it pass for you?

Kaustubh-Khedkar: The test does not run. I have mentioned this in the earlier comments.

boyceg: > The test does not run. I have mentioned this in the earlier comments.

Ah, I misunderstood when you said that you had added a test case. For the test to be effective, the output in the repo needs to be the same as whatever is generated when the test is run by `attest`.

What goes wrong when you run your test?

boyceg: (Also, I would suggest not to check off "Does the test suite pass?" if the test suite is not working. 😁 )

boyceg: @drwells, if you use the output above, does the test suite work for you?

drwells: > (Also, I would suggest not to check off "Does the test suite pass?" if the test suite is not working. grin )

I think that was me, actually, when we had all the tests passing a few days ago.

We will have to do some more debugging of the test runner but if `make tests; ./attest` doesn't work then that's beyond the scope of this PR. The right fix for this case (where we need an extra PETSc option) is to just set the option directly in the source file. Do you know how to do this?

Kaustubh-Khedkar: > We will have to do some more debugging of the test runner but if `make tests; ./attest` doesn't work then that's beyond the scope of this PR. The right fix for this case (where we need an extra PETSc option) is to just set the option directly in the source file. Do you know how to do this?

No. Could you show me how to do that please?

boyceg: Is the extra PETSc flag needed to get the right result?

boyceg: (I mean is the answer with lower tolerance “wrong” or simply less accurate?)

drwells: You can add `PetscOptionsSetValue(nullptr, "-stokes_ksp_tol", "1e-10");` to the top of `main()` after you make the `IBTKInitialize` object (which initializes PETSc).

boyceg: > You can add `PetscOptionsSetValue(nullptr, "-stokes_ksp_tol", "1e-10");` to the top of `main()` after you make the `IBTKInitialize` object (which initializes PETSc).

`PetscOptionsSetValue(nullptr, "-stokes_ksp_rtol", "1e-10");`?

drwells: Oh yeah, I forgot the r.

Kaustubh-Khedkar: > You can add `PetscOptionsSetValue(nullptr, "-stokes_ksp_rtol", "1e-10");` to the top of `main()` after you make the `IBTKInitialize` object (which initializes PETSc).

I added the above and the output is as follows:

Printing level set at cell center closest to point (4.2396, 0.52) 
0.01	        -0.03998368677
0.02	-0.03995058906
0.03	-0.03991532614
0.04	-0.03988282362
0.05	-0.0398549999
0.06	-0.03983235022
0.07	-0.03981472417
0.08	-0.03980168783
0.09	-0.03979270488
0.1	        -0.03978722355

Kaustubh-Khedkar: I discussed this with Amneet and he recommended to add the following lines in the nwt.mpirun=8.input file:

   // Solver parameters
   stokes_solver_type   = "VC_STAGGERED_STOKES_PETSC_KRYLOV_SOLVER"
   stokes_solver_db {
      options_prefix    = "stokes_"
      ksp_type          = "fgmres"
      rel_residual_tol  = 1e-14
      abs_residual_tol  = 1e-14
      max_iterations    = 1000
   }

This will make a tighter tolerance and should not create any difference in values in the output file.
I was also able to verify the test in one of the lab workstations using ./attest -R nwt and the test passed. Should I push these changes or do we need to investigate more regarding this issue?

boyceg: I agree — I think that putting these tolerances in the input file is better than hard coding the command line option in the source code.

drwells: Oh yeah, that's a way better way to do it. Nice catch!

drwells: In general, since tests are small enough to run quickly, its a good idea to use very tight tolerances so that we get consistent output even when our dependencies or algorithms change.

drwells (approved):

drwells:
Thanks a lot for working through the fixes to the test! I reran the test suite and pushed a fix to the build scripts - I'll merge once the CI is happy.

PR #1104, opened 2020-09-10
Multilevel ibfe 1

drwells:
We need to fundamentally change the way we set up the patch-to-element mapping to support multilevel interaction in IBFE. It's not backwards-compatible, but we can make our lives a lot easier by just getting rid of all the extra `FEDataManager::getManager()` overloads so that we can set up the block id to patch level mapping just once.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/incompatibilities/20200910DavidWells
+++ b/doc/news/changes/incompatibilities/20200910DavidWells
@@ -1,0 +1,4 @@
+Changed: IBTK::FEData now requires that the relevant libMesh::EquationSystems
+object be passed to its constructor.
+<br>
+(David Wells, 2020/09/10)
--- a/doc/news/changes/incompatibilities/20200910DavidWells-2
+++ b/doc/news/changes/incompatibilities/20200910DavidWells-2
@@ -1,0 +1,4 @@
+Changed: The overloads of IBTK::FEDataManager::getManager() have been
+consolidated into a single function.
+<br>
+(David Wells, 2020/09/10)
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -169,7 +169,7 @@ public:
      * Constructor. Registers the object with the restart database: i.e.,
      * inheriting classes should not also register themselves.
      */
-    FEData(std::string object_name, const bool register_for_restart);
+    FEData(std::string object_name, libMesh::EquationSystems& equation_systems, const bool register_for_restart);
 
     /*!
      * Destructor.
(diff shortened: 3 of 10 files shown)


boyceg (commented):
General comment --- I have not checked every instance carefully, but I would prefer to use `max_level_number` to identify the maximum level number, and `max_levels` to indicate the maximum _number_ of levels. Note that `max_levels = max_level_number + 1`. It looks like there are a few places where `max_level_number` was changed to `max_levels`.

drwells:
Good point - let me fix that.

boyceg:
It looks like this is done consistently, so I'm not sure if there is anything to fix.

drwells:
I improved the status quo but some of the function argument names are wrong.

drwells:
That should fix things - the first commit now should always use `max_levels` but internally store `d_finest_level_number`. I think `d_finest_level_number` is more commonly used than `d_max_level_number` but its not consequential - we can change it if we want.

boyceg (approved):

boyceg:
I usually think of `finest_level_number` as the finest level currently in the patch hierarchy, and `max_level_number` as the finest possible level number. In most use cases, these are the same. However, if we let folks assign different structures (or parts of structures) to different levels, then we could have models where the structures are not on the finest level of the patch hierarchy.

boyceg:
Specifically, you could set up something where feature detection criteria dynamically add or remove AMR grid levels that are finer than the finest level that a structure lives on.

drwells:
that's a good distinction to make (and its relevant here since FEDataManager will, currently, tag cells when we set up the hierarchy and there are no patches on `d_finest_level_number`). Let me update those names to make this clearer.

boyceg:
If users actually specify which levels they want to assign structures to, then there actually is not need to pass in `max_levels` or `max_level_number` to `FEDataManager`.

boyceg:
OTOH, because `FEDataManager` is partially responsible for setting up the AMR grid, it needs to know how "deep" the structures should go in the patch hierarchy.

drwells:
> OTOH, because FEDataManager is partially responsible for setting up the AMR grid, it needs to know how "deep" the structures should go in the patch hierarchy.

Yes - I am pretty sure that we need to have this value available in the case where we tag cells and the user does not specify which patch levels elements should be on.

boyceg:
Anyway, you may notice that there are a number of places in the code that update the equivalents of `coarsest_patch_level_number` and `finest_patch_level_number` based on the actual configuration of the `PatchHierarchy`.

drwells:
I added one more commit where I rename things, when relevant, to `max_patch_level`.

boyceg (approved):

PR #1105, opened 2020-09-10
Fix --with-zlib

drwells:
It looks like @cpuelz encountered this recently. Coincidentally I found some publicly-available notes from Michael Waldron which also contained a workaround.

Almost every Unix machine has zlib installed in a standard location so this flag is basically never necessary. This patch fixes the link flags so that it actually works correctly in the very unlikely case that it is explicitly needed.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/config/IBAMR_config.h.tmp.in
+++ b/config/IBAMR_config.h.tmp.in
@@ -130,9 +130,6 @@
 /* Define if you have the libSAMRAI3d_xfer library. */
 #undef HAVE_LIBSAMRAI3D_XFER
 
-/* Define to 1 if you have the `z' library (-lz). */
-#undef HAVE_LIBZ
-
 /* Define to 1 if you have the <memory.h> header file. */
 #undef HAVE_MEMORY_H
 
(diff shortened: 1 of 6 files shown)


drwells:
@cpuelz would you please verify that this fixes the problem?

boyceg (approved):

boyceg:
@drwells shall we merge?

drwells:
Hm. We could merge but this patch is not urgent - I would prefer if @cpuelz or someone else who actually encountered the bug can verify that this fixes the problem so we know this is the right fix.

boyceg:
@cpuelz can you verify that this fixes your zlib issue?

cpuelz:
sorry for the delay on this.  I somewhat recently rebuilt IBAMR master and had no problems with zlib.  I can try rebuiding this branch too.

drwells:
If you want to test it be sure to explicitly provide a flag to `./configure` specifying with zlib is - the default (no zlib flag) already worked.

drwells:
This is probably fine.

PR #1024, opened 2020-06-26
add a static pressure (with optional local projection stabilization) to FEMechanicsBase

boyceg:
### IBAMR Pull Request Checklist
- [ ] Does the test suite pass?
- [x] Was clang-format run?
- [x] Were relevant issues cited? Please remember to add `Fixes #12345` to close
      the issue automatically if we are fixing the problem.
- [x] Is this a change others will want to know about? If so, then has a
      changelog entry been added?
- [x] If new data structures have been added to a class then have they been set
      up appropriately for restarts? If so, ensure that the restart version
      number is incremented.
- [ ] Does this change include a bug fix or new functionality? If so, a new test
      or tests should be added. New tests should run quickly (less than a minute
      in release mode). If possible, an older test should gain a new option so
      that we do not need to compile more test executables.
- [x] Did you (if your account has permission to do so) set relevant labels on
      GitHub for the pull request?

PR #1106, opened 2020-09-11
add local pressure stabilization

boyceg:
This adds a "local projection" stabilizer to the projections provided by `FEProjector`. We are experimenting with this in different branches for different purposes, and so it would be good to get it into master to keep those in synch.

This local stabilization approach augments the standard projection, `(p, q) = (f, q)`, with a stabilization of the form `(p, q) + epsilon (p - Pi p, q - Pi q) = (f, q)`. Currently the only supported local projection is to piecewise constant functions.

This is not yet used anywhere in the library, but will be an option to be used in `IBFEMethod` and/or the new explicit mechanics solvers in the near future.

Ref: https://epubs.siam.org/doi/abs/10.1137/S0036142905444482

- [X] Does the test suite pass?
- [X] Was clang-format run?
- [X] Were relevant issues cited? Please remember to add `Fixes #12345` to close
- [X] Is this a change others will want to know about? If so, then has a
      changelog entry been added?
- [X] If new data structures have been added to a class then have they been set
      number is incremented.
- [X] Does this change include a bug fix or new functionality? If so, a new test
- [X] Did you (if your account has permission to do so) set relevant labels on
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEProjector.h
+++ b/ibtk/include/ibtk/FEProjector.h
@@ -87,6 +87,13 @@ public:
     /*!
      * \return Pointers to a linear solver and sparse matrix corresponding to a
      * L2 projection operator with a local projection stabilization term.
+     *
+     * This local stabilization approach augments the standard projection with a
+     * stabilization of the form (p, q) + epsilon (p - Pi p, q - Pi q) = (f, q).
+     * Currently the only supported local projection is to piecewise constant
+     * functions.
+     *
+     * Ref: https://epubs.siam.org/doi/abs/10.1137/S0036142905444482
      */
     std::pair<libMesh::PetscLinearSolver<double>*, libMesh::PetscMatrix<double>*>
     buildStabilizedL2ProjectionSolver(const std::string& system_name, double epsilon);
@@ -117,6 +124,13 @@ public:
     /*!
      * \brief Set U to be the L2 projection of F with a local projection
      * stabilization term.
+     *
+     * This local stabilization approach augments the standard projection with a
+     * stabilization of the form (p, q) + epsilon (p - Pi p, q - Pi q) = (f, q).
+     * Currently the only supported local projection is to piecewise constant
+     * functions.
+     *
+     * Ref: https://epubs.siam.org/doi/abs/10.1137/S0036142905444482
      */
     bool computeStabilizedL2Projection(libMesh::PetscVector<double>& U,
                                        libMesh::PetscVector<double>& F,


ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -86,17 +88,21 @@ get_dirichlet_bdry_ids(const std::vector<boundary_id_type>& bdry_ids)
 /////////////////////////////// PUBLIC ///////////////////////////////////////
 
 FEProjector::FEProjector(EquationSystems* equation_systems, const bool enable_logging)
-    : d_fe_data(new FEData("FEProjector", *equation_systems, /*register_for_restart*/ false)),
-      d_enable_logging(enable_logging)
+    : d_fe_data(new FEData("FEProjector", /*register_for_restart*/ false)), d_enable_logging(enable_logging)
```
drwells, line 91: In #1104 we changed `FEData` to take an `EquationSystems` in its constructor - that's causing the build failure.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -357,6 +365,164 @@ FEProjector::buildLumpedL2ProjectionSolver(const std::string& system_name)
...
+                                                 /*assymetric_constraint_rows*/ false);
+                prune_roundoff_entries(M_e);
+                M_mat->add_matrix(M_e, dof_id_scratch);
+            }
+        }
+
+        // Flush assemble the matrix.
+        M_mat->close();
+
+        // Reset values at Dirichlet boundaries.
```
drwells, line 464: I am still pretty sure that this code block (which we have in multiple places) doesn't actually do anything since we are now applying constraints correctly. Do we have any examples where we actually use it?

drwells:
This is mostly done. @boyceg do you have time to wrap it up today? If not I can fix a few things and merge myself.

boyceg:
Whoops! Fixed (I think).

ibtk/src/lagrangian/FEProjector.cpp
boyceg: I am not sure that we do.

ibtk/src/lagrangian/FEProjector.cpp
boyceg: Fixed I think.

ibtk/src/lagrangian/FEProjector.cpp
boyceg: I don't know if we always set up proper Dirichlet constraints from these tags. (In the past, at least, I know that we did _not_.)

drwells: That's something else that we should add to the set of constraints. Let's just keep the status quo for now.

drwells (changes requested):
The implementation looks fine - could we add a short test and also write down the weak form and citation in the documentation?

boyceg:
I can add a comment. Tests could come once it is used in solvers. I don't know how to really test outside of the context of the solvers.

boyceg:
Added a comment.

boyceg:
(For instance, this can be tested in #1024.)

drwells (approved):

PR #1118, opened 2020-09-15
Fix a problem with printing the partitioning in tests.

drwells:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/tests/spread/spread_01.cpp
+++ b/tests/spread/spread_01.cpp
@@ -376,17 +376,11 @@ main(int argc, char** argv)
             half_f_vector.close();
         }
 
-        std::ostringstream out;
+        // the partitioning isn't relevant when we only have one processor
         if (IBTK_MPI::getNodes() != 1)
         {
-            // partitioning is only relevant when there are multiple processors
-            Pointer<PatchLevel<NDIM> > patch_level =
-                patch_hierarchy->getPatchLevel(patch_hierarchy->getFinestLevelNumber());
-            const BoxArray<NDIM> boxes = patch_level->getBoxes();
-            plog << "hierarchy boxes:\n";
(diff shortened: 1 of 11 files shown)


tests/spread/spread_01.cpp
```diff
@@ -376,17 +376,11 @@ main(int argc, char** argv)
...
         }
 
-        std::ostringstream out;
+        // the partitioning isn't relevant when we only have one processor
         if (IBTK_MPI::getNodes() != 1)
         {
-            // partitioning is only relevant when there are multiple processors
-            Pointer<PatchLevel<NDIM> > patch_level =
-                patch_hierarchy->getPatchLevel(patch_hierarchy->getFinestLevelNumber());
-            const BoxArray<NDIM> boxes = patch_level->getBoxes();
```
drwells, line 385: Here is the mistake - @boyceg pointed out that this gets all boxes (i.e., across all processors). Fortunately a second bug here (printing boxes to plog when I intended to print to out) mostly canceled out the first bug, so we print the correct boxes in the correct order but are just missing the labels.

tests/tests.h
```diff
@@ -69,6 +71,30 @@ print_strings_on_plog_0(const std::string& out)
         MPI_Send(out.data(), size, MPI_CHAR, 0, 0, IBTK_MPI::getCommunicator());
 }
 
+/**
+ * Print the parallel partitioning (i.e., the boxes) on all processes to plog
+ * stream on processor 0.
+ */
+template <int DIM>
+inline void
+print_partitioning_on_plog_0(const Pointer<PatchHierarchy<DIM> >& patch_hierarchy,
```
boyceg, line 80: 👍

boyceg (approved):

PR #1121, opened 2020-09-16
add RT0 fortran kernels

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
I think we may want to use RT0 to connect levels in multilevel IBFE. It looks like the Fortran kernels for doing it were never implemented (???).

I need to setup a test.

boyceg:
Hrmmm, I don't know what I am doing, it seems like this is already in the library.

boyceg:
I am clearly thinking about RT0 right now. 😁 

I think I found an old RT0 refine operator branch that has already been merged on my workstation and forgot that it had already been merged. 😢

PR #1122, opened 2020-09-16
fix RT0 coarsen operator at physical boundaries

boyceg:
The RT0 coarsen operator (discrete adjoint of the RT0 refine operator) should ignore ghost values along physical boundaries.

(It also should ignore ghost values along coarse-fine interfaces, but that seems harder to enforce within the `CoarsenOperator` interface.)

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
@drwells, if I did not screw stuff up, I think that this will fix the issue that you ran into in #584.

drwells:
Great. How about I move the tests to this branch?

boyceg:
Sure.

boyceg:
Note, however, that this will not fix any issues with the routine also doing the wrong thing on c-f interfaces. That seems harder to sort out...

boyceg:
I factored out a bunch of nearly identical code into separate subroutines and tried to simplify the implementation a bit.

drwells:
This is a lot closer than where we were before - if you just look at the graphical output then things work perfectly. However, it looks like there is still an O(dx) error in an interpolation on the boundary - values at `x = 0` or `y = 0` are off but everything else is fine. I suspect this is from incorrectly reading the bounding boxes since side-centered data is wacky and `CartSideDoubleRT0Coarsen` doesn't use `toSideBox` at all.

@boyceg do you have time to look at the Fortran? I can probably fix it but it will take me a lot longer than it would take you.

drwells:
I force-pushed an update with the test (whose output demonstrates the current problem).

drwells:
One more thing - for comparison, if we run the same test on master now we get `O(1)` errors instead of `O(dx)`.

boyceg:
There is probably an off-by-one error on the lower side of the patches. I can try to investigate.

boyceg:
@drwells, I think maybe there is a bug in the tester. I think it is missing comparable error on the "upper" sides of the patches.

That said, I think that maybe this coarsening operator is not exact for linear functions at the boundaries.

This coarsening operator is the discrete adjoint of the RT0 refine operator. In practice, this means that the coarsening weights are the same as the refining weights (up to a rescaling).

At a physical boundary, this means that we only use boundary + interior values from the fine patch to determine the boundary value on the coarse patch. Unfortunately, this will generally not be exact for linear functions.

Consider a linear function that takes the value zero at the boundary. Because the weights are positive, the weighted average of the fine interior values will be either positive or negative, but it cannot be zero. (To get zero, you would need both positive and negative weights.)

So I think that we should in fact see first-order accuracy at the physical boundary. (Alas.)

I think that we should instead check that it captures constants exactly, and that it captures linear functions exactly in the interior.

boyceg:
> Consider a linear function that takes the value zero at the boundary. Because the weights are positive, the weighted average of the fine interior values will be either positive or negative, but it cannot be zero. (To get zero, you would need both positive and negative weights.)

(I should have said --- this follows because the interior fine values will either be all positive or all negative.)

boyceg:
BTW, I just tried putting random values on all boundaries, and the tester only shows the errors along the lower sides, not the upper sides. It appears that the physical boundary values are only being "fixed up" along the lower sides of the patch. E.g., if I just print out the location index information where the boundary fix-up is applied, it is only being applied on the lower sides of the patch.

I am not sure if this is a bug in the coarsen operator implementation (it would be in the C++ code, not the Fortran subroutine) or the tester.

boyceg:
OK, I continue to be unsure whether this is a bug in the coarsen operator, the tester, or possibly SAMRAI, but if I look at the boundary boxes on the coarse patch instead of the fine patch, then I get all of the boundaries.

boyceg:
@drwells, I just updated the code to use the coarse patch instead of the fine patch to determine where the boundaries are. We now get O(dx) errors on all boundaries, which I think is actually what we expect to see.

I am a bit nervous about this, however, because if you look at the comments in SAMRAI in `CoarsenSchedule.C`, you see:
```
/*
 * ************************************************************************
 *                                                                        *
 * Execute the stored communication schedule that copies data into the    *
 * the destination patch data components of the destination level from    *
 * the source patch data components of the source level.  The steps       *
 * to the algorithm are as follows:                                       *
 *                                                                        *
 *      (1) Allocate the source space on the temporary patch level.       *
 *      (2) Coarsen the data from the fine patch level to the temporary   *
 *          patch level (local operation).                                *
 *      (3) Copy data from the source space of the temporary patch        *
 *          level into the destination space of the destination patch     *
 *          level (requires interprocessor communication).                *
 *      (4) Deallocate the source space on the temporary patch level.     *
 *                                                                        *
 * ************************************************************************
 */
```
This means that the "fine" patch should be a real patch but the "coarse" patch may be a temporary patch, and I am not sure the coarse patch will be guaranteed to have boundary information set up correctly or not.

I wonder if this is a bug in the tester.

boyceg:
@abarret, can you help out here? 😀 

I tried asking the fine patch if it touches the regular boundary on each side, and it says it does:
```
    for (int axis = 0; axis < NDIM; ++axis)
      for (int upperlower = 0; upperlower <= 1; ++upperlower)
        pout << "axis = " << axis << " upperlower = " << upperlower << " touches bdry = " << fine.getPatchGeometry()->getTouchesRegularBoundary(axis, upperlower) << "\n";
```
prints out:
```
axis = 0 upperlower = 0 touches bdry = 1
axis = 0 upperlower = 1 touches bdry = 1
axis = 1 upperlower = 0 touches bdry = 1
axis = 1 upperlower = 1 touches bdry = 1
```
but I only get boundary boxes on the fine patch for the lower side, not the upper side.

boyceg:
I think that the `CoarsenOperator` implementation was actually correct, and so I am going to revert most of the last commit.

I think that the reason that the "upper" coarse patch boundary values are not being set "correctly" is that something is going wrong about the way that the `PatchGeometry` is being set up on the fine patch in the test.

Ironically, setting the coarse patch boundary values "correctly" is not as accurate as setting them "incorrectly". The reason is that, as described above, the adjoint of the RT0 `RefineOperator` cannot capture linear functions exactly on the physical boundary, because at such boundaries.

I suppose that there may be some use cases where it is better to lose the discrete adjoint property at physical boundaries and maintain higher accuracy.

boyceg:
@drwells (and maybe also @abarret ?? 😀), I think that the operator implementation is now correct, but there remains the issue with the test that the "upper" boundaries are not showing up in the `PatchGeometry` object associated with the fine patch. I think this is not a bug in the `CoarsenOperator`.

abarret:
The problem seems to be that the fine patch is too large. It covers the coarse patch extended by one cell. Outputting the boxes from the coarse and fine box, plus the boundary boxes on the fine patch gives
```
coarse boxes: [(0,0),(31,31)]
fine boxes: [(0,0),(131,131)]
fine boundary box: [(-1,0),(-1,127)]
fine boundary box: [(0,-1),(127,-1)]
```
The fine box should only by `[(0,0),(127,127)]`. I don't know why it's larger than that.

boyceg:
Got it:
```
StandardTagAndInitialize {
   tagging_method = "REFINE_BOXES"
   RefineBoxes {
      level_0 = [(0, 0), (N, N)]
   }
}
```
should be
```
StandardTagAndInitialize {
   tagging_method = "REFINE_BOXES"
   RefineBoxes {
      level_0 = [(0, 0), (N - 1, N - 1)]
   }
}
```
Do you have a minute to update and squash?

abarret:
After fixing that, are there any other problems with this branch? I'm not familiar with and haven't looked at anything other than the boundary boxes.

boyceg:
It would be good to add a 3D test like the 2D one, and to have variants of both tests that use either one patch or multiple patches.

This is all uniform grid stuff for now.

abarret:
Squashed and force pushed

boyceg:
Can you add 3D tests?

boyceg:
And also multi-patch tests?

abarret:
For some reason, if I force more patches by decreasing the largest patch size, I'm getting NaN's that appear on the coarse grid. This will take some time to debug

abarret:
Ah. I think it's because the fine grid doesn't have ghost cells set.

boyceg:
Yes needs ghost cells to work.

boyceg:
Could you just add the 3D single-patch test?

abarret:
I'm almost done with the multipatch test.

abarret:
I'm still getting something weird happening with multiple patches on the fine level. If I don't fill ghost cells outside the physical domain, I see NaNs appearing on the patch edges. This doesn't occur when there is a single patch on the fine level.

If I do fill ghost cells outside the physical domain, the max norm "error" is three orders of magnitude larger than on a single patch.

abarret:
I've pushed the 3D single patch test

boyceg:
Hmmm. What happens with periodic boundaries?

abarret:
The max error is located at an interior patch boundary, so switching to periodic boundaries does not seem to fix anything. I will take a closer look tomorrow. I suspect there must be an off by one index somewhere, but I'm not sure.

drwells:
Nice work. @abarret can you finish this? If so I'll move on to something else.

abarret:
This branch should be ready to go. I've added a multiple patch test for 2D and 3D. I didn't realize the test wasn't set up for multiple patches, so that was the cause of the large errors. I've squashed and fixed the conflicts. I'm rerunning the tests on a different system to make sure they work.

drwells:
Could you make the 3D tests smaller? Those are some huge output files for what we need to demonstrate (that errors are zero inside the domain)

abarret:
> Could you make the 3D tests smaller? Those are some huge output files for what we need to demonstrate (that errors are zero inside the domain)

I'll do something to change it. Although the errors are not zero entirely inside the domain, as Boyce pointed out earlier. This isn't "true" RT0, but the discrete adjoint of the RT0 refine operator. In particular, they are not zero on patch boundaries. Perhaps we can just print out a max error norm per component?

drwells:
I'd still prefer to print out the pointwise values.

> This isn't "true" RT0, but the discrete adjoint of the RT0 refine operator. In particular, they are not zero on patch boundaries.

Ah. This isn't a problem as long as we know what to expect.

boyceg:
The errors should be zero on interior patch boundaries!

abarret:
> I'd still prefer to print out the pointwise values.

In that case, we can drastically decrease the size of the domain.

> The errors should be zero on interior patch boundaries!

The errors are only on the boundary, but since we're using 32 grid points in each direction, for one 3D test, we print out 6000 numbers.

abarret:
@drwells, I decreased the size of the 3D tests. They're about 300 lines per test.

drwells (approved):

boyceg:
I want to move the RT0 coarsen tests into `tests/coarsen` (which seems to have been only partially set up) before merging. Right now `make tests` fails because there is not a `coarsen` test directory.

boyceg:
OK, test suite passes except for `IBFE/explicit_ex4_2d.multilevel.d.input`, which I think had some problems already.

drwells:
I know, I know... see #1173.

boyceg:
@drwells, @abarret, are we happy with this PR?

abarret (approved):

boyceg:
@drwells can you check out the changes to `CMakeLists.txt`?

drwells (approved):

drwells:
The CMakeLists changes are good but there's a really dumb CMake bug that I need to fix (that is totally unrelated to this patch).

PR #1123, opened 2020-09-16
Multilevel ibfe 5

drwells:
This is enough to get velocity interpolation working correctly - we get about the same (they are a factor of 2 better) errors as we do when the finest grid level is one level lower and all the elements are on that level.

`FEDataManager::collectActivePatchElements()` is getting a bit messy but I think it would be best for now to leave refactoring that for later - this is already pretty complicated.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -294,6 +294,22 @@ public:
         return it->second;
     }
 
+    /*!
+     * Return whether or not there are any elements on a given level.
+     */
+    bool levelHasElements(const int level_number) const
+    {
+        const auto array_it = std::find(d_fixed.begin(), d_fixed.end(), level_number);
+        if (array_it != d_fixed.end())
+        {
+            return true;
+        }
+        else
+        {
+            return d_map.count(level_number) != 0;
+        }
+    }
+
 private:
     /*!
      * Return a writable reference.
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -1129,10 +1129,29 @@ private:
      * are allowed to move no more than one cell width between regridding
      * operations).
      *
+     * The parameters refer to the levels of different objects:
+     * <ol>
+     *   <li>@p level_number - the level number in the patch hierarchy on which
+     *     we are identifying intersections.</li>
+     *   <li>@p coarsest_elem_ln - The minimum level number of elements we should
+     *     consider (see the main documentation of this class for an explanation
+     *     on how elements are assigned to particular levels)</li>
+     *   <li>@p finest_elem_ln - The maximum level number of elements we should
+     *     consider.</li>
+     * </ol>
+     *
+     * All three parameters are necessary because we use this function both to
+     * tag cells for refinement (i.e., we want to refine cells containing
+     * elements on levels higher than the present level) and to do IB
+     * calculations (where all three numbers will be the same).
+     *
      * In this method, the determination as to whether an element is local or
      * not is based on the position of the bounding box of the element.
      */
-    void collectActivePatchElements(std::vector<std::vector<libMesh::Elem*> >& active_patch_elems, int level_number);
+    void collectActivePatchElements(std::vector<std::vector<libMesh::Elem*> >& active_patch_elems,
+                                    int level_number,
+                                    int coarsest_elem_ln,
+                                    int finest_elem_ln);
 
     /*!
      * Collect all of the nodes of the active elements that are located within a
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -296,7 +296,7 @@ public:
 
 private:
     /*!
-     * Return a writeable reference.
+     * Return a writable reference.
      */
     int& get(const libMesh::subdomain_id_type id)
     {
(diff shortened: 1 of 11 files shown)


ibtk/include/ibtk/FEDataManager.h
```diff
@@ -294,9 +294,25 @@ class SubdomainToLevelTranslation
         return it->second;
     }
 
+    /*!
+     * Return whether or not there are any elements on a given level.
+     */
+    bool levelHasElements(const int level_number) const
```
boyceg, line 300: 👍

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2543,7 +2531,7 @@ FEDataManager::applyGradientDetector(const Pointer<BasePatchHierarchy<NDIM> > hi
         // level.
         std::vector<std::vector<Elem*> > active_level_elem_map;
         const IntVector<NDIM> ghost_width = 1;
-        collectActivePatchElements(active_level_elem_map, level_number);
+        collectActivePatchElements(active_level_elem_map, level_number, level_number + 1, d_max_level_number);
```
drwells, line 2534: This more precise API call replaces `if (getPatchLevel(elem) <= level_number) continue;`

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -564,21 +564,9 @@ FEDataManager::reinitElementMappings()
...
-        d_active_patch_elem_map[ln].resize(level->getProcessorMapping().getNumberOfLocalIndices());
-        d_active_patch_node_map[ln].resize(level->getProcessorMapping().getNumberOfLocalIndices());
-    }
-
-    // TODO: collectActivePatchElements still assumes every element is on the
-    // same patch level.
-    collectActivePatchElements(d_active_patch_elem_map.back(), d_max_level_number);
     for (int ln = 0; ln <= d_max_level_number; ++ln)
     {
+        collectActivePatchElements(d_active_patch_elem_map[ln], ln, ln, ln);
```
drwells, line 569: This is the main change in the patch - we now actually assign elements to their requested levels.

ibtk/include/ibtk/FEDataManager.h
drwells: Unless there is a bug in the implementation the check that uses this should prevent us from calculating the global set of bounding boxes any more times than we currently do on `master`.

drwells:
@boyceg are you happy with this? If so lets merge so we can move on to spreading :)

boyceg:
I got sidetracked. I will check it out tonight.

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -1129,10 +1145,29 @@ class FEDataManager : public SAMRAI::tbox::Serializable
      * are allowed to move no more than one cell width between regridding
      * operations).
      *
+     * The parameters refer to the levels of different objects:
+     * <ol>
+     *   <li>@p level_number - the level number in the patch hierarchy on which
+     *     we are identifying intersections.</li>
+     *   <li>@p coarsest_elem_ln - The minimum level number of elements we should
```
boyceg, line 1152: For input parameters, I wonder if it is better to be more verbose, e.g.:
```suggestion
     *   <li>@p coarsest_elem_level_num - The minimum level number of elements we should
```

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -1129,10 +1145,29 @@ class FEDataManager : public SAMRAI::tbox::Serializable
...
+     *     consider (see the main documentation of this class for an explanation
+     *     on how elements are assigned to particular levels)</li>
+     *   <li>@p finest_elem_ln - The maximum level number of elements we should
```
boyceg, line 1155: Similarly:
```suggestion
     *   <li>@p finest_elem_level_num - The maximum level number of elements we should
```

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -1129,10 +1145,29 @@ class FEDataManager : public SAMRAI::tbox::Serializable
...
+     * elements on levels higher than the present level) and to do IB
+     * calculations (where all three numbers will be the same).
+     *
      * In this method, the determination as to whether an element is local or
      * not is based on the position of the bounding box of the element.
      */
-    void collectActivePatchElements(std::vector<std::vector<libMesh::Elem*> >& active_patch_elems, int level_number);
+    void collectActivePatchElements(std::vector<std::vector<libMesh::Elem*> >& active_patch_elems,
+                                    int level_number,
+                                    int coarsest_elem_ln,
```
boyceg, line 1169:
```suggestion
                                    int coarsest_elem_level_num,
```

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -1129,10 +1145,29 @@ class FEDataManager : public SAMRAI::tbox::Serializable
...
+                                    int finest_elem_ln);
```
boyceg, line 1170:
```suggestion
                                    int finest_elem_level_num);
```
...although I guess this is in the private interface, so maybe it does not matter.

boyceg (approved):
Looks good to me.

drwells:
I went with `_ln` since that is what we use in `updateQuadPointCountData` and a couple of other functions - should we use `level_num` everywhere in the future?

drwells:
Since this is approved I will merge to keep the ball rolling and open an issue to track style changes.

boyceg:
> I went with `_ln` since that is what we use in `updateQuadPointCountData` and a couple of other functions - should we use `level_num` everywhere in the future?

I think `_ln` is fine. It may make sense to standardize on a suffix at some point.

boyceg:
I can't wait for _Multilevel ibfe 6_. 😀

drwells:
I didn't have to fix any bugs in `interpWeighted` for this to work so I am cautiously hopeful that multilevel spreading won't be too bad.

PR #1125, opened 2020-09-17
Improve the documentation of SubdomainToLevelTranslation.

drwells:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -261,30 +261,33 @@ protected:
 };
 
 /*!
- * Class that can translate libMesh subdomain IDs into levels.
+ * Class that can translate libMesh subdomain IDs into patch level numbers.
  *
- * The overwhelming majority of subdomain IDs used with IBAMR come from
- * block IDs set by ExodusII - these are numbered sequentially from zero.
- * However, in principle, a subdomain ID could be any signed 64-bit integer
- * so we cannot use a small fixed length array.
- *
- * Since we look up element levels a lot optimize for the common case by
- * using a fixed-length array and a map for everything else.
+ * The primary use of this class is multilevel IBFE - i.e., enabling a finite
+ * element mesh to interact with multiple patch levels.
  */
-class SubdomainToLevelTranslation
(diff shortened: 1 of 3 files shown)


ibtk/include/ibtk/FEDataManager.h
```diff
@@ -274,6 +274,18 @@ class FEData : public SAMRAI::tbox::Serializable
 class SubdomainToLevelTranslation
 {
 public:
+    /*!
+     * Constructor. Takes as argument the maximum level number and an input
```
boyceg, line 278:
```suggestion
     * Constructor. Takes as argument the maximum level number in the Cartesian grid patch hierarchy and an input
```

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -283,8 +295,8 @@ class SubdomainToLevelTranslation
     static constexpr int fixed_array_size = 256;
 
     /*!
-     * Return a constant reference. Returns d_max_level_number if no level
-     * is set
+     * Given a libMesh subdomain id, return the patch level on which that
```
boyceg, line 298:
```suggestion
     * Given a libMesh subdomain id, return the patch level of the Cartesian grid hierarchy on which that
```

boyceg (approved):
Looks good. I think that libMesh may have a notion of level also, and so it might be helpful to specify "level of the Cartesian grid hierarchy" or maybe "level number within the Cartesian grid hierarchy" to be crystal clear.

drwells:
Oh yeah, we even support multilevel FE now so we should fix that.

drwells:
Fixed!

boyceg (approved):
Looks good to me!

boyceg:
BTW, this is outside of the scope of this PR, but thinking about how "level" is an overloaded term, I wonder if it would be more clear to make the class name `SubdomainToPatchLevelTranslation`.

drwells:
That's easy enough to fix here - lets just rename it while we are at it.

drwells:
I also did another pass on documentation improvements.

Issue #1112, opened 2020-09-12
Multilevel IBFE support


drwells:
We talked about this a bit on slack and it's worth summarizing the status quo here.

The main goal is to let different `Elem`s interact with different levels in the patch hierarchy. The main application of this at the moment is putting valves in the heart model on some extra-special level while big structures are on a coarser level.

Users should be able to enable this feature by adding a new section to their input file:
```
IBFEMethod {
   // ... normal stuff goes here

   // this part is new
   FEDataManager {
      block_ids_on_levels {
         level_0 = // nothing on level 0
         level_1 = 100, 110 // big structures that aren't that dynamic are in the middle
         level_2 = // everything else on level 2
      }
   }
}
```
the idea is that we can keep backwards compatibility by assigning everything to the finest level by default. Unlike sidesets, block ids are unique to an element.

Here is our progress so far:
- [x] Give FEDataManager an input database (#1089)
- [x] Cleanup `FEDataManager` so that we do not need to worry about almost half a dozen constructors (#1093, #1104)
- [x] Modify `FEDataManager`'s data structures so that we loop over all levels instead of just the finest level (#1108)
- [x] Implement the `block_ids_on_levels` (or `subdomain_ids_on_levels`: the terminology isn't that important) input database and set up an appropriate data structure inside `FEDataManager` to store it (#1113)
- [x] Implement a function in FEDataManager that looks up what level an element should be on from its subdomain id (#1113)
- [x] Reimplement cell tagging based on what level an element should be on (#1119).
- [x] Reimplement the way we associate patches with elements (#1123).

At this point velocity interpolation should work automatically.
- [x] Correctly prolong quadrature point data so that levels are tagged correctly (#1127)
- [x] Implement some method for prolonging forces spread onto coarser patch levels up to finer patch levels as appropriate. This is the only step I don't know how to do yet (#1129)
- [x] Make sure things work when the mesh is *not* on the finest mesh level. (#1140)
- [x] Update the scratch hierarchy in IBFEMethod to work with multiple levels (this might be hard) (#1129, #1136)

boyceg:
@drwells, here is the algorithm for force spreading:
```
f := 0  /// on all levels of the composite grid

for each level (coarsest to finest) {
    if (!coarsest level) {
        f_{level} := prolong f_{level-1}   /// prolong from next coarser level (if it exists)
    }
    f_{level} += S_{level} F_{level}   /// spread the local forces
}

boyceg:
[...]

Issue #1124, opened 2020-09-17
Style Guide


drwells:
Not critical, but using consistent naming conventions makes code a lot easier to navigate. A few examples:

- all our headers are named after the class (like `IBFEMethod.h`) or have a name like `libmesh_utilities.h` to indicate their contents
- use `patch_num` or `level_num` instead of `pn` or `ln`
- use clang-format

If other people can think of other things we can add them here

drwells:
See also #1123.

drwells:
From #1126 - let's see if we can teach clang-format to put the function name at the end of functions like it does for namespaces.

PR #1126, opened 2020-09-17
Multilevel ibfe 6

drwells:
This PR is mostly a cleanup of a bunch of different FEDataManager functions - it no longer makes sense to ask for the level on which the patch hierarchy resides, nor does it make sense to assign this with a function call (its set up in the constructor by the input database).

I removed a bunch of functions so this isn't backwards compatible but since this feature didn't exist in the past I think it's not a big deal to break it.

Part of #1112.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -619,27 +619,14 @@ public:
     SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > getPatchHierarchy() const;
 
     /*!
-     * \brief Reset range of patch levels over which operations occur.
-     *
-     * The levels must exist in the hierarchy or an assertion failure will
-     * result.
+     * Get the coarsest patch level number on which elements are assigned.
      */
-    void setPatchLevels(int coarsest_ln, int finest_ln);
+    int getCoarsestPatchLevelNumber() const;
 
     /*!
-     * \brief Get the range of patch levels used by this object.
-     *
-     * \note Returns [coarsest_ln,finest_ln+1).
+     * Get the finest patch level number on which elements are assigned.
      */
-    std::pair<int, int> getPatchLevels() const;
-
-    //\}
-
-    /*!
-     * \return The level number to which the equations system object managed by
-     * the FEDataManager is assigned.
-     */
-    int getLevelNumber() const;
+    int getFinestPatchLevelNumber() const;
 
     /*!
      * \return The ghost cell width used for quantities that are to be
@@ -1243,7 +1230,6 @@ private:
      * Grid hierarchy information.
      */
     SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > d_hierarchy;
-    int d_coarsest_ln = IBTK::invalid_level_number, d_finest_ln = IBTK::invalid_level_number;
 
     /*!
      * Maximum possible level number in the patch hierarchy.
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -518,7 +518,9 @@ IBFEMethod::setupTagBuffer(Array<int>& tag_buffer, Pointer<GriddingAlgorithm<NDI
     for (unsigned int part = 0; part < d_meshes.size(); ++part)
     {
         const int gcw = d_primary_fe_data_managers[part]->getGhostCellWidth().max();
-        const int tag_ln = d_primary_fe_data_managers[part]->getLevelNumber() - 1;
+        // We need to refine cells up to, but not including, those on
+        // FEDataManager's finest patch level used for interaction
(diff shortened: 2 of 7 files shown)


boyceg:
I am not sure if it is a good idea to remove the method that sets the range of patch level numbers — although it is possibly just fine. Note that the number of levels will change as data structures are being initialized at the beginning of the simulation, and could conceivably include finer levels than the finest that is involved in the IBFE part of the computation.

Similarly we may want at some point to restrict the coarsest level to be finer than level 0.

Is there any harm in leaving them in?

drwells:
The main advantage to removing them is that we set the patch levels corresponding to each element exactly once (in the constructor) - there is no way to modify this information once its set.

> Note that the number of levels will change as data structures are being initialized at the beginning of the simulation, and could conceivably include finer levels than the finest that is involved in the IBFE part of the computation.

True, but the only thing we call in `FEDataManager` prior to the hierarchy being completely set up is applyGradientDetector which (at the first time step) reconstructs all the data it needs each time it is called.

drwells:
In a perfect world I would like to resolve this wart by splitting FEDataManager into two classes - one that deals with tagging cells at the initial point in time and another that does everything else since the two processes are quite different (they use the same functions but not the same data structures).

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -619,27 +619,14 @@ class FEDataManager : public SAMRAI::tbox::Serializable
...
-    /*!
-     * \brief Get the range of patch levels used by this object.
-     *
-     * \note Returns [coarsest_ln,finest_ln+1).
+     * Get the coarsest patch level number on which elements are assigned.
      */
-    std::pair<int, int> getPatchLevels() const;
-
-    //\}
+    int getCoarsestPatchLevelNumber() const;
```
boyceg, line 624: This seems fine, although
```suggestion
    std::pait<int,int> getPatchLevelNumberRange() const;
```
also could work?

src/IB/IBFEMethod.cpp
```diff
@@ -1772,8 +1789,9 @@ IBFEMethod::doInitializeFEEquationSystems()
         // object we only have to do this assignment once
         d_active_fe_data_managers[part]->COORDINATES_SYSTEM_NAME = COORDS_SYSTEM_NAME;
     }
+
     return;
-}
+} // doInitializeFEEquationSystems
```
boyceg, line 1794: Can we rip all of these out? 😀

(Of course I put most of them there in the first place.)

boyceg (approved):
👍

boyceg:
> The main advantage to removing them is that we set the patch levels corresponding to each element exactly once (in the constructor) - there is no way to modify this information once its set.
> 
> > Note that the number of levels will change as data structures are being initialized at the beginning of the simulation, and could conceivably include finer levels than the finest that is involved in the IBFE part of the computation.
> 
> True, but the only thing we call in `FEDataManager` prior to the hierarchy being completely set up is applyGradientDetector which (at the first time step) reconstructs all the data it needs each time it is called.

If it simplifies things to remove them, it is OK with me.

Right now we are usually operating on the full range of patch level numbers, and so this is redundant information.

ibtk/include/ibtk/FEDataManager.h
boyceg: I guess it is more clear to have the separate functions, since when you get a pair, is `second` the finest level number, or one past the finest level number, or ... ? 😀

src/IB/IBFEMethod.cpp
drwells: I actually kinda like these - it makes it easier, when you are at the end of some massive function, to see where you are.

If you don't like it then we can argue about it in #1124 :)

ibtk/include/ibtk/FEDataManager.h
drwells: They ultimately do the same thing (I think I would prefer to use one-past-the-end values everywhere in the library but we don't) - I went with this interface since it is more like what all the SAMRAI objects offer and it uses the word patch to be clear that we are not talking about libMesh FE levels.

drwells:
> Right now we are usually operating on the full range of patch level numbers, and so this is redundant information.

Yes but once we get force spreading going (hopefully today) it won't be :)

src/IB/IBFEMethod.cpp
boyceg: Yeah, that is the intention!

What I don't like about them is that they are maintained manually.

ibtk/include/ibtk/FEDataManager.h
boyceg: 👍

boyceg: I _think_ that one change in SAMRAI 3/4 is that they consistently use "one-past-the-end" semantics.

Issue #715, opened 2019-10-07
Simplify the relationship between IBFEMethod and FEDataManager


drwells:
These two classes know too much about each other which makes it difficult to make major changes to either. We should aim to reduce the interpedendencies between these classes.

The end goal of this issue is to move most of the initialization logic done by FEDataManager into free functions: it should be possible for IBFEMethod to set itself up without calling into FEDataManager at all. This makes things a lot simpler since FEDataManager will not be in a semivalid state during initialization.

PR #1127, opened 2020-09-18
Multilevel ibfe 7

drwells:
Posted here so that @boyceg can take a look. I think that this is most of what we need though there are still a few TODOs.

I am getting tired of having to wait for all the tests to relink so this PR includes the CMake build system - I'll take it out once everything works (that's not yet ready to be merged but in practice works)

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEDataManager.h
+++ b/ibtk/include/ibtk/FEDataManager.h
@@ -610,6 +610,11 @@ public:
 
     /*!
      * \brief Reset patch hierarchy over which operations occur.
+     *
+     * The patch hierarchy must be fully set up (i.e., contain all the levels it
+     * is expected to have) at the point this function is called. If you need to
+     * tag cells for refinement to create the initial hierarchy then use
+     * applyGradientDetector, which does not use the stored patch hierarchy.
      */
     void setPatchHierarchy(SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy);
 
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1283,24 +1283,13 @@ IBFEMethod::initializePatchHierarchy(Pointer<PatchHierarchy<NDIM> > hierarchy,
     // Initialize the FE data managers.
     reinitElementMappings();
 
-    // Set up the Eulerian data caches.
-    int coarsest_fe_level_num = std::numeric_limits<int>::max();
-    int finest_fe_level_num = std::numeric_limits<int>::min();
-    for (unsigned int part = 0; part < d_meshes.size(); ++part)
-    {
-        coarsest_fe_level_num =
-            std::min(coarsest_fe_level_num, d_primary_fe_data_managers[part]->getCoarsestPatchLevelNumber());
-        finest_fe_level_num =
-            std::max(finest_fe_level_num, d_primary_fe_data_managers[part]->getFinestPatchLevelNumber());
-    }
-
-    // Set up the scratch data cache to work on levels that actually have elements.
+    // FEDataManager needs all scratch data for tagging cells:
     d_primary_eulerian_data_cache->setPatchHierarchy(hierarchy);
-    d_primary_eulerian_data_cache->resetLevels(coarsest_fe_level_num, finest_fe_level_num);
+    d_primary_eulerian_data_cache->resetLevels(0, getFinestPatchLevelNumber());
     if (d_use_scratch_hierarchy)
     {
         d_scratch_eulerian_data_cache->setPatchHierarchy(d_scratch_hierarchy);
-        d_scratch_eulerian_data_cache->resetLevels(coarsest_fe_level_num, finest_fe_level_num);
+        d_scratch_eulerian_data_cache->resetLevels(0, getFinestPatchLevelNumber());
     }
 
     d_is_initialized = true;
@@ -2815,6 +2804,30 @@ IBFEMethod::imposeJumpConditions(const int f_data_idx,
     return;
 } // imposeJumpConditions
 
+int
+IBFEMethod::getCoarsestPatchLevelNumber() const
+{
+    int level_number = std::numeric_limits<int>::max();
+    for (unsigned int part = 0; part < d_meshes.size(); ++part)
+    {
+        level_number = std::min(d_primary_fe_data_managers[part]->getCoarsestPatchLevelNumber(), level_number);
+    }
+    TBOX_ASSERT(level_number != std::numeric_limits<int>::max());
+    return level_number;
+} // getCoarsestPatchLevelNumber
+
+int
+IBFEMethod::getFinestPatchLevelNumber() const
+{
+    int level_number = std::numeric_limits<int>::min();
+    for (unsigned int part = 0; part < d_meshes.size(); ++part)
+    {
+        level_number = std::max(d_primary_fe_data_managers[part]->getFinestPatchLevelNumber(), level_number);
+    }
+    TBOX_ASSERT(level_number != std::numeric_limits<int>::min());
+    return level_number;
+} // getFinestPatchLevelNumber
+
 SAMRAI::xfer::RefineSchedule<NDIM>&
 IBFEMethod::getPrimaryToScratchSchedule(const int level_number,
                                         const int primary_data_idx,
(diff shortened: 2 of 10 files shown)


src/IB/IBFEMethod.cpp
```diff
@@ -1017,6 +1022,41 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+    if (d_use_scratch_hierarchy)
+    {
+        // not yet implemented
+        assertStructureOnFinestLevel();
+    }
+    else
+    {
+        // TODO - we should save these schedules in the same manner that we save
+        // the schedules for the scratch hierarchy.
+        for (int ln = getCoarsestPatchLevelNumber(); ln < getFinestPatchLevelNumber(); ++ln)
```
drwells, line 1035: This is the relevant change.

src/IB/IBFEMethod.cpp
```diff
@@ -1017,6 +1022,41 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+        {
+            Pointer<RefineAlgorithm<NDIM> > refine_algorithm = new RefineAlgorithm<NDIM>();
+            Pointer<CartesianGridGeometry<NDIM> > geometry = d_hierarchy->getGridGeometry();
+            // TODO: this is probably the wrong refinement operator
+            Pointer<RefineOperator<NDIM> > refine_op =
+                geometry->lookupRefineOperator(f_var, "CONSERVATIVE_LINEAR_REFINE");
+            // TODO: do we need another scratch data index? We should be able to
+            // prolong directly into the output index
+            refine_algorithm->registerRefine(
+                f_prolong_scratch_data_idx, f_scratch_data_idx, f_prolong_scratch_data_idx, refine_op);
```
boyceg, line 1045:
```suggestion
                f_prolong_scratch_data_idx, f_prolong_scratch_data_idx, f_prolong_scratch_data_idx, refine_op);
```

src/IB/IBFEMethod.cpp
```diff
@@ -1017,6 +1022,41 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+    else
+    {
+        // TODO - we should save these schedules in the same manner that we save
+        // the schedules for the scratch hierarchy.
+        for (int ln = getCoarsestPatchLevelNumber(); ln < getFinestPatchLevelNumber(); ++ln)
+        {
+            Pointer<RefineAlgorithm<NDIM> > refine_algorithm = new RefineAlgorithm<NDIM>();
+            Pointer<CartesianGridGeometry<NDIM> > geometry = d_hierarchy->getGridGeometry();
+            // TODO: this is probably the wrong refinement operator
+            Pointer<RefineOperator<NDIM> > refine_op =
```
boyceg, line 1040:
```suggestion
            Pointer<RefineOperator<NDIM> > refine_op = d_f_prolong_op;
```

src/IB/IBFEMethod.cpp
```diff
@@ -1017,6 +1022,41 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+            // prolong directly into the output index
+            refine_algorithm->registerRefine(
+                f_prolong_scratch_data_idx, f_scratch_data_idx, f_prolong_scratch_data_idx, refine_op);
+            Pointer<PatchLevel<NDIM> > coarse_level = d_hierarchy->getPatchLevel(ln);
+            Pointer<PatchLevel<NDIM> > fine_level = d_hierarchy->getPatchLevel(ln + 1);
+            // We can ignore the fourth argument since we don't need to read
+            // forces outside the physical domain
+            Pointer<RefineSchedule<NDIM> > schedule = refine_algorithm->createSchedule(
+                "FILL_INTERIORS_ONLY", fine_level, coarse_level, nullptr, false, nullptr);
+            schedule->fillData(data_time);
```
boyceg, line 1052: I would do this all in `FEDataManager`...

boyceg: ...and @drwells and I just discussed on Slack why this is not a good thing to do for multi-part models.

src/IB/IBFEMethod.cpp
```diff
@@ -1017,6 +1022,41 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+            Pointer<RefineOperator<NDIM> > refine_op =
+                geometry->lookupRefineOperator(f_var, "CONSERVATIVE_LINEAR_REFINE");
+            // TODO: do we need another scratch data index? We should be able to
+            // prolong directly into the output index
+            refine_algorithm->registerRefine(
+                f_prolong_scratch_data_idx, f_scratch_data_idx, f_prolong_scratch_data_idx, refine_op);
+            Pointer<PatchLevel<NDIM> > coarse_level = d_hierarchy->getPatchLevel(ln);
+            Pointer<PatchLevel<NDIM> > fine_level = d_hierarchy->getPatchLevel(ln + 1);
+            // We can ignore the fourth argument since we don't need to read
+            // forces outside the physical domain
```
boyceg, line 1049: perhaps add a comment here to the effect that dealing with stuff being spread "outside of the domain" should have already been handled by `FEDataManager`?

drwells: Sure (though we now do this inside IBFEMethod itself)

src/IB/IBFEMethod.cpp
drwells: I don't understand this change. The second argument should be the source index - that's definitely `f_scratch_data_idx`. What am I missing?

src/IB/IBFEMethod.cpp
boyceg: I clearly have forgotten the changes that we made to try to improve multi-part performance.

src/IB/IBFEMethod.cpp
```diff
@@ -947,14 +947,18 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
     // zero data.
     Pointer<hier::Variable<NDIM> > f_var;
     VariableDatabase<NDIM>::getDatabase()->mapIndexToVariable(f_data_idx, f_var);
     auto f_active_data_ops = HierarchyDataOpsManager<NDIM>::getManager()->getOperationsDouble(f_var, hierarchy);
-    f_active_data_ops->resetLevels(ln, ln);
+    f_active_data_ops->resetLevels(getCoarsestPatchLevelNumber(), getFinestPatchLevelNumber());
     f_active_data_ops->setToScalar(f_scratch_data_idx,
                                    0.0,
                                    /*interior_only*/ false);
+    f_active_data_ops->setToScalar(f_prolong_scratch_data_idx,
```
boyceg, line 959: If this will only show up as the scratch context in a `RefineSchedule`, I don't think you need to zero it out.

src/IB/IBFEMethod.cpp
boyceg: I misunderstood what is going on here --- I think that the current code is right.

src/IB/IBFEMethod.cpp
drwells: In summary - we don't want to communicate inside FEDataManager when looping over parts since it leads to big workload imbalances.

boyceg: It might be good someday to have something that sits between `IBFEMethod` and the single-part `FEDataManager` so that `IBFEMethod` doesn't need to know so much about what `FEDataManager` does.

drwells: They have a very unhealthy codependent relationship :( see also #715.

src/IB/IBFEMethod.cpp
drwells: At first I agreed but when I made this change the tests started to fail (it looks like we end up reading uninitialized data) so we need it for reasons I don't fully understand.

src/IB/IBFEMethod.cpp
drwells: I'm stuck on this one - I am still looking for a way to pull this out of `IBExplicitHierarchyIntegrator`. As far as I can tell `RefineOperator`s get assigned to `RefineAlgorithm`s and then cannot be retrieved again.

boyceg: If you have a `RefineSchedule`, you can eventually get to the `RefineOperator` through `RefineSchedule::getEquivalenceClasses()`.

boyceg: You can do the same with `RefineAlgorithm::getEquivalenceClasses()`.

boyceg: But it might be better to add something to the `IBHierarchyIntegrator` interface to make it easier to access.

The model that I had in mind when setting up `IBHierarchyIntegrator` was that `IBStrategy` implementations would only need the `RefineSchedule`s and not need to know any details.

boyceg: Could you use the existing interface if you first made a copy of the level-by-level force data and then use the `RefineSchedule`s passed in via `IBStrategy::spreadForce()`?

If you don't like the idea of copying data, SAMRAI also makes it easy to "swap" storage associated with two patch data indices.

src/IB/IBFEMethod.cpp
boyceg: Yeah, but the way that you are using it here, I think you _do_ need to zero out --- you are not just using it as scratch data.

src/IB/IBFEMethod.cpp
boyceg: BTW, looking at `IBHierarchyIntegrator`, the prolongation and refine operators are hard coded anyway. Probably they should be made to be run-time options.

drwells:
Prolongation appears to work but there are some additional patch-to-element mapping bugs that need to be fixed.

drwells:
Working on multilevel support revealed a cascade of latent bugs in the way we set up Eulerian scratch data that needed to be fixed. Hence this PR has morphed from "finish multilevel IBFE by implementing prolongation" to "fix the way we tag cells for refinement by restricting/coarsening quadrature point count data".

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -610,6 +610,11 @@ class FEDataManager : public SAMRAI::tbox::Serializable
 
     /*!
      * \brief Reset patch hierarchy over which operations occur.
+     *
+     * The patch hierarchy must be fully set up (i.e., contain all the levels it
+     * is expected to have) at the point this function is called. If you need to
+     * tag cells for refinement to create the initial hierarchy then use
+     * applyGradientDetector, which does not used the stored patch hierarchy.
```
boyceg, line 617:
```suggestion
     * applyGradientDetector, which does not use the stored patch hierarchy.
```

boyceg (approved):
👍

PR #1129, opened 2020-09-22
Multilevel ibfe 8

drwells:
This builds on #1127 (and for now includes its commits) - these changes let us spread forces onto multiple levels in FEDataManager and then prolong them appropriately afterwards. There are still some things left to do in this PR:
- [x] Store the communication schedules so that we do not need to keep recomputing them.
- [x] Add some spreading tests that verify the correct order of convergence (perhaps we can spread forces and then interpolate them again and see what we get)
- [x] Add some tests that execute 'real' programs to make sure all the pieces work
- [x] Find a cleaner API for propagating 'what coarsening algorithm should we use with forces' to the call site (see the discussion thread here: https://github.com/IBAMR/IBAMR/pull/1127#discussion_r491117137)

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
This patch also temporarily includes the CMake build system since using it saves huge amounts of time when recompiling/relinking the test suite.

drwells:
This seems to work in practice. Here is my favorite example, where we only put the top part of the circle on level 2 and the rest is on level 0:

visit0001

drwells:
I fixed some more bugs but prolongation doesn't seem to work correctly. Here are the force magnitudes of an IBFE example (the boundary of the structure is on level 3, the interior is on level 2) where I set the zero value to pink to make it clear where there are no forces:

visit0001
The forces computed on the grid levels themselves look right but the prolongation goes to the wrong place.

By comparison, if we totally disable prolongation we get
visit0002
which shows that the prolongation is responsible for filling in that region outside the structure.

boyceg:
@drwells, do things seem to work if you put the whole structure on level 0 but also have a level 1 underneath it? You can do this by using `REFINE_BOXES` in the input file.

drwells:
No, for mostly unrelated reasons - there is a bunch of code that breaks if the structure is not at least partly on the finest grid level.

boyceg:
Have you looked at what you get if you spread the constant function?

drwells:
That's worth a shot - let me try that.

boyceg:
To me, this looks like the prolonged force is not getting added in on the finer level.

src/IB/IBFEMethod.cpp
```diff
@@ -1011,12 +1025,47 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+            refine_algorithm->registerRefine(
+                f_prolong_scratch_data_idx, f_scratch_data_idx, f_prolong_scratch_scratch_data_idx, refine_op);
+            Pointer<PatchLevel<NDIM> > coarse_level = d_hierarchy->getPatchLevel(ln);
+            Pointer<PatchLevel<NDIM> > fine_level = d_hierarchy->getPatchLevel(ln + 1);
+            // We can ignore the fourth argument since we don't need to deal with
+            // forces outside the physical domain (this was handled above by f_phys_bdry_op)
+            Pointer<RefineSchedule<NDIM> > schedule = refine_algorithm->createSchedule(
+                "FILL_INTERIORS_ONLY", fine_level, coarse_level, nullptr, false, nullptr);
+            schedule->fillData(data_time);
+            // Add the values prolonged from the level below onto the scratch index.
```
boyceg, line 1062:
```suggestion
            // Add the values prolonged from the coarser level onto the scratch index on the finer level.
```

src/IB/IBFEMethod.cpp
```diff
@@ -1011,12 +1025,47 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+                geometry->lookupRefineOperator(f_var, "CONSERVATIVE_LINEAR_REFINE");
+            TBOX_ASSERT(refine_op);
+            refine_algorithm->registerRefine(
+                f_prolong_scratch_data_idx, f_scratch_data_idx, f_prolong_scratch_scratch_data_idx, refine_op);
+            Pointer<PatchLevel<NDIM> > coarse_level = d_hierarchy->getPatchLevel(ln);
+            Pointer<PatchLevel<NDIM> > fine_level = d_hierarchy->getPatchLevel(ln + 1);
+            // We can ignore the fourth argument since we don't need to deal with
+            // forces outside the physical domain (this was handled above by f_phys_bdry_op)
+            Pointer<RefineSchedule<NDIM> > schedule = refine_algorithm->createSchedule(
+                "FILL_INTERIORS_ONLY", fine_level, coarse_level, nullptr, false, nullptr);
```
boyceg, line 1060: Try omitting the fill pattern for now. Usually these are just optimizations.

src/IB/IBFEMethod.cpp
```diff
@@ -1011,12 +1025,47 @@ IBFEMethod::spreadForce(const int f_data_idx,
...
+            // TODO: this is probably the wrong refinement operator
+            Pointer<RefineOperator<NDIM> > refine_op =
+                geometry->lookupRefineOperator(f_var, "CONSERVATIVE_LINEAR_REFINE");
+            TBOX_ASSERT(refine_op);
+            refine_algorithm->registerRefine(
+                f_prolong_scratch_data_idx, f_scratch_data_idx, f_prolong_scratch_scratch_data_idx, refine_op);
+            Pointer<PatchLevel<NDIM> > coarse_level = d_hierarchy->getPatchLevel(ln);
+            Pointer<PatchLevel<NDIM> > fine_level = d_hierarchy->getPatchLevel(ln + 1);
+            // We can ignore the fourth argument since we don't need to deal with
+            // forces outside the physical domain (this was handled above by f_phys_bdry_op)
```
boyceg, line 1058: Maybe --- maybe not --- depending on the refine operator you may need boundary values close to the boundary. This is one reason to prefer to use RT0 here.

boyceg:
BTW, you have looked at forces just on the coarser level and they look OK? It is just the prolonged coarse forces that are not coming through somehow?

drwells:
> BTW, you have looked at forces just on the coarser level and they look OK? It is just the prolonged coarse forces that are not coming through somehow?

I am not sure how to make SAMRAI drop the finest patch level but as far as I can tell they look right.

boyceg:
You should be able to look in VisIt.

abarret:
> > BTW, you have looked at forces just on the coarser level and they look OK? It is just the prolonged coarse forces that are not coming through somehow?
> 
> I am not sure how to make SAMRAI drop the finest patch level but as far as I can tell they look right.

@drwells you can use the Index Select operator to show data only on specific levels.

drwells:
Even if the forces are not spread correctly there is still a bug related to prolongation. If we add the following code
```cpp
        auto f_data_ops = HierarchyDataOpsManager<NDIM>::getManager()->getOperationsDouble(f_var, d_hierarchy);
        f_data_ops->resetLevels(0, getFinestPatchLevelNumber());
        f_data_ops->setToScalar(f_scratch_data_idx, std::sqrt(2)/2.0, /*interior_only*/ false);
```
right before the loop over levels (two levels, structure lives on the finest level, so we only have two patches) we get

visit0003

which doesn't make any sense - its like we are only prolonging into part of the finer patch.

src/IB/IBFEMethod.cpp
drwells: Good point. I'll get back to this later once the prolongation is working.

drwells:
I counted the cells and it looks like we are only prolonging a quarter of the patch - how bizarre!

boyceg:
Can we try to mimic what is done in the FAC preconditioner implementations?
```
    // TODO: Here we take a pessimistic approach and are recreating refine schedule for
    // (coarsest_reset_ln - 1) level as well.
    for (int dst_ln = std::max(d_coarsest_ln + 1, coarsest_reset_ln - 1); dst_ln <= finest_reset_ln; ++dst_ln)
    {
        d_prolongation_refine_schedules[dst_ln] =
            d_prolongation_refine_algorithm->createSchedule(d_hierarchy->getPatchLevel(dst_ln),
                                                            Pointer<PatchLevel<NDIM> >(),
                                                            dst_ln - 1,
                                                            d_hierarchy,
                                                            d_prolongation_refine_patch_strategy.getPointer());
    }

    for (int dst_ln = coarsest_reset_ln; dst_ln < std::min(finest_reset_ln + 1, d_finest_ln); ++dst_ln)
    {
        d_restriction_coarsen_schedules[dst_ln] = d_restriction_coarsen_algorithm->createSchedule(
            d_hierarchy->getPatchLevel(dst_ln), d_hierarchy->getPatchLevel(dst_ln + 1));
    }
```

boyceg:
In the way that you are setting this up, I don't think that SAMRAI knows that the two levels have different index spaces.

(`RefineAlgorithm`/`RefineSchedule` arguably should be broken up into two different sets of classes: one that only does stuff related to _refining_, and another than only does stuff related to ghost filling within a single level.)

boyceg:
Let me annotate this:
```
refine_algorithm->createSchedule(dst_level,               // dst level to fill to
                                 dst_level_old,           // another version of the dst level to copy from --- if available --- can be NULL
                                 dst_ln - 1,              // src level number to refine from in filling dst level
                                 patch_hierarchy,         // patch hierarchy to look in to find the src level
                                 refine_patch_strategy);  // routines for handling physical boundaries, c-f interfaces, etc 
```
You will want to do:
```
refine_algorithm->createSchedule(dst_level,
                                 nullptr,
                                 dst_ln - 1,
                                 patch_hierarchy,
                                 refine_patch_strategy);
```

drwells:
> In the way that you are setting this up, I don't think that SAMRAI knows that the two levels have different index spaces.

I just had the same thought - it is very suspicious that with a refinement ratio of 2 I end up with a quarter of the domain. I checked and we get 1/16th of the domain with a refinement ratio of 4 so this is probably it.

boyceg:
I think you are using this version of `createSchedule()`: https://ibamr.github.io/IBAMR-docs/samrai/2.4.4/html/classSAMRAI_1_1xfer_1_1RefineAlgorithm.html#f43c19270e614c958f83b8d87e9ed64d

Note that this is intended to be used with two levels with the same index space.

drwells:
>  Note that both levels must reside in the same AMR hierarchy index space, or in index spaces that represent the same level of mesh refinement. No spatial interpolation is performed.

alas, if only I knew how to read.

boyceg:
This complexity is one reason that [`HierarchyGhostCellInterpolation`](https://ibamr.github.io/IBAMR-docs/ibamr/html/class_i_b_t_k_1_1_hierarchy_ghost_cell_interpolation.html) exists.

drwells:
Yup, that fixes the problem (or at least most of them).

It looks like we are spuriously getting extra forces at the coarse-fine interface:
visit0006

that thin arc in the middle corresponds exactly with where elements go from level 3 to level 1.

drwells:
I think I figured out the problem in the picture above: values spread into ghost regions of fine patches need to be coarsened into the corresponding coarse patches below them. This could be tricky.

boyceg:
I think that should only impact plotting and not numerics.

You can always coarsen out after prolonging to synch up everything.

boyceg:
In particular, values on coarse levels that are covered by fine values (in the “invalid” region of the coarse grid) have no physical effect in the numerics.

drwells:
I think this is still relevant in the case of force spreading. Consider a point `X` with a nine-point stencil:

```
+----+----+----+----+
|    |    |    |    |
|    |    |    |    | 
|    |    |    |    |
+----x----x----x----+
|    |    |    |    |
|    |    |    |    | 
|    |  X |    |    |
+----x----x----x----+
|    |    |    |    |
|    |    |    |    | 
|    |    |    |    |
+----x----x----x----+
|    |    |    |    |
|    |    |    |    | 
|    |    |    |    |
+----+----+----+----+
```
here we spread to nine `x`s. On the other hand, if we are at a coarse-fine interface, then
```

+----+----+----+----+
|    |    |         |
|    |    |         | 
|    |    |         |
+----x----x         |
|    |    |         |
|    |    |         | 
|    |  X |         |
+----x----x---------+
|    |    |         |
|    |    |         | 
|    |    |         |
+----x----x         |
|    |    |         |
|    |    |         | 
|    |    |         |
+----+----+----+----+
```
i.e., the effect of spreading forces into the rightmost three points is no longer accounted for: we need to restrict that force onto the next finer level.

boyceg:
Yes but there is no way for the fine grid stuff to effect the coarse grid forces in the “valid” region.

Keep in mind that spreading happens on each element as if it is on a uniform grid. All of the linkage between levels is via refinement/prolongation.

boyceg:
> Yes but there is no way for the fine grid stuff to effect the coarse grid forces in the “valid” region.
> 
> Keep in mind that spreading happens on each element as if it is on a uniform grid. All of the linkage between levels is via refinement/prolongation.

Specifically, if “X” is talking to the fine grid, then the fine grid should have already been constructed so that It cannot spread outside of the coarse grid.

You didn’t change the way tagging works by any chance? 😀

drwells:
I didn't change how tagging works.

> Specifically, if “X” is talking to the fine grid, then the fine grid should have already been constructed so that It cannot spread outside of the coarse grid.

You're right - it must be something else doing this (or just some numerical artifact). I need to write some method of manufactured solutions tests next to convince myself this does what I think it does.

boyceg:
Whoops, emailing while trying to get kids to sleep causes typos. I meant: it cannot spread outside of the fine grid.

Also, if you are worried about hitting the coarse side of the coarse-fine interface, you can always increase the tag buffer to be sure that there are enough fine grid cells.

drwells:
It's late and I read it correctly anyway :)

drwells:
There is definitely a numerical artifact of some sort with RT0 prolongation. Here is what we get when we spread a force of 1 from a structure (here the physical boundary is far away):

visit0000
The line separating the blue and red regions is also the region separating elements interacting with the fine grid (left) from those interacting with the coarse grid (right).

The red is overshoot and the blue is undershoot. The amount of each appears to be independent of grid resolution and MFAC.
 I guess the good news is that the amount of overshoot equals the amount of undershoot (the values appear to average to the right answer).

drwells:
Essentially the same thing happens with the conservative linear refine.

boyceg:
There should not be any undershoot or overshoot with RT0 prolongation.

drwells:
I agree. I still cannot find a good explanation for why this occurs only at the interface.

boyceg:
Here is a thought. Is this from prolongation or is it related to spreading from multiple levels? Can you zero-out the spread forces on the fine level to ensure that you are really looking at prolongation-related artifacts?

boyceg:
It also occurs to me that we might benefit from blending the kernel function instead of having an abrupt change. Let's worry about that later.

drwells:
Yes - I tried that and I get a gradient.

> is it related to spreading from multiple levels?

I don't think so - if we use a refinement ratio of 1 then we get the correct values.

boyceg:
> Yes - I tried that and I get a gradient.

The same kind of gradient with under/overshoots?

drwells:
No - it goes monotonically down to zero. The problem is the slope - it should go down faster so that we do not have an undervalued region and then an overvalued region as we go from coarse to fine.

boyceg:
> No - it goes monotonically down to zero. The problem is the slope - it should go down faster so that we do not have an undervalued region and then an overvalued region as we go from coarse to fine.

Can you post an image of what this looks like if you zero out the fine grid values and then prolong from the coarse grid?

drwells:
Definitely, though I find the color scale hard to interpret. 

Rather than zeroing data, I prolong the values and then I swap (so the value we end up with is exactly what we got from prolongation).

visit0001
I have no idea what color scheme to use. I modified the test to use a fine grid to the right of 0.5 and coarse grid to the left of 0.5.

boyceg:
Can you just do a line-out?

boyceg:
I cannot make sense out of the colors 😁

drwells:
Yeah they are really hard to read. How do I do a line-out?

boyceg:
There is a little button in the viewer window: https://visit-sphinx-github-user-manual.readthedocs.io/en/latest/Quantitative/Lineout.html.

drwells:
My line-out doesn't look nearly as good as theirs - give me a minute.

boyceg:
You also can specify the endpoints by manually applying the operator.

drwells:
Here are the values solely resulting from prolongation:
visit0002
here is the line:
visit0003

boyceg:
OK this looks fine.

I think that this is not related to prolongation _per se_, but rather about the abrupt change in kernel size in the spreading operator.

Let's just see what happens when we use this. It would be a surprise if there are no artifacts when doing this.

drwells:
I just finished two very long runs that demonstrate exactly that.

Here I have a color scheme that shows forces in certain brackets on the interior of the domain. Due to the differences in the spread operator we can see where we switch from interacting with the fine and interacting with the coarse levels:
visit0000
visit0001

boyceg:
Looks pretty good actually. I am less worried about what the forces look like, and more worried about structural deformations and stresses.

boyceg:
There are always artifacts at coarse-fine interfaces.

drwells:
Is there an easy scalar we can plot to check the quality?

boyceg:
We could look at something like von Mises stresses.

drwells:
Is there an easy way to compute that from the components of the deviatoric stress tensor? You can tell that I haven't tried to di this before.

boyceg:
This is what the post processor is supposed to help with. I don’t think we have a built in expression for those, although it would be easy enough to add.

IIRC, the post processor basically can evaluate either the L2 projection or the direct interpolation of simple C-style functions like what are used in evaluating pointwise stresses, forces, etc.

drwells:
Yeah, but I already postprocessed the deviatoric stresses so I want to use them :)

boyceg:
You can make an expression in visit to evaluate them if you have the right stuff in the viz file.

boyceg:
You also could evaluate the principal stress.

src/IB/IBFEMethod.cpp
drwells: I thought about this some more (and I had to reread some other comments I left in IBFEMethod after #896): since we spread into ghost regions that f_phys_bdry_op would in turn accumulate into another ghost region we need to handle this part prior to the global 'sum forces spread into ghost regions into the correct patch interiors' step. Hence - at this point in the routine we have already accumulated all forces so we should deliberately ignore these values.

drwells:
OK that should do it - I fixed the remaining bugs, added tests, and got everything working with the scratch hierarchy. I also added a note to `FEDataManager` about the artifacts.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -1137,6 +1137,7 @@ FEDataManager::spread(const int f_data_idx,
                 for (unsigned int e_idx = 0; e_idx < num_active_patch_elems; ++e_idx)
                 {
                     Elem* const elem = patch_elems[e_idx];
+                    TBOX_ASSERT(getPatchLevel(elem) == ln);
```
boyceg, line 1140: Do we want to include a debugging guard? I think that `TBOX_ASSERT` is _not_ a no-op in SAMRAI 2 (but that it is in SAMRAI 3/4).

include/ibamr/IBFEMethod.h
```diff
@@ -892,6 +892,14 @@ class IBFEMethod : public FEMechanicsBase, public IBStrategy
                                 int scratch_data_idx,
                                 SAMRAI::xfer::RefinePatchStrategy<NDIM>* patch_strategy = nullptr);
 
+    /*!
+     * Get the schedule used to prolong force values. Data is read from @p
+     * coarse_data_idx on level @p level_number and written into fine_data_idx
+     * on level level_number + 1.
```
boyceg, line 898: 👍

src/IB/IBFEMethod.cpp
```diff
@@ -2871,6 +2898,49 @@ IBFEMethod::getScratchToPrimarySchedule(const int level_number,
...
+        {
+            Pointer<CartesianGridGeometry<NDIM> > geometry = hierarchy->getGridGeometry();
+            refine_op = geometry->lookupRefineOperator(f_var, "CONSERVATIVE_LINEAR_REFINE");
+        }
+        else
+            refine_op = new CartSideDoubleRT0Refine();
+        TBOX_ASSERT(refine_op);
+        refine_algorithm->registerRefine(fine_data_idx, coarse_data_idx, fine_data_idx, refine_op);
+        Pointer<PatchLevel<NDIM> > fine_level = hierarchy->getPatchLevel(level_number + 1);
+        // We can ignore the fourth argument since we don't need to deal with
```
boyceg, line 2935: Fifth argument?

Might be more clear to make a variable set to `nullptr` that is passed in as an argument to `createSchedule`.

I could imagine being confused by this when revisiting the code some day. 😀

boyceg (approved):
Looks good. Some minor comments.

boyceg:
Thinking about `TBOX_ASSERT` --- we should possible have an `IBTK_ASSERT` and/or `IBAMR_ASSERT` that does the right thing depending on the status of `NDEBUG`, and use that instead of `TBOX_ASSERT`.

drwells:
We might want to move that discussion somewhere else but I think its generally OK to run cheap asserts (i.e., asserts not in deeply nested loops) all the time.

ibtk/src/lagrangian/FEDataManager.cpp
drwells: Yeah, lets do this. This check should always pass so in principle it should only be run when we are debugging IBAMR itself.

src/IB/IBFEMethod.cpp
drwells: Oops - wrong comments are bad. Fixed.

boyceg (approved):

boyceg:
I am not proposing to change the way we handle asserts in this particular PR. 😁

drwells:
I'd hope so, but we should open an issue for it so we remember to discuss is later :)

PR #1130, opened 2020-09-22
Cleanup initializeFEEquationSystems

drwells:
It turns out that we can call this function in the constructor (and hence simplify a bunch of code) in most cases - the drawback at the moment is that `IBFEMethod::setInterpSpec` at al don't work anymore with this patch (those are used in the TAVR model and also IBFE example 6).

It should be possible to work around this in two ways:
1. Rewrite those functions to pass the information along to the FEDataManager object.
2. Expand the `IBFEMethod` input database to take `FEDataManager_0`, `FEDataManager_1`, or some other naming scheme that lets us set these things up in the constructor.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/include/ibamr/FEMechanicsBase.h
+++ b/include/ibamr/FEMechanicsBase.h
@@ -403,11 +403,6 @@ public:
 
 protected:
     /*!
-     * Do the actual work in initializeFEEquationSystems.
-     */
-    virtual void doInitializeFEEquationSystems() = 0;
-
-    /*!
      * Do the actual work in reinitializeFEData and initializeFEData. if @p
      * use_present_data is `true` then the current content of the solution
      * vectors is used: more exactly, the coordinates and velocities (computed
@@ -546,12 +541,6 @@ protected:
      * Object managing access to libMesh system vectors for the pressure.
      */
     std::unique_ptr<IBTK::LibMeshSystemVectors> d_P_vecs;
-
-    /*!
-     * Whether or not the libMesh equation systems objects have been initialized
-     * (i.e., whether or not initializeFEEquationSystems has been called).
-     */
-    bool d_fe_equation_systems_initialized = false;
 
     /*!
      * Whether or not all finite element data (including that initialized by
--- a/include/ibamr/IBFEMethod.h
+++ b/include/ibamr/IBFEMethod.h
@@ -345,12 +345,10 @@ class IBFEDirectForcingKinematics;
  * up if restart information is not available.
  *
  * This code is based on an IBFE example and assumes that the IBAMR objects
- * objects are already set up in the usual way.
+ * objects are already set up in the usual way and IBFEMethod has been
+ * constructed (but no further functions have been called).
  *
  * @code
- * // Actually create the EquationSystems objects.
- * ib_method_ops->initializeFEEquationSystems();
- *
  * // This code assumes we only have one part, so there is only one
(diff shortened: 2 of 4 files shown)


drwells:
@jabrown893 as far as I know you are the only person who uses this feature. I won't break your code but I would like to understand if there is a more convenient way to set multiple kernel functions. Would writing an input file that looks like
```
IBFEMethod {
   // blablabla

   FEDataManager {
       // set some stuff common to all FEDataManagers
   }

   FEDataManager_0 {
      // override for the 0th FEDataManager
      ib_kernel = "PIECEWISE_LINEAR"
   }

   FEDataManager_1 {
      // etc.
      ib_kernel = "IB_4"
   }
}
```
be more or less convenient than what you are doing now?

jabrown893:
@drwells, I think @xxjaehoxx might use this as well?

drwells:
I only have the TAVR code - if @xxjaehoxx's stuff is doing similar things then yes.

jabrown893:
The only potential issue I can think of is that my codes are often set up to load meshes conditionally. So, something like this:
`// Configure the IBFE solver.
        pout << "\nConfiguring the solver...\n";
        if(ROOT)
        {
            string leaflet_kernel_fcn = input_db->getStringWithDefault("LEAFLET_KERNEL_FCN", "IB_3");
            FEDataManager::InterpSpec leaflet_interp_spec = ibfe_method_ops->getDefaultInterpSpec();
            leaflet_interp_spec.kernel_fcn = leaflet_kernel_fcn;
            ibfe_method_ops->setInterpSpec(leaflet_interp_spec, LEAFLET_PART);
            FEDataManager::SpreadSpec leaflet_spread_spec = ibfe_method_ops->getDefaultSpreadSpec();
            leaflet_spread_spec.kernel_fcn = leaflet_kernel_fcn;
            ibfe_method_ops->setSpreadSpec(leaflet_spread_spec, LEAFLET_PART);

            string housing_kernel_fcn = input_db->getStringWithDefault("HOUSING_KERNEL_FCN", "PIECEWISE_LINEAR");
            FEDataManager::InterpSpec housing_interp_spec = ibfe_method_ops->getDefaultInterpSpec();
            housing_interp_spec.kernel_fcn = housing_kernel_fcn;
            ibfe_method_ops->setInterpSpec(housing_interp_spec, HOUSING_PART);
            FEDataManager::SpreadSpec housing_spread_spec = ibfe_method_ops->getDefaultSpreadSpec();
            housing_spread_spec.kernel_fcn = housing_kernel_fcn;
            ibfe_method_ops->setSpreadSpec(housing_spread_spec, HOUSING_PART);
        }`
Would that cause the code to break -- if there is an `FEDataManager_1` in the input file but none is loaded in main?

boyceg:
@drwells I think it would be useful to be able to set this stuff up programatically.

At least with the ways that we are doing things right now, using something like `FEDataManager_0`, `FEDataManager_1` in the input files seems to require that the order that the parts are set up in the code corresponds to the order that things are listed in the input file, which seems brittle.

boyceg:
@drwells do you want to pursue this one?

drwells:
This PR is nice in that it simplifies code but it makes assumptions that may not be valid about user workflows, so lets not go forward with it.

PR #1136, opened 2020-10-01
Fix scratch and multilevel tests

drwells:
I am working on a new IBFE example and I noticed that the scratch multilevel tests aren't right - they don't actually use multiple levels. It turns out the other tests I added don't either. Fortunately you can see that they don't make a big difference in the values we compute (which implies that the multilevel stuff works).

Another fix: we were not regridding all scratch hierarchy levels which lead to some poor load balancing. This is also fixed.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1379,7 +1379,7 @@ IBFEMethod::addWorkloadEstimate(Pointer<PatchHierarchy<NDIM> > hierarchy, const 
         const int current_rank = IBTK_MPI::getRank();
 
         std::vector<double> workload_per_processor(n_processes);
-        HierarchyCellDataOpsReal<NDIM, double> hier_cc_data_ops(hierarchy);
+        HierarchyCellDataOpsReal<NDIM, double> hier_cc_data_ops(hierarchy, 0, hierarchy->getFinestLevelNumber());
         workload_per_processor[current_rank] = hier_cc_data_ops.L1Norm(workload_data_idx, IBTK::invalid_index, true);
 
         const auto right_padding = std::size_t(std::log10(n_processes)) + 1;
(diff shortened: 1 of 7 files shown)


src/IB/IBFEMethod.cpp
```diff
@@ -1506,12 +1506,11 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
...
             // any sort of tagging based on the current time I think we can
             // just put in a bogus (nonnegative) value.
-            int finest_level = d_scratch_hierarchy->getFinestLevelNumber();
-            if (finest_level == 0)
+            if (getCoarsestPatchLevelNumber() == 0)
                 d_scratch_gridding_algorithm->makeCoarsestLevel(d_scratch_hierarchy, 0.0 /*d_current_time*/);
             else
                 d_scratch_gridding_algorithm->regridAllFinerLevels(
-                    d_scratch_hierarchy, finest_level - 1, 0.0 /*d_current_time*/, tag_buffer);
+                    d_scratch_hierarchy, getCoarsestPatchLevelNumber() - 1, 0.0 /*d_current_time*/, tag_buffer);
```
boyceg, line 1513:
```suggestion
                    d_scratch_hierarchy, std::max(getCoarsestPatchLevelNumber() - 1, 0), 0.0 /*d_current_time*/, tag_buffer);
```
?

boyceg (approved):
Looks good.

Only comment is that I am not sure if `regridAllFinerLevels()` needs to have a valid level number. It seems like the level number could be invalid if `getCoarsestPatchLevelNumber()` is `0`.

src/IB/IBFEMethod.cpp
drwells: See like 1509 :)

boyceg: Oh right.

boyceg: Should there be a subsequent call to `regridAllFinerLevels()` if `getCoarsestPatchLevelNumber()` returns `0`?

drwells: I think you're right. Let me take a look.

drwells:
I fixed the problem with the coarsest level - the scratch hierarchy now works with all supported possibilities.

drwells:
Lets merge this before we go on to #1137.

boyceg (approved):

PR #1137, opened 2020-10-01
synch coarse grid values with overlying fine grid data in IBFEMethod::interpolateVelocity()

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -658,12 +658,21 @@ IBFEMethod::postprocessIntegrateData(double current_time, double new_time, int n
 
 void
 IBFEMethod::interpolateVelocity(const int u_data_idx,
-                                const std::vector<Pointer<CoarsenSchedule<NDIM> > >& /*u_synch_scheds*/,
+                                const std::vector<Pointer<CoarsenSchedule<NDIM> > >& u_synch_scheds,
                                 const std::vector<Pointer<RefineSchedule<NDIM> > >& u_ghost_fill_scheds,
                                 const double data_time)
 {
     IBAMR_TIMER_START(t_interpolate_velocity);
     const std::string data_time_str = get_data_time_str(data_time, d_current_time, d_new_time);
+
+    // Ensure coarse grid data are consistent with any overlying fine grid data.
+    for (unsigned int ln = getFinestPatchLevelNumber(); ln > getCoarsestPatchLevelNumber(); --ln)
+    {
+        if (ln < u_synch_scheds.size() && u_synch_scheds[ln])
+        {
+            u_synch_scheds[ln]->coarsenData();
+        }
+    }
 
     if (d_use_scratch_hierarchy)
     {


drwells:
I looked at this earlier and I came to the conclusion that "if the scratch hierarchy works without doing this then its probably not necessary" but we should definitely test that assumption.

boyceg:
I think that it is good practice to resynch just in case a different coarsen operator was used. We don't require coarsen operators to be consistent in different places, so it is possible (albeit unlikely) that the coarse grid data were set using some other coarsen operator.

As a specific example --- we are about to try to switch to using RT0-based coarsening for multilevel IBFE, but that is not what we want to use for the incompressible flow solvers.

boyceg:
I just confirmed that the current IBFE tests. (I know that you are updating the multilevel tests.)

boyceg:
@drwells would you prefer that we handle restricting internally in `IBFEMethod`, as is now done for prolongation?

boyceg:
I am happy to do it either way, but I am about to start mucking around with the way that we do coarsening for multilevel IBFE.

drwells:
That's a good question. I would prefer that the Navier-Stokes solver correctly set the velocity on all levels whenever it sets a value in the hierarchy so that its not up to a class like IBFEMethod which should, in principle, just be reading those values. If we want to do something else inside IBFEMethod we should use a scratch index.

boyceg:
We already do use a scratch index --- the velocity is interpolated from a copy of the velocity from the fluid solver, since in general the fluid solver does not have the right ghost cell widths.

boyceg:
(Similarly, forces are spread to a scratch index, for similar reasons.)

boyceg:
The way this was supposed to work was that `IBHierarchyIntegrator` (or a subclass) handles setting up communication schedules, so that the IB implementations don't need to mess with them --- but instead just use them "in the right way".

`IBFEMethod` effectively maintains its own patch hierarchy, however, and so it cannot use the provided refine schedule for spreading.

I guess I am wondering if it would be clearer for it also not to use the provided coarsen schedule for interpolating. Using the provided coarsen schedule splits up the specification of the interlevel transfer operators between the `IBHierarchyIntegrator` and the `IBStrategy`, which does not seem ideal.

drwells:
> IBFEMethod effectively maintains its own patch hierarchy, however, and so it cannot use the provided refine schedule for spreading.

Even if it did not due to the way we now accumulate forces (which is necessary for efficiency reasons) we really do need to put all the IBFE-specific forces into a separate privately managed index.

> Using the provided coarsen schedule splits up the specification of the interlevel transfer operators between the IBHierarchyIntegrator and the IBStrategy, which does not seem ideal.

It seems to me that `IBHierarchyIntegrator` should be in charge here. In particular, if we have multiple IBStrategy objects, they are going to each repeat the same communication, or (like in this case) overwrite values in a difficult to understand way.

drwells:
I rebased this against master and all the tests still pass. I'm not convinced that this is necessary but it doesn't hurt - it also makes sense to do this if we are also going to execute the other schedules.

drwells (approved):
Let's add this for the sake of consistency. If it shows up a lot in profiling (which is unlikely) we can easily revert it later.

PR #1139, opened 2020-10-07
Added a functionality to handle variable C coef in CCPoissonPointRela…

Raamkrishnan:
…xationFACOperator class and the test

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?
I have added functionality to handle the variable C coefficient as well in ```CCPoissonPointRelaxationFACOperator``` class. I have mainly added the Fortran routines which take a cell-centered C coefficient for Gauss-seidel sweep in ```smooth_error``` function. I have also added a test where I solve the variable C coefficient Helmholtz equation with FAC preconditioner.

### DIFF
--- a/ibtk/src/solvers/impls/CCPoissonPointRelaxationFACOperator.cpp
+++ b/ibtk/src/solvers/impls/CCPoissonPointRelaxationFACOperator.cpp
@@ -66,96 +66,184 @@
 
 // FORTRAN ROUTINES
 #if (NDIM == 2)
-#define GS_SMOOTH_FC IBTK_FC_FUNC(gssmooth2d, GSSMOOTH2D)
-#define RB_GS_SMOOTH_FC IBTK_FC_FUNC(rbgssmooth2d, RBGSSMOOTH2D)
-#define VC_CELL_GS_SMOOTH_FC IBTK_FC_FUNC(vccellgssmooth2d, VCCELLGSSMOOTH2D)
-#define VC_CELL_RB_GS_SMOOTH_FC IBTK_FC_FUNC(vccellrbgssmooth2d, VCCELLRBGSSMOOTH2D)
+#define SMOOTH_GS_CONST_DC_FC IBTK_FC_FUNC(smooth_gs_const_dc_2d, SMOOTH_GS_CONST_DC_2D)
+#define SMOOTH_GS_RB_CONST_DC_FC IBTK_FC_FUNC(smooth_gs_rb_const_dc_2d, SMOOTH_GS_RB_CONST_DC_2D)
+#define SMOOTH_GS_VAR_D_CONST_C_FC IBTK_FC_FUNC(smooth_gs_var_d_const_c_2d, SMOOTH_GS_VAR_D_CONST_C_2D)
+#define SMOOTH_GS_RB_VAR_D_CONST_C_FC IBTK_FC_FUNC(smooth_gs_rb_var_d_const_c_2d, SMOOTH_GS_RB_VAR_D_CONST_C_2D)
+#define SMOOTH_GS_CONST_D_VAR_C_FC IBTK_FC_FUNC(smooth_gs_const_d_var_c_2d, SMOOTH_GS_CONST_D_VAR_C_2D)
+#define SMOOTH_GS_RB_CONST_D_VAR_C_FC IBTK_FC_FUNC(smooth_gs_rb_const_d_var_c_2d, SMOOTH_GS_rb_CONST_D_VAR_C_2D)
+#define SMOOTH_GS_VAR_DC_FC IBTK_FC_FUNC(smooth_gs_var_dc_2d, SMOOTH_GS_VAR_DC_2D)
+#define SMOOTH_GS_RB_VAR_DC_FC IBTK_FC_FUNC(smooth_gs_rb_var_dc_2d, SMOOTH_GS_RB_VAR_DC_2D)
 #endif
 #if (NDIM == 3)
-#define GS_SMOOTH_FC IBTK_FC_FUNC(gssmooth3d, GSSMOOTH3D)
-#define RB_GS_SMOOTH_FC IBTK_FC_FUNC(rbgssmooth3d, RBGSSMOOTH3D)
-#define VC_CELL_GS_SMOOTH_FC IBTK_FC_FUNC(vccellgssmooth3d, VCCELLGSSMOOTH3D)
-#define VC_CELL_RB_GS_SMOOTH_FC IBTK_FC_FUNC(vccellrbgssmooth3d, VCCELLRBGSSMOOTH3D)
+#define SMOOTH_GS_CONST_DC_FC IBTK_FC_FUNC(smooth_gs_const_dc_3d, SMOOTH_GS_CONST_DC_3D)
+#define SMOOTH_GS_RB_CONST_DC_FC IBTK_FC_FUNC(smooth_gs_rb_const_dc_3d, SMOOTH_GS_RB_CONST_DC_3D)
+#define SMOOTH_GS_VAR_D_CONST_C_FC IBTK_FC_FUNC(smooth_gs_var_d_const_c_3d, SMOOTH_GS_VAR_D_CONST_C_3D)
+#define SMOOTH_GS_RB_VAR_D_CONST_C_FC IBTK_FC_FUNC(smooth_gs_rb_var_d_const_c_3d, SMOOTH_GS_RB_VAR_D_CONST_C_3D)
+#define SMOOTH_GS_CONST_D_VAR_C_FC IBTK_FC_FUNC(smooth_gs_const_d_var_c_3d, SMOOTH_GS_CONST_D_VAR_C_3D)
+#define SMOOTH_GS_RB_CONST_D_VAR_C_FC IBTK_FC_FUNC(smooth_gs_rb_const_d_var_c_3d, SMOOTH_GS_rb_CONST_D_VAR_C_3D)
+#define SMOOTH_GS_VAR_DC_FC IBTK_FC_FUNC(smooth_gs_var_dc_3d, SMOOTH_GS_VAR_DC_3D)
+#define SMOOTH_GS_RB_VAR_DC_FC IBTK_FC_FUNC(smooth_gs_rb_var_dc_3d, SMOOTH_GS_RB_VAR_DC_3D)
 #endif
 
 // Function interfaces
 extern "C"
 {
-    void GS_SMOOTH_FC(double* U,
-                      const int& U_gcw,
-                      const double& alpha,
-                      const double& beta,
-                      const double* F,
-                      const int& F_gcw,
-                      const int& ilower0,
-                      const int& iupper0,
-                      const int& ilower1,
-                      const int& iupper1,
+    void SMOOTH_GS_CONST_DC_FC(double* U,
+                               const int& U_gcw,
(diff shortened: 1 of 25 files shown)


tests/IBTK/var_c_helmholtz_2d.cpp
```diff
@@ -0,0 +1,261 @@
...
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+// Config files
+#include <IBTK_config.h>
+#include <SAMRAI_config.h>
```
drwells, line 16:
```suggestion
#include <ibtk/config.h>
```
The headers in the PR work fine with autotools but are deprecated. `ibtk/config.h` works with both autotools and the (in-progress) CMake build system.

drwells (changes requested):
Looks good to me. I ramped up `N` in the tests to verify that we really do get second-order convergence in the infinity norm.

tests/IBTK/var_c_helmholtz_2d.cpp
```diff
@@ -0,0 +1,261 @@
...
+                    const SAMRAI::hier::Index<NDIM>& patch_lower_index = patch_box.lower();
+                    const double* const patch_dx = patch_geom->getDx();
+                    for (unsigned int d = 0; d < NDIM; ++d)
+                    {
+                        cell_coord[d] = patch_X_lower[d] + patch_dx[d] * (ci(d) - patch_lower_index(d) + 0.5);
+                    }
+                    (*C_data)(ci) = std::pow(cell_coord[0], 2.0) + std::pow(cell_coord[1], 2.0);
+                }
+            }
+        }
```
amneetb, line 182: @Raamkrishnan : Can you use muParserCartGridFunction here instead of coding `C coef`?

tests/IBTK/var_c_helmholtz_2d.input
```diff
@@ -0,0 +1,98 @@
...
+
+StandardTagAndInitialize {
+   tagging_method = "REFINE_BOXES"
+   RefineBoxes {
+//    level_0 = [( N/4 , 0 ),( 3*N/4 - 1 , N - 1 )]
+//    level_0 = [( 0 , N/4 ),( N - 1 , 3*N/4 - 1 )]
+//    level_0 = [( N/4 , N/4 ),( 3*N/4 - 1 , 3*N/4 - 1 )]
+//    level_0 = [( N/4 , N/4 ),( 3*N/4 - 1 , N/2 - 1 )] , [( N/4 , N/2 ),( N/2 - 1 , 3*N/4 - 1 )]
+//    level_0 = [( N/4 , N/4 ),( N/2 - 1 , 3*N/4 - 1 )] , [( N/2 , N/4 ),( 3*N/4 - 1 , N/2 - 1 )]
+      level_0 = [( N/4 , N/4 ),( N/2 - 1 , N/2 - 1 )] , [( N/2 , N/4 ),( 3*N/4 - 1 , N/2 - 1 )] , [( N/4 , N/2 ),( N/2 - 1 , 3*N/4 - 1 )]
```
amneetb, line 84: Remove these commented lines also.

ibtk/src/solvers/impls/SCPoissonPointRelaxationFACOperator.cpp
```diff
@@ -74,36 +74,37 @@
 
 // FORTRAN ROUTINES
 #if (NDIM == 2)
-#define GS_SMOOTH_FC IBTK_FC_FUNC(gssmooth2d, GSSMOOTH2D)
+#define CONST_DC_GS_SMOOTH_FC IBTK_FC_FUNC(const_dc_gssmooth2d, CONST_DC_GSSMOOTH2D)
```
boyceg, line 77: Let's stick with the original names for the original functions. Although unlikely, it is possible that they are used in other places.

ibtk/src/solvers/impls/CCPoissonPointRelaxationFACOperator.cpp
```diff
@@ -65,96 +65,184 @@
 
 // FORTRAN ROUTINES
 #if (NDIM == 2)
-#define GS_SMOOTH_FC IBTK_FC_FUNC(gssmooth2d, GSSMOOTH2D)
-#define RB_GS_SMOOTH_FC IBTK_FC_FUNC(rbgssmooth2d, RBGSSMOOTH2D)
-#define VC_CELL_GS_SMOOTH_FC IBTK_FC_FUNC(vccellgssmooth2d, VCCELLGSSMOOTH2D)
-#define VC_CELL_RB_GS_SMOOTH_FC IBTK_FC_FUNC(vccellrbgssmooth2d, VCCELLRBGSSMOOTH2D)
+#define CONST_DC_GS_SMOOTH_FC IBTK_FC_FUNC(const_dc_gssmooth2d, CONST_DC_GSSMOOTH2D)
```
boyceg, line 68: Let's stick with the original names for the original functions. Although unlikely, it is possible that they are used in other places.

ibtk/src/solvers/impls/SCPoissonPointRelaxationFACOperator.cpp
amneetb: @boyceg: This will cause confusion. The renaming was done to make function context clearer. There was various combinations of var C, const C, var D, const D.

ibtk/src/solvers/impls/CCPoissonPointRelaxationFACOperator.cpp
```diff
@@ -65,96 +65,184 @@
...
+#define CONST_DC_RB_GS_SMOOTH_FC IBTK_FC_FUNC(const_dc_rbgssmooth2d, CONST_DC_RBGSSMOOTH2D)
+#define VAR_D_CONST_C_GS_SMOOTH_FC IBTK_FC_FUNC(var_d_const_c_gssmooth2d, VAR_D_CONST_C_GSSMOOTH2D)
```
boyceg, line 70: Suggest just using variable C and variable D to the VC implementation, instead of adding new specialized Fortran kernels for the 3 cases of variable C / constant D, constant C / variable D, and variable C / variable D. I am not sure it is worth the extra code to handle all of the cases.

If one of the coefficients is constant,

boyceg: (If one of the coefficients is constant, you can just copy the constant value into a properly sized array.)

ibtk/src/solvers/impls/SCPoissonPointRelaxationFACOperator.cpp
drwells: I also like the renames and I don't see any other usages of `gssmooth2d` in the library. Since the function prototypes are in a source file and not a header these functions are de facto internal anyway.

ibtk/src/solvers/impls/CCPoissonPointRelaxationFACOperator.cpp
boyceg: (OTOH, if you know that it makes a big performance difference, that is a good reason to have different specializations.)

ibtk/src/solvers/impls/SCPoissonPointRelaxationFACOperator.cpp
boyceg: OK, well, if they are getting renamed, let's put "GS smooth" in the front instead of the back.

ibtk/src/solvers/impls/CCPoissonPointRelaxationFACOperator.cpp
amneetb: We should always strive to reduce memory usage in IBAMR, wherever possible. It will unnecessary create and then destroy temp (stack or heap) memory to hold constant values of C/D.

ibtk/src/solvers/impls/SCPoissonPointRelaxationFACOperator.cpp
boyceg: Also, suggest to rename "alpha" and "beta" to "D" and "C" (or is it "C" and "D") in the list of arguments both in the C prototype and in the Fortran routines.

amneetb: I'm fine either way.

ibtk/src/solvers/impls/CCPoissonPointRelaxationFACOperator.cpp
boyceg: Since @amneetb and @drwells also like the new names, I will mark this as resolved.

ibtk/src/solvers/impls/SCPoissonPointRelaxationFACOperator.cpp
boyceg: My usual rationale in naming is to go from "more general" to "more specific" for these long function names. I guess one could argue that it would better follow that pattern to write `smooth_gs...`...

boyceg:
Looks good -- make sure to add a changelog entry! 😀

ibtk/src/solvers/impls/SCPoissonPointRelaxationFACOperator.cpp
amneetb: @Raamkrishnan can change the names (`smooth_gs` as prefix) along with my other suggestions. He can also add the changelog entry.

ibtk/src/solvers/impls/CCPoissonPointRelaxationFACOperator.cpp
drwells: We already have a ton of repeated code, unfortunately, in the FORTRAN routines (e.g., 90% of every interpolation and spread function). It would be great if we could get rid of it but I support the simple solution here since its consistent with what we have and avoids having to allocate lots of extra memory for no reason.

boyceg:
👍

boyceg:
Are there tests for all of the smoother options in both 2D and 3D? It would be helpful to test all of the cases explicitly. You should be able to setup tests with the same driver and just have different input/output files.

Raamkrishnan:
I have incorporated the requested changes. Please take a look.

Raamkrishnan:
Yes. I can add the tests for other smoother options as well. I will work on it tomorrow.

drwells:
@Raamkrishnan Let us know if you are stuck on something or need assistance finishing the PR :)

drwells:
Aside from the merge conflict I think the only thing left to do is make sure we have 3D test coverage - since we already have a 2D test driver this should be easy to add.

Raamkrishnan:
> Are there tests for all of the smoother options in both 2D and 3D? It would be helpful to test all of the cases explicitly. You should be able to setup tests with the same driver and just have different input/output files.

I have added the tests for all smoother options in both 2D and 3D. Please take a look and let me know your suggestions.

Raamkrishnan:
@drwells I have pushed the 2D and 3D tests. I have checked the convergence rate for all the tests and I got second-order accuracy in space. I have also resolved the conflict in MakeFile.am. Please take a look.

drwells:
Looks good - I modified the commit to include the necessary changes (two lines) to get the test working with CMake.

drwells (approved):

PR #1140, opened 2020-10-08
Fix multilevel IBFE when the structure is not on the finest level.

### LINKED ISSUE #1112: Multilevel IBFE support

drwells:
Closes #1112.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -179,6 +179,20 @@ collect_unique_elems(std::vector<Elem*>& elems, const ContainerOfContainers& ele
     return;
 } // collect_unique_elems
 
+std::set<libMesh::subdomain_id_type>
+collect_subdomain_ids(const libMesh::MeshBase& mesh)
+{
+    std::set<libMesh::subdomain_id_type> subdomain_ids;
+    // Get all subdomain ids present, not just local elements.
+    const auto el_begin = mesh.elements_begin();
+    const auto el_end = mesh.elements_end();
+    for (auto el_it = el_begin; el_it != el_end; ++el_it)
+    {
+        subdomain_ids.insert((*el_it)->subdomain_id());
+    }
+    return subdomain_ids;
+}
+
 /**
  * A difficulty with FEDataManager is that it needs to work with both volumetric
  * and surface meshes, even though for performance reasons we template FEMapping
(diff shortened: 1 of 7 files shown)


ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -3025,6 +3050,8 @@ FEDataManager::collectActivePatchElements(std::vector<std::vector<Elem*> >& acti
     const MeshBase& mesh = d_fe_data->d_es->get_mesh();
     System& X_system = d_fe_data->d_es->get_system(COORDINATES_SYSTEM_NAME);
 
+    // We may
```
boyceg, line 3053: ??

drwells: I was going to write a comment here but it ended up not being necessary. Let me delete the stub.

boyceg (approved):
Looks good.

Should we check somewhere that the finest patch hierarchy level is >= the max level number?

drwells:
I don't think so. In the case of the test I added we may not necessarily even have patches on the finest patch level until later in the time evolution since the structure isn't on it.

boyceg:
You can use REFINE_BOXES to make sure there are finer levels.

drwells:
Good idea. I'll add another test that does exactly that.

PR #1145, opened 2020-10-16
Fix the memory leak with the scratch hierarchy.

drwells:
It turns out that unless we request a unique `HierachyDataOps` then we store a copy of that object indefinitely, and that object stores a pointer to the scratch hierarchy - i.e., we never deallocate the scratch hierarchies and we get a new one every time we regrid.

I figured this out by setting a breakpoint in `SAMRAIDataCache::setPatchHierarchy()`. At the point we are supposed to destroy the hierarchy (because its reference count is zero) we in fact have a reference count of 3. To verify that this works I rigged up a copy of an IBFE example to regrid at every time step - without this patch it more than triples in memory and with this patch the memory usage is nearly constant (in both cases I had 1000 timesteps with 1000 regrids).

Since these are not really memory leaks (we destroy the objects at exit) I don't know of a good way to test that we really fixed the problem. Hence I added an item to the release checklist to remind us to look for such problems manually.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/release-tasks.md
+++ b/doc/release-tasks.md
@@ -41,6 +41,10 @@ GitHub.
 - [ ] Check that we don't call functions like
   `ReplicatedMesh::active_local_elements_end()` inside `for`-loop declarations
   since the end iterator is expensive to compute.
+- [ ] Check that we always get unique `HierarchyDataOps` by passing `true` as
+  the third argument to `getOperationsDouble` et al. This prevents memory leaks
+  where a statically stored `HierarchyDataOps` object keeps a patch hierarchy
+  alive that is otherwise inaccessible.
 - [ ] Run clang-format on the entire code base.
 
 ## testing
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -962,7 +962,7 @@ IBFEMethod::spreadForce(const int f_data_idx,
     // zero data.
     Pointer<hier::Variable<NDIM> > f_var;
     VariableDatabase<NDIM>::getDatabase()->mapIndexToVariable(f_data_idx, f_var);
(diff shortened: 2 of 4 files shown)


boyceg:
Have you looked to see if these kinds of failures to release memory show up in other places (e.g. in the fluid solvers)?

doc/release-tasks.md
```diff
@@ -41,6 +41,10 @@ GitHub.
 - [ ] Check that we don't call functions like
   `ReplicatedMesh::active_local_elements_end()` inside `for`-loop declarations
   since the end iterator is expensive to compute.
+- [ ] Check that we always get unique `HierarchyDataOps` by passing `true` as
```
boyceg, line 44: This was originally set up as a way to reduce memory usage...

If we always want unique `HierarchyDataOps` objects should we get rid of the option to share them?

boyceg: Oh, whoops, this is an internal SAMRAI thing, so we just need to make sure we cache these ourselves as needed to reduce memory usage.

boyceg (approved):
👍

doc/release-tasks.md
drwells: I don't see the point of caching these objects at all - the size of each one can be measured in bytes since all they do is call the correct functions (i.e., cell-centered, side-centered, or whatever) for the specified variable.

Either way, since its a SAMRAI function, we should just try to avoid it.

drwells:
> Have you looked to see if these kinds of failures to release memory show up in other places (e.g. in the fluid solvers)?

We get pretty consistent memory usage now so I think the fluid solver is fine. This particular problem is due to us creating and destroying lots of `PatchHierarchy` objects so its unlikely to show up elsewhere in the library (I don't know of any other place where we do this).

PR #1147, opened 2020-10-27
Implement an alternative regrid request strategy.

drwells:
We can measure the displacement of the structure directly instead of relying on
estimates from the fluid solver.

As requested by @boyceg.

This is only implemented fully for `IBFEMethod` - using it in other classes will result in the program being aborted. If people know how to implement this for other IBStrategy objects they should write patches for them.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/ibtk_utilities.h
+++ b/ibtk/include/ibtk/ibtk_utilities.h
@@ -20,6 +20,7 @@
 
 #include <ibtk/config.h>
 
+#include "PatchHierarchy.h"
 #include "tbox/MathUtilities.h"
 #include "tbox/PIO.h"
 #include "tbox/Utilities.h"
@@ -136,6 +137,12 @@ get_data_time_str(const double data_time, const double current_time, const doubl
         return "unknown";
     }
 }
+
+/*!
+ * Get the smallest cell width on the specified level. This operation is
+ * collective.
+ */
+double get_min_patch_dx(const SAMRAI::hier::PatchLevel<NDIM>& patch_level);
 
 template <class T, unsigned N>
 inline std::array<T, N>
--- a/include/ibamr/IBHierarchyIntegrator.h
+++ b/include/ibamr/IBHierarchyIntegrator.h
@@ -86,6 +86,47 @@ namespace IBAMR
  * integrator for various versions of the immersed boundary method on an AMR
  * grid hierarchy, along with basic data management for variables defined on
  * that hierarchy.
+ *
+ * <h2>Options Controlling Regridding</h2>
+ *
+ * Most IBAMR applications involve structure meshes defined on the finest level
+ * of the patch hierarchy. These structures move: in particular, they can
+ * potentially move off the finest grid level, causing the interaction routines
+ * to no longer work.
+ *
+ * This class offers two different strategies for calculating how much the
+ * structure has moved (which it then uses to specify that a regrid is
+ * required):
+ *
+ * <ol>
+ *   <li><em>Estimation based on the fluid</em>: The structure is assumed to
+ *     move at the maximum velocity of the fluid at each time step. This class
+ *     will regrid once a single fluid point has moved at least
+ *     <code>regrid_fluid_cfl_interval</code> cell widths since the last
+ *     regrid.</li>
+ *
+ *   <li><em>Estimation based on the structure</em>: The structure's
+ *     displacement is calculated directly from its position vector and that
+ *     value is used to determine the number of cell widths the structure has
+ *     moved. In this case we will regrid once a point on the structure has
+ *     moved at least <code>regrid_structure_cfl_interval</code> cell widths
+ *     since the last regrid.</li>
+ * </ol>
+ *
+ * Both <code>regrid_fluid_cfl_interval</code> and
+ * <code>regrid_structure_cfl_interval</code> can be specified in the input
+ * database. For backwards compatibility the value
+ * <code>regrid_cfl_interval</code> is equivalent to
+ * <code>regrid_fluid_cfl_interval</code>. <em>At the present time
+ * <code>regrid_structure_cfl_interval</code> is not implemented for all
+ * IBStrategy classes.</em>
+ *
+ * Alternatively, one can request that the solver (regardless of any computed
+ * displacement or velocity) regrid every time a fixed number of timesteps have
+ * occurred by specifying <code>regrid_interval</code> in the input database.
+ * <em>If either <code>regrid_structure_cfl_interval</code> or
+ * <code>regrid_fluid_cfl_interval</code> are provided in the input database
+ * then <code>regrid_interval</code> is ignored.</em>
  */
 class IBHierarchyIntegrator : public IBTK::HierarchyIntegrator
 {
@@ -299,15 +340,37 @@ protected:
      */
     SAMRAI::tbox::Pointer<INSHierarchyIntegrator> d_ins_hier_integrator;
 
+    /**
+     * The regrid CFL interval indicates the number of meshwidths a particle may
+     * move in any coordinate direction between invocations of the regridding
+     * process.
+     *
+     * @note Currently, when the CFL-based regrid interval is specified, it is
+     * always used instead of the fixed-step regrid interval.
+     */
+    double d_regrid_fluid_cfl_interval = -1.0;
+
     /*
      * The regrid CFL interval indicates the number of meshwidths a particle may
      * move in any coordinate direction between invocations of the regridding
      * process.
      *
-     * NOTE: Currently, when the CFL-based regrid interval is specified, it is
+     * @note Currently, when the CFL-based regrid interval is specified, it is
      * always used instead of the fixed-step regrid interval.
      */
-    double d_regrid_cfl_interval = 0.0, d_regrid_cfl_estimate = 0.0;
+    double d_regrid_structure_cfl_interval = -1.0;
+
+    /**
+     * Estimation on the maximum fraction of fluid cells the structure has
+     * moved based on the maximum fluid velocity.
+     */
+    double d_regrid_fluid_cfl_estimate = 0.0;
+
+    /**
+     * Estimation on the maximum fraction of fluid cells the structure has
+     * moved based on the infinity norm of the structure's displacement.
+     */
+    double d_regrid_structure_cfl_estimate = 0.0;
 
     /*
      * IB method implementation object.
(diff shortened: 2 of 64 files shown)


include/ibamr/IBHierarchyIntegrator.h
```diff
@@ -86,6 +86,39 @@ namespace IBAMR
...
+ * <h2>Options Controlling Regridding</h2>
+ *
+ * Most IBAMR applications involve structure meshes defined on the finest level
+ * of the patch hierarchy. These structures move: in particular, they can
+ * potentially move off the finest grid level, causing the interaction routines
+ * to no longer work. This class will indicate that it needs a regrid when the
+ * structure has moved <code>regrid_cfl_interval</code> (which is specified in
+ * the input database) mesh widths since the last regrid.
+ *
+ * This class offers two different strategies for calculating how much the
```
boyceg, line 99: Can we use both CFL and displacement-based criteria?

E.g. you could set up tagging criteria to let you use a relatively large CFL-based criterion (CFL ~ 2 or 4 or something) along with a displacement-based criterion.

boyceg: (E.g. you might allow for large CFL-based regridding by choosing a larger tag buffer.)

drwells: I don't understand what you are asking. This PR adds an option to change how we calculate `d_regrid_cfl_estimate` - that's totally independent of tagging (in fact the tag buffer set up by these objects never reads any of this information).

boyceg: I am asking if this is set up to all for both strategies.

drwells: Oh, I see - then no.

I think we could add another option to do it easily.

boyceg: This input entry could be a string array so that multiple options can be set.

boyceg: @drwells just checking on this. Comments above were sent while also watching kids; let me know if it is still unclear what I was trying to say. Basically I think that we should be able to set one, the other, or both criteria --- if both are active, whichever is triggered first "wins".

drwells: Sorry, it's me, not you. I'll wrap this up this afternoon.

drwells:
Rewriting this took awhile but its now done.

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -137,6 +138,13 @@ get_data_time_str(const double data_time, const double current_time, const doubl
     }
 }
 
+/*!
+ * Get the smallest cell width on the specified level. This operation is
+ * collective.
+ */
+double get_min_patch_dx(const SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy,
```
boyceg, line 145: Not crucial, but it would be a bit more general to take in a `PatchLevel` (since one can have a `PatchLevel` that is not in a `PatchHierarchy`).

src/IB/IBFEMethod.cpp
```diff
@@ -535,6 +508,21 @@ IBFEMethod::setupTagBuffer(Array<int>& tag_buffer, Pointer<GriddingAlgorithm<NDI
...
+    double max_displacement = 0.0;
+    for (unsigned int part = 0; part < d_meshes.size(); ++part)
+    {
+        PetscVector<double>& last = d_X_vecs->get("last_regrid", part);
+        PetscVector<double>& diff = d_X_vecs->get("tmp", part);
+        diff = d_X_vecs->get("solution", part);
+        diff -= last;
+        max_displacement = std::max(max_displacement, diff.linfty_norm());
+    }
+    return max_displacement / get_min_patch_dx(d_hierarchy, getFinestPatchLevelNumber());
```
boyceg, line 523: Not a common use case, but for stretched grids, could we benefit from handling this dimension-by-dimension?

src/IB/IBHierarchyIntegrator.cpp
```diff
@@ -192,6 +193,77 @@ IBHierarchyIntegrator::preprocessIntegrateHierarchy(const double current_time,
...
+
+    // Not all IBStrategy objects implement this so make it optional
+    if (d_regrid_structure_cfl_interval != std::numeric_limits<double>::max())
+        d_regrid_structure_cfl_estimate = d_ib_method_ops->getMaxPointDisplacement();
+
+    if (d_enable_logging)
+    {
+        plog << d_object_name << "::postprocessIntegrateHierarchy(): CFL number = " << cfl_max << "\n";
+        plog << d_object_name
+             << "::postprocessIntegrateHierarchy(): estimated (fluid) upper bound on IB "
```
boyceg, line 253: Let's not call this "(fluid)", because in the solid region, the Eulerian velocity is the solid velocity, not the fluid velocity.

Something like _Eulerian_ and _Lagrangian_?

boyceg (changes requested):
Let's not call the Cartesian grid velocity-based estimate a "fluid velocity"-based estimate. (The velocity in a solid location is the solid velocity, not the fluid velocity! 😀 ) Suggest something like "Eulerian" or "background grid" velocity.

src/IB/IBFEMethod.cpp
drwells: In principle yes and this wouldn't be that hard to do (we just need to subdivide the finite element vector into x, y, and z components) but I don't think its worth doing now since we have no such examples (and I have no idea what else might break in the library if we do this).

src/IB/IBHierarchyIntegrator.cpp
drwells: How about "estimaded (from Eulerian)"?

drwells:
The first two commits in this PR are taken from #1158 - we can do everything at once but that PR is logically separate (though this one is dependent on the changes made there).

src/IB/IBHierarchyIntegrator.cpp
boyceg: "Eulerian estimate of upper bound on ..." ?

drwells:
> The velocity in a solid location is the solid velocity, not the fluid velocity!

Yeah, but we are using the fluid velocity, it's just inside the structure :)

drwells:
Either way - its now fixed.

drwells:
@boyceg can you take a look soon? I don't want this to get stale.

include/ibamr/IBHierarchyIntegrator.h
```diff
@@ -299,15 +333,37 @@ class IBHierarchyIntegrator : public IBTK::HierarchyIntegrator
      */
     SAMRAI::tbox::Pointer<INSHierarchyIntegrator> d_ins_hier_integrator;
 
+    /**
+     * The regrid CFL interval indicates the number of meshwidths a particle may
+     * move in any coordinate direction between invocations of the regridding
+     * process.
+     *
+     * @note Currently, when the CFL-based regrid interval is specified, it is
```
boyceg, line 341: @drwells a quick question --- if you request to regrid after each step, does this setting override that setting?

include/ibamr/IBStrategy.h
```diff
@@ -225,6 +225,14 @@ class IBStrategy : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM>, public
     virtual bool getLagrangianStructureIsActivated(int structure_number = 0,
                                                    int level_number = std::numeric_limits<int>::max()) const;
 
+    /*!
+     * Get the ratio of the maximum point displacement of all the structures
+     * owned by the current class to the cell width of the grid level on which
+     * the structure is assigned. This value is usually used to determine if the
```
boyceg, line 231:
```suggestion
     * the structure is assigned. This value is useful for determining if the
```

include/ibamr/IBStrategy.h
```diff
@@ -225,6 +225,14 @@ class IBStrategy : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM>, public
...
+     * Eulerian patch hierarchy needs to be regridded.
```
boyceg, line 232:
```suggestion
     * Eulerian patch hierarchy needs to be regridded and/or if the assignment of Lagrangian data to Cartesian grid patches needs to be updated.
```

boyceg (approved):
Looks good. A few minor suggested changes in wording in the comments. It might be more clear also to be able to allow users to explicitly specify a "fixed regrid interval" regridding strategy, and then the code use which ever triggers first?

boyceg:
> Looks good. A few minor suggested changes in wording in the comments. It might be more clear also to be able to allow users to explicitly specify a "fixed regrid interval" regridding strategy, and then the code use which ever triggers first?

Specifically, I can imagine being confused if I were to set a fixed regrid interval and then don't get it because a CFL-like criterion also is set.

Since we now can specify multiple strategies for triggering regridding, would it make sense to add that one too?

include/ibamr/IBHierarchyIntegrator.h
drwells: Yes.

include/ibamr/IBStrategy.h
drwells: No, this isn't true - `IBFEMethod` deals with that problem (reassociating Lagrangian structures with the Eulerian grid) in an independent way. See `IBFEMethod::preprocessIntegrateData()`.

drwells:
> Specifically, I can imagine being confused if I were to set a fixed regrid interval and then don't get it because a CFL-like criterion also is set.

The current behavior is to ignore the fixed regrid interval when the CFL-like condition is also set. I'll update the documentation to reflect the status quo. Do you want to change the way this works? That might break what people expect to happen.

boyceg:
> > Specifically, I can imagine being confused if I were to set a fixed regrid interval and then don't get it because a CFL-like criterion also is set.
> 
> The current behavior is to ignore the fixed regrid interval when the CFL-like condition is also set. I'll update the documentation to reflect the status quo. Do you want to change the way this works? That might break what people expect to happen.

I guess people may already be confused. 😀

include/ibamr/IBStrategy.h
drwells: In particular: the displacement since the last regrid does not necessarily equal the displacement since the last patch-element association.

boyceg: Could you help future me by adding to this comment by pointing to where this is handled and how this is distinct from the re-association trigger?

boyceg: I am stuck in the past, when regridding was the only way to update the associations between patches and elements.

drwells: That was only three months ago :)

boyceg (approved):
👍

PR #1154, opened 2020-11-02
Cmake support static builds

drwells:
As usual the checklist doesn't apply.

This change also changes the link order to better handle our dependencies that are statically linked and depend on HDF5.


CMakeLists.txt
```diff
@@ -745,6 +748,21 @@ ENDMACRO()
 # everything but the source files)
 FUNCTION(IBAMR_SETUP_TARGET_LIBRARY target_library)
   MESSAGE(STATUS "setting up target ${target_library}")
+
+  # Figure out if we are a 2D or a 3D library:
```
boyceg, line 752: Can one configuration build both 2D and 3D? Or does a particular configuration now have its spatial dimension fixed at configure time?

doc/cmake.md
```diff
@@ -79,6 +83,16 @@ problems with CMake, a good way to get out of trouble is to delete
 `CMakeCache.txt`, which will delete the cache. Deleting the cache is harmless -
 the subsequent initial configuration run will just take a few more seconds.
 
+### Configuring the build
+- If you want to build IBAMR with static libraries then pass the argument
+  `-DBUILD_SHARED_LIBS=ON` to the initial call to `cmake`. IBAMR defaults to
```
boyceg, line 88: ON or OFF?

drwells: oops :blush:

CMakeLists.txt
drwells: The default is to build both but we still compile IBAMR2d, IBAMR3d, IBTK2d, IBTK3d as independent libraries.

drwells: This particular bit of logic refers to if we should link to the 2D or 3D versions of SAMRAI.

boyceg (approved):

PR #1160, opened 2020-11-09
Fischer

drwells:
Every once-in-a-while I take a look at using the algorithm Paul Fischer came up with in _Projection techniques for iterative solution of Ax=b with successive right-hand sides_ and try to get it to work with IBAMR. The implementation in PETSc is too slow for our purposes since it always wants to collect the full set of vectors, orthogonalize w.r.t. the induced inner product, and then finally compute an initial guess. PETSc's implementation is probably fine for harder problems but since we just do L2 projections its overkill for us.

As a break from the news last week I went ahead and implemented a much worse version of this algorithm that completely skips orthogonalization (instead we solve with the SVD so we can tolerate singular matrices), doesn't use the right inner product (since our problem only requires O(10) solver iterations adding more matvecs is prohibitively expensive), and uses a window of RHSs instead of accumulating and then restarting. This technique actually works surprisingly well for L2 projections - it lowers the total number of KSP iterations we use in the heart model for finite element stuff by 50% while requiring only about 0.5% of the total runtime (the required inner products are much cheaper than matvecs) (very roughly, this lowers the amount of time spent doing L2 projections from 20% to 10%).

I don't have time to wrap this up immediately but I wanted to make it public - if a student knows some linear algebra and wants an easy project this is a good candidate since there are a few other things worth testing if we want to summarize our results in a little paper.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
@drwells this might be more important for the stabilization schemes, because the solvers are not as effective and so will eat up a larger fraction of runtime.

This might also help with local FE mesh refinement, since those cases are much slower than I would anticipate for reasons that I do not currently understand.

drwells:
> this might be more important for the stabilization schemes, because the solvers are not as effective and so will eat up a larger fraction of runtime.

As long as the stabilized schemes are well-conditioned (like the mass matrix is) we should be able to get away with the current approach.

> This might also help with local FE mesh refinement, since those cases are much slower than I would anticipate for reasons that I do not currently understand.

Do we have any solid numbers on the effect of FE-AMR on the number of solver iterations?

boyceg:
> > this might be more important for the stabilization schemes, because the solvers are not as effective and so will eat up a larger fraction of runtime.
> 
> As long as the stabilized schemes are well-conditioned (like the mass matrix is) we should be able to get away with the current approach.

I know that it is not as well conditioned as the mass matrix --- but I have not looked very carefully yet.

> > This might also help with local FE mesh refinement, since those cases are much slower than I would anticipate for reasons that I do not currently understand.
> 
> Do we have any solid numbers on the effect of FE-AMR on the number of solver iterations?

No! But it is very surprising to me that in the heart model, local h-refinement for some small structures using P1 elements is about 2X as expensive as global p-refinement (from P1 to P2).

drwells:
As long as we aren't taking derivatives the conditioning is probably fine.

I'm also surprised by Charles' timings - it might be due to the way libMesh implements constraints (the diagonal value ends up being O(1) instead of O(h)).

drwells:
Getting this ready took some work but its now done.

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -395,6 +395,11 @@ class SubdomainToPatchLevelTranslation
  * it produces some artifacts at the coarse-fine interface, but that these
  * generally don't effect the overall solution quality.
  *
+ * <code>num_fischer_vectors</code>: The number of vectors used to compute
+ * initial guesses in FischerGuess. Defaults to 5. Using more vectors lowers the
```
boyceg, line 399: Is this enabled or disabled by default? If disabled, should this be something like "defaults to 5 when enabled"?

ibtk/include/ibtk/FischerGuess.h
```diff
@@ -0,0 +1,96 @@
...
+ *   <li>Since this class does not orthogonalize, it can compute the projection
+ *     matrix incrementally. This means (unlike the PETSc implementation) we can
+ *     always use the full set of available vectors to compute a guess.</li>
+ *   <li>The projection is done with the standard $R^n$ inner product, which may
+ *     not be optimal, but is much faster than doing sparse matrix-vector
+ *     multiplications.</li>
+ * </ol>
+ *
+ * Since the systems we solve typically require low iteration counts, these
+ * tricks to get a less optimal guess at a lower cost are benefitial.
```
boyceg, line 55: beneficial

(for some reason I cannot get to the "selection" button in GitHub)

ibtk/include/ibtk/FEDataManager.h
drwells: It's enabled by default since the optimization helps a lot.

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -2783,7 +2783,7 @@ FEDataManager::FEDataManager(std::string object_name,
                              std::shared_ptr<FEData> fe_data,
                              bool register_for_restart)
     : d_fe_data(fe_data),
-      d_fe_projector(new FEProjector(d_fe_data)),
+      d_fe_projector(new FEProjector(d_fe_data, false, input_db->getIntegerWithDefault("num_fischer_vectors", 5))),
```
boyceg, line 2786: Ah I guess it defaults to enabled. 😀

ibtk/include/ibtk/FEProjector.h
```diff
@@ -45,10 +47,10 @@ class FEProjector
 {
 public:
     /// Constructor.
-    FEProjector(libMesh::EquationSystems* equation_systems, bool enable_logging = true);
+    FEProjector(libMesh::EquationSystems* equation_systems, bool enable_logging = true, int num_fischer_vectors = 5);
```
boyceg, line 50: Suggest keeping `enable_logging` as last argument --- if we are still doing this consistently.

Should `FEProjector` take in a `Database` for settings...?

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -665,12 +666,19 @@ FEProjector::computeL2Projection(PetscVector<double>& U_vec,
         IBTK_CHKERRQ(ierr);
         ierr = KSPSetFromOptions(solver->ksp());
         IBTK_CHKERRQ(ierr);
+
+        auto pair = d_initial_guesses.emplace(system_name, d_num_fischer_vectors);
+        FischerGuess& fischer_guess = (pair.first)->second;
+
+        fischer_guess.guess(U_vec, F_vec);
```
boyceg, line 673: Should the documentation make it clear this is only done for "regular" L2 projections?

I guess the modified projection is not yet in the main branch...

boyceg: Also --- should we bypass if the number of vectors is 0?

boyceg (commented):
This seems like a case where we might want also to allow users to set at the command line, like with other PETSc solver options.

For lumped mass matrix w/ AMR, does the Fisher guess help?

drwells:
I'll address all of this next week :)

> For lumped mass matrix w/ AMR, does the Fisher guess help?

Good question. There's nothing in the method itself that is specific to the conditioning of the matrix, but I suspect that the method works better when the matrix has a narrow spectrum. Doing AMR correctly should double the condition number for every AMR level so I don't think it will have a substantial impact (though we still need to fix libMesh's implementation of constraints to realize this).

ibtk/include/ibtk/FEProjector.h
drwells: I think everything should take an input database - it makes dealing with these solver settings way easier.

ibtk/src/lagrangian/FEProjector.cpp
drwells: We do - see https://github.com/drwells/IBAMR/blob/Fischer/ibtk/src/lagrangian/FischerGuess.cpp#L102

ibtk/include/ibtk/FEProjector.h
drwells: I added an input database. We don't enable logging when this object is set up in `FEDataManager` - I preserved that behavior.

drwells:
I would prefer not to add command line options for this since its not how we configure any other IBAMR object - only PETSc objects are configured with such things and IBAMR always uses the input database.

boyceg (approved):
👍

PR #1161, opened 2020-11-10
Correctly write IBAMRTargets.cmake in the build directory.

drwells:
This correctly exports targets from the build directory so that other libraries can use a built (but not installed) copy of IBAMR. Since this just modifies the build system the normal list doesn't apply.


boyceg (approved):
Looks OK to me, but I don't know CMake (yet?).

drwells:
It's a learning process for all of us.

Essentially, instead of `LIBS` and whatnot we now export targets so users can request to use `IBAMR2d`, `IBAMR3d`, `IBTK2d`, or `IBTK3d` as a dependency and all of our dependencies will be correctly propagated along. This patch sets up target information in the build directory (hence the funny looking `export(export ...)` call) in addition to the one already correctly installed in the installation directory.

drwells:
there's no reason to run the CI since it hasn't yet learned how to deal with CMake so I will go ahead and merge.

PR #1165, opened 2020-11-17
Periodic correction for the constraintsIB method

hongk45:
This patch corrects the calculations of the COM, MOI and related quantities in the presence of periodic boundary for constraintIB

- [x] Was clang-format run? (see https://github.com/hongk45/IBAMR/pull/1)
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/ConstraintIBMethod.cpp
+++ b/src/IB/ConstraintIBMethod.cpp
@@ -13,85 +13,34 @@
 
 /////////////////////////////// INCLUDES /////////////////////////////////////
 
-#include "ibamr/ConstraintIBKinematics.h"
 #include "ibamr/ConstraintIBMethod.h"
-#include "ibamr/IBHierarchyIntegrator.h"
-#include "ibamr/IBStrategy.h"
-#include "ibamr/INSHierarchyIntegrator.h"
 #include "ibamr/INSVCStaggeredHierarchyIntegrator.h"
-#include "ibamr/StokesSpecifications.h"
-#include "ibamr/app_namespaces.h" // IWYU pragma: keep
+#include "ibamr/namespaces.h"
 
 #include "ibtk/CCLaplaceOperator.h"
 #include "ibtk/CCPoissonPointRelaxationFACOperator.h"
 #include "ibtk/FACPreconditioner.h"
-#include "ibtk/FACPreconditionerStrategy.h"
-#include "ibtk/HierarchyMathOps.h"
-#include "ibtk/IBTK_CHKERRQ.h"
-#include "ibtk/IBTK_MPI.h"
 #include "ibtk/IndexUtilities.h"
-#include "ibtk/LData.h"
-#include "ibtk/LDataManager.h"
-#include "ibtk/LIndexSetData.h"
-#include "ibtk/LMesh.h"
-#include "ibtk/LNode.h"
 #include "ibtk/LNodeSetData.h"
-#include "ibtk/LSet.h"
-#include "ibtk/LSetData.h"
-#include "ibtk/LSetDataIterator.h"
-#include "ibtk/LinearOperator.h"
-#include "ibtk/LinearSolver.h"
+#include "ibtk/PETScKrylovLinearSolver.h"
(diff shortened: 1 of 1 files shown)


drwells:
Thanks for taking the time to submit the fix as a PR!

There are a lot of changes in this commit that aren't related to fixing the center of mass calculation. Do you know how they got into the patch?

I think that you fixed an older version of IBAMR - if so we can generate a `.patch` file with exactly those changes which will be easier to work with.

hongk45:
Oh, sorry, I was working on the older version of IBAMR. I just updated the changes. It looks much cleaner now

amneetb:
@hongk45 : Thanks Hong! For some reason I still see a lot of unnecessary `#includes` in the .cpp file. Perhaps you can attach the .cpp file here and @drwells can generate a .patch based upon changes specific to COM and MOI calculations.

drwells:
@amneetb Are you looking at the second commit? The updated version looks fine to me.

amneetb:
I'm looking at https://github.com/hongk45/IBAMR/commit/7ac63ebf1e7274370d1bd12a570fad432e1b8c60, which I believe is the second commit

It is showing some changes in `#include` area and `IBTK_MPI`

drwells:
Yes, thats because the second commit already contains all of the requested updates. Try looking directly at https://github.com/IBAMR/IBAMR/blob/7ac63ebf1e7274370d1bd12a570fad432e1b8c60/src/IB/ConstraintIBMethod.cpp

drwells:
@hongk45 It looks like you forgot to update `ConstraintIBMethod` with the new array declarations. Can you add them?

src/IB/ConstraintIBMethod.cpp
```diff
@@ -1271,17 +1321,18 @@ ConstraintIBMethod::calculateMomentumOfKinematicsVelocity(const int position_han
                 {
                     const int local_idx = node_idx->getLocalPETScIndex();
                     const double* const X = &X_data[local_idx][0];
+                    const Vector &displacement = node_idx->getPeriodicDisplacement(); // HONG ADDED
```
drwells, line 1324:
```suggestion
                    const Vector &displacement = node_idx->getPeriodicDisplacement();
```
We have better ways of giving credit

drwells (changes requested):
Unfortunately some parts of our patch process aren't yet automated - I'll need to run the test suite manually for now.

Do you have clang-format? We require that it is run on patches. I can send you a patch for your patch if you don't have access to it.

The way we give credit for changes is by adding changelog entries: see, e.g., https://github.com/IBAMR/IBAMR/blob/master/doc/news/changes/minor/20200901AaronBarrett for an example of one. Can you write one and add it so you receive credit for fixing this case?

hongk45:
> 
> 
> @hongk45 It looks like you forgot to update `ConstraintIBMethod` with the new array declarations. Can you add them?

This should be added to include\ibamr\ConstraintIBMethod.h

std::vector<std::vector<double> >  d_center_of_mass_unshifted_current, d_center_of_mass_unshifted_new;

Is this enough for you to proceed? Please let me know

hongk45:
> 
> 
> Unfortunately some parts of our patch process aren't yet automated - I'll need to run the test suite manually for now.
> 
> Do you have clang-format? We require that it is run on patches. I can send you a patch for your patch if you don't have access to it.

Unfortunately, I don't. Can you send me one and let me know what I need to do with it. Sorry, I am not quite familiar with patch process (in general).

hongk45:
@drwells I was not able to attach the *.h file to this PR. How should I proceed now? Should I open a new PR for just a little change on the *.f file. Yes, I am working on this patch.

drwells:
@hongk45 It should be possible to add changes to the header to this PR. Try the following on this branch in the top-level IBAMR directory:
```
> git add include/ibamr/ConstraintIBMethod.h
> git commit -m "Added changes to the header."
> git push origin periodic-correction-constraintIB-1
```
where the last command may be different if you gave the repository on GitHub a different name (just use whatever command you use to update the branch). Does this make sense?

PR #1170, opened 2020-12-01
Periodic correction for the constraintsIB method

hongk45:
corrects the calculations of the COM, MOI and related quantities in the presence of periodic boundary for constraintIB

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/ConstraintIBMethod.cpp
+++ b/src/IB/ConstraintIBMethod.cpp
@@ -219,6 +219,8 @@ ConstraintIBMethod::ConstraintIBMethod(std::string object_name,
       d_omega_com_def_new(d_no_structures, std::vector<double>(3, 0.0)),
       d_center_of_mass_current(d_no_structures, std::vector<double>(3, 0.0)),
       d_center_of_mass_new(d_no_structures, std::vector<double>(3, 0.0)),
+      d_center_of_mass_unshifted_current(d_no_structures, std::vector<double>(3, 0.0)),
+      d_center_of_mass_unshifted_new(d_no_structures, std::vector<double>(3, 0.0)),
       d_moment_of_inertia_current(d_no_structures, Eigen::Matrix3d::Zero()),
       d_moment_of_inertia_new(d_no_structures, Eigen::Matrix3d::Zero()),
       d_tagged_pt_lag_idx(d_no_structures, 0),
@@ -896,11 +898,23 @@ ConstraintIBMethod::calculateCOMandMOIOfStructures()
     const int coarsest_ln = 0;
     const int finest_ln = d_hierarchy->getFinestLevelNumber();
 
+    Pointer<CartesianGridGeometry<NDIM> > grid_geom = d_hierarchy->getGridGeometry();
+    const double* const domain_x_lower = grid_geom->getXLower();
+    const double* const domain_x_upper = grid_geom->getXUpper();
+    double domain_length[NDIM];
+    for (int d = 0; d < NDIM; ++d)
+    {
+        domain_length[d] = domain_x_upper[d] - domain_x_lower[d];
+    }
+    const IntVector<NDIM>& periodic_shift = grid_geom->getPeriodicShift();
+
     // Zero out the COM vector.
     for (int struct_no = 0; struct_no < d_no_structures; ++struct_no)
     {
         for (int d = 0; d < 3; ++d)
         {
+            d_center_of_mass_unshifted_current[struct_no][d] = 0.0;
+            d_center_of_mass_unshifted_new[struct_no][d] = 0.0;
             d_center_of_mass_current[struct_no][d] = 0.0;
             d_center_of_mass_new[struct_no][d] = 0.0;
         }
@@ -949,12 +963,13 @@ ConstraintIBMethod::calculateCOMandMOIOfStructures()
                 if (lag_idx_range.first <= lag_idx && lag_idx < lag_idx_range.second)
                 {
                     const int local_idx = node_idx->getLocalPETScIndex();
+                    const Vector& displacement = node_idx->getPeriodicDisplacement();
                     const double* const X_current = &X_data_current[local_idx][0];
                     const double* const X_new = &X_data_new[local_idx][0];
                     for (unsigned int d = 0; d < NDIM; ++d)
                     {
-                        X_com_current[d] += X_current[d];
-                        X_com_new[d] += X_new[d];
+                        X_com_current[d] += X_current[d] + displacement[d];
+                        X_com_new[d] += X_new[d] + displacement[d];
                     }
                     if (lag_idx == d_tagged_pt_lag_idx[location_struct_handle])
                     {
@@ -964,8 +979,8 @@ ConstraintIBMethod::calculateCOMandMOIOfStructures()
             }
             for (int d = 0; d < NDIM; ++d)
             {
-                d_center_of_mass_current[location_struct_handle][d] += X_com_current[d];
-                d_center_of_mass_new[location_struct_handle][d] += X_com_new[d];
+                d_center_of_mass_unshifted_current[location_struct_handle][d] += X_com_current[d];
+                d_center_of_mass_unshifted_new[location_struct_handle][d] += X_com_new[d];
             }
         }
         ptr_x_lag_data_current->restoreArrays();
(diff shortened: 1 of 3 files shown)


hongk45:
@drwells: just decided to create a brand-new PR to make all look cleaner

hongk45:
@drwells : how do I make a new entry to IBAMR/doc/news/changes/? create a new PR?

drwells:
Just add a new file into `doc/news/changes/minor` to this PR (it should look like https://github.com/IBAMR/IBAMR/blob/master/doc/news/changes/minor/20201016DavidWells, which is also a bug fix). Make sure the file is named in the same way (YYYYMMDDName).

src/IB/ConstraintIBMethod.cpp
```diff
@@ -217,6 +217,8 @@ ConstraintIBMethod::ConstraintIBMethod(std::string object_name,
       d_vel_com_def_new(d_no_structures, std::vector<double>(3, 0.0)),
       d_omega_com_def_current(d_no_structures, std::vector<double>(3, 0.0)),
       d_omega_com_def_new(d_no_structures, std::vector<double>(3, 0.0)),
+      d_center_of_mass_unshifted_current(d_no_structures, std::vector<double>(3, 0.0)),
+      d_center_of_mass_unshifted_new(d_no_structures, std::vector<double>(3, 0.0)),
```
drwells, line 221: These need to be after `d_center_of_mass_new` to match the declaration order in the header.

drwells (changes requested):
Looks good but we need to fix a compiler warning.

drwells:
All the tests pass so this is probably fine. @amneetb - could you take a second look since you understand this class best?

drwells (approved):

boyceg:
@drwells, is it worth squashing some commits before merging?

drwells:
Cleaning up history is nice but ultimately optional. If @hongk45 knows how to squash commits that would be great - otherwise we can go with it as-is.

hongk45:
@drwells, @boyceg: I tried a bit to clean up the history. Hope it appears cleaner now.

drwells:
@hongk45 I'm not sure things are the way you intended for them to be - we went from six commits to thirteen.

The current state of the history is now problematic since it contains a merge commit. In particular, if we were to merge this PR as it stands, the commits here would be intercalated with those on `master`, which would make `git bisect` no longer work. More exactly: I'm fine with not squashing commits but we cannot have 'merge into master' commits like this in PRs since it makes using other git tools impossible.

The best way to fix this situation is to completely rewrite history and force push. More exactly:
1. Do a soft reset so that the commits are gone but the changes are still present.
2. Add the changes again and write a single commit containing all of them.

the following git commands should resolve this problem:
```
git checkout periodic-correction-constraintIB
git merge master
git reset --soft master
git commit -m "Periodic correction for the ConstraintIB method"
```
In order:
1. Move to the branch
2. If you have updated master then we merge those changes in, in order. This gets around the problem with intercalated changes to `master` making `git rebase` not work.
3. Keep the changes but reset `HEAD` to master (i.e., remove all commits but not the changes)
4. Recommit all the changes into a single commit

Do this and then force push and we should be set :+1:

drwells (changes requested):
No merge commits, please.

amneetb:
Do we have an example to test the new functionality?

hongk45:
@drwells: followed your steps, but the problem seems to be persistent. Here is what I did:
---------------
git checkout periodic-correction-constraintIB
git merge master
git reset --soft master
git commit -m "Periodic correction for the ConstraintIB method"
git add doc/news/changes/minor/20201201HongNguyen
git add src/IB/ConstraintIBMethod.cpp
git add src/IB/ConstraintIBMethod.cpp
git push origin periodic-correction-constraintIB
--------------------
Am I missing anything?

boyceg:
> Do we have an example to test the new functionality?

This is a good point. If it has not already been done, we should verify that there are not any artifacts when structures "pass through" the periodic boundary. I think that ConstraintIB will be less susceptible to this than formulations designed for flexible structures, but it is valuable to check.

One thing that could be done is to compare lift/drag forces for a body that is dragged through the domain. In a fully periodic domain, the traces should be identical if the structure is offset by a distance that is an integer number of Cartesian grid cells in any direction.

boyceg:
> ## @drwells: followed your steps, but the problem seems to be persistent. Here is what I did:
> ## git checkout periodic-correction-constraintIB
> git merge master
> git reset --soft master
> git commit -m "Periodic correction for the ConstraintIB method"
> git add doc/news/changes/minor/20201201HongNguyen
> git add src/IB/ConstraintIBMethod.cpp
> git add src/IB/ConstraintIBMethod.cpp
> git push origin periodic-correction-constraintIB
> Am I missing anything?

@hongk45: Sorry for dragging you into this git mess. 😬 

@drwells: At this point, would you be able to fix up the history for @hongk45?

amneetb:
> One thing that could be done is to compare lift/drag forces for a body that is dragged through the domain. In a fully periodic domain, the traces should be identical if the structure is offset by a distance that is an integer number of Cartesian grid cells in any direction.

This is a very good example. The simulation can be set following sec 5.1.2 of:
https://www.sciencedirect.com/science/article/pii/S0021999117305016
https://arxiv.org/pdf/1704.00239.pdf

Another example would be 2d eel swimming. One can set the eel to cross the periodic boundary during its swimming path, and check that if its velocity or displacement remains the same or not, when it does not cross the periodic boundary.

boyceg:
> Another example would be 2d eel swimming. One can set the eel to cross the periodic boundary during its swimming path, and check that if its velocity or displacement remains the same or not, when it does not cross the periodic boundary.

I think that you would want to do the same basic check, however --- have the initial position vary by some integer number of grid cells and verify that the kinematics are identical.

boyceg:
One quick comment on "identical" results.

In principle, when using periodic boundaries and a uniform (not adaptive!) Cartesian grid, you should be able to offset the structure by an integer number of Cartesian grid cells and get identical results.

It is important to use a uniform Cartesian grid because the AMR grid generator may not generate identical (up to a translation) locally refined regions for the offset positions of the structure, and it _will not_ generate identical locally refined grids as the structure approaches / passes through the periodic boundary.

Further, even on a uniform Cartesian grid, there are some ordering-dependent parts of the solvers that could generate slightly different results for the offset configurations. These differences should be small compared to the overall forces and/or trajectories.

drwells:
@hongk45 Did you force push when you updated the branch on GitHub? The last command shouldn't work since you rewrote the git history in a conflicting way. Let's hold off on fixing this until we determine what sort of example or test we should add.

boyceg:
@hongk45, were you able to add a test to verify correctness? Note that the test does not need to be particularly well resolved --- we just want something to confirm that the method does not "see" the periodic boundary.

hongk45:
@boyceg: Yeah, I actually simulated free swimming of a  single deforming structure in the regime of not too high Re where there is no vortex shedding. When I chose the physical box sufficiently large, then nothing peculiar appeared when the structure crossed the periodic box, for example the swimming velocity. 

I prescribed both shape and (deforming) velocity and found the same swimming velocity with "CONSTRAINT_VELOCITY" or "CONSTRAINT_POSITION". However, if the calculation of COM was not fixed and "CONSTRAINT_POSITION" was applied, when the structure was getting close to the boundary and wanted to pass, it could not do that because a wrong center of mass was used to update its position. 

In this example, I only have translational degree of freedom. If the structure is allowed to rotate, then, of course, we need a correct value of MOI.

drwells:
This was on pause over the holiday but I'd like to get it in ASAP.

@hongk45 by 'test' we mean an integration test that doesn't necessarily run with physical parameters but still demonstrates that some part of the code should work - in particular it should run in less than a minute and use the smallest grid it can to get the point across. Can you modify your code to do this? If so, are you willing to add it to IBAMR to make sure that ConstraintsIB continues to work with periodic BCs?

hongk45:
> 
> 
> This was on pause over the holiday but I'd like to get it in ASAP.
> 
> @hongk45 by 'test' we mean an integration test that doesn't necessarily run with physical parameters but still demonstrates that some part of the code should work - in particular it should run in less than a minute and use the smallest grid it can to get the point across. Can you modify your code to do this? 
>
@drwells: sorry for my slow reply. Actually, I don't mind modifying my example to serve as a test for this additional functionality.  
>If so, are you willing to add it to IBAMR to make sure that ConstraintsIB continues to work with periodic BCs?
>
You mean it is possible for my example to be added to the master branch if it works well, or something else. Sorry I am a bit confused here.

Anyway, please help me to understand why the current example codes in examples/ConstraintIB/ can not be used to test this. I think 'impulsively_started_cylinder', 'falling_sphere' or 'oscillating_rigid_cylinder' might be a good tester with a little change of input parameters.

drwells:
We would add the test to this PR so that we merge both the bug fix and the test that verifies the bug fix at the same time.

> Anyway, please help me to understand why the current example codes in examples/ConstraintIB/ can not be used to test this. I think 'impulsively_started_cylinder', 'falling_sphere' or 'oscillating_rigid_cylinder' might be a good tester with a little change of input parameters.

If all we need to verify that the bug is fixed is a changed input file then that's even better. Adding a new test or example requires modifying the build system - I can take care of that. Could you attach the input file to a GitHub comment in this PR?

drwells:
@hongk45 Lets try to wrap this up soon. Do you still have the input file you used to show that these changes work? If so, would you be willing to send it to me (either by email or by posting it here)?

hongk45:
> 
> 
> @hongk45 Lets try to wrap this up soon. Do you still have the input file you used to show that these changes work? If so, would you be willing to send it to me (either by email or by posting it here)?

hi @drwells, i am currently fully occupied by something else. I will get back to this by the end of this week or next.

drwells:
Not a problem - let me know when you're ready to wrap this up.

hongk45:
> 
> 
> Not a problem - let me know when you're ready to wrap this up.

hi @drwells, just want to let you know that i will get back to this ASAP

hongk45:
Hi @drwells, I have here 2 input files that might be useful to verify that this fix actually works.
These are for oscillating_rigid_cylinder example:
1) [input2d_oscillating_rigid_cylinder_nonPBC.txt](https://github.com/IBAMR/IBAMR/files/5974146/input2d_oscillating_rigid_cylinder_nonPBC.txt)
This file only applies PBC along y, but not along x. The cylindrer is placed at the center of the box which is large enough to be considered as infinite.

2) [input2d_oscillating_rigid_cylinder_PBC.txt](https://github.com/IBAMR/IBAMR/files/5974148/input2d_oscillating_rigid_cylinder_PBC.txt)
This file applies PBC along both x and y. The cylinder is placed right at the boudary at the beginning. The point here is to make the structure to continually cross the periodic boundary. The box size is the same as in previous case. 


We would expect that the 2 setups will produce identical results, e.g the drag force. And indeed, this is the case.
You can also check that this fix will produce the correct result for the cylinder's COM while previous unfixed code cannot.

Let me know if you need more example.

drwells (approved):

drwells:
All the tests pass and this looks good to me so lets merge.

drwells:
@hongk45 Thanks again for fixing this!

PR #1175, opened 2020-12-07
add .cache (from clangd) to .gitignore

boyceg:
- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
It seems like `clangd` has changed the location where it stores cache files.

drwells:
I don't use `clangd` but this change seems perfectly reasonable to me.

drwells (approved):

Issue #1177, opened 2020-12-08
Approximate floating point comparison with 0.0


abarret:
The `SAMRAI::tbox::MathUtilities<double>::equalEps()` function checks relative differences between numbers. This gives a more stringent check for values close to zero than for larger values. In particular:

`MathUtilities<double>::equalEps(0.0, std::numeric_limits<double>::epsilon())` returns `false`
while
`MathUtilities<double>::equalEps(1.0, 1.0 + std::numeric_limits<double>::epsilon())` returns `true`.

A better comparison (in some cases), would be an absolute comparison. We should implement a `rel_equal_eps()` and a `abs_equal_eps()`, and make sure these are used appropriately in the library.

drwells:
@abarret do you want to push anything else for the 0.8.0 release?

abarret:
> @abarret do you want to push anything else for the 0.8.0 release?

When do you plan on putting out 0.8.0? I can probably get to it tomorrow.

The only thing left is to switch over from using `MathUtilities<double>::equalEps` to the correct `IBTK` version. It shouldn't (fingers crossed) actually change the code paths, but if it does, determining what the "right" behavior is might be difficult.

drwells:
> When do you plan on putting out 0.8.0? I can probably get to it tomorrow.

I want to be done by the start of the semester so ideally all patches need to be in this week (so I have a week to fix anything that fell through the cracks).

drwells:
@abarret will you get this done by January?

abarret:
It is on my to-do list.

I ran into issues before because changing to the `IBTK` versions caused some tests to fail. I will revisit this and see what's going on.

abarret:
@drwells, replacing the SAMRAI `equalEps` with the appropriate `IBTK` version causes a change in the test `explicit_ex4_2d.displace.mpirun=4`. The difference is only in one line in calculating the maximum node displacement:
```
----------------
##4404    #:7   <== 0.002533931815
                ==> 0.002533934630
@ Absolute error = 2.8150000000e-9, Relative error = 1.1109217633e-6
```

drwells:
That's unexpected. Is that branch publicly available?

abarret:
Yes, you can find it here
https://github.com/abarret/ibamr/tree/replace_equal_eps

There's potential that very small values could now be ignored with the patch, but it's surprising that it's only one line of one test.

[...]

PR #1178, opened 2020-12-08
Added functions to check for absolute and relative equality between doubles.

abarret:
Part one of #1177. This adds functions to compare relative and absolute equality between doubles. I've also added a test for the functions. There's an optional argument for a tolerance, which defaults to `sqrt(numeric_limits<double>epsilon())`. This is the same value that SAMRAI uses to check for equality.

There is a slight difference between SAMRAI's function and `rel_equal_eps`. SAMRAI regularizes the division, while this version checks first if both numbers are exactly zero. This gives `rel_equal_eps` a more "accurate" relative difference, but allows for slightly different answers, mainly:

`rel_equal_eps(0.0, std::numeric_limits<double>::min())` returns `false`
while
`MathUtilities<double>::equalEps(0.0, std::numeric_limits<double>::min())` returns `true`

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
You'll also need to add the test to `tests/CMakeLists.txt` - do you see how to do that?

abarret:
Whoops. I meant to add this to CMake, but forgot before I pushed. It should be fixed.

ibtk/src/utilities/ibtk_utilities.cpp
```diff
@@ -48,4 +48,19 @@ get_min_patch_dx(const PatchLevel<NDIM>& patch_level)
 
     return result;
 } // get_min_patch_dx
+
+bool
+rel_equal_eps(const double a, const double b, const double eps)
```
boyceg, line 53: I think these should be inline functions.

ibtk/src/utilities/ibtk_utilities.cpp
```diff
@@ -48,4 +48,19 @@ get_min_patch_dx(const PatchLevel<NDIM>& patch_level)
...
+{
+    if (a == b) return true;
```
boyceg, line 55:
```suggestion
    return (a == b) || ((std::abs(a - b) / std::max(std::abs(a), std::abs(b))) < eps);
```
This will "short circuit" if `a == b` and not evaluate the RHS of the `||`.

ibtk/src/utilities/ibtk_utilities.cpp
```diff
@@ -48,4 +48,19 @@ get_min_patch_dx(const PatchLevel<NDIM>& patch_level)
...
+    double abs_max = std::max(std::abs(a), std::abs(b));
+    return (std::abs(a - b) / abs_max) < eps;
+}
+
+bool
+abs_equal_eps(const double a, const double b, const double eps)
+{
+    return std::abs(a - b) < eps;
```
boyceg, line 63:
```suggestion
    return (a == b) || (std::abs(a - b) < eps);
```

ibtk/src/utilities/ibtk_utilities.cpp
boyceg: BTW, I am not sure if the extra equality check is worth doing or not (in either of these functions) --- might be worthwhile to do some timing checks.

abarret: The equality check here is to prevent dividing by zero. I don't think it's worth doing in the other function.

If speed is truly a concern, we could implement a check to see if `a` is within some number of representable numbers away from `b`. This would remove the division but still be similar to a relative difference check.

boyceg (approved):
Looks good to me. The only question I have is whether it would make sense to template this over the input types.

abarret:
> Looks good to me. The only question I have is whether it would make sense to template this over the input types.

I considered that, but I think the only other type we need comparisons for would be `float`s, which we don't use in the library. I can add them if you think that would be helpful.

boyceg:
Templating it properly is a little tricky if you want to properly handle mixed type comparisons. Let’s not worry about if for now.

@drwells, what do you think?

tests/IBTK/equal_eps.cpp
```diff
@@ -0,0 +1,74 @@
+#include <ibtk/AppInitializer.h>
```
drwells, line 1: You're missing the copyright blurb

ibtk/src/utilities/ibtk_utilities.cpp
```diff
@@ -48,4 +48,5 @@ get_min_patch_dx(const PatchLevel<NDIM>& patch_level)
 
     return result;
 } // get_min_patch_dx
+
```
drwells, line 51: Lets remove the blank line unless clang-format put it there.

drwells (changes requested):

drwells:
I think the current version is fine - using `float` in this way is pretty uncommon.

abarret:
Fixed and squashed.

drwells (approved):

PR #1184, opened 2020-12-17
Add CMake files.

drwells:
This is a PR onto #1132, not master.

@Raamkrishnan I added the necessary changes to this branch to get things compatible with CMake (the current branch is fine with autotools). We can add this whenever you are ready to do so - no rush.


drwells:
I'm going to add these changes directly into #1132 so we don't need this PR any more.

PR #1188, opened 2021-01-02
enable users to control whether quadrature rules are allowed to use negative weights

### LINKED ISSUE #1187: Add input options to control whether quadrature rules can use negative weights

boyceg:
Resolves #1187.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
@drwells, should we leave the current default behavior (allowing for negative weights) until after the 0.8 release?

drwells:
We need an easy way to turn things off (which this patch does and is good) but this change is not backwards-compatible: adding an extra boolean at the front like this means that we will get different behavior for the same constructor calls which we really shouldn't do.

Unfortunately this isn't going to do what you want with spreading and interpolation since `FEDataManager::updateQuadratureRule()` is not actually used in those functions (see #1189). We assume `IBTK::QuadratureCache` has all the information necessary to pick quadrature rules and its not informed of the whole positive/negative weights thing.

Minimally, to address the first problem, we should deprecate the non-default constructor (the one that takes all the arguments) and pass in an input database instead. That will let us add parameters in a backwards-compatible way. To address the second problem we will need to modify the constructor of `IBTK::QuadatureCache` to take `allow_q_rules_with_negative_weights` as an input.

boyceg:
This could be a coarser-grained control (e.g. state variable of `FEDataManager`).

drwells:
That would work but it's going to get confusing if we have settings for quadrature rules both inside the `Spec` structs and in a state variable of `FEDataManager` - we would also need a way to link `FEDataManager` to the quadrature caches.

boyceg:
How about I just move the new argument to the end of the constructor interface, default it to `false` (for now --- change to `true` after 0.8), and do the right thing w.r.t. `QuadratureCache`?

drwells:
That does solve both problems so lets go for it.

boyceg:
What do you think about leaving the current default for now? (Which I originally reversed in the comment above...)

drwells:
It's an incompatible change so I'd prefer to preserve the status quo right now - besides, this change is going to make a lot of test outputs (particularly for spreading) different so I don't want to deal with that right before we put out a release

boyceg:
@drwells, please take a look.

Note that I have a single definition for the `key_type` for quadrature rules (in `libmesh_utilities.h`) that is referenced as needed elsewhere.

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -57,6 +57,8 @@ IBTK_ENABLE_EXTRA_WARNINGS
 
 namespace IBTK
 {
+using quadrature_key_type = std::tuple<libMesh::ElemType, libMesh::QuadratureType, libMesh::Order, bool>;
```
drwells, line 60: Good idea - it's not ideal that I defined this tuple in a dozen places.

ibtk/include/ibtk/libmesh_utilities.h
```diff
@@ -514,14 +516,14 @@ void apply_transposed_constraint_matrix(const libMesh::DofMap& dof_map, libMesh:
...
-                 const boost::multi_array<double, 2>& X_node,
-                 const double dx_min);
+quadrature_key_type getQuadratureKey(libMesh::QuadratureType quad_type,
+                                     libMesh::Order order,
+                                     bool use_adaptive_quadrature,
+                                     double point_density,
+                                     bool allow_rules_with_negative_weights,
+                                     const libMesh::Elem* elem,
+                                     const boost::multi_array<double, 2>& X_node,
+                                     double dx_min);
```
drwells, line 526: This is an incompatible change (so is the bounding box change) but it's really unlikely that any users will notice so I don't mind. We do need to update the documentation of this function.

drwells (approved):
Looks good - we'll need to run the tests though.

boyceg:
Weird, a bunch of IBFE tests are now failing. 😞

boyceg:
Ah, it is because of the change in the arguments to the `XXXSpec` constructors.

drwells:
There is something wrong in how this is getting set up: I see errors like

```
Quadrature rule 44 not supported!
```

boyceg:
> There is something wrong in how this is getting set up: I see errors like
> 
> ```
> Quadrature rule 44 not supported!
> ```

I forgot to update the arguments to the `InterpSpec` and `SpreadSpec` constructors in `IBFEMethod` and `IBFESurfaceMethod`. I am re-running tests now.

boyceg:
@drwells I think I fixed the errors. I am re-running the test suite now.

drwells:
Did you forget to patch `tests/IBTK/jacobian_calc_01.cpp`?

boyceg:
Apparently. Could you push a fix?

drwells:
Sure.

drwells:
The only other failing tests are the ones based on `IBFE/explicit_ex5`.

boyceg:
Hmmm, I though I fixed those. Let me look again...

drwells:
Its not obvious what the problem is. That test does use IBFESurfaceMethod, though.

boyceg:
@drwells, found it. I'm re-running tests now.

boyceg:
@drwells, test suite now passes for me. Do you want to give this another look before merging?

boyceg:
Squashed.

PR #1192, opened 2021-01-05
Explicitly compile example grid generators with C++11.

drwells:
This is necessary for some compilers (like clang on my mac) that don't yet
default to C++11 or C++14.

Since this only pertains to examples that have grid generation programs that don't use IBAMR or IBTK the checklist doesn't apply.


drwells:
Clarification: this doesn't actually apply to any examples (the original title was wrong). It only applies to things like `bargeGen` which we compile like examples but do not depend on IBAMR or IBTK.

boyceg (approved):

PR #1195, opened 2021-01-06
use QNODAL to build lumped mass matrices

boyceg:
(for libMesh versions < 1.5.0, manually use either QTRAP or QSIMPSON, as appropriate)

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
I tentatively approve, based on the observation from the nodal quadrature paper that Hinton's rule gives us meaningless (but positive) values for P2 but is otherwise equal to or inferior to doing nodal quadrature or row sums in all other circumstances. By 'meaningless' I mean "it doesn't converge with very basic static test problems like projecting the divergence of the PK1 stress". If it works for other specific problems that's fine but I don't think we should endorse it as a default.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -317,34 +320,31 @@ FEProjector::buildLumpedL2ProjectionSolver(const std::string& system_name)
         for (MeshBase::const_element_iterator el_it = el_begin; el_it != el_end; ++el_it)
         {
             const Elem* const elem = *el_it;
+#if NDEBUG && LIBMESH_VERSION_LESS_THAN(1, 5, 0)
+            {
+                auto elem_type = elem->elem_type();
+                TBOX_ASSERT(((fe_type.order == FIRST) && ((dim == 2 && (elem_type == TRI3 || elem_type == QUAD4)) ||
+                                                          (dim == 3 && (elem_type == TET4 || elem_type == HEX8)))) ||
```
drwells, line 327: this isn't right for surface meshes.

drwells: Actually, maybe it is - let me check.

boyceg:
An alternative would be to support the choice of either Hinton's rule or nodal quadrature-based constructions.

ibtk/src/lagrangian/FEProjector.cpp
boyceg: No you are right --- need to allow for `dim == 1` for surface-based models.

drwells: I think we need to explicitly write out `dim` and `spacedim` or whatever libmesh calls it.

This check also forbids us from using mixed meshes - I don't know if anyone is interested in this case but we might want to support it.

boyceg: Mixed meshes are disallowed by checks in `FEMechanicsBase`.

boyceg: @drwells I think I fixed this.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -322,11 +322,24 @@ FEProjector::buildLumpedL2ProjectionSolver(const std::string& system_name)
...
+                TBOX_ASSERT(elem->default_order() == fe_order);
                 auto elem_type = elem->elem_type();
-                TBOX_ASSERT(((fe_type.order == FIRST) && ((dim == 2 && (elem_type == TRI3 || elem_type == QUAD4)) ||
-                                                          (dim == 3 && (elem_type == TET4 || elem_type == HEX8)))) ||
-                            ((fe_type.order == SECOND) && ((dim == 2 && (elem_type == TRI6 || elem_type == QUAD9)) ||
-                                                           (dim == 3 && (elem_type == TET10 || elem_type == HEX27)))));
+                switch (fe_order)
+                {
+                case FIRST:
+                    TBOX_ASSERT((dim == 1 && (elem_type == EDGE2)) ||
```
boyceg, line 331: I guess I can simplify this and just list out the `FIRST` and `SECOND` order elements that are supported, and not worry about `dim`.

I will do that now...

drwells:
> An alternative would be to support the choice of either Hinton's rule or nodal quadrature-based constructions.

I don't know of any advantage to Hinton's method so I don't see any reason to do this inside IBAMR.

boyceg:
There are some diffs in the `TRI6` nodal interpolation tests. I will try to investigate.

boyceg:
Hmmm, this is confusing to me.

There are some nontrivial differences in `interpolate_velocity_01_2d.nodal_quadrature.tri6` and `interpolate_velocity_01_3d.nodal_quadrature.tet10`.

For nodal coupling, it seems like it should be the case that the form of the diagonal mass matrix does not matter for IBFE interpolation.

boyceg:
Oh, wait, I misread the test. It also prints out the diagonal mass matrix entries, which are different for the two schemes.

boyceg:
The interpolation max errors agree to at least 12 digits. I will just update the tests.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -317,34 +320,43 @@ FEProjector::buildLumpedL2ProjectionSolver(const std::string& system_name)
         for (MeshBase::const_element_iterator el_it = el_begin; el_it != el_end; ++el_it)
         {
             const Elem* const elem = *el_it;
+#if NDEBUG && LIBMESH_VERSION_LESS_THAN(1, 5, 0)
```
drwells, line 323: I think you mixed up `NDEBUG` and `DEBUG` here.

I am always in favor of running these sorts of checks - anything thats not in an inner loop is negligibly cheap.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -550,7 +562,7 @@ FEProjector::buildDiagonalL2MassMatrix(const std::string& system_name)
         FEData::SystemDofMapCache& dof_map_cache = *d_fe_data->getDofMapCache(system_name);
         dof_map.compute_sparsity(mesh);
         FEType fe_type = dof_map.variable_type(0);
-        std::unique_ptr<QBase> qrule = fe_type.default_quadrature_rule(dim);
+        std::unique_ptr<QBase> qrule = QBase::build(QNODAL, dim);
```
drwells, line 565: This needs to be version-guarded.

ibtk/src/lagrangian/FEProjector.cpp
boyceg: Fixed.

ibtk/src/lagrangian/FEProjector.cpp
boyceg: Fixed.

tests/IBFE/interpolate_velocity_01_2d.nodal_quadrature.tri6.input
```diff
@@ -1,59 +1,686 @@
-print_diagonal_mass_matrix = TRUE
+Number of elements: 160
```
drwells, line 1: I think you replaced the input file with the new output file.

drwells:
Only one of the calls to `QNODAL` has a version check - can you add a utility function in the anonymous namespace so we don't have repeated code?

drwells:
I fixed the QNodal code. We get different diagonal entries now with libMesh 1.1.0 so there is something wrong with the switch.

boyceg:
> Only one of the calls to `QNODAL` has a version check - can you add a utility function in the anonymous namespace so we don't have repeated code?

Both should have a guard using `LIBMESH_VERSION_LESS_THAN` --- I force pushed, maybe you need to reset to origin?

boyceg:
> We get different diagonal entries now with libMesh 1.1.0 so there is something wrong with the switch.

Hmmm. Looking at the libMesh source code, `QTRAP` and `QSIMPSON` do not seem to have been changed in any important way since 1.1.0 (I think that the libMesh commit is 941e959).

drwells (changes requested):
I must have screwed something else up - everything works (once we fix `elem_type` -> `type`) with libMesh 1.1.0.

boyceg:
@drwells can you push a fix since I am not testing with older libMesh versions?

boyceg:
> @drwells can you push a fix since I am not testing with older libMesh versions?

Never mind, I finished it. Take a look?

drwells:
This one still doesn't work with older versions of libMesh. I'll fix it in the morning.

drwells:
There are several problems with QNODAL, for our use case:
1. It doesn't work with monomials.
2. It doesn't exist in older versions of libMesh
3. Related to 1: it sets the quadrature rule based on the nodes of the `Elem`, which are not necessarily the interpolation points of the actual FE we are using (e.g., we can put `QUAD4` elements on a `QUAD9` mesh, but then we don't want to use `QSIMPSON`). We do this in the postprocessors.

Hence I rewrote this to pick the correct quadrature rule based on the order of the actual finite element field under consideration rather than the `Elem`'s number of nodes.

I'm still not convinced this doesn't do bad things to postprocessors (though the tests do now pass). I'd prefer not to merge it in this release until we know what's going on.

boyceg:
Looks good to me. I think that it should be fine to put this in for the release, but if you want to wait, that is fine too.

drwells:
True - we could add extra checks for serendipity elements.

drwells:
nodal quadrature is (in the documentation) still marked as experimental so I think its fine if we skip supporting unusual elements for now. I added another commit that removes usage of the lumped mass matrix in the standard L2 projection solver - that fixes the problem for arbitrary element types with non-nodal quadrature so I'm content that there won't be any odd incompatibilities.

drwells (approved):

boyceg:
Sorry, what is the change in behavior for the standard L2 projection?

drwells:
We compute the preconditioner from the system matrix, instead of the lumped one. I thought at one point that the later would be better but it turns out there is no significant difference (particularly since we have very low iteration counts). Put another way: the lumped mass preconditioner might reduce the iterations from 62 to 58, but that effect is gone now that we have good initial guesses and only need 1 or 2 iterations most of the time.

PR #1197, opened 2021-01-07
add input option controlling if negative weights are allowed in IBFE coupling

boyceg:
This option should have been included in #1188.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -2966,6 +2966,7 @@ IBFEMethod::getFromInput(const Pointer<Database>& db, bool /*is_from_restart*/)
         d_default_interp_spec.quad_order = FIRST;
         d_default_interp_spec.use_adaptive_quadrature = false;
         d_default_interp_spec.use_consistent_mass_matrix = false;
+        d_default_interp_spec.allow_rules_with_negative_weights = false;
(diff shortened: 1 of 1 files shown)


src/IB/IBFEMethod.cpp
```diff
@@ -2966,6 +2966,7 @@ IBFEMethod::getFromInput(const Pointer<Database>& db, bool /*is_from_restart*/)
         d_default_interp_spec.quad_order = FIRST;
         d_default_interp_spec.use_adaptive_quadrature = false;
         d_default_interp_spec.use_consistent_mass_matrix = false;
+        d_default_interp_spec.allow_rules_with_negative_weights = false;
```
drwells, line 2969: Since this changes the default behavior I would prefer to wait on merging this until after the release.

boyceg: It does not change the default behavior because the nodal rules only have positive weights.

drwells: oh, I see - this is for `nodal_quadrature`. I should have scrolled up.

boyceg: Most of these settings do not have any impact for the specialized nodal interaction code path.

drwells (approved):

Issue #1171, opened 2020-12-03
IBAMR 0.8


drwells:
It's that time of year again: lets put out an official release by ~the end of the month~ January 15. Unless someone is willing to volunteer to help with the release tasks we need a pretty conservative timeline so that I can get everything done. In particular:

- All non-bugfix patches must be in by ~December 15~ January 10
- We should try to get all bugfix patches in by ~the 18th~ January 10, but depending on their severity we might want to include some extra ones later
- Branch and put out a release candidate on ~the 23rd~ January 11
- Finalize the release by ~the 28th~ January 15

# IBAMR release tasks

## static analysis

- [x] verify that neither `namespaces.h` nor `app_namespaces.h` are in IBTK or
  IBAMR headers
- [x] clear basic compilation warnings with GCC and clang on linux (#1194)
- [x] compile with the latest libMesh release (which should be configured with
  `--disable-deprecated`) to verify that we do not use any deprecated
  functionality (this will cause compilation errors if we do use them). Add
  version checks: for example, here is one (lightly edited) in
  `IBFEInstrumentPannel.cpp` that uses a macro defined in `libmesh_version.h`:
```cpp
    // new API in 1.4.0
#if LIBMESH_VERSION_LESS_THAN(1, 4, 0)
    boundary_info.build_node_list(nodes, bcs);
#else
    const auto node_list = boundary_info.build_node_list();
    for (const auto& pair : node_list)
    {
        nodes.push_back(std::get<0>(pair));
        bcs.push_back(std::get<1>(pair));
    }
#endif
```
  around APIs to keep backwards compatibility with older versions of libMesh.
- [x] check the test suite with the oldest (3.7.0) and newest (latest release)
  versions of PETSc. Also check with the oldest (1.1.0) and newest (latest
  release) versions of libMesh.
- [x] clear basic compilation warnings with GCC and clang on linux and macOS (#1193)
- [x] run cppcheck via
```
  cppcheck --force -j4 --enable=all -I./include/ -I./ibtk/include/  \
  $(find ./src ./ibtk/src -name '*.cpp') >cppcheck-results.txt 2>&1
```
- [x] Check that we don't call functions like
  `ReplicatedMesh::active_local_elements_end()` inside `for`-loop declarations
  since the end iterator is expensive to compute.
- [x] Check that we always get unique `HierarchyDataOps` by passing `true` as
  the third argument to `getOperationsDouble` et al. This prevents memory leaks
[...]

PR #1205, opened 2021-01-12
Update copyright 0.8

drwells:
Part of #1171.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


scripts/maintenance/update_copyright.sh
```diff
@@ -0,0 +1,58 @@
...
+  # (it should be enough to look at the last 2 messages since
+  # ideally no two successive commits should have updated the
+  # copyright year. let's err on the safe side and take the last
+  # 3 commits.)
+  last_year=`git log -n 3 --date=short --format="format:%cd %s" $FILE | \
+             egrep -i -v "update.*copyright|copyright.*update" | \
+             head -n 1 | \
+             perl -p -e 's/^(\d\d\d\d)-.*/\1/g;'`
+
+  # get the first year this file was modified from the actual
```
boyceg, line 48: I wonder if it makes sense to pick either the min or the max here. Like it could be that the first year listed is later than the earliest commit (because of typos).

boyceg (approved):
I think that I am going to have to take your word for it. 😀

scripts/maintenance/update_copyright.sh
drwells: There is some ambiguity here so I am content to let the author decide on copyright.

drwells:
A lot of the updated copyrights are for 2019 - it looks like whatever we did then didn't work correctly. This should be smaller in the future.

PR #1207, opened 2021-01-12
Bump VERSION to 0.9.0-pre

drwells:
Before merging, we should decide on a few things:
1. Will we move to SAMRAI 3 before the next release?
2. How long should we keep autotools around for?

I don't know what the state of the SAMRAI 3 transition is. I would like to keep autotools around for at least two more releases, something like:
- 0.8.0: unofficial CMake support (we are here now)
- 0.9.0: (circa May 2021) CMake is official, autotools is deprecated and prints warnings about its upcoming removal
- 0.10.0: (circa January 2022) same
- 1.0.0: (circa May 2022) autotools is gone

We should increment the major version number if we switch to CMake or if we switch to SAMRAI 3 - otherwise I recommend we keep at `0.X.Y`.


boyceg:
1. Is it worth considering alternatives to SAMRAI 3/4?
2. If we go with SAMRAI 3/4 (as currently planned), I think that we should try to make this move whenever @abarret's scripts are ready. 😀 
3. I think that whenever we get away from SAMRAI 2 is when we should bump up to IBAMR 1.0.
4. I think keeping autotools around until IBAMR 1.0.

abarret:
> 2. If we go with SAMRAI 3/4 (as currently planned), I think that we should try to make this move whenever @abarret's scripts are ready.

SAMRAI 4 is a lot further away than I had hoped. Turns out there's a lot of things about IBAMR that I didn't know existed.

boyceg:
> SAMRAI 4 is a lot further away than I had hoped. Turns out there's a lot of things about IBAMR that I didn't know existed.

IBAMR is large, IBAMR contains multitudes.

drwells:
> Is it worth considering alternatives to to SAMRAI 3/4?

I'd say no. We have a very small number of files that only depend on libMesh but everything else in the library is based on SAMRAI. We'd basically have to rewrite the whole library.

Besides - SAMRAI does a really good job at the whole 'structured grid FDM/FVM in parallel'. I'm not aware of any competitor in this space that's significantly better (parthenon is cool but doesn't support MAC-like discretizations, amrex is really different so I don't know if that's even on the table). Parthenon is built on kokkos which would be awesome to use but not at the cost of rewriting everything.

Even if we were to replace SAMRAI, libMesh is a bigger bottleneck for the things we do now, so I consider `IBFEDEALMethod` to be a higher priority right now (we need to fix those scaling issues before we will run into scaling issues with SAMRAI).

drwells:
I know it sounds like more work but I'd bet its easier to move to SAMRAI 3 (or really any intermediate release) first and just step our way to the end. Now that we have the test suite we should be able to check automatically if things broke particularly badly.

boyceg:
> Even if we were to replace SAMRAI, libMesh is a bigger bottleneck for the things we do now, so I consider `IBFEDEALMethod` to be a higher priority right now (we need to fix those scaling issues before we will run into scaling issues with SAMRAI).

BTW, instead of making `IBFEDEALMethod`, I wonder if it would be possible to have a generic `IBFEMethod` interface that has libMesh and deal.II back ends.

drwells:
> BTW, instead of making IBFEDEALMethod, I wonder if it would be possible to have a generic IBFEMethod interface that has libMesh and deal.II back ends.

It would be great if users could just load their data into either library in an agnostic manner but I don't think this is possible - we leak too much stuff (like postprocessing, stress function pointers, etc) that depends on libMesh outside of the class. I have some plans but they are at the sketches phase - we can talk about this later but I don't want to dive in quite yet.

drwells:
It sounds like the next release should not be 1.0 so lets go ahead and merge this. Would someone please approve?

boyceg (approved):

PR #1211, opened 2021-01-25
Convert libMesh::DenseMatrix to Eigen::Matrix.

drwells:
Thes solves a few problems:
1. We assume (in FischerGuess) that libMesh links against LAPACK - this might not be true so we shouldn't use its SVD solve.
2. Even in OPT mode, libMesh still does some bounds checking that shows up in profiling (e.g., in IBTK::PointMap). We check the bounds outside the loop anyway so switching to Eigen (which won't bounds-check) is safe and more efficient.
3. Eigen offers a cleaner API.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/FEMapping.h
+++ b/ibtk/include/ibtk/FEMapping.h
@@ -25,7 +25,6 @@
 
 #include "tbox/Utilities.h"
 
-#include <libmesh/dense_matrix.h>
 #include <libmesh/elem.h>
 #include <libmesh/enum_elem_type.h>
 #include <libmesh/enum_order.h>
@@ -35,6 +34,10 @@
 
 IBTK_DISABLE_EXTRA_WARNINGS
 #include <boost/multi_array.hpp>
+IBTK_ENABLE_EXTRA_WARNINGS
+
+IBTK_DISABLE_EXTRA_WARNINGS
+#include <Eigen/Core>
 IBTK_ENABLE_EXTRA_WARNINGS
 
 #include <array>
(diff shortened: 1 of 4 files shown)


boyceg:
Outside of the scope of this PR, but I wonder if we should consider deprecating libMesh-based APIs in favor of Eigen-based APIs, to facilitate switching FE "back-ends" without substantial changes to user code?

drwells:
We already do this in some places - all of our mapping classes use `Matrix<double, dim, spacedim>` since libMesh doesn't really have a notion of a 2 x 3 tensor.

That being said - the way we would do things with deal.II would be completely different than the way we do things with libMesh since the library has totally different data structures. For example - no code where a user adds a system to a mesh has an analogue in deal.II since dofs are not in a `Triangulation` (they are in `DoFHandler`) and `System` doesn't exist. Hence any such compatibility layer would have very limited applicability since it could only be used for very basic problems and nothing else (and if people are only doing very basic things then they probably don't care about the differences in the backend since they never see it anyway).

boyceg:
> Hence any such compatibility layer would have very limited applicability since it could only be used for very basic problems and nothing else (and if people are only doing very basic things then they probably don't care about the differences in the backend since they never see it anyway).

For a lot of use cases, users do not need to know anything about the mesh data structures --- just need to know how to evaluate forces/stresses/etc --- so the intention would be to make those kinds of codes back-end-oblivious.

boyceg (approved):

drwells:
If all we need to do is support defining forces and stresses then this is achievable in IBFEMethod.

We should talk about plans for improved parallelization via deal.II at some point but I want to wrap some other projects up first, so lets shelve that discussion for now.

PR #1228, opened 2021-03-16
Remove some deprecated libMesh code.

drwells:
`boundary_info` is now protected and thus IBAMR won't compile unless deprecated code is turned back on in libMesh.

We are currently incompatible with libMesh's development branch - see https://github.com/libMesh/libmesh/issues/2883.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/FEMechanicsBase.cpp
+++ b/src/IB/FEMechanicsBase.cpp
@@ -579,7 +579,7 @@ FEMechanicsBase::doInitializeFEData(const bool use_present_data)
         // Set up boundary conditions.  Specifically, add appropriate boundary
         // IDs to the BoundaryInfo object associated with the mesh, and add DOF
         // constraints for the nodal forces and velocities.
-        const MeshBase& mesh = equation_systems.get_mesh();
+        MeshBase& mesh = equation_systems.get_mesh();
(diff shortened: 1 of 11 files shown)


src/IB/FEMechanicsBase.cpp
```diff
@@ -579,7 +579,7 @@ FEMechanicsBase::doInitializeFEData(const bool use_present_data)
         // Set up boundary conditions.  Specifically, add appropriate boundary
         // IDs to the BoundaryInfo object associated with the mesh, and add DOF
         // constraints for the nodal forces and velocities.
-        const MeshBase& mesh = equation_systems.get_mesh();
+        MeshBase& mesh = equation_systems.get_mesh();
```
boyceg, line 582: Do we need to lose the `const` here?

boyceg (approved):

src/IB/FEMechanicsBase.cpp
drwells: Yes, because we call `boundary_info.add_node(node, dirichlet_bdry_ids)` below.

Issue #1235, opened 2021-04-05
Refactor out the scratch hierarchy from IBFEMethod


drwells:
This might be useful to have if we want to scale up `IBMethod` - it would also be nice for writing more classes like IBFEMethod (e.g., the deal.II equivalent). Cleaning up this interface will also make #1226 easier to understand since it will be clearer where the data goes with an explicit interface (instead of IBFEMethod owning everything.

drwells:
@boyceg not urgent, but hopefully you know the answer - one of the next things to do is to calculate the workload, store it in the *primary* hierarchy, regrid the *primary* hierarchy, and then copy it over to the secondary hierarchy. Hence, a question: how do we add variables that are going to be regrided? I read through `HierarchyIntegrator::registerVariable()` but it isn't explicitly set anywhere.

boyceg:
[This version of `HierarchyIntegrator::registerVariable()`](https://ibamr.github.io/IBAMR-docs/ibamr/html/class_i_b_t_k_1_1_hierarchy_integrator.html#a2b2ce04d072210438e34aab7de9f97e1) is intended to be used for persistent variables (i.e. that are automatically transferred during a regridding operation).

drwells:
That works inside `HierarchyIntegrator` but I cannot figure out how to do it independently of our own code: do you know what exactly triggers the association?

Some things in particular that don't work:
- registering for restart
- calling `grid_geom->lookupCoarsenOperator` and `grid_geom->lookupRefineOperator` to do some kind of background registration
- either regridding just the coarsest level or all finer levels

i.e., we can do the regrid independently of everything else in IBTK (good) but it doesn't propagate data. See https://github.com/drwells/IBAMR/blob/088d5ec143fe93ce07b61c02e71cb8f514fbf9bf/tests/IBTK/scratch_hierarchy_01.cpp for my attempts.

drwells:
In all cases, patch level data is no longer allocated after that level is regridded.

boyceg:
You need to allocate and communicate the data you want to retain after regrid in an implementation of `initializeLevelData()`. See `HierarchyIntegrator::initializeLevelData()`.

boyceg:
This function is from `StandardTagAndInitStrategy`.

boyceg:
Basically, when you regrid, you need to have a list of the patch data indices that you want to retain, along with suitable transfer operators for each to handle coarsening/refining.

drwells:
[...]

PR #1243, opened 2021-04-13
Refactor scratch hierarchy 4

drwells:
More progress towards #1235 - this PR moves some initialization into that class.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1517,7 +1517,7 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
 
             // Use this class' buffer requirements when regridding
             Array<int> tag_buffer;
-            setupTagBuffer(tag_buffer, d_secondary_hierarchy->d_gridding_algorithm);
+            setupTagBuffer(tag_buffer, d_secondary_hierarchy->getGriddingAlgorithm());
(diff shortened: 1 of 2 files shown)


src/IB/IBFEMethod.cpp
```diff
@@ -675,7 +673,7 @@ IBFEMethod::interpolateVelocity(const int u_data_idx,
         for (int ln = 0; ln <= getFinestPatchLevelNumber(); ++ln)
         {
             d_secondary_hierarchy
-                .getPrimaryToScratchSchedule(ln, u_data_idx, u_data_idx, d_ib_solver->getVelocityPhysBdryOp())
+                ->getPrimaryToScratchSchedule(ln, u_data_idx, u_data_idx, d_ib_solver->getVelocityPhysBdryOp())
```
boyceg, line 676: I'm surprised that this is clang-format approved. 😀

boyceg (approved):

src/IB/IBFEMethod.cpp
drwells: I think you're right - it looks like I didn't run it after all. I need to spend some real time getting the CI set up at some point.

drwells: That change is OK, but others were not 😆

PR #935, opened 2020-03-23
add full IIM support to IBFESurfaceMethod (part 4)

boyceg:
### IBAMR Pull Request Checklist
- [ ] Does the test suite pass?
- [x] Was clang-format run?
- [x] Were relevant issues cited? Please remember to add `Fixes #12345` to close
      the issue automatically if we are fixing the problem.
- [ ] Is this a change others will want to know about? If so, then has a
      changelog entry been added?
- [x] If new data structures have been added to a class then have they been set
      up appropriately for restarts? If so, ensure that the restart version
      number is incremented.
- [ ] Does this change include a bug fix or new functionality? If so, a new test
      or tests should be added. New tests should run quickly (less than a minute
      in release mode). If possible, an older test should gain a new option so
      that we do not need to compile more test executables.
- [x] Did you (if your account has permission to do so) set relevant labels on
      GitHub for the pull request?

PR #1244, opened 2021-04-13
IIM master commit

drwells:
Supercedes #935 et al.

This is still missing a couple of things - lets set up PRs onto this PR to get the rest of the features in. As discussed elsewhere, the plan is to rename the implementation here as IIMethod. Aside from a bug inside FEDataManager (disabled with 30925bb) everything seems to work. Here's ex12:

visit0000

This is rebased and everything compiles cleanly, works with CMake, no deprecated libMesh features, etc.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
In particular, now that we have rebased, please don't work any more on `boyceg/merge-surface-jumps-X`: enough internals have changed for various reasons over the last two years that getting this branch up to speed took several hours, so patches need to be ported to this branch (which is on top of `master`)

boyceg:
@drwells, are there things that need to be done for this PR that @AminKolah or I could help with?

AminKolah:
I'll be available if help is needed for this PR. Perhaps, it can be made simpler by focusing to merge the IIM implementation first, and then worrying about getting the rest of the features in separately.

drwells:
I updated #12345 - that's the roadmap for the whole thing. For this PR we need to get the following done:

- [ ] fix some bug where nodes are not associated with patches - i.e., delete the workaround in 30925bbf6f4eb10752f27ac415b0ad7d3a86aa50
- [ ] add tests. The example is a good start but it needs to be reconfigured to run in about 30 seconds or less. If we have more benchmarks we can also turn them into tests by dialing down the grid resolution.

boyceg:
@drwells should we make PRs into this branch?

drwells:
We can do that but I am open to whatever is easiest for everyone else. Would you prefer to push to this branch directly?

AminKolah:
For tests, we can select examples from published papers (currently on manuscript repository) that can be added to the library. Additionally, I can revive simpler tests to individually check different parts of implementation(interpolation, traction etc.). Should the examples be added as a new push to this branch? It's fine with me however we decide to do it.

drwells:
Lets just set up PRs onto this branch (which is on my fork) if its the same to everyone else.

src/IB/IIMethod.cpp
```diff
@@ -0,0 +1,4444 @@
...
+} // getFEDataManager
+
+void
+IIMethod::registerDisconElemFamilyForJumps(const unsigned int part)
+{
+    TBOX_ASSERT(!d_fe_equation_systems_initialized);
+    TBOX_ASSERT(part < d_num_parts);
+    d_use_discon_elem_for_jumps[part] = true;
+    return;
+} // registerTangentialVelocityMotion
```
AminKolah, line 235:
```suggestion
} // registerDisconElemFamilyForJumps
```

boyceg: Actually, if we are updating this anyway, I'd suggest `setJumpFEFamily` and then pass in the FE family.

Also, I think we usually set up things like `FEFamily` choices in the input file. I am not sure that we can switch the `FEFamily` after everything has been initialized. I wonder if it is better to only allow this to be done in the input file?

AminKolah: Okay, but what if this is only an attribute for certain part(s)? I think for smooth geometries we would prefer to continue using continuous Lagrange family. This function allows for registering the discontinuous FE family for a certain part in the main file.

boyceg: Fair enough. Is there logic to confirm that the settings are being changed prior to initializing everything?

Does it make sense to allow users to pass in the `FEFamily`?

I guess the alternative is to let users specify the continuous and discontinuous `FEFamily` choices in the input file, and then switch between them for a given part. Suggest something like `setJumpConditionFEFamily{Continuous,Discontinuous}`.

AminKolah: So currently there is an assertion to confirm that FE equation systems are not initialized prior to registering with parts. Additionally, the user passes the following in the input file which only gets activated if the registration is also done with the part in the main file, otherwise everything is done using regular Lagrange family:
```
   pressure_jump_fe_family = "L2_LAGRANGE"
   velocity_jump_fe_family = "L2_LAGRANGE"
   wss_fe_family = "L2_LAGRANGE"
   tau_fe_family = "L2_LAGRANGE"
```
I realize though that this can actually be done more neatly all in `registerDisconElemFamilyForJumps` maybe

drwells: These are fixes that are worth doing but they are beyond the scope of this PR - could you open a separate issue for this so we can follow up later?

AminKolah: done.

drwells:
I rebased and fixed a few other minor things, so if anyone is working off this branch they will need to update them.

drwells:
All right, this is now:
- rebased
- working with restarts
- working with the latest libMesh release
- working in parallel (the node distribution bug is fixed)
- has an example
- has tests

i.e., ready to merge. We should ultimately add more examples but what we have now is good for starters.

AminKolah (approved):

drwells:
Thanks! These are all fixed so I will merge.

boyceg:
💥

drwells:
Hopefully we will have many more interface methods added in the future!

PR #1246, opened 2021-04-15
Move the SecondaryHierarchy to its own file in IBTK.

drwells:
Yet another part of #1235. We're getting closer - the main things left now are adding some more accessor functions and then cleaning up the load balancing logic (which will put a big dent in #1126, since we will avoid calculating workloads two more times).

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/SecondaryHierarchy.h
+++ b/ibtk/include/ibtk/SecondaryHierarchy.h
@@ -1,0 +1,207 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2021 - 2021 by the IBAMR developers
+// All rights reserved.
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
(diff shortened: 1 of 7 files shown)


ibtk/include/ibtk/SecondaryHierarchy.h
```diff
@@ -0,0 +1,207 @@
...
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace IBTK
+{
+/**
+ * Class encapsulating the idea of a secondary hierarchy. In practice, this
+ * is an equivalent (same index space) hierarchy to a given PatchHierarchy,
```
boyceg, line 46: Is it equivalent in the sense that corresponding levels cover the same region?

boyceg (approved):

ibtk/include/ibtk/SecondaryHierarchy.h
drwells: yup!

drwells: My intent with that statement was to state that this should be the same patches in a different order.

boyceg: Is that what we actually want foe the typical use case?

PR #1256, opened 2021-05-17
Refactor scratch hierarchy 7

### LINKED ISSUE #1235: Refactor out the scratch hierarchy from IBFEMethod

drwells:
Fixes #1235. This patch is longer than the others since it moves both tagging and load balancing into the SecondaryHierarchy (which requires that IBFEMethod no longer do those things in some contexts).

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/SecondaryHierarchy.h
+++ b/ibtk/include/ibtk/SecondaryHierarchy.h
@@ -138,8 +138,10 @@ protected:
     int d_finest_patch_level_number;
 
     /**
-     * Tag strategy. Always does nothing since this class only repartitions and
-     * does not refine or coarsen cells.
+     * Tag strategy. Since this class does not have any notion of refinement or
+     * coarsening this object will simply tag cells in such a way that we end up
+     * with exactly the same refinement pattern on the secondary hierarchy as
+     * the primary hierarchy.
      */
     std::unique_ptr<SAMRAI::mesh::StandardTagAndInitStrategy<NDIM> > d_tag_strategy;
 
--- a/ibtk/include/ibtk/SecondaryHierarchy.h
+++ b/ibtk/include/ibtk/SecondaryHierarchy.h
@@ -75,7 +75,8 @@ public:
     void reinit(int coarsest_patch_level_number,
                 int finest_patch_level_number,
                 SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > patch_hierarchy,
-                const SAMRAI::tbox::Array<int>& tag_buffer);
+                const SAMRAI::tbox::Array<int>& tag_buffer,
+                int workload_idx);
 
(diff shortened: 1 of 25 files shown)


ibtk/src/utilities/SecondaryHierarchy.cpp
```diff
@@ -40,19 +42,100 @@
...
+                TBOX_ASSERT(check_box == finer_box);
+            }
+
+            for (SAMRAI::hier::PatchLevel<NDIM>::Iterator p(level); p; p++)
+            {
+                const Pointer<Patch<NDIM> > patch = level->getPatch(p());
+                const Box<NDIM> box = patch->getBox();
+                Pointer<CellData<NDIM, int> > tag_data = patch->getPatchData(tag_index);
+
+                // SAMRAI's intersection code is wrong and returns nonsense
```
boyceg, line 113: Really??

ibtk/include/ibtk/SecondaryHierarchy.h
```diff
@@ -148,6 +137,12 @@ struct SecondaryHierarchy
      */
     int d_finest_patch_level_number;
 
+    /**
+     * Tag strategy. Always does nothing since this class only repartitions and
```
boyceg, line 141: Can we omit it?

ibtk/src/utilities/SecondaryHierarchy.cpp
drwells: Unfortunately yes:
```cpp
template<int DIM> inline
Box<DIM> Box<DIM>::operator*(const Box<DIM>& box) const
{
   Box<DIM> both(*this);
   both.d_lo.max(box.d_lo);
   both.d_hi.min(box.d_hi);
   return(both);
}
```
this produces invalid boxes if the two boxes don't actually intersect. This causes some loops to behave very weirdly inside SAMRAI itself.

drwells: I don't think SAMRAI has a notion of an empty box, but it should create one here.

ibtk/include/ibtk/SecondaryHierarchy.h
drwells: I should reword this comment - it doesn't really do nothing, it just tags cells to exactly duplicate the tagging done to create the primary hierarchy.

drwells: Fixed!

PR #1259, opened 2021-05-24
Clarify that we need fPIC with SAMRAI.

### LINKED ISSUE #1258: CMake build fails to link unless SAMRAI is compiled with -fPIC
When compiling IBAMR, I get an error linking with SAMRAI. The error messages look like
```
relocation R_X86_64_PC32 against symbol `_ZN6SAMRAI4tbox12InputManager18s_manager_instanceE' can not be used when making a shared object; recompile with -fPIC
```
Recompiling SAMRAI with `-fPIC` fixes the issue.

drwells:
Normal checklist isn't relevant.

Fixes #1258, unless something odd is happening: @abarret this is only with CMake, correct?

### DIFF
--- a/doc/cmake.md
+++ b/doc/cmake.md
@@ -29,6 +29,11 @@ A few things are now possible with CMake that were not possible before:
 
 ## How to use the CMake build system
 
+By default, IBAMR will compile itself with shared libraries. *This requires that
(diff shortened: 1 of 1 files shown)


abarret:
Yes that's correct.

abarret (approved):

PR #1261, opened 2021-05-27
Child integrators are now traversed correctly in regridHierarchy().

### LINKED ISSUE #1260: Children integrators are not properly traversed by parent integrators
When an implementation of `INSHierarchyIntegrator` has an `AdvDiffHierarchyIntegrator` registered and is also registered with `IBHierarchyIntegrator`, the `AdvDiffHierarchyIntegrator` is not seen as a child of `IBHierarchyIntegrator`. Functions such as `AdvDiffHierarchyIntegrator::regridHierarchyEndSpecialized()` (and others) are not called.

This can be fixed in some ways by registering a callback function with the topmost hierarchy integrator, but that does not work with every function currently (`regridHierarchyEndSpecialized()` in particular) and is cumbersome if you need to register multiple callbacks.

The issue seems to be that `HierarchyIntegrator` is only set up for a tree of height one. We could extend this to build a proper tree.

abarret:
Fixes #1260. This fixes `HierarchyIntegrator` so that it properly traverses the tree of child integrators during `regridHierarchy`. I still need to add a test.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/minor/20210527AaronBarrett
+++ b/doc/news/changes/minor/20210527AaronBarrett
@@ -1,0 +1,4 @@
+Fixed: IBTK::HierarchyIntegrator now properly traverses child integrators
+during regridHierarchy().
+<br>
+(Aaron Barrett, 2021/05/27)
--- a/ibtk/src/utilities/HierarchyIntegrator.cpp
+++ b/ibtk/src/utilities/HierarchyIntegrator.cpp
@@ -398,10 +398,16 @@ HierarchyIntegrator::regridHierarchy()
     bool check_volume_change = !d_parent_integrator && d_hierarchy_is_initialized;
     const double old_volume = check_volume_change ? d_hier_math_ops->getVolumeOfPhysicalDomain() : 0.0;
 
-    regridHierarchyBeginSpecialized();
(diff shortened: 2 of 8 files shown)


boyceg (approved):

boyceg:
Looks good to me. I wonder if the pattern of looping over all of the child integrators is something that could / should be abstracted out, or if a list of all of the integrators is something that should be maintained as a state variable.

abarret:
> Looks good to me. I wonder if the pattern of looping over all of the child integrators is something that could / should be abstracted out, or if a list of all of the integrators is something that should be maintained as a state variable.

I considered this. I'm not sure if it's worth it to maintain as a state variable, but we could include a function to gather all the integrators. Gathering the children only appears four different times in `HierarchyIntegrator`. Usually, the children are handled recursively, but there are four times where it's not really feasible to handle it recursively, so we need the list.

There aren't very many integrators. I don't see at the moment anyone using more than three.

abarret:
I'm going to go ahead and merge this. If we start adding more `HierarchyIntegrator`s then we can revisit how we handle the tree.

PR #1262, opened 2021-05-28
Avoid printing the initial workload.

drwells:
For reasons already documented this isn't portable before timestepping.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

Followup to #1256 - I forgot that printing out the initial workload isn't portable, so we need an extra check for that.

@abarret thanks again for finding this. Could you please double-check that all these tests pass?

### DIFF
--- a/src/IB/IBFEMethod.cpp
+++ b/src/IB/IBFEMethod.cpp
@@ -1528,7 +1528,14 @@ void IBFEMethod::endDataRedistribution(Pointer<PatchHierarchy<NDIM> > /*hierarch
         d_F_IB_vecs->reinit();
         if (d_Q_IB_vecs) d_Q_IB_vecs->reinit();
 
-        if (d_use_scratch_hierarchy && d_do_log)
+        // This condition is complex - we only want to log the workload if we are
+        // - logging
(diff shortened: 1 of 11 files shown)


abarret:
Now `explicit_ex4_2d.displace.mpirun=4` fails because the INS solver's residual is a slightly different zero during the first timestep. I don't know why it's suddenly failing now and not before.
```
----------------
##30      #:7   <== 1.7552086389515313075e-15
                ==> 1.7652553544022779834e-15
@ Absolute error = 1.0046715451e-17, Relative error = 5.7239437112e-3
----------------
##31      #:7   <== 1.9063795106268790827e-15
                ==> 1.6805202181334061029e-15
@ Absolute error = 2.2585929249e-16, Relative error = 1.3439843809e-1
```

abarret:
Nevermind. I reran it and it passed. I don't know what happened? I think my machine is haunted.

abarret (approved):

drwells:
Those residuals aren't a problem since the absolute tolerance limit ignores them.

PR #1263, opened 2021-05-28
Remove an extra data redistribution.

drwells:
Since the IBStrategy object isn't set up at this point this doesn't do anything.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg (commented):
I think this might be necessary for parallel data initialization to finish with `IBMethod` / `LDataManager`.

drwells:
Good point - I'll see if any examples run differently with this change.

boyceg:
@drwells, did you ever determine if this is the right/wrong thing to do?

drwells:
No - I should try this again when I have time. I think its OK. We have better test coverage now too which should help make it easier to figure out.

boyceg:
@drwells : did you get a chance to check this out?

On the one hand, it would be good to eliminate this if it really isn't necessary. On the other hand, it only happens at the first time step, and so would seem to hardly make a difference in any practical computation.

boyceg:
@drwells can we re-run tests to see if this causes any problems?

drwells:
Sure - I rebased and I'm running things locally. I'm pretty sure this is fine.

boyceg (approved):

drwells:
We've also written a lot more tests over the last year (and the CI)

drwells:
Let me rebase and see what happens.

drwells:
All the tests pass locally for me. Lets make sure the CI is happy and merge.

Issue #1203, opened 2021-01-11
Processor-dependent Brinkman results


drwells:
There's something off with the `bp_adv_diff_02_2d.mpirun=8.input` test: I get very different results on longleaf with mvapich2. On my desktop, if we use `N` processes then the output keeps changing:
```
N = 2                                   
Error in q (only in the fluid region) :
  L1-norm:  0.09483302851
  L2-norm:  0.02611110191
  max-norm: 0.04508115626
N = 4
Error in q (only in the fluid region) :
  L1-norm:  0.08363466228
  L2-norm:  0.02390712478
  max-norm: 0.04407465121
N = 6
Error in q (only in the fluid region) :
  L1-norm:  0.09230859941
  L2-norm:  0.02609431656
  max-norm: 0.04450106521
N = 8
Error in q (only in the fluid region) :
  L1-norm:  0.09049706675
  L2-norm:  0.02532477419
  max-norm: 0.04455027832
N = 10
Error in q (only in the fluid region) :
  L1-norm:  0.1089298953
  L2-norm:  0.03024700779
  max-norm: 0.04547011798
N = 12
Error in q (only in the fluid region) :
  L1-norm:  0.1117547867
  L2-norm:  0.03154663636
  max-norm: 0.0419741536
N = 14
Error in q (only in the fluid region) :
  L1-norm:  0.1089298953
  L2-norm:  0.03024700779
  max-norm: 0.04547011798
N = 16
Error in q (only in the fluid region) :
  L1-norm:  0.1121401814
  L2-norm:  0.03050099456
  max-norm: 0.0431874644
```
We aren't using AMR so these values should not change as we vary `N`, hence I'm pretty sure this is a bug - @Raamkrishnan can you take a look?

drwells:
@abarret pointed out to me earlier that the error values computed with a single processor are 100x smaller than those computed with multiple processors - however, he still gets different results even in serial. This is definitely a bug. @Raamkrishnan *ping*

PR #1265, opened 2021-05-30
Filling ghost cell in demarcateBrinkmanZone function

Raamkrishnan:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

@drwells @amneetb I am filling ghost cells for level set in demarcateBrinkmanZone function in ```BrinkmanPenalizationRigidBodyDynamics``` class. I modified the tests/wave_tank/nwt_cylinder.mpirun=8.output also. Please take a look.


amneetb (approved):
@drwells : This was my fault, wherein I was doing ghost cell filling optimization in two routines of BPRigidBodyDyanmics, which ended up being an order-dependent thing.

src/IB/BrinkmanPenalizationRigidBodyDynamics.cpp
```diff
@@ -367,14 +367,29 @@ BrinkmanPenalizationRigidBodyDynamics::demarcateBrinkmanZone(int u_idx, double t
...
+    const int ls_solid_idx = var_db->mapVariableAndContextToIndex(d_ls_solid_var, d_adv_diff_solver->getNewContext());
     const int ls_scratch_idx =
         var_db->mapVariableAndContextToIndex(d_ls_solid_var, d_adv_diff_solver->getScratchContext());
     Pointer<PatchHierarchy<NDIM> > patch_hierarchy = d_adv_diff_solver->getPatchHierarchy();
     int finest_ln = patch_hierarchy->getFinestLevelNumber();
     Pointer<PatchLevel<NDIM> > finest_level = patch_hierarchy->getPatchLevel(finest_ln);
+
+    // Ghost fill the level set values.
+    typedef HierarchyGhostCellInterpolation::InterpolationTransactionComponent InterpolationTransactionComponent;
+    std::vector<InterpolationTransactionComponent> phi_transaction_comps(1);
```
boyceg, line 380: No need to change, but for this kind of code, initialization lists can make things slightly more compact:
```
    std::vector<InterpolationTransactionComponent> phi_transaction_comps = { ... }
```

boyceg:
If there isn't one already, would it make sense to add a test case for a single processor run as well?

@Raamkrishnan, please also finish up the other items in the checklist before this is merged --- if nothing else, I think that there probably should be a changelog entry for this since it is a bug fix.

drwells:
Thanks for the fix. I'm surprised that this doesn't seem to make a difference for the other tests that use Brinkman penalization.

boyceg:
> Thanks for the fix. I'm surprised that this doesn't seem to make a difference for the other tests that use Brinkman penalization.

This might be the only test that exercises that part of the code. I think that the other Brinkman-related tests are for concentration fields.

amneetb:
> > Thanks for the fix. I'm surprised that this doesn't seem to make a difference for the other tests that use Brinkman penalization.
> 
> This might be the only test that exercises that part of the code. I think that the other Brinkman-related tests are for concentration fields.



> > Thanks for the fix. I'm surprised that this doesn't seem to make a difference for the other tests that use Brinkman penalization.
> 
> This might be the only test that exercises that part of the code. I think that the other Brinkman-related tests are for concentration fields.

Yes.

(We verified with our WEC simulations that the new and old results are pretty much the same)

Raamkrishnan:
I have added a new test for a single processor. The output for both 8 processors and a single processor is the same. I have also added a changelog. Please take a look.

amneetb:
Thanks! Looks good.

boyceg:
Does this fix #1203?

Raamkrishnan:
No. Issue #1203 is related Brinkman advection-diffusion test case which is not related to this bug. BrinkmanPenalizationRigidBodyDynamics class is used to impose solid velocity for the momentum equation.

drwells:
Thanks for adding the serial test - its good to see that we get exactly the same output with 1 or 8 processors.

drwells (approved):

PR #1283, opened 2021-07-08
Silence some warnings if Silo isn't present.

drwells:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/tests/IB/explicit_ex0.cpp
+++ b/tests/IB/explicit_ex0.cpp
@@ -42,9 +42,9 @@
 #include <ibtk/muParserRobinBcCoefs.h>
 
 // Set up application namespace declarations
-#include <ibamr/app_namespaces.h>
-
 #include <array>
+
+#include <ibamr/app_namespaces.h>
 
 int finest_ln;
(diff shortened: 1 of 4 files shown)


tests/IB/explicit_ex0.cpp
```diff
@@ -42,10 +42,10 @@
 #include <ibtk/muParserRobinBcCoefs.h>
 
 // Set up application namespace declarations
-#include <ibamr/app_namespaces.h>
-
 #include <array>
 
+#include <ibamr/app_namespaces.h>
```
boyceg, line 47: Should we do a pass of `clang-format` before the next release?

boyceg (approved):

tests/IB/explicit_ex0.cpp
drwells: Yeah, definitely - it's in the release tasks document.

We should also set up another CI runner to error out if clang-format complains because things like this sometimes slip through, but I want to get the test suite running first.

PR #1285, opened 2021-07-12
Test build on pull-request and push

vibraphone:
### Summary

This adds a github workflow that runs on pull-requests and pushes to the `master` branch of IBAMR (and only the master branch). Each PR or push to master runs a build of IBAMR inside a docker container. We use sccache and github's cache action to accelerate compilation of unchanged files. Tests are compiled but not run yet due to an issue with at least one hanging test. The docker image used for testing is [here](https://hub.docker.com/repository/docker/dcthomp/ibamr) and generated by [this](https://gitlab.kitware.com/dcthomp/ibamr-docker/-/blob/master/0.9/build-deps/Dockerfile) docker file. Ideally, these images would be built by github actions themselves.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


.github/test-build
```diff
@@ -0,0 +1,57 @@
...
+mkdir /bld
+cd /bld
+echo "::group::Configure"
+cmake \
+  -DCMAKE_C_COMPILER_LAUNCHER:STRING=sccache \
+  -DCMAKE_CXX_COMPILER_LAUNCHER:STRING=sccache \
+  -DCMAKE_Fortran_COMPILER_LAUNCHER:STRING=sccache \
+  -DCMAKE_C_FLAGS="-O3 -fPIC -march=native" \
+  -DCMAKE_CXX_FLAGS="-O3 -fPIC -march=native" \
+  -DCMAKE_Fortran_FLAGS="-O3 -fPIC -march=native" \
```
drwells, line 29: Why was adding `-fPIC` necessary? The default for CMake is to build with shared libraries (so this flag should always be on by default).

.github/test-build
```diff
@@ -0,0 +1,57 @@
...
+echo "::endgroup::"
+echo "::group::Test install"
+time make -j4 install
+echo "::endgroup::"
+echo "::group::Build tests"
+time (cd tests; make -j4 tests)
+sccache -s
+echo "::endgroup::"
+echo "::group::Run tests"
+echo "Tests disabled until hanging tests are fixed."
```
drwells, line 51: I'll add a timeout to `attest` so we can get this in.

.github/test-build
vibraphone: It may not be... I have pushed a new version without it (and the build is running now). It was unclear whether SAMRAI or IBAMR needed the flag, so I added it to both.

drwells:
Compiling with `-O3 -march=native` is nice since the tests will run faster and we will catch some weird compiler vectorization bugs, but it is significantly slower than `O1` or `O2`. Do you have any intuition for which optimization level is useful for something like this?

Unfortunately, some parts of the test suite have extremely bad behavior in debug mode (like a test will take 20x longer) so we do need some optimization level enabled.

.github/test-build
drwells: SAMRAI definitely needs it (and thats in the docker images) but we should not.

vibraphone:
I had omitted optimization at first since I was aiming only to test compilation and it goes slower with optimization enabled. If `-O1` works around bad behavior, I would suggest it to [keep compile time at a minimum](http://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html). I'm happy to put in `-O3` to get this merged.

Adding `-march=native` should not be a problem as long as we run the tests in the same job as the build.

.github/test-build
drwells: Fixed by #1287.

vibraphone: I've rebased to catch the fix, enabled the tests, and pushed; we'll see what happens with the tests.

drwells: We can remove that last `echo` now that tests are back on.

.github/test-build
```diff
@@ -0,0 +1,57 @@
...
+time ${src}/attest \
+  --mpi-executable `which mpiexec` \
+  --numdiff-executable /numdiff/bin/numdiff \
+  --test-directory=/tests \
+  -j2
```
drwells, line 56: Lets use `-j4` here.

I don't know how the very parallel MPI tests will work - can we requisition more cores?

drwells (approved):
I have two more comments but overall this looks great. We should merge soon so we can try it out.

There is a lot going on in the github workflow script that I don't understand. Can I send you a list of questions? I will also check out GitHub's own documentation.

.github/test-build
vibraphone: Oops, I'll fix it after the action runs to completion on the [other PR](https://github.com/vibraphone/IBAMR/runs/3049643028?check_suite_focus=true)...

drwells: Oh, cool, it's actually running somewhere!

vibraphone:
@drwells I'm happy to field questions about it, especially while it's still in my mind.

.github/test-build
vibraphone: Yes, if you look at the actions tab in my fork, you can see the long string of failures it took to get here. 😄

boyceg:
Since the builds and tests are so slow, should we try to set up our own physical CI server? (We actually have one that could be used, but it is not currently provisioned.)

drwells:
Thanks. I'll dump them here.

1. `sccache` is 'shared ccache' - where do these objects end up living since the test setup is for the most part ephemeral?
2. How did you pick the key for the cache? Also - what does it do?
3. Similarly - what are restore keys?
3. What does the `@v2` signify?
4. Do we really want to turn on running the CI for pushes to master? Where would we view the results in that case?

vibraphone:
@boyceg It would be nice to have a faster CI server, but [be aware of the security issues](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#self-hosted-runner-security-with-public-repositories). I'm not sure whether randomly-submitted PRs would get to run code on your machine. While they would run in a container... nothing is guaranteed.

drwells:
Unfortunately this PR takes us further away from running our own CI - in particular we would have to find some free version of GitHub's CI and run that locally in order to run the scripts here on our own hardware.

vibraphone:
@drwells No, the link in my reply to Boyce makes it clear Github allows you to add your own self-hosted runners. I believe you register them with the repo and tell it to use them instead of Github's runners.

drwells:
Like I mentioned on Friday, for deal.II we do have some stuff running on custom hardware (it's in Timo's office at Clemson) but I don't know how that is set up to talk to the rest of our GitHub CI. In particular, you can see a bunch of octocats and then three timos at the end of our CI status check on PRs, so integrating them *is* possible, it's just not what this PR is doing.

vibraphone:
> * `sccache` is 'shared ccache' - where do these objects end up living since the test setup is for the most part ephemeral?

[Here](https://github.com/IBAMR/IBAMR/pull/1285/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R24-R37) you can see the github action that populates the `/sccache` directory at the start of the run and – if the build is successful – copies the contents from `/sccache` at the end of the run to permanent storage associated with the given key.
>
> * How did you pick the key for the cache? Also - what does it do?

The key's purpose is to differentiate caches so the most appropriate one (most likely to generate cache hits) can be retrieved. The key name includes the platform/architecture, the purpose (to hold object files from sccache), the reference (pull request or branch name), and a hash of 2 include files that, if changed, will invalidate most of the cache. The caches are LRU and cleaned after disuse for 7 days.

> * Similarly - what are restore keys?

The restore keys are a list of strings to match if an exact key name doesn't exist. They are ordered from most specific to least since the order is the order in which they are fetched (if they exist).

> * What does the `@v2` signify?

That is a tag name on the repo that holds the github action's source. That way actions can change their API without breaking the builds of repos that use them.
 
> * Do we really want to turn on running the CI for pushes to master? Where would we view the results in that case?

Yes. The Actions tab will have them. It will let you compare test results on PRs to what's currently in master and verify that a merged PR didn't cause a failure.

boyceg:
Seems like self-hosted runners with GitHub Actions would be a bad idea: https://stackoverflow.com/questions/64553739/how-to-prevent-github-actions-workflow-being-triggered-by-a-forked-repository-ev

.github/test-build
drwells: It looks like we can only get 2 cores from GitHub so lets leave this as-is for now.

drwells:
The version of this over on vibraphone/IBAMR completed successfully. Lets merge this and I'll put up another PR to quickly test it out.

abarret:
> Seems like self-hosted runners with GitHub Actions would be a bad idea: https://stackoverflow.com/questions/64553739/how-to-prevent-github-actions-workflow-being-triggered-by-a-forked-repository-ev

If we want to use a self-hosted runner, it's possible to set up the action to only execute when e.g. a label is applied to a pull request. I believe we can set it up (if it's not already set up) to only allow labels to be applied by those with certain privileges. 

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

boyceg:
One of the issues, as I understand it, is that forks can also use the same actions. I may be misunderstanding, but I think that anyone who forked the repo would be able to run the actions.

vibraphone:
Surely there's some shared secret used to authenticate to UNC's runner? Would it be shared with forks?

vibraphone:
One alternative might be to make a private IBAMR repo that mirrors the public one and do at least some of the PR workflow on it? Is there a way to mirror some subset of PRs to a fork?

boyceg:
> Surely there's some shared secret used to authenticate to UNC's runner? Would it be shared with forks?

I don't know. I was doing some reading yesterday that made it seem like it is hard to securely use private runners with public repos. It seems hard to imagine there is not a solution. 🤷‍♂️

boyceg:
Looks like this is possible: https://github.com/actions/runner/issues/631

boyceg:
Some other references:
* https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/
* https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

boyceg:
* https://github.blog/changelog/2020-12-15-github-actions-environments-environment-protection-rules-and-environment-secrets-beta/
* https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

boyceg:
> One alternative might be to make a private IBAMR repo that mirrors the public one and do at least some of the PR workflow on it? Is there a way to mirror some subset of PRs to a fork?

For CI, it seems useful to have any failed tests show up on the public repo, so that it is clear that merging is blocked because of a failed test.

It would be even better if logs are publicly available, so that folks can see what needs to be done to fix things. We never achieved that with the Jenkins CI server --- only authorized users could see the logs to see why things failed.

vibraphone:
I started working on a branch that uses cdash now that we have a dashboard to submit to. The logs will appear there. Example: https://open.cdash.org/index.php?project=IBAMR&date=2021-07-13

There's a few more things that need to happen before things get submitted regularly; tests should be up on the dashboard some time next week.

PR #1286, opened 2021-07-12
Remove some unused includes.

drwells:
These are used by the base class only. I think they were added by iwyu.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg (approved):
Is it worth applying IWYU in an effort to reduce compilation time? (https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md)

drwells:
Theoretically yes, but in practice running IWYU was such a painful experience (see #917 and #920) that I would really rather not bother with it ever again.

boyceg:
I guess it might be worth trying again with CMake? (https://github.com/include-what-you-use/include-what-you-use#using-with-cmake)

drwells:
Its worth a try, but I don't think that will fix the more fundamental problems of IWYU (like including private PETSc headers, internal compiler headers, etc). Trying to minimize the header count just does weird things.

PR #1289, opened 2021-07-12
Attest cleanup

drwells:
Lets see what the CI does!

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/attest
+++ b/attest
@@ -1,16 +1,16 @@
 #!/usr/bin/env python3
-## ---------------------------------------------------------------------
-##
-## Copyright (c) 2019 - 2020 by the IBAMR developers
-## All rights reserved.
-##
-## This file is part of IBAMR.
-##
-## IBAMR is free software and is distributed under the 3-clause BSD
-## license. The full text of the license can be found in the file
-## COPYRIGHT at the top level directory of IBAMR.
-##
-## ---------------------------------------------------------------------
+# ---------------------------------------------------------------------
+#
+# Copyright (c) 2019 - 2020 by the IBAMR developers
+# All rights reserved.
+#
+# This file is part of IBAMR.
+#
+# IBAMR is free software and is distributed under the 3-clause BSD
+# license. The full text of the license can be found in the file
+# COPYRIGHT at the top level directory of IBAMR.
+#
+# ---------------------------------------------------------------------
 
 """Test runner for autotools-based tests.
 """
@@ -28,7 +28,6 @@ import subprocess
 import sys
 import tempfile
 import threading
-import time
 
 # We rely on subprocess.run, which is new in 3.5
 assert sys.version_info >= (3, 5)
@@ -43,6 +42,8 @@ def interrupt_handler(sig, frame):
     del sig, frame
     print("SIGINT caught. Exiting.")
     sys.exit(1)
+
+
 signal.signal(signal.SIGINT, interrupt_handler)
 
 
@@ -51,8 +52,8 @@ class Parameters:
 
     Parameters read from input arguments or the input file:
 
-    test_directory: relative path to the directory containing tests. Defaults to
-                    "/tests/".
+    test_directory: relative path to the directory containing tests. Defaults
+                    to "/tests/".
 
     mpiexec: program used to launch MPI applications. Defaults to "mpiexec".
 
@@ -73,9 +74,11 @@ class Parameters:
 
     test_timeout: maximum length, in seconds, to run a test. Tests that do not
     finish in time are considered as failing.
+
+    verbose: If verbose is True then, if a test fails, the first couple of
+    lines of stderr or the failing diff will be printed to stdout.
     """
     def __init__(self, input_arguments):
-        # TODO: finish implementing verbose mode
         self.keep_work_directories = input_arguments.keep_work_directories
         self.mpiexec = input_arguments.mpiexec
         self.numdiff = input_arguments.numdiff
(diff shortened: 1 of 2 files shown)


drwells:
Hm, actually running the test suite has surprisingly poor performance. Let me investigate further before we merge.

vibraphone:
@drwells It looks like you may need to kill an action that is still running? I don't think old actions are killed when new ones are created...

drwells:
I think you're right. Let me manually kill any old ones.

drwells:
Maintaining my own test runner is expensive but it's going to be necessary to keep up while we support autotools. I added a verbose mode - if you look at some of the old runs the majority of tests fail and I don't see why some would pass and others would fail. My best guess is that we are underlinked.

drwells:
We get a lot of illegal instruction errors - that's our answer for `-march=native`.

drwells:
@vibraphone we need to enable tirpc in the docker images so that we get XDR support. It's not documented but our libMesh tests rely on that for restart data.

drwells:
We have tests running but the output for a lot of them is completely wrong. For example:
```
          IBFE/interpolate_velocity_01_2d.nodal_quadrature.b.mpirun=4.input  DIFF FAILED
  test failed with output:
      ----------------
      ##9       #:4   <== 8.7554918579035856396e-05
      ##9       #:4   ==> 1.6131799264655186654
      @ Absolute error = 1.6130923715e+0, Relative error = 1.8423777873e+4
      ##9       #:5   <== 0.00021638268562762519309
      ##9       #:5   ==> 3.0727148287267458393
      @ Absolute error = 3.0724984460e+0, Relative error = 1.4199372917e+4
```
The errors are really high in a way that doesn't make sense.

drwells:
Furthermore, some of these timeouts don't make sense:
```
          IBFE/explicit_ex4_2d.scratch_hier.levels=1.mpirun=4.input ------ TIMEOUT
```
this test runs in 1.5 seconds on my 4-core desktop computer - it doesn't seem right that the virtual machine is 40x slower than my 8 year old workstation.

vibraphone:
> @vibraphone we need to enable tirpc in the docker images so that we get XDR support. It's not documented but our libMesh tests rely on that for restart data.

@drwells OK. Do I need to pass libmesh's `./configure ...` line additional arguments or just make sure `tirpc-devel` is installed?

drwells:
I think installing the package should be enough - libMesh doesn't have an extra flag to specify an install location.

vibraphone:
> We get a lot of illegal instruction errors - that's our answer for `-march=native`.

Yes, the architecture of my linux box (that builds the 3rd party packages) is not necessarily the same as the runner that builds IBAMR. I can change the flags on the docker image (since I'm rebuilding libmesh and adding tirpc); do you have any suggestions on the flags to use? On our dashboard machines, we use `-march=core2 -mno-avx512f` if that helps.

drwells:
I don't see any `-march=native` flags in the dockerfile - I don't think we need to change any compiler flags there.

There is something weird going on with libMesh that I do not understand - no parallel libMesh test passed. I'm not sure why.

vibraphone:
> We have tests running but the output for a lot of them is completely wrong. For example ...

Have you been able to replicate by building in a local container (i.e., running `docker run -ti dcthomp/ibamr:0.9-build-deps bash` and building yourself)? If it helps, once the container is running, you can run the build script to build IBAMR.

drwells:
Not yet - that's a good idea.

vibraphone:
> I don't see any `-march=native` flags in the dockerfile - I don't think we need to change any compiler flags there.

@drwells The absence of `-march=native` in the Dockerfile doesn't mean the presumed architecture matches the github runners. We ran into that with our packaging process for modelbuilder.

> There is something weird going on with libMesh that I do not understand - no parallel libMesh test passed. I'm not sure why.

Well, as I just re-ran libmesh's configure step I see that it built without MPI. That is probably why.

drwells:
That makes a lot of sense, actually - the partitioning tests all output the global range on every processor.

I'll put up some more libMesh checks in our `CMakeLists.txt` to try and catch these errors faster.

vibraphone:
It's odd, because the Dockerfile passes mpicc/mpicxx to libmesh as the compilers to use.

drwells:
libMesh's configuration script is somewhat buggy. For example, `--enable-petsc-required` isn't guaranteed to work - based on how you install PETSc it may not find all the files it needs and not end up including PETSc even with that flag.

drwells:
Yet more reasons to switch to CMake, but you already probably agree with that sentiment :)

vibraphone:
CMake is nice, but it doesn't prevent bugs.

drwells:
Of course, but it's a lot harder to write bugs in CMake with targets et al than it is with m4 soup.

drwells:
This branch got away from its original purpose to implement some necessary fixes (add a missing configuration option, implement verbose mode, do some general tidying up) of `attest`. @boyceg could you give this a quick look and approve? I'll move the CI script changes into a new PR.

attest
```diff
@@ -1,16 +1,16 @@
...
-## This file is part of IBAMR.
-##
-## IBAMR is free software and is distributed under the 3-clause BSD
-## license. The full text of the license can be found in the file
-## COPYRIGHT at the top level directory of IBAMR.
-##
-## ---------------------------------------------------------------------
+# ---------------------------------------------------------------------
+#
+# Copyright (c) 2019 - 2020 by the IBAMR developers
```
boyceg, line 4: Do we still need to update to 2021?

boyceg (approved):

attest
drwells: the update copyright release step will take care of that.

Issue #1291, opened 2021-07-13
Document the XDR and ExodusII requirements in libMesh


drwells:
I think we require ExodusII - we definitely require XDR for restarts.

PR #1292, opened 2021-07-13
Add some more checks for libMesh features.

drwells:
We use XDR in restarts so require it. Similarly, we always use MPI so make sure we have it on.

The normal checklist doesn't apply. Additionally, this will probably not pass with the current CI since (as discussed in #1289) we managed to configure libMesh without MPI in the docker image.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -553,6 +553,43 @@ The version of PETSc detected by libMesh differs from the version of PETSc \
 detected by IBAMR. This is not allowed.")
     ENDIF()
 
+    # Verify that libMesh uses MPI:
+    CHECK_CXX_SOURCE_COMPILES(
+      "
+      #include <libmesh/libmesh_config.h>
+      #ifdef LIBMESH_HAVE_MPI
+      // OK
+      #else
+      #error
+      #endif
+      int main() {}
+      "
+      LIBMESH_WITH_MPI
+      )
+    IF(NOT "${LIBMESH_WITH_MPI}")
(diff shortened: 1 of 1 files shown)


drwells:
```

  CMake Error at CMakeLists.txt:570 (MESSAGE):
    IBAMR requires that libMesh be compiled with MPI.
  
  
  -- Performing Test LIBMESH_WITH_MPI - Failed
```
yup, we failed correctly.

drwells:
Part of #1291. We should document that we don't absolutely require ExodusII (its just in the examples) somewhere else.

CMakeLists.txt
```diff
@@ -553,6 +553,43 @@ The version of PETSc detected by libMesh differs from the version of PETSc \
...
+      int main() {}
+      "
+      LIBMESH_WITH_XDR
+      )
+    IF(NOT "${LIBMESH_WITH_XDR}")
+      MESSAGE(FATAL_ERROR "IBAMR requires that libMesh be compiled with XDR.")
+    ENDIF()
+
+    # Note that we don't technically require ExodusII - even though we have
+    # AppInitializer::getExodusIIFilename() we never actually use ExodusII APIs.
```
boyceg, line 591: Should we emit a warning, since the examples are currently set up to use ExodusII?

boyceg (approved):

CMakeLists.txt
drwells: A warning here is unlikely to be seen among all the other configuration messages. Since libMesh defaults to compiling and installing its own copy of ExodusII I don't think its worth worrying about in practice.

PR #1293, opened 2021-07-13
Improve test build

drwells:
Migrated from #1289. Depends on #1289 (we need verbose mode).

@vibraphone These changes are needed to get the test suite running - I also disabled all the very expensive tests since the virtual machine is pretty slow.

### DIFF
--- a/.github/test-build
+++ b/.github/test-build
@@ -20,13 +20,20 @@ echo "::endgroup::"
 mkdir /bld
 cd /bld
 echo "::group::Configure"
+# In some basic profiling I measured that running the test suite with 4 cores
+# takes 18 minutes with -O3 -march=native and 19 minutes with O1. Hence compile
+# C++ with lower optimization. Fortran is orders of magnitude faster to compile
+# than C++ so keep it at O3.
+#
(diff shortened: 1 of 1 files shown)


.github/test-build
```diff
@@ -1,10 +1,11 @@
 #!/bin/sh
+set -eou pipefail
 
 # Remember where we started; this is where our source checkout lives.
-src=`pwd`
+src=$(pwd)
 
 # Load the module command and use it to load MPI.
-. /etc/profile.d/modules.sh
+./etc/profile.d/modules.sh
```
drwells, line 8: I'm about 99% sure this was a typo but we should verify that.

vibraphone, line 8: This is not correct? `. script` is equivalent to `source script`. Is there even a `./etc` directory that exists?

drwells: You're right - let me fix this again.

drwells: I got this wrong (see below).

drwells:
I think this has been superceded by https://github.com/vibraphone/IBAMR/pull/2 so it doesn't need to be reviewed.

boyceg (approved):

Issue #390, opened 2018-11-10
use std::vector<T>::data() where possible


boyceg:
We get a pointer to the data in `std::vector<T> x` using the idiom `&x[0]` all over the place. In C++11, this can be replaced by the much better `x.data()`.

Let's use `std::vector<T>::data()` where possible.

boyceg:
This also allows use to remove a bunch of checks that `x.size() > 0`. E.g. it is common to see horrible code like `x.empty() ? nullptr : &x[0]` or the like. We can get rid of this too, since `x.data()` works even if `x` is empty.

boyceg:
The only thing to be on the lookout for is that `x.data()` is not guaranteed to be `nullptr` if `x` is empty --- so anywhere that we rely on the pointer being null if `x` is empty, we still need to be careful.

PR #1296, opened 2021-07-14
Use std::vector::data().

drwells:
Part of #390.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/utilities/RNG.cpp
+++ b/src/utilities/RNG.cpp
@@ -75,7 +75,7 @@ namespace IBAMR
 /* Period parameters */
 #define N 624
 #define M 397
-#define MATRIX_A 0x9908b0df /* constant vector a */
+#define MATRIX_A 0x9908b0df   /* constant vector a */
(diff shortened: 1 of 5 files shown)


src/utilities/RNG.cpp
```diff
@@ -75,7 +75,7 @@ namespace IBAMR
 /* Period parameters */
 #define N 624
 #define M 397
-#define MATRIX_A 0x9908b0df /* constant vector a */
+#define MATRIX_A 0x9908b0df   /* constant vector a */
```
boyceg, line 78: Is clang-format aligning comments now?

boyceg (approved):

src/utilities/RNG.cpp
drwells: I'm not totally positive where that one came from - I guess so?

boyceg: I like it! I used to do this stuff by hand and was sad that clang-format washed it away.

PR #1303, opened 2021-07-20
Export dependency cmake paths

### LINKED ISSUE #1302: IBAMRConfig.cmake doesn't export location of bundled Boost or Eigen
When IBAMR is built with the bundled version of Boost or Eigen, the location of the libraries is not propagated to `IBAMRConfig.cmake`. This can cause a conflict when there exists a system installed Boost or Eigen.

drwells:
Fixes #1302, and a lot of other silly CMake bugs.

Normal checklist doesn't apply.

@abarret could you give this a try?


abarret (approved):
This works for me.

PR #1304, opened 2021-07-20
Remove an exported boost preprocessor symbol.

drwells:
This is nice to have during compilation (since we do not officially require anything more than the boost headers) but it really should not be set when users use IBAMR since they may want to use other parts of boost.

An example: https://github.com/drwells/fiddle uses IBAMR + deal.II, and deal.II uses the boost serialization library which is not header-only. For reasons I don't fully understand things seem to work, but it doesn't make sense to define this symbol in this circumstance, so we just shouldn't define it and let users do whatever they think is appropriate.

As usual, the list doesn't apply.

@abarret it would help if you could try this one out too before merging.


abarret (approved):
Works for me.

PR #1309, opened 2021-07-26
CMake: Add install target for examples

kwryankrattiger:
- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ SET(IBAMR_DEPENDENCY_LINK_ARGUMENTS "")
 
 # Build tests?
 option(IBAMR_ENABLE_TESTING "Should tests be compiled and configured to run with ctest?" ON)
-
+option(IBAMR_ENABLE_EXAMPLES "Enable building examples" ON)
 # Do we want dynamic or static linking?
 OPTION(BUILD_SHARED_LIBS "Whether or not to build shared libraries." ON)
 # Build the libraries in one place:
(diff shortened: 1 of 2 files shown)


drwells (changes requested):
We cannot use cmake_path since it was added in CMake 3.20. In particular, Ubuntu 20.04 ships CMake 3.16 - we really cannot use a newer version than that, and I would prefer to stick to 3.15.

boyceg:
I'm not sure that I follow what is going on here. Does this mean that you have to decide whether or not to build examples at configure time?

drwells:
That is part of the change in this PR, though its not in the CMake build system right now. Presently examples are always set up (i.e., we unconditionally generate Makefiles for them) but you have to type `make examples` to compile them: they aren't part of the `all` target (what gets compiled when you just type `make`).

Our colleagues at kitware pointed out here and in other patches that I set things up in a more autotools-like way rather than idomatic CMake: with autotools we still write makefiles and defining separate `make tests` and `make examples` targets makes sense. With CMake, though, it is more common to do `cmake -DBUILD_EXAMPLES=ON` or `cmake -DBUILD_EXAMPLES=OFF` and then `make` will compile or not compile examples (i.e., no separate `make examples`). I like excluding these from `all` but there are good reasons to do it this way.

boyceg:
I think I would prefer to keep separate make targets for examples and tests by default, but possibly allow users to include flags at configure time to "add" examples or tests to `make all` if they want.

boyceg:
@drwells should we finish this one too?

boyceg:
@drwells: you are the CMake boss, so it is really up to you.

drwells:
In that case I will close for now, at least until we get a compelling use-case.

PR #1311, opened 2021-07-26
Check that PETSc does not use 64bit indices.

drwells:
The usual checklist doesn't apply since this doesn't change the library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -352,19 +352,38 @@ ENDFOREACH()
 #
 MESSAGE(STATUS "")
 MESSAGE(STATUS "Setting up PETSc")
-FIND_FILE(PETSC_VARIABLES_FILE petscvariables HINTS ${PETSC_ROOT} PATH_SUFFIXES conf lib/petsc/conf)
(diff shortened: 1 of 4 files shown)


drwells:
I tested this manually by putting the failing regex in `petscconf.h` myself.

boyceg:
It might not be that much work to make things work correctly with both 32- and 64-bit integers by compiling against a PETSc built with 64-bit integers and swapping `int`s for `PetscInt` as needed.

(Of course, we would need to have the CI test against both going forward to avoid regressions.)

drwells:
I don't think it's going to be that straightforward - even if we can compile and run programs there are probably limits in SAMRAI and IBAMR that we will run into only when we get into the billions of dofs range. This is a constant source of trouble in deal.II even though we have supported it for years.

drwells:
In fact, I don't think SAMRAI can handle this, since it doesn't support `long int` - we'd have to patch it if we want to use `PetscInt` for things like `constructPatchLevelDOFIndices`.

boyceg (changes requested):
Can we also do this in the legacy autotools configure script?

drwells:
Sure

drwells:
Fixed!

boyceg (approved):

PR #1312, opened 2021-07-26
Cleanup test regex

drwells:
Another minor build system patch.

I prefer to use capital letters for all built ins so that they stick out. Whether or not that is a good choice is debatable (I think its deprecated behavior) but no matter the style we should be consistent.


boyceg (approved):

drwells:
I think this might be in another PR - lets hold off on merging for now.

drwells:
I cleaned this up and added another test. This is ready for review again.

PR #1313, opened 2021-07-28
Fix child integrator test that fails in debug mode.

abarret:
Fixes a failing test pointed out in #1307. First, `d_current_cycle_num` was not being incremented correctly by the dummy integrator. This causes it to fail when `NDEBUG` is not defined. Second, the input and output files were not named correctly, so the test never ran in `attest`.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg (approved):

kwryankrattiger:
I am seeing difference in the output for this case after a rebase. Can you confirm?

[LastTest.log](https://github.com/IBAMR/IBAMR/files/6902972/LastTest.log)

abarret:
> I am seeing difference in the output for this case after a rebase. Can you confirm?
> 
> [LastTest.log](https://github.com/IBAMR/IBAMR/files/6902972/LastTest.log)

It appears that the tester has been built without support for Silo, is that correct? I'm seeing warnings about not having Silo available.

drwells:
@abarret could you apply the fix from #1283 also here?

drwells:
or just disable plotting output completely - both work.

kwryankrattiger:
SILO should be enabled for this build, everything should be turned on AFIAK. It is strange that it is printing an error for this case but none of the other tests that use SILO.

PR #1317, opened 2021-08-02
Release 0.9 cleanups

drwells:
A small number of fixes for the release.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/tests/level_set/fe_surface_distance.cpp
+++ b/tests/level_set/fe_surface_distance.cpp
@@ -186,7 +186,6 @@ main(int argc, char* argv[])
 
         // Get various standard options set in the input file.
         const bool dump_viz_data = app_initializer->dumpVizData();
-        const bool uses_visit = dump_viz_data && !app_initializer->getVisItDataWriter().isNull();
 #ifdef LIBMESH_HAVE_EXODUS_API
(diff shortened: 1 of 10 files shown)


tests/level_set/fe_surface_distance.cpp
```diff
@@ -403,7 +402,9 @@ main(int argc, char* argv[])
                 << E_interface / volume_near_interface << std::endl;
             out << "Number of points used to compute the interface error = " << num_interface_pts << std::endl;
         }
-        // Output data for plotting.
+        // Disable plotting by default in the test suite:
+#if 0
```
boyceg, line 406: Should we just remove this code?

drwells: It might be useful to turn on in the future if we want to debug something, so I would prefer to keep it.

boyceg: OK.

How about using a regular conditional instead of a preprocessor macro, so that the code block continues to get compiled even if it isn't used?

drwells: Sure.

boyceg: I will go ahead and approve now.

boyceg (approved):

PR #1327, opened 2021-08-09
Recommend disabling tests.

drwells:
We should mention how to disable tests in `cmake.md` which contains the rest of the configuration documentation for CMake.


abarret (approved):
Should we make disabled tests be the default behavior?

drwells:
Maybe, but I don't want to change that right now since the CI depends on it. I'd rather go for the simpler fix in the release.

PR #1338, opened 2021-08-13
Use consistent variables for muParser in CMake.

drwells:
Same as #1337 but for `master`.


CMakeLists.txt
```diff
@@ -210,13 +210,13 @@ ENDIF()
...
-  FIND_LIBRARY(MUPARSER_LIBRARIES NAMES muparser HINTS ${MUPARSER_ROOT}/lib /usr/lib)
-  IF("${MUPARSER_INCLUDE_DIRS}" STREQUAL "MUPARSER_INCLUDE_DIRS-NOTFOUND" OR
-      "${MUPARSER_LIBRARIES}" STREQUAL "MUPARSER_LIBRARIES-NOTFOUND")
+  FIND_PATH(muParser_INCLUDE_DIRS NAMES muParser.h HINTS ${muParser_ROOT}/include /usr/include/)
+  FIND_LIBRARY(muParser_LIBRARIES NAMES muparser HINTS ${muParser_ROOT}/lib /usr/lib)
+  IF("${muParser_INCLUDE_DIRS}" STREQUAL "muParser_INCLUDE_DIRS-NOTFOUND" OR
+      "${muParser_LIBRARIES}" STREQUAL "muParser_LIBRARIES-NOTFOUND")
     SET(IBAMR_USE_BUNDLED_muParser TRUE)
   ELSE()
     SET(IBAMR_USE_BUNDLED_muParser FALSE)
```
drwells, line 222: The bug is on line 225: we should be using `muParser` here.

It is somewhat arbitrary, but muParser should have a CMake package so I am using its proper name instead of all caps (like I do for libmesh).

CMakeLists.txt
```diff
@@ -96,7 +96,7 @@ SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
 # With modern CMake, each find module uses the environment variable ending in
 # _ROOT to look for an installation.
 SET(_required_roots "Boost_ROOT" "Eigen3_ROOT" "HDF5_ROOT" "HYPRE_ROOT"
-  "MUPARSER_ROOT" "MPI_ROOT" "PETSC_ROOT" "SAMRAI_ROOT")
+  "muParser_ROOT" "MPI_ROOT" "PETSC_ROOT" "SAMRAI_ROOT")
```
boyceg, line 99: And also PETSC ---> PETSc?

CMakeLists.txt
boyceg: What is the rationale for the difference between "proper name" and "always all caps"?

drwells: Good question. I should make this consistent. I prefer all caps but I went with `Boost` and `Eigen3` since CMake uses those names in its own code - the result only makes sense if you know which modules CMake provides.

I'll write a followup commit that changes the defaults to be all-caps but also accepts common capitalizations (`libMesh`, `PETSc`, `Eigen`, etc) as alternatives.

CMakeLists.txt
drwells: I'll fix this with the other one.

CMakeLists.txt
drwells: Fixed!

PR #1345, opened 2021-08-27
update mean velocity calculations in INSStaggeredHierarchyIntegrator

boyceg:
- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
+++ b/src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
@@ -550,10 +550,40 @@ INSStaggeredHierarchyIntegrator::INSStaggeredHierarchyIntegrator(std::string obj
     }
     d_P_bc_coef = new INSStaggeredPressureBcCoef(this, d_bc_coefs, d_traction_bc_type);
 
-    // Check to see whether to track mean flow quantities and turbulent kinetic
-    // energy.
+    // Check to see whether to track mean flow quantities, and determine the number of distinct flow snapshots to
+    // retain.
     if (input_db->keyExists("flow_averaging_interval"))
         d_flow_averaging_interval = input_db->getInteger("flow_averaging_interval");
+    if (input_db->keyExists("cyclic_flow_averaging_period"))
+    {
+        d_cyclic_flow_averaging_period = input_db->getInteger("cyclic_flow_averaging_period");
+        if (!input_db->keyExists("flow_averaging_interval"))
+        {
+            d_flow_averaging_interval = 1;
+        }
+    }
+    if (input_db->keyExists("num_flow_averaging_cycles"))
+        d_num_flow_averaging_cycles = input_db->getInteger("num_flow_averaging_cycles");
+    else
+        d_num_flow_averaging_cycles = std::numeric_limits<unsigned int>::max();
+
+    if (d_flow_averaging_interval)
+    {
+        if (d_cyclic_flow_averaging_period)
+        {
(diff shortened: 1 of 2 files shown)


xxjaehoxx:
@boyceg  - I thought the way you implemented wasn't wrong per se, but it was assuming that you have a constant time step size. Maybe I misunderstood.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -550,10 +550,40 @@ INSStaggeredHierarchyIntegrator::INSStaggeredHierarchyIntegrator(std::string obj
...
+        d_cyclic_flow_averaging_period = input_db->getInteger("cyclic_flow_averaging_period");
+        if (!input_db->keyExists("flow_averaging_interval"))
+        {
+            d_flow_averaging_interval = 1;
+        }
+    }
+    if (input_db->keyExists("num_flow_averaging_cycles"))
+        d_num_flow_averaging_cycles = input_db->getInteger("num_flow_averaging_cycles");
+    else
+        d_num_flow_averaging_cycles = std::numeric_limits<unsigned int>::max();
```
drwells, line 568: We typically put default values in the header.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -550,10 +550,40 @@ INSStaggeredHierarchyIntegrator::INSStaggeredHierarchyIntegrator(std::string obj
...
+        if (d_cyclic_flow_averaging_period)
+        {
+            d_num_flow_snapshots = d_cyclic_flow_averaging_period / d_flow_averaging_interval;
+        }
+        else
+        {
+            d_num_flow_snapshots = 1;
+        }
+    }
+    if (d_num_flow_averaging_cycles)
```
drwells, line 581:
```suggestion
    if (d_cyclic_flow_averaging_period > 0)
```

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -939,47 +958,21 @@ INSStaggeredHierarchyIntegrator::initializeHierarchyIntegrator(Pointer<PatchHier
...
-                         d_U_mean_scratch_idx,
-                         d_U_mean_var,
-                         cell_ghosts,
+        registerVariable(d_U_mean_current_idxs[k],
+                         d_U_mean_new_idxs[k],
+                         d_U_mean_scratch_idxs[k],
+                         d_U_mean_vars[k],
+                         side_ghosts,
                          d_U_mean_coarsen_type,
                          d_U_mean_refine_type);
```
drwells, line 973: I suppose that this approach is necessary but it means that regrids will continue to get slower and slower since every snapshot will be regridded.

drwells (commented):
I have some comments.

It is difficult to tell what is going on since we are
- deleting wrong code
- adding new turbulence statistics
- adding cycling

all in the same PR. If the current TKE code is wrong then we should get rid of it on `master` first.

boyceg:
> I have some comments.

I guess I am really looking for comments on the general approach at this point --- particularly, the `if`/`else` for switching between computing dynamics and interpolating from precomputed values, and the logic for handling both steady flow conditions and cyclic flow conditions. Is this the right way to be doing these things?

boyceg:
Superseded by #1408 and #1428.

PR #1349, opened 2021-08-31
Start work on CMake + SAMRAI 3.

drwells:
### *note that this PR is onto `samrai_main` and not `master`*

SAMRAI doesn't export its targets so we still need to set things up manually. The major changes are:

1. libSAMRAI.a is gone
2. libraries are dimension-independent
3. headers are now in `${SAMRAI_ROOT}/include/SAMRAI` instead of `${SAMRAI_ROOT}/include/`

A major difficulty in developing the SAMRAI 3 branch is that we cannot apply the SAMRAI 3 updates on the branch itself. If we do then it will create virtually impossible to reconcile merge conflicts with `master` when we add new features there. Hence this branch is not going to compile with either one on GitHub for awhile but we should still move forward when we can.

Another issue with SAMRAI 3 is that several of the m4 includes are gone. To avoid copyright problems I will reimplement them without looking at the originals so that we can put them unambiguously under our own copyright and license. I'll make that change on master later today.


abarret:
> Another issue with SAMRAI 3 is that several of the m4 includes are gone.

Do you mean the m4's for the Fortran arrays? They are located in `pdat/fortran`.

abarret:
Also, I'm able to link to SAMRAI with just `FIND_PACKAGE(SAMRAI)`. Are you targeting the main branch, or a specific version of SAMRAI?

drwells:
Are you working with SAMRAI 3 or a newer version?

abarret:
I'm targeting the latest release

drwells:
That explains it - I was working off of samrai 3. I'll check out the other m4 includes and figure out if we need to reimplement them or not.

drwells:
Should we target the latest release? I don't remember if we agreed on it or not.

abarret:
I thought we had decided just to target the latest release.

drwells:
We probably did and I forgot. I'll update this to target that instead.

drwells:
> > Another issue with SAMRAI 3 is that several of the m4 includes are gone.
> 
> Do you mean the m4's for the Fortran arrays? They are located in `pdat/fortran`.

I do. Those files are no longer installed AFAICT so we should reimplement them.

boyceg:
Can the original m4's be included as a patch to SAMRAI 3/4?

abarret:
> Can the original m4's be included as a patch to SAMRAI 3/4?

The original m4's are still present. They are not install targets though. To use them, we'd need to know where the source directory is located.

boyceg:
In that case, could we just patch SAMRAI's build system to make sure they are installed?

drwells:
We could put up a PR for them to include it. Honestly these files are not that complicated (and I am almost done reimplementing our own copies). Let me check to see which is easier.

drwells:
OK, adding them was pretty easy:

https://github.com/LLNL/SAMRAI/pull/178

It only took three years but I am finally writing some samrai patches!

drwells:
OK, I updated the branch to use the SAMRAI CMake modules - much nicer this way.

boyceg:
> Should we target the latest release? I don't remember if we agreed on it or not.

I don't know if we agreed on this. I think that I had advocated for jumping straight to the current release, but that @drwells had been thinking in terms of working up along the SAMRAI 3 releases until we get to the current release.

I think that the main reason to migrate first to SAMRAI 3 and then to SAMRAI 4 is that the changes will be more incremental.

I think that the main reason to migrate straight to the current SAMRAI release is that the most painful changes already show up in the transition from SAMRAI 3 to SAMRAI 4. Also, there were bug fixes and API changes/improvements that were made since the first release of SAMRAI 3, and so instead of working around those bugs and/or evolutions in the APIs as we update IBAMR, it seems like it would be less work to jump straight to the current SAMRAI release.

boyceg:
One other thought --- if there are modifications to SAMRAI that we want to propose (via PRs), it seems easiest to do those with the current SAMRAI release instead of past releases.

drwells:
Lets see how responsive they are to the present PR. The primary advantage of working off the latest SAMRAI release is that we can fix whatever needs to be fixed properly. However, that assumes they accept our patches, so the reverse is also true (if we want to fix things correctly we could be waiting a very long time). One reason to create our own copies of these `.m4` files is to retain control of the timeline.

@abarret are the scripts written for SAMRAI 4? If so we should do SAMRAI 4. Otherwise I'd rather do 3 first.

drwells:
@abarret *ping*

abarret (approved):

abarret:
> Lets see how responsive they are to the present PR. The primary advantage of working off the latest SAMRAI release is that we can fix whatever needs to be fixed properly. However, that assumes they accept our patches, so the reverse is also true (if we want to fix things correctly we could be waiting a very long time). One reason to create our own copies of these `.m4` files is to retain control of the timeline.
> 
> @abarret are the scripts written for SAMRAI 4? If so we should do SAMRAI 4. Otherwise I'd rather do 3 first.

They are written for 4. Although it shouldn't be too hard to separate them out. The scripts can do a lot of the more tedious changes, like `Pointer` to `shared_ptr`, namespace changes, removing `NDIM`. But it will be impossible to write a script for every change, partly because not all the required changes are documented.

drwells:
I still think we should go to SAMRAI 3 first - if nothing else we will get something working faster that way (the amount of time we will spend with a completely broken IBAMR will be much shorter). @abarret we need to make a decision on this. Do you agree or disagree? If you agree I'll revert this PR to https://github.com/IBAMR/IBAMR/commit/5f98ef6c0a3c90a3df37e52e4761e14eb306f4c6 and then we can split off any scripts for SAMRAI 4 into a different folder.

boyceg:
Why do you think it will be faster to start with SAMRAI 3?

drwells:
It is always less work to only jump one major revision at a time since IBAMR will be in a completely broken state until we fix every single API change. The fewer large changes we have to deal with the better.

This conversion has dragged on for years with minimal progress - we shouldn't make the transition any harder.

abarret:
I agree. Having attempted and failed one large transition several times, I think it's a good idea to try something different. 

Do we need a strategy for helping update user codes?

Should we try to jump versions as documented in the release notes in SAMRAI? Most of the scripts are pointed towards the release notes in 3.1.0-beta.

boyceg:
It might be a good idea to take a look at the release notes when picking a version to target. I think there are some things that we currently used that changed or were broken in early 3.x releases and then changed again — might be easier to pick an intermediate release.

drwells:
That's not something I anticipated so yeah, lets do it. It would certainly help if they rolled back some of the changes to periodic BCs.

@abarret how exhaustive is the list in 

https://github.com/IBAMR/IBAMR/blob/60f6474b4157d5787d819b40b2acab0e70824067/scripts/samrai3_convert/README

? That document implies that it is for SAMRAI 3, not SAMRAI 4, and it doesn't really mention which version of SAMRAI 3 it targets. If you don't have this information we should work to compile it.

abarret:
That document was originally created before SAMRAI 4 existed. I haven't updated any documentation to say SAMRAI 4.

As far as I know, there are no incompatibilities going from the last SAMRAI 3 release to SAMRAI 4. The only things they changed were RAJA/Umpire support (which are optional) and removed the autoconf build system.

abarret:
The README does document all the scripts though.

boyceg:
I think the main difference between SAMRAI 3 and 4 is replacing fortran kernels with RAJA. I don’t think there are big changes in the interface.

drwells:
The current release of SAMRAI also has proper CMake find modules (i.e., the difference between https://github.com/IBAMR/IBAMR/commit/5f98ef6c0a3c90a3df37e52e4761e14eb306f4c6 and https://github.com/IBAMR/IBAMR/commit/b2a880cbd93a63ac5818ca8c65a31210a21412b0).

drwells:
I read through the release notes for every SAMRAI 3.x version. I am now convinced that we should get this to work incrementally. In particular, 3.1 removes a lot of stuff, 3.2 requires boost, and 3.3 gets rid of `touchesPeriodicBoundary()`. These are all pretty huge changes - even if we end up doing more work overall we can make much steadier progress if we just focus on one minor version at a time.

boyceg:
Yes, but the first publicly available SAMRAI 3.X release is 3.3.2-beta. 😁

drwells:
rats!

drwells:
Even though it isn't tagged, commit 109671252090

https://github.com/LLNL/SAMRAI/commit/1096712520903e559058f55fb6a688c9ad5b6f1f

still has `PatchGeometry::getTouchesPeriodicBoundary()` (but is not is templated on the dimension) so we can work backwards from the git history to recover the releases.

drwells:
@boyceg You can use this branch to mess around with new versions of SAMRAI and UMPIRE - it successfully compiles all of the Fortran in IBAMR and IBTK. I disabled every C++ file since they all use SAMRAI.

drwells:
@boyceg does this work? If so, lets merge (note that this is a PR onto the SAMRAI-3 branch, not `master`)

boyceg:
I am still messing around with setting up and testing SAMRAI master. Will start trying to link to it soon.

Do you want to wipe out all of the autotools stuff in this branch? Let's just use CMake to link to SAMRAI and friends.

drwells:
This works so lets merge.

PR #1352, opened 2021-09-03
Cleanup GitHub ci

drwells:
This only effects the CI, so the usual checklist doesn't apply.

This PR fixes two things:
- We wrote `REGEX_REPLACE` when we should have written `REGEX REPLACE`
- Our convention is to put CMake keywords in ALL CAPS to make them conspicuous


drwells:
In particular, if you look at our merge commits, they all fail on the tester because of the `REGEX_REPLACE` bug.

boyceg (approved):

PR #1353, opened 2021-09-03
Use Eigen, not LAPACK, for the Cholesky and PLU decompositions.

### LINKED ISSUE #1215: Make LAPACK an optional dependency
At some point in the past we acquired an implicit dependency on LAPACK - we can probably replace this with Eigen. If not we should make this an optional dependency since not everyone has it installed.

drwells:
Fixes #1215.

More progress on #1215: it is pretty easy to use Eigen's own solvers instead.

I added a new test that verifies that we get the same results (up to the usual absolute and relative numerical tolerances) for the Cholesky option in `DirectMobilitySolver`.

Edit: add fix tag.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/minor/20210903DavidWells
+++ b/doc/news/changes/minor/20210903DavidWells
@@ -1,0 +1,3 @@
+Improved: IBAMR no longer depends on LAPACK.
+<br>
+(David Wells, 2021/09/03)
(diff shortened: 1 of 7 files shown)


drwells:
Writing the same for LU was pretty easy so I went ahead and did it. Hooray, no more implicit dependencies!

drwells:
@amneetb like the other one - take a look when you can and let me know if there is anything extra we should check.

amneetb (approved):

PR #1356, opened 2021-09-07
Fix usage of Eigen::Index to work with older versions of Eigen.

drwells:
I don't really understand this problem - the `typedef` is clearly in an included Eigen header (`Meta.h`). Nevertheless relying on the type stored by the matrix works for both recent versions of Eigen and the one bundled with IBAMR.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -568,37 +568,32 @@ boost support or an external boost library.")
       ENDIF()
     ENDIF()
 
+    # Use slightly different error messages for bundled and non-bundled Eigen
     IF(EXISTS ${LIBMESH_ROOT}/include/Eigen)
       IF("${EIGEN3_ROOT}" STREQUAL "${LIBMESH_ROOT}/include/Eigen")
         # OK
+      ELSEIF(${IBAMR_USE_BUNDLED_EIGEN3})
+        MESSAGE(FATAL_ERROR "\
+libMesh appears to have been compiled with a bundled copy of Eigen3 (i.e., the \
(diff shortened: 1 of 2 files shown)


drwells:
FYI: the only place libMesh uses Eigen is inside a radial basis interpolation class - its not something we use in IBAMR and its not something most users would want to do in a solid mechanics code.

CMakeLists.txt
```diff
@@ -568,37 +568,32 @@ boost support or an external boost library.")
       ENDIF()
     ENDIF()
 
+    # Use slightly different error messages for bundled and non-bundled Eigen
     IF(EXISTS ${LIBMESH_ROOT}/include/Eigen)
       IF("${EIGEN3_ROOT}" STREQUAL "${LIBMESH_ROOT}/include/Eigen")
         # OK
+      ELSEIF(${IBAMR_USE_BUNDLED_EIGEN3})
```
boyceg, line 575: Can we add this check to the legacy build system as well?

drwells: We don't need this check for the autotools-based build system. In particular, if we detect libMesh + bundled Eigen, then we will switch to libMesh's version of Eigen - i.e., its designed to not error-out but instead reconfigure itself. We can't do that with CMake since libMesh doesn't install the CMake module files.

drwells: This patch is needed for people who use the copy of Eigen bundled with libMesh - if there aren't any other things that need to be changed here then could someone please approve and merge?

boyceg (approved):

PR #1362, opened 2021-09-24
Update bundled boost version to 1.76.

drwells:
Its time to update our bundled version of boost. We need to do this for two reasons:
1. The new version of libMesh (1.7.0) defaults to C++17 if the compiler can handle it. Clang removes `std::auto_ptr` in C++17, so our bundled version of boost fails to compile with that configuration.
2. The boost replacement for GSL we need (Gauss-Kronrod quadrature) was first added in 1.66.

The latest release of boost has a bunch of weird bugs in it and a lot of distributions are not yet shipping it, so I went with one version back.

I have tried to use a smaller subset of boost to keep downloads reasonably small - after all, we only use four headers from boost (multi_array, roots, round, and demangle).

To compute the minimal subset, I copied and pasted the four headers we use and then ran the following shell script:

    #!/bin/bash
    set -e

    for i in $(seq 0 1000)
    do
        echo "round ${i}"
        for FILE in $(bash ./test-boost-2.sh)
        do
            mkdir -p $(dirname "boost/${FILE}")
            echo "file = ${FILE}"
            cp temp/boost_1_76_0/${FILE} ./boost/${FILE}
        done
    done

which depends on this shell script:

    #!/bin/bash
    # set -e

    for FILE in $(find ./boost -type f)
    do
        /usr/local/bin/g++-11 -DBOOST_ALL_NO_LIB -I$(pwd)/boost -E ${FILE} -o test.o 2>&1 | grep 'No such file' | awk -F: '{print $5}'
    done | grep boost | sort -u | grep -v 'MPL_PREPROCESSED_HEADER'

to ensure that we can at least preprocess (i.e., all dependent headers are present) each required header. If we want to add more headers in the future then we should rerun the first script to add its minimal dependencies. The process took about 400 'iterations' to converge.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

If we need to add more headers in the future we should use those scripts to pick up all their dependencies.


drwells:
@IBAMR/ibamr-admin-team does anyone know of external projects that use other stuff from boost? I'd prefer to bundle the most minimal subset we can but if people are dependent on features we bundle we can add them back in.

boyceg:
I don't think we need to include anything from Boost that is not used internally. If folks need a bigger Boost, it is not hard to use an external one.

boyceg:
IIRC, the Boost thing that we depend on most is `boost::multi_array`. I haven't looked in a while, but I am not aware of a good alternative to that class.

drwells:
`boost::multi_array` is pretty good - I don't see any reason to replace it.

PR #1363, opened 2021-10-01
Remove gsl

### LINKED ISSUE #1358: Convert the GSL integrals to use boost
boost 1.66 was released in 2017 and it includes support for Gauss-Kronrod quadrature (the quadrature we use in eel3d):

https://www.boost.org/users/history/version_1_66_0.html

I'd like to update our minimum and bundled versions of boost so we can use this instead of GSL - this way users don't need one more dependency just to run this example.

AFAICT we only use GSL in this one place and we should be able to convert it pretty easily to use the boost version instead. Thoughts?

drwells:
Fixes #1358. Depends on #1362 (we don't want to bundle a version of boost that is too old).

I ran a thousand or so timesteps with the eel3d code to verify that this doesn't introduce large changes - the transverse velocity changes at the solver tolerance level (about O(10^-11)) but otherwise things are identical.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


amneetb:
Looks good. Another test could be comparing the vertex file generated by new and old eelgenerator.cpp. Also, I think we should keep gsl.m4 just in case somebody wants to use GSL in an application?

boyceg:
> Looks good. Another test could be comparing the vertex file generated by new and old eelgenerator.cpp. Also, I think we should keep gsl.m4 just in case somebody wants to use GSL in an application?

I think that I would prefer to drop `gsl.m4` if we don't use it anywhere in the library, examples, or tests --- seems easy to accidentally break.

amneetb:
Perhaps boost already supports most of the gsl features, so we can remove it.

drwells:
@amneetb I think we now have this - I added a test

tests/external/eelgenerator3d.cpp

in this PR that checks that we really do get the same results for both.

PR #1366, opened 2021-10-18
Improve kernel selection

drwells:
Incredibly, `fdl::IFEDMethod` spends about 6-7% of its time doing string comparisons inside `LEInteractor`. This patch rewrites those string comparison functions to check the minimal number of characters and instead moves the 'verify the kernel' checks to debug mode and also object constructors.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/lagrangian/LEInteractor.cpp
+++ b/ibtk/src/lagrangian/LEInteractor.cpp
@@ -1440,14 +1440,6 @@ LEInteractor::isKnownKernel(const std::string& kernel_fcn)
 int
 LEInteractor::getStencilSize(const std::string& kernel_fcn)
 {
-    // Some IBAMR applications want to call this function *a lot*, e.g., ideally
-    // we would call this on every element in a finite element mesh instead of
-    // batching with huge temporary arrays. Hence doing all these string
-    // comparisons can get surprisingly expensive. Chop them up first with a
-    // switch statement.
-    //
-    // For readability, the old and not hacky version is included at the end of
-    // the function.
     switch (string_to_kernel(kernel_fcn))
     {
     case BSPLINE_3:
--- a/ibtk/src/lagrangian/LEInteractor.cpp
+++ b/ibtk/src/lagrangian/LEInteractor.cpp
@@ -1311,6 +1311,102 @@ spread_data(const int stencil_sz,
     }
 #endif
 } // spread_data
+
+// For internal use - convert to an enumeration. This may make it into the
+// public interface one day but for backwards compatibility we accept only
+// strings for now.
+enum KernelType
+{
+    BSPLINE_3,
+    BSPLINE_4,
+    BSPLINE_5,
+    BSPLINE_6,
+    DISCONTINUOUS_LINEAR,
+    PIECEWISE_CONSTANT,
+    PIECEWISE_LINEAR,
+    PIECEWISE_CUBIC,
+    IB_3,
+    IB_4,
+    IB_4_W8,
+    IB_5,
+    IB_6,
+    USER_DEFINED,
+    INVALID
+};
+
+KernelType
+string_to_kernel(const std::string& kernel_fcn)
+{
+    // Some IBAMR applications want to call this function *a lot*, e.g., ideally
+    // we would call this on every element in a finite element mesh instead of
+    // batching with huge temporary arrays. Hence doing all these string
+    // comparisons can get surprisingly expensive. Chop them up first with a
+    // switch statement.
+#ifndef NDEBUG
+    if (!LEInteractor::isKnownKernel(kernel_fcn))
+    {
+        TBOX_ERROR("Unknown kernel function " << kernel_fcn << std::endl);
+        return INVALID;
+    }
+#endif
+    switch (kernel_fcn.front())
+    {
+    // BSPLINE family
+    case 'B':
+    {
+        switch (kernel_fcn.back())
+        {
+        case '3':
+            return BSPLINE_3;
+        case '4':
+            return BSPLINE_4;
+        case '5':
+            return BSPLINE_5;
+        case '6':
+            return BSPLINE_6;
+        }
+    }
+    // DISCONTINUOUS family
+    case 'D':
+    {
+        return DISCONTINUOUS_LINEAR;
+    }
+    // PIECEWISE family
+    case 'P':
+    {
+        switch (kernel_fcn[11])
+        {
+        case 'O':
+            return PIECEWISE_CONSTANT;
+        case 'I':
+            return PIECEWISE_LINEAR;
+        case 'U':
+            return PIECEWISE_CUBIC;
+        }
+    }
+    // classic IB family
+    case 'I':
+    {
+        switch (kernel_fcn[3])
+        {
+        case '3':
+            return IB_3;
+        case '4':
+            return kernel_fcn.size() == 4 ? IB_4 : IB_4_W8;
+        case '5':
+            return IB_5;
+        case '6':
+            return IB_6;
+        }
+    }
+    // user defined
+    case 'U':
+        return USER_DEFINED;
+    }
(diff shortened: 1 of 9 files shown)


drwells:
Well, strings are the poor man's data structure...

I am ambivalent about enums but yeah, we need an extension interface (like function pointers) for LEInteractor. The mask code and other cool stuff would be a lot easier to implement if we could use something like inheritance and teach LEInteractor to keep some state around.

drwells:
@boyceg Is there anything you would like to change? I want to get this merged soon.

boyceg:
How about adding a new interface that takes in the `KernelName` (maybe `KernelType`?) enum and use that interface internally while still supporting (but deprecating) the string-based interface externally?

(Or, maybe we should go ahead and make things more extensible and performant at the same time...?) 

Also, are you calling these functions with small batches of interaction points? It might be better anyway to call these functions for larger collections of points --- ideally the interaction points are ordered so that the Cartesian and Lagrangian data move through memory in parallel.

drwells:
> How about adding a new interface that takes in the KernelName (maybe KernelType?) enum and use that interface internally while still supporting (but deprecating) the string-based interface externally?


I would prefer not to do this. `LEInteractor` has 44 member functions that take `std::string &kernel_fcn` as a parameter. We would need to write a massive amount of code here for no real gain now that we have cheap conversion functions between the string and enums.


> (Or, maybe we should go ahead and make things more extensible and performant at the same time...?)


I would really like to write a new version of `LEInteractor` that actually stores some state so we can do things with variable masks, custom kernels, etc. in a much simpler way. The performance of the current version is OK - it would be nice if we could skip the 'check that the point really is in the patch' part but it isn't massively expensive.


I don't have time to lead it but I have some ideas on how we could vectorize all the IB kernels - we need a graduate student who loves assembly code more than biomechanics for this project.


> Also, are you calling these functions with small batches of interaction points? It might be better anyway to call these functions for larger collections of points --- ideally the interaction points are ordered so that the Cartesian and Lagrangian data move through memory in parallel.


I found it more convenient to call `LEInteractor` on each element:

https://github.com/drwells/fiddle/blob/master/source/interaction/interaction_utilities.cc#L628

I suspect that this approach may be faster since we don't have to set up huge arrays, write them to memory, and then read them back in again (we also avoid recomputing some FE data). I didn't profile it though.

boyceg:
> I would really like to write a new version of `LEInteractor` that actually stores some state so we can do things with variable masks, custom kernels, etc. in a much simpler way. The performance of the current version is OK - it would be nice if we could skip the 'check that the point really is in the patch' part but it isn't massively expensive.
> 
> I don't have time to lead it but I have some ideas on how we could vectorize all the IB kernels - we need a graduate student who loves assembly code more than biomechanics for this project.

@knepley do you have anyone who might be interested in this kind of thing? 😁

ibtk/src/lagrangian/LEInteractor.cpp
```diff
@@ -1311,6 +1311,108 @@ spread_data(const int stencil_sz,
     }
 #endif
 } // spread_data
+
+// For internal use - convert to an enumeration. This may make it into the
+// public interface one day but for backwards compatibility we accept only
+// strings for now.
+enum KernelName
```
boyceg, line 1318: Not critical since this is internal to this implementation file, but suggest:
```suggestion
enum KernelType
```
which (I think) is consistent with other `enum`s.

ibtk/src/lagrangian/LEInteractor.cpp
```diff
@@ -1311,6 +1311,108 @@ spread_data(const int stencil_sz,
...
+KernelName
+string_to_kernel(const std::string& kernel_fcn)
+{
+    // Some IBAMR applications want to call this function *a lot*, e.g., ideally
+    // we would call this on every element in a finite element mesh instead of
+    // batching with huge temporary arrays. Hence doing all these string
+    // comparisons can get surprisingly expensive. Chop them up first with a
+    // switch statement.
+#ifndef NDEBUG
+    if (!LEInteractor::isKnownKernel(kernel_fcn)) goto unknown_kernel;
```
boyceg, line 1346: `goto`??? 😁

boyceg, line 1346:
```suggestion
    if (LEInteractor::isKnownKernel(kernel_fcn))
```

ibtk/src/lagrangian/LEInteractor.cpp
```diff
@@ -1311,6 +1311,108 @@ spread_data(const int stencil_sz,
...
+#endif
+    switch (kernel_fcn.front())
```
boyceg, line 1348:
```suggestion
    {
        switch (kernel_fcn.front())
```

ibtk/src/lagrangian/LEInteractor.cpp
```diff
@@ -1311,6 +1311,108 @@ spread_data(const int stencil_sz,
...
+        }
+    }
+    // user defined
+    case 'U':
+        return USER_DEFINED;
+    default:
+        goto unknown_kernel;
+    }
+
+unknown_kernel:
```
boyceg, line 1411:
```suggestion
    }
```

ibtk/src/lagrangian/LEInteractor.cpp
```diff
@@ -1311,6 +1311,108 @@ spread_data(const int stencil_sz,
...
+            return IB_6;
+        default:
+            goto unknown_kernel;
+        }
+    }
+    // user defined
+    case 'U':
+        return USER_DEFINED;
+    default:
+        goto unknown_kernel;
```
boyceg, line 1408:
```suggestion
```

boyceg (approved):
Looks good. Suggest dropping the `goto`s.

knepley:
> > I would really like to write a new version of `LEInteractor` that actually stores some state so we can do things with variable masks, custom kernels, etc. in a much simpler way. The performance of the current version is OK - it would be nice if we could skip the 'check that the point really is in the patch' part but it isn't massively expensive.
> > I don't have time to lead it but I have some ideas on how we could vectorize all the IB kernels - we need a graduate student who loves assembly code more than biomechanics for this project.
> 
> @knepley do you have anyone who might be interested in this kind of thing? 😁

Let me talk to @drwells and see how similar it is to the FE vectorization in CEED, or to the particle vectorization we have in DMSwarm.

boyceg:
In fact, it might make sense to think about reimplementing `IBMethod` using `DMSwarm`.

Moreover, as we are moving more towards fully nodal coupling schemes, it might make sense to have a common `DMSwarm` implementation for the coupling part that is used everywhere.

knepley:
This is a great time to talk about that. My student who is working on it will graduate in May, and is currently doing a large rewrite of Swarm to incorporate some new things we needed for plasma physics. If we could get the requirements together, we can roll everything in.

ibtk/src/lagrangian/LEInteractor.cpp
drwells: It's a reasonable way to handle failure in this case!

I try not to write more than one per decade.

drwells:
> Let me talk to @drwells and see how similar it is to the FE vectorization in CEED

@knepley I will need some time to put notes together. The main difficulty with parallelization right now is handling IB points near the boundaries of patches. I also don't know a ton about SIMD programming (about 90% of what I know comes from reading Martin Kronbichler's code) and I don't know how to do efficient gathers/scatters of SIMD data to/from our SAMRAI data structures.

> In fact, it might make sense to think about reimplementing IBMethod using DMSwarm.

That would be very nice!

drwells:
Lets tentatively plan on meeting in January - I have enough other stuff to do this year.

I got rid of the `goto`s.

PR #1371, opened 2021-11-04
Fix ibtk mpi

drwells:
This came up on the mailing list and, unfortunately, the permanent fix is quite long.

The issue we have is that we made the MPI communicator argument implicit. Some versions of MPI have
```cpp
typedef unsigned int MPI_Comm;
```
which means that, with
```cpp
template <typename T>
static T sumReduction(T, IBTK_MPI::comm commiunicator = getCommunicator());
template <typename T>
static void sumReduction(T* x, const int n = 1, IBTK_MPI::comm communicator = getCommunicator());
```
if we do
```cpp
unsigned int size = 100;
double *values = array.begin();
sumReduction(values, size);
```
we end up calling the *first* function instead of the second one (due to C++ overload resolution rules its a better fit), which is wrong.

As far as I can tell, we have two reasonable options:
1. Audit every call site from now forward to make sure we always use the correct integer type.
2. Ban default arguments for communicators.

I am in favor of the second one because it is the permanent fix - if there aren't any overloads with optional arguments then we will always call the correct function and things won't randomly break. I marked this as WIP since I still need to run the full test suite and audit all the changes again.

If people agree with this approach then lets just change `IBTK_MPI` to not use default arguments any more. Minimally, we could keep the default communicator argument, but every other optional argument really needs to go (passing a default array length is not a great idea).

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/incompatibilities/20211215DavidWells
+++ b/doc/news/changes/incompatibilities/20211215DavidWells
@@ -1,0 +1,6 @@
+Changed: IBTK_MPI's unused MPI type aliases have been removed.
+Additionally, the optional communicator arguments have been removed.
+This change was necessary to prevent a type conversion error that caused
+compilation issues with versions of MPI that use unsigned int for MPI_Comm.
+<br>
+(David Wells, 2021/12/15)
--- a/ibtk/include/ibtk/IBTK_MPI.h
+++ b/ibtk/include/ibtk/IBTK_MPI.h
@@ -20,11 +20,11 @@
 
 #include <ibtk/config.h>
 
+#include <mpi.h>
+
 #include <type_traits>
 #include <utility>
 #include <vector>
-
-#include <mpi.h>
 
 namespace IBTK
 {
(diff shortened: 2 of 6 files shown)


boyceg:
Appending the communicator to each call makes things so verbose...

Here is a thought --- what about changing the functions so that they are not `static`, so one could write something like:
```
IBTK_MPI mpi;
mpi.sum_reduction(...);
```
and then have the default constructor use the default communicator? (This is moving closer to the way that SAMRAI handles this starting in SAMRAI 3...)

drwells:
Hm, maybe. I'll have to think about it some more. I would prefer to be explicit in the communicators - not being able to use subcommunicators is an unfortunate SAMRAI-ism that we have inherited. Creating a scoped `IBTK_MPI` would fix that but most of our call sites are isolated - we'll just do a single MPI reduction and return, so we don't save much typing by setting that up instead.

This patch needs to be done this year (in time for the next release) so we have some time to figure it out.

drwells:
Lets just delete the communicator as an argument instead of worrying about if it should or should not have a default value. Right now we are dependent on setting the communicator as a global variable since SAMRAI does that. Hence, it doesn't make sense to make specifying the communicator mandatory when 100% of the time it will have the same value.

We can revisit this when we support SAMRAI 3 but I think this is the right choice at the moment.

abarret (approved):

abarret:
This changes the public interface. It should be noted in the changelog.

drwells:
Fixed!

PR #1377, opened 2021-11-11
Fix silo linkage

drwells:
Since this change just modifies the build system the normal checklist doesn't apply. I noticed this while working on autoibamr.

We need to set up Silo as a private dependency so that it is not included in the publicly available link line. This should prevent duplicate symbol problems.


CMakeLists.txt
```diff
@@ -1009,10 +1022,7 @@ FUNCTION(IBAMR_SETUP_TARGET_LIBRARY target_library)
   ENDIF()
   # Silo is underlinked and depends on HDF5, so do it first:
   IF(${IBAMR_HAVE_SILO})
-    TARGET_LINK_LIBRARIES(${target_library} PRIVATE "${SILO_LIBRARIES}")
-    TARGET_INCLUDE_DIRECTORIES(${target_library} PUBLIC "${SILO_INCLUDE_DIRS}")
```
drwells, line 1013: What we had already technically worked but its weird to include the headers but not the libraries - this could fail in difficult to fix ways if a user somehow wanted parts of Silo not used inside IBAMR.

boyceg (approved):

PR #1379, opened 2021-11-19
Add infrastructure to solve linear systems with augmented DOFs

abarret:
This creates two new classes, `PETScAugmentedLinearSolver` and `PETScLinearAugmentedOperator` that allow for specifying an additional `Vec` to be included in the linear solve. This allows for solving linear systems with more than just degrees of freedom on the Cartesian grid. There's also an example that solves a Helmholtz problem outside of a disk with degrees of freedom on the disk using a RBF-FD method.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


abarret:
@drwells and @boyceg, if you have design suggestions, I'd be interested in hearing them.

boyceg:
Does this work with a coupled ODE model? Like a 0D circulation model?

amneetb:
I did something like this over here:

https://github.com/IBAMR/IBAMR/blob/master/src/IB/CIBStaggeredStokesSolver.cpp#L136-L223

abarret:
> I did something like this over here:
> 
> https://github.com/IBAMR/IBAMR/blob/master/src/IB/CIBStaggeredStokesSolver.cpp#L136-L223

I think this implementation works in a very similar way to this.

> Does this work with a coupled ODE model? Like a 0D circulation model?

Yes. It takes in a `Vec`, which can be whatever you want. You just need an implementation of `PETScLinearAugmentedOperator` to compute the action of the operator.

boyceg:
A couple quick questions/thoughts:

1) Is there a reason for "augmented linear" in one place and "linear augmented" in the other?

2) Are you expecting to have an implementation of the solver that doesn't use PETSc? If not, does there need to be another class? The solver class hierarchy is useful for Cartesian grid solvers / preconditioners since those actually are implemented using different frameworks, but if we only expect to have a PETSc-based implementation, would it be better just to make it easier to set up and pass in operators and preconditioners to a PETSc-based solver? I guess maybe it would make sense in some places to override the current Cartesian grid solvers with an augmented solver, although in those cases I worry about how to set the RHS and solution vectors within the time stepping loop.

abarret:
> 1. Is there a reason for "augmented linear" in one place and "linear augmented" in the other?

Laziness on my part.

> 2. Are you expecting to have an implementation of the solver that doesn't use PETSc? If not, does there need to be another class? The solver class hierarchy is useful for Cartesian grid solvers / preconditioners since those actually are implemented using different frameworks, but if we only expect to have a PETSc-based implementation, would it be better just to make it easier to set up and pass in operators and preconditioners to a PETSc-based solver? I guess maybe it would make sense in some places to override the current Cartesian grid solvers with an augmented solver, although in those cases I worry about how to set the RHS and solution vectors within the time stepping loop.

My intended application is to use PETSc based solvers.

I think the ideal ("correct?") way to implement this is to replace all `SAMRAIVectorReal` with a `IBTKVector` base class and `IBTK_SAMRAIVectorReal` that wraps `SAMRAIVectorReal`. This approach would require changing the public interface (although it should just be a simple search and replace). The only other alternative that I see to this is to allow users to register a `Vec` directly to `PETScKrylovLinearSolver`, and copy all necessary data to that `Vec`. However, this would require two copies per iteration of the Krylov method.

I agree that the usefulness of this class is suspect when it comes to preexisting integrators. Although, in principle, you would be able to set up the augmented RHS and solution vector by registering a call to `registerPreprocessIntegrateHierarchyCallback` or something similar.

boyceg:
I would bet that @drwells would support replacing `SAMRAIVectorReal` with _Something Else_. (For instance, it is too hard to instantiate new Cartesian grid data, although I am not sure whether the vector interface is the right place to address that.)

drwells:
It is hard to instantiate new Cartesian grid data but that's independent of the linear algebra infrastructure. I don't know how to fix this in a sane way aside from adding our own convenience SAMRAI functions.

This patch has some really useful stuff - being able to do these coupled solves is a nice generalization of our current setups. I want to spend some more time thinking about it before I commit to one answer or another. In particular, I'd like to better understand how PETSc handles blocked systems - perhaps we can rely on that to do more of this work. Either way I'll dig in after the holiday.

knepley:
@drwells PETSc has you tell it the decomposition using lists of dofs (ISes), but some it can figure out (every kth dof). Once you have this, you can manipulate everything in terms of blocks.

abarret:
> This patch has some really useful stuff - being able to do these coupled solves is a nice generalization of our current setups. I want to spend some more time thinking about it before I commit to one answer or another. In particular, I'd like to better understand how PETSc handles blocked systems - perhaps we can rely on that to do more of this work. Either way I'll dig in after the holiday.

One potential issue is that PETSc doesn't handle any of the Eulerian data directly. All of the `Vec` operations are replaced by those from SAMRAI. In particular, the PETSc vector is given a dummy number of DOFs.

boyceg:
To amplify @abarret's point a bit --- for most Cartesian AMR linear and nonlinear solvers, our approach is based off of the original PETSc-SAMRAI interface, which only defines the action of the operators along with standard vector space operations for vectors and a minimal collection of pointwise operations. We do not provide PETSc with proper DOF indexing information (nor do we compute that indexing information in most cases). It shouldn't be too hard to create the DOF indexing automatically, in case we wanted to create a `DM`.

The main exceptions are implementations of `PETScLevelSolver`, which use PETSc in a more typical way.

boyceg:
FWIW, IIRC (which I probably don't), when I first started messing around with the linear solvers, it was difficult to synchronize "shared" DOF indices along patch boundaries, which occur for all data centerings other than `CellData`/`CellVariable`. I also really wanted to avoid creating matrix-based representations of coarse-fine interface discretizations.

drwells:
PETSc is general enough that I think we can get things to work when we bolt together our own `Vec` and some other notion of a `Vec` together but I'll have to look.

I'd rather not explicitly compute or store DoFs but we can do it for side-centered data - we have `PETScVecUtilities::constructPatchLevelDOFIndices()`.

boyceg:
The need to give extra information before calling `solveSystem()` seems brittle.

On the other hand, that seems like the easiest way to inject additional DOFs into the existing time steppers.

boyceg:
I like the idea of having a more generic vector class that can be a collection of different kinds of vectors, and then to have a specialization to AMR data that is roughly equivalent to `SAMRAIVectorReal` that is specific to Cartesian field data. However, if you do this, you run into the problem that if you want to inject additional application-specific DOFs into a model, then you need to somehow rewrite the call to `solveSystem()`. Being able to modify the behavior through the pre/post-solve hooks seems less invasive.

PR #1388, opened 2021-12-23
Fix floating point comparisons

### LINKED ISSUE #1177: Approximate floating point comparison with 0.0

abarret:
Fixes #1177. This replaces all `SAMRAI::equalEps` with an appropriate call to either `IBTK::abs_equal_eps` or `IBTK::rel_equal_eps`.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/minor/20211223AaronBarrett
+++ b/doc/news/changes/minor/20211223AaronBarrett
@@ -1,0 +1,3 @@
+Fixed: IBAMR now uses appropriate comparisons that respect finite precision.
+<br>
+(Aaron Barrett, 2021/12/23)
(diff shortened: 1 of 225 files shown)


drwells (changes requested):
OK once we fix the unused variable warning.

abarret:
Fixed the unused variable warning.

drwells (approved):

PR #1391, opened 2022-01-06
Exclude tests from all

drwells:
Most of the time I want to compile `tests` independently of the library - i.e., `tests` should not be part of the `all` target. This isn't the standard CMake practice but 

This should work with the GitHub CI infrastructure - lets see if the tests still run with these changes.


drwells:
The tests run correctly so this is fine.

abarret (approved):

PR #1395, opened 2022-01-12
Fix IIM restarts with older versions of libMesh.

drwells:
We encounter the same issue here as we do in IBFEMethod - serialized vectors are
not set up as ghosted when they should be. It looks like newer versions of
libMesh fixed this, but it is definitely present in 1.5.1.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/libmesh_utilities.h
+++ b/ibtk/include/ibtk/libmesh_utilities.h
@@ -435,8 +435,8 @@ void setup_system_vectors(libMesh::EquationSystems* equation_systems,
  *   the normal way.</li>
  *   <li>If @p from_restart is true then the vector stored in the System
  *   corresponding to the given name is overwritten and has its type changed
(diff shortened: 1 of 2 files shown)


drwells:
Since we cannot use IIMethod restarts without this patch we should get it in before we create the release branch. @abarret do you have time to take a quick look?

abarret:
I don't see how this fixes restarts. It appears to me that `setup_system_vectors` does the exact same thing that the deleted code did.

drwells:
I thought the same thing at first, but if you look at the brackets, you'll see that we originally had
```cpp
if (from_restart)
{
  // do stuff
}
else
{
  // code that was deleted goes here
}
```
Instead, the new version sets up the vector outside this check and relies on the IBTK utility function to implement the correct logic w.r.t. restarts.

abarret:
Ahh.. The `files changed` tab was very misleading :)

abarret (approved):

PR #1400, opened 2022-01-19
Work around an rpath problem on macOS.

drwells:
Another problem caught by autoibamr.

I think this is a CMake bug - if hypre and PETSc are installed in the same place then CMake puts calls like

    execute_process(COMMAND /usr/bin/install_name_tool
      -delete_rpath "/Users/drwells/autoibamr/packages/petsc-3.13.1/lib"
      "$ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}/lib/libIBAMR2d.dylib")

inside src/cmake_install.cmake. This doesn't make any sense: that path doesn't need to be removed from the library prior to installation because it is for an external dependency. Furthermore, those paths are duplicated, which leads to errors of the form

    error: /Library/Developer/CommandLineTools/usr/bin/install_name_tool: 
    for: /Users/drwells/autoibamr/packages/IBAMR-0.10.0-rc1/lib/libIBAMR3d.dylib
    (for architecture x86_64) option "-add_rpath /Users/drwells/autoibamr/packages/petsc-3.13.1/lib"
    would duplicate path, file already has LC_RPATH for: /Users/drwells/autoibamr/packages/petsc-3.13.1/lib

which cause installation to fail.

We can work around this by relying on PETSc to transitively link to hypre for us in this case.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,12 +47,10 @@ SET(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib/")
 SET(IBAMR_DIMENSIONS "2" "3")
 
 # Start building up RPATH:
-SET(CMAKE_MACOSX_RPATH 1)
 SET(CMAKE_BUILD_RPATH "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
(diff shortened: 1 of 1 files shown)


drwells:
It might still be an issue on our end - see also https://gitlab.kitware.com/cmake/cmake/-/issues/17995

abarret (approved):

PR #1401, opened 2022-01-19
Cmake rpath macos workaround 0.10

drwells:
Same as #1400 but for 0.10.


drwells:
Whoops, wrong base branch. Let me try again.

PR #1404, opened 2022-01-26
Update the documentation for CMake.

### LINKED ISSUE #1365: CMake doesn't seem to respect loaded compiler modules
If you load a module then `which gcc` returns the correct value, but CMake will default to using /usr/bin/cc - that is technically correct but its not what we want. We should add some instructions to make this clearer and specify that CMake really does need us to set an environment variable, `CMAKE_CXX_COMPILER`, or equivalent.

drwells:
Fixes #1365: we need to be explicit.


doc/cmake.md
```diff
@@ -115,9 +111,42 @@ which contains all of the files generated by CMake. Deleting the cache is
 harmless - the subsequent initial configuration run will just take a few more
 seconds.
 
+### Working with modules
+If you are on a cluster then you may want to use compilers and an MPI
+implementation which are not installed in standard locations, but are managed
```
boyceg, line 116:
```suggestion
implementation that are not installed in standard locations, but are managed
```

boyceg (approved):

drwells:
This change is totally independent of the CI so I'll merge it as-is.

Issue #1219, opened 2021-02-27
investigate the feasibility of creating an application-independent convergence tester


boyceg:
It seems like it would not be too hard to generalize the application-specific convergence testers to allow for generic variables.

PR #1408, opened 2022-02-08
Added class that can store snapshots and compute flow statistics.

abarret:
This adds infrastructure to store flow snapshots, compute flow statistics over periodic intervals, and output intermediate time velocities. The intention of this class is two fold:

1. Compute flow statistics over a period,
2. Skip solving the momentum equation when we are at a periodic steady state.

This supersedes #1345. It is a less invasive approach, although to solve 2 above, it will necessarily become invasive.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


abarret:
@boyceg, this is partially what I was imagining. We can put this class in the `INSStaggeredHierarchyIntegrator` and switch between interpolated snapshots and solving the equations.

abarret:
This is WIP. If you have feedback on the design choice, let me know.

boyceg:
Looks like this is on the right track.

There are two basic things to compute:

1. The mean velocity field. This is simply the average of velocity snapshots, perhaps captured with some interval greater than 1. Note that you do not need to keep individual snapshots to update the mean velocity.
2. The cycle-averaged velocity field --- this is the average of velocity fields at the same time in the period. Again, you don't need all of the snapshots, just the means at each time in the interval.

Note that these are both flow fields --- it looks like maybe the current interface is set up for them to be simple scalar values?

Once you have the means, you can then compute higher-order quantities like TKE and Reynolds stress. These are defined in terms of the fluctuating part of the field (which you only know once you have the mean).

Another thing that will need to be done is to handle regridding. It might be nice to support a separate patch hierarchy, so that (e.g.) it is possible just to store on a uniform grid and then transfer to/from the uniform grid when storing/looking up snapshots/averages.

abarret:
> There are two basic things to compute:
> 
> 1. The mean velocity field. This is simply the average of velocity snapshots, perhaps captured with some interval greater than 1. Note that you do not need to keep individual snapshots to update the mean velocity.
> 2. The cycle-averaged velocity field --- this is the average of velocity fields at the same time in the period. Again, you don't need all of the snapshots, just the means at each time in the interval.

These both seem like they should be handled separately from this class. This class stores snapshots and computes quantities based on those snapshots. The decision of when to store snapshots is separate from this class and should be handled by the INS integrator (or user code).
> 
> Note that these are both flow fields --- it looks like maybe the current interface is set up for them to be simple scalar values?

This is specialized to side centered data, although, with a few changes, this could be used with any node centering to store snapshots of pressure, concentration fields, etc.

> 
> Another thing that will need to be done is to handle regridding. It might be nice to support a separate patch hierarchy, so that (e.g.) it is possible just to store on a uniform grid and then transfer to/from the uniform grid when storing/looking up snapshots/averages.

Another possibility is to also store snapshots of the patch hierarchy and then transfer from each snapshot as needed.

boyceg:
> > There are two basic things to compute:
> > 
> > 1. The mean velocity field. This is simply the average of velocity snapshots, perhaps captured with some interval greater than 1. Note that you do not need to keep individual snapshots to update the mean velocity.
> > 2. The cycle-averaged velocity field --- this is the average of velocity fields at the same time in the period. Again, you don't need all of the snapshots, just the means at each time in the interval.
> 
> These both seem like they should be handled separately from this class. This class stores snapshots and computes quantities based on those snapshots. The decision of when to store snapshots is separate from this class and should be handled by the INS integrator (or user code).

I guess I am confused about what this is doing. If this is for storing snapshots, it seems like the class should have a different name. If it is computing means, then --- as mentioned --- you don't need the individual snapshots to update the mean. (I think that what one would want to do in practice is to keep averaging until the mean converges --- and you typically don't know a priori how many snapshots will be necessary to achieve convergence.)

I would suggest an interface in which each time you "set" a new snapshot, the class updates the appropriate mean.

> > Note that these are both flow fields --- it looks like maybe the current interface is set up for them to be simple scalar values?
> 
> This is specialized to side centered data, although, with a few changes, this could be used with any node centering to store snapshots of pressure, concentration fields, etc.

What is the plan for `computeTimeAvgVelocity()`? I don't understand its return value. Why is it a `VectorNd`?

Also, note that you can look up hierarchy data ops with the correct centering automatically using `HierarchyDataOpsManager::getOperationsDouble()`, and so specializing to side-centered data should be unnecessary.

> > Another thing that will need to be done is to handle regridding. It might be nice to support a separate patch hierarchy, so that (e.g.) it is possible just to store on a uniform grid and then transfer to/from the uniform grid when storing/looking up snapshots/averages.
> 
> Another possibility is to also store snapshots of the patch hierarchy and then transfer from each snapshot as needed.

Yes, that is what I did in my half-finished branch. At any rate, if the values are stored on a locally refined grid, it seems like it will be necessary to provide a way to trigger regridding.

boyceg:
Is this class intended to store the results obtained by averaging over cycles and then doing time interpolation to return the averaged results at arbitrary times? That seems good too --- but if so it seems like the class should get a different name --- something like `HierarchyTimeInterpolator`??

abarret:
> Is this class intended to store the results obtained by averaging over cycles and then doing time interpolation to return the averaged results at arbitrary times?

This was my intention. You're right though, it is a poor name choice. I'll switch it.

> 
> What is the plan for `computeTimeAvgVelocity()`? I don't understand its return value. Why is it a `VectorNd`?
> 

The time averaged velocity field should have NDIM components, shouldn't it? It's not an average over space. Although, given your above comments, this should also be considered something that could/should be an input to the class.

boyceg:
> > Is this class intended to store the results obtained by averaging over cycles and then doing time interpolation to return the averaged results at arbitrary times?
> 
> This was my intention. You're right though, it is a poor name choice. I'll switch it.
> 
> > What is the plan for `computeTimeAvgVelocity()`? I don't understand its return value. Why is it a `VectorNd`?
> 
> The time averaged velocity field should have NDIM components, shouldn't it? It's not an average over space. Although, given your above comments, this should also be considered something that could/should be an input to the class.

The time-averaged velocity field is a velocity field. What is `computeTimeAvgVelocity()` returning? Also, from the above, it sounded like your plan is to leave the actual time averaging operations for another class. I guess I am unclear about the purpose you intend for this class. There are a few possibilities:

1. Infrastructure for storing and looking up "snapshots". As mentioned above, this can be made to be very generic (not tied in to specific data centerings) using `HierarchyDataOpsManager::getOperationsDouble()`.
2. Infrastructure for storing and looking up temporal averages. As in (1), this can be done in a generic way (but doesn't seem to be what you have in mind from comments above).
3. Infrastructure for tracking mean flows and quantities that describe the fluctuating part of the flow, like turbulence kinetic energy. The details here will be tied to specific discretizations (e.g. TKE requires evaluating the fluctuating components of the flow field in a co-located way).

Clearly I am confused about what the plan is here. 😁

When computing time-averaged quantities, I don't think that you want to keep copies of each one of the fields that goes into the average --- I think that you just want to track & update the running average(s). Under steady conditions, you would probably just want to track a single time-averaged field that is updated at some interval (may not need to update it at each time step for very long simulations). Under time-periodic conditions, I anticipate that one would generally want to keep track of cycle averages at multiple points within the period.

Generic infrastructure for tracking snapshots could also be used for other analyses (POD comes to mind).

abarret:
Ok I completely reworked things. Currently I only have header files. I will add implementations if we're on the same page :D. 

There are now two classes, written for general data spacing and depth. Part of the trouble is that we're storing snapshots at time points, but there's no guarantee that after a full period later, we will exactly match up with that previous time point (because the time step can change size). So I plan on storing snapshots at slightly fuzzy time points (a tolerance that should be within a time step size or two).

`SnapshotCache` stores snapshots of patch indices at specified time points on a snapshot of the patch hierarchy. You can retrieve snapshots transferred onto a new hierarchy. This is all it does. It has no notion of periodic steady states. This is potentially useful in #1219

`HierarchyTimeInterpolator` stores a specified number of mean snapshots over a period (or perhaps a steady state snapshot with "period" 0). It has mechanisms to update the average snapshot and determine if the snapshots are at a periodic steady state. It can also interpolate between snapshots. This class stores snapshots via the `SnapshotCache`.

In practice, `INSStaggeredHierarchyIntegrator` will feed `HierarchyTimeInterpolator` flow fields, which will in turn keep track of the average flow field at that time. When we hit a (periodic) steady state, we can switch from solving equations to find flow fields to interpolating flow fields.

We can add functions that access the data in `HierarchyTimeInterpolator` to calculate e.g. the TKE.

include/ibamr/HierarchyTimeInterpolator.h
```diff
@@ -0,0 +1,182 @@
...
+     */
+    const std::set<double>& getSnapshotTimePts()
+    {
+        return d_snapshot_time_pts;
+    }
+
+    /*!
+     * Returns true if a flow snapshot is being stored within the specified tolerance.
+     */
+    bool timePtIsStored(double time, double tol);
```
boyceg, line 111: Add default tolerances?

include/ibamr/HierarchyTimeInterpolator.h
```diff
@@ -0,0 +1,182 @@
...
+     */
+    void clearSnapshots();
+
+    /*!
+     * Fill a patch index at a particular time point. This will perform a time interpolation of the snapshots stored in
+     * the SnapshotCache object. Note if there is only one snapshot (i.e. the period is 0 and we are assumed to be at a
+     * steady state), a copy will be performed
+     */
+    void
+    fillSnapshotAtTime(int u_idx, double time, SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy);
```
boyceg, line 81: Should there be a tolerance for the `time`?

include/ibamr/HierarchyTimeInterpolator.h
```diff
@@ -0,0 +1,182 @@
...
+
+    /*!
+     * Update the time averaged snapshot. The update is computed through the relation
+     * ``` u_avg = u_avg + 1/N * (u - u_avg)```
```
boyceg, line 85: state what N is here?

boyceg: Also, this averaging process is also useful for models that are not periodic but instead give a steady mean value. Can/should we support both in the same class?

boyceg:
BTW, "steady" here really means that we expect to be able to decompose u as u(x,t) = u_bar(x) + u'(x,t), where u_bar is the mean flow, which is defined in principle by averaging over all time, and u' is fluctuating.

In the time periodic case, we would expect to get u(x,t) = u_bar(x,t) + u'(x,t), in which u_bar(x,t) is a periodic function of t that is defined by cycle averages.

Does this make sense?

include/ibamr/HierarchyTimeInterpolator.h
abarret: This class should take care of that. You can store one snapshot and set the "period" to be 0.

include/ibamr/HierarchyTimeInterpolator.h
abarret: This function interpolates between snapshots. The idea being to be able to drive simulations with interpolated snapshots.

include/ibamr/HierarchyTimeInterpolator.h
```diff
@@ -0,0 +1,182 @@
...
+    bool updateTimeAveragedSnapshot(int u_idx,
+                                    double time,
+                                    SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy);
+
+    /*!
+     * Return true if all the tracked mean fields are at a steady state.
+     */
+    bool isAtPeriodicSteadyState();
+
+    /*!
```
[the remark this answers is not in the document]
abarret: I had originally set this up so that the number of snapshots and time points must be set in the constructor. I think we can get rid of that requirement, as long as the time points that snapshots are stored aren't too close together.

include/ibamr/SnapshotCache.h
```diff
@@ -0,0 +1,150 @@
...
+{
+public:
+    /*!
+     * The constructor for class SnapshotCache sets some default values, reads in configuration information from input
+     * databases, and registers some variable/context pairs with the VariableDatabase.
+     */
+    SnapshotCache(std::string object_name,
+                  SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db,
+                  SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > var,
+                  SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > patch_hierarchy);
```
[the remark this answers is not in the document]
abarret: We don't need the patch hierarchy in the constructor.

The idea is to store data on snapshots of the patch hierarchy, and we only transfer data as needed (i.e. when we need to use the snapshot).

abarret:
> BTW, "steady" here really means that we expect to be able to decompose u as u(x,t) = u_bar(x) + u'(x,t), where u_bar is the mean flow, which is defined in principle by averaging over all time, and u' is fluctuating.
> 
> In the time periodic case, we would expect to get u(x,t) = u_bar(x,t) + u'(x,t), in which u_bar(x,t) is a periodic function of t that is defined by cycle averages.
> 
> Does this make sense?

I believe so. This class should be storing u_bar, whether it's a function of time or not.

include/ibamr/SnapshotCache.h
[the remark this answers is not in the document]
abarret: When we store a snapshot, we create a new patch hierarchy that is a snapshot of the original hierarchy. Data is stored on the snapshot hierarchy. Then we can transfer from the snapshot as needed, similar to how we do convergence testing. As long as we don't do something crazy with the original patch hierarchy (like change the coarsest level), we should be able to transfer from a snapshot copy back to the original without too much trouble.

abarret:
I've added implementations. `HierarchyTimeInterpolator` and `SnapshotCache` are now templated over the variable type, although currently only `double`s are allowed. If we need to add greater type support, it shouldn't be hard to template over data type as well.

- [ ] Add tests.
- [ ] Incorporate this with `INSStaggeredHierarchyIntegrator`.

@boyceg, take a look and let me know what you think.

include/ibamr/SnapshotCache.h
```diff
@@ -67,30 +73,51 @@ class SnapshotCache
...
      */
-    void setSnapshot(int u_idx, double time, SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy);
+    void
+    setSnapshot(int u_idx, double time, SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > current_hierarchy);
 
     /*!
      * Update a snapshot. The data from this patch index is copied into an array of indices controlled by this
      * object. This outputs an error if a snapshot with that time point within the specified tolerance can not be found.
      */
     void updateSnapshot(int u_idx,
```
boyceg, line 83: Same comment as above.

include/ibamr/SnapshotCache.h
```diff
@@ -67,30 +73,51 @@ class SnapshotCache
...
 
     /*!
      * Transfer the snapshot at the time value within a given tolerance to the patch index u_idx on the supplied patch
      * hierarchy.
      *
      * Note this can cause a significant amount of communication if the supplied patch hierarchy has a different
      * configuration of patches than the snapshot patch hierarchy.
+     *
+     * This function does not synchronize the hierarchy (i.e. no coarsening is performed). It copies the snapshot as it
+     * was stored.
```
boyceg, line 96: Does this mean that the caller needs to have the original hierarchy to copy back onto?

abarret: I don't know about the specifics of copying data from one hierarchy to another, but whatever hierarchy you want to copy back onto has to be "valid" to copy to. I think that's just that the snapshotted hierarchy and the one to copy to have the same number of levels, their coarsest levels cover the same index space, and they have the same refinement ratios.

I'll figure out the details when I test.

boyceg: It seems like there are two likely use cases:

1. The caller wants the data on the original patch hierarchy layout; in this case, maybe the function should give access to a pointer to the internal clone of the patch hierarchy? (Most models will only have one `PatchHierarchy` object, and so recovering the hierarchy state at some previous time point seems annoying.)
2. The caller wants the data on their current patch hierarchy layout; in this case, maybe there could be a different function that uses something (`HierarchyGhostCellInterpolator`?) to transfer to the desired hierarchy layout.

abarret: > 2\. The caller wants the data on their current patch hierarchy layout; in this case, maybe there could be a different function that uses something (`HierarchyGhostCellInterpolator`?) to transfer to the desired hierarchy layout.

This is what the function `getSnapshot` currently does: transfers from the snapshot hierarchy onto whatever patch hierarchy is given as the argument to the function. Although I realize there's an issue in the implementation that I'll fix soon.


> 1. The caller wants the data on the original patch hierarchy layout; in this case, maybe the function should give access to a pointer to the internal clone of the patch hierarchy? (Most models will only have one `PatchHierarchy` object, and so recovering the hierarchy state at some previous time point seems annoying.)

We can easily return the snapshotted patch hierarchy, perhaps in a function `getSnapshotAndHierarchy`.

drwells: If you want to copy between hierarchies you might want to take a look at `IBTK::SecondaryHierarchy`. The original point of that class was to implement an alternative parallel partitioning but I think it implements what you need here too.

abarret:
Well here's a potential problem. `RefineSchedule` requires that destination and source patch data be stored at the same time points.

boyceg:
You can manually reset the time as needed.

abarret:
> You can manually reset the time as needed.

What a pain.

boyceg:
`RefineSchedule` also knows how to do time interpolation, and so I think that this is to avoid the case where you should be doing time interpolation but are not using the correct method.

You also could copy to a temporary patch data index.

abarret:
It has been fixed. We now store, retrieve, and interpolate snapshots correctly.

What remains to be tested is whether we can compute time averages.

boyceg:
I would suggest doing a test like Taylor vortices with forcing to get a time-periodic solution. You could start with zero initial conditions and run until you get the periodic steady state.

ibtk/include/ibtk/HierarchyTimeInterpolator.h
```diff
@@ -0,0 +1,226 @@
...
+#include <string>
+#include <vector>
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBTK
+{
+/*!
+ * \brief Class HierarchyTimeInterpolator provides a method of determining and storing average fields over periodic
+ * intervals. This class can also be used to determine intermediate velocity fields when at a periodic steady state.
```
boyceg, line 40: Since this is doing more than just interpolating data in time, I'd suggest considering a more specific class name. I am not crazy about these suggestion, but something like `HierarchyAveragedDataManager` or `AveragedSnapshotDataManager`?

This might be too complicated, but you could also make a little class hierarchy --- `SnapshotTimeInterpolator` or `SnapshotDataManager` for interpolating snapshots in time, and a subclass like `SnapshotMeanTimeInterpolator` or `SnapshotMeanDataManager` for tracking mean values.

abarret: We actually don't need a class to interpolate between snapshots. It would probably be cleaner to make those stand alone functions.

PR #1410, opened 2022-02-08
fix IIM ex0 compilation for autotools-based configurations

drwells:
It's a build system change so the normal checklist doesn't really apply.


boyceg:
Just noticed --- should/do we update the copyright in the Makefiles? The file I copied lists 2019.

boyceg (approved):

PR #1414, opened 2022-02-08
Explicitly remind us to check example compilation.

drwells:
Apparently I forgot.


boyceg (approved):

boyceg:
This is good to check, of course, but probably wouldn't have caught the error from #1410 because it was an interstitial Makefile that was missing --- you had to go all the way down to IIM ex0 to encounter the error.

drwells:
It's better than nothing but yeah, the examples are difficult to deal with for this reason.

PR #1416, opened 2022-02-11
Fix compilation of examples that depend just on boost.

drwells:
The fix in the last CMake PR wasn't quite right - since we support compiling examples that don't depend on IBAMR but do depend on boost we need to set the C++ version in a different way.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -74,10 +74,14 @@ MACRO(IBAMR_ADD_EXAMPLE)
       "${EX_OUTPUT_NAME}")
     TARGET_INCLUDE_DIRECTORIES("${EX_TARGET_NAME}" PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}")
 
-    IF("${EX_LINK_TARGETS}" STREQUAL "")
-      # If we have no target link libraries we need to set up the C++ standard ourselves
+    # If we have no target link libraries or rely on just boost (which is
+    # header-only and, until CMake 3.20 or later, we cannot set a C++ standard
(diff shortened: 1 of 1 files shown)


examples/CMakeLists.txt
```diff
@@ -74,10 +74,14 @@ MACRO(IBAMR_ADD_EXAMPLE)
...
     TARGET_INCLUDE_DIRECTORIES("${EX_TARGET_NAME}" PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}")
 
-    IF("${EX_LINK_TARGETS}" STREQUAL "")
-      # If we have no target link libraries we need to set up the C++ standard ourselves
+    # If we have no target link libraries or rely on just boost (which is
+    # header-only and, until CMake 3.20 or later, we cannot set a C++ standard
+    # on) we need to set up the C++ standard ourselves
+    IF("${EX_LINK_TARGETS}" STREQUAL "" OR
+       "${EX_LINK_TARGETS}" STREQUAL "BOOST_INTERFACE")
       TARGET_COMPILE_FEATURES("${EX_TARGET_NAME}" PUBLIC cxx_std_11)
```
boyceg, line 82: Random question --- is the Cmake build system set up so that users can request  a standard other than C++11? Like can folks request C++17 or GNU++11?

boyceg (approved):

examples/CMakeLists.txt
drwells: Not directly. Setting this manually with `-DCMAKE_CXX_FLAGS="-std=c++17"` should work.

PR #1417, opened 2022-02-11
Fix another MPI compiler wrapper case.

drwells:
Another build system fix - thanks to @jabrown893 and @labdala for helping me track this down.

It is well-known that CMake does not handle MPI compiler wrappers well. In particular, if we provide our own compiler wrappers and have a system copy of MPI installed, then CMake will use both, causing a disaster. This patch tries to detect if we are using an MPI compiler wrapper and if so it sets `MPI_ROOT` to the directory up from the executables.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -156,6 +156,35 @@ ENDIF()
 #
 MESSAGE(STATUS "")
 MESSAGE(STATUS "Setting up MPI")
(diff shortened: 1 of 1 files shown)


drwells:
While this is a complex fix, it preserves the present `check that all dependencies link against the same version of MPI' check with the MPI compiler wrappers, which is definitely work preserving.

boyceg (approved):

PR #1428, opened 2022-02-19
Added a snapshot cache that can store snapshots.

abarret:
This introduces a class that can store snapshots, and adds a function that can interpolate in time between snapshots.

This is part of #1408. I'm breaking up what would be a much larger commit into smaller PRs.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,203 @@
...
+#include <string>
+#include <vector>
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBTK
+{
+/*!
+ * \brief Class SnapshotCache provides a method of storing snapshots. This class can also be used to determine
+ * intermediate velocity fields when at a periodic steady state.
```
boyceg, line 45:
```suggestion
 * intermediate velocity fields between snapshots. A periodic time interval can be specified for time-periodic cases.
```

boyceg: BTW, I don't see in the class definition where the periodic interval is specified.

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,203 @@
...
+                     double time,
+                     SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > current_hierarchy,
+                     const std::string& snapshot_refine_type,
+                     double tol = 1.0e-8);
+    //@}
+
+    /*!
+     * Get the number of stored snapshots.
+     */
+    size_t getNumSnapshots() const
```
boyceg, line 122:
```suggestion
    inline size_t getNumSnapshots() const
```

ibtk/include/ibtk/snapshot_utilities.h
```diff
@@ -0,0 +1,63 @@
...
+/*!
+ * This function interpolates in time between two snapshots. We find the two closest time points stored in the snapshot
+ * cache and linear interpolate in time between them. Each snapshot will be transferred onto the provided hierarchy. A
+ * scratch patch index with the same patch layout as u_idx must be provided, and data for both must be allocated. The
+ * refine_type must be a valid refinement operation for the data layout.
+ *
+ * This assumes that the time is between two snapshots stored in the cache. An error will occur if two time points can
+ * not be found or if there is a single snapshot, if the time value does not match what's stored in the cache.
+ *
+ * The hier_data_ops object, if provided, must match the same variable type used in for the snapshots.
```
boyceg, line 47: Should this class handle time periodic cases?

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,251 @@
...
+SnapshotCache<VariableType>::updateSnapshot(const int u_idx,
+                                            const double time,
+                                            Pointer<PatchHierarchy<NDIM> > current_hierarchy,
+                                            const double tol)
+{
+    // Make sure we are on a stored snapshot.
+    auto it =
+        std::find_if(d_idx_time_map.begin(),
+                     d_idx_time_map.end(),
+                     [time, tol](const std::pair<int, double>& t) -> bool { return std::abs(t.second - time) < tol; });
```
boyceg, line 111: Should we have a standard "fuzzy equals" (with absolute and relative tolerance variants) that we use in these situations?

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,251 @@
...
+        Pointer<PatchLevel<NDIM> > cur_level = current_hierarchy->getPatchLevel(ln);
+        Pointer<PatchLevel<NDIM> > snp_level = snapshot_hierarchy->getPatchLevel(ln);
+
+        // RefineSchedule requires that patch data be stored at the same points.
+        // We reset the snapshot patch data time to that of the current time.
+        double allocated_time =
+            cur_level->getPatch(*PatchLevel<NDIM>::Iterator(cur_level))->getPatchData(u_idx)->getTime();
+        // Note we need scratch allocated on the current hierarchy. Use the snapshot for that.
+        cur_level->allocatePatchData(snapshot_idx, allocated_time);
+        for (PatchLevel<NDIM>::Iterator p(snp_level); p; p++)
```
boyceg, line 180:
```suggestion
        snp_level->setTime(allocated_time, snapshot_idx);
```

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,251 @@
...
+        {
```
boyceg, line 181:
```suggestion
```

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,251 @@
...
+            Pointer<Patch<NDIM> > patch = snp_level->getPatch(p());
```
boyceg, line 182:
```suggestion
```

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,251 @@
...
+            Pointer<PatchData<NDIM> > data = patch->getPatchData(snapshot_idx);
```
boyceg, line 183:
```suggestion
```

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,251 @@
...
+            data->setTime(allocated_time);
```
boyceg, line 184:
```suggestion
```

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,251 @@
...
+        }
```
boyceg, line 185:
```suggestion
```

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+        // Now copy the data.
+        Pointer<RefineAlgorithm<NDIM> > refine_alg = new RefineAlgorithm<NDIM>();
+        Pointer<CartesianGridGeometry<NDIM> > grid_geom = current_hierarchy->getGridGeometry();
+        Pointer<RefineOperator<NDIM> > refine_op = grid_geom->lookupRefineOperator(snapshot_var, snapshot_refine_type);
+        refine_alg->registerRefine(u_idx, snapshot_idx, snapshot_idx, refine_op);
+        Pointer<RefineSchedule<NDIM> > schedule =
+            refine_alg->createSchedule(cur_level, snp_level, ln - 1, current_hierarchy);
```
boyceg, line 193: If these get re-used, it might be worth saving them.

boyceg:
Some minor comments. A more major one is: should these classes handle time periodic cases?

ibtk/src/utilities/SnapshotCache.cpp
abarret: The though has crossed my mind. If we're pulling a lot of snapshots out of this cache, we'll potentially be doing a lot of transfers which could be expensive. I'm not sure of a good way to save the schedule's. We would have to determine if `current_hierarchy` hasn't changed it's patch configuration.

I'd rather tackle this problem later if it becomes an issue.

ibtk/include/ibtk/SnapshotCache.h
abarret: This class knows nothing of periodicity. Apparently I forgot to remove the comments about periodic intervals.

ibtk/include/ibtk/snapshot_utilities.h
abarret: This function just interpolates between two snapshots. Although, as this is written, we wouldn't be able to interpolate between the last and first snapshot. We'll need a function that knows about periodicity.

ibtk/src/utilities/SnapshotCache.cpp
abarret: Oh yeah. I wrote those back in #1177. We should just use that.

abarret:
I added an optional parameter to the function that interpolates between snapshots. It now treats periodic steady states correctly.

boyceg (approved):
👍

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,247 @@
...
+SnapshotCache<VariableType>::fillSnapshot(const int u_idx,
+                                          Pointer<PatchHierarchy<NDIM> > hierarchy,
+                                          const int snapshot_idx,
+                                          const double time)
+{
+    // Create a snapshot of the patch hierarchy.
+    Pointer<PatchHierarchy<NDIM> > snapshot_hierarchy = hierarchy->makeRefinedPatchHierarchy(
+        d_object_name + "::SnapshotHierarchy_" + std::to_string(d_num_snapshots_stored),
+        1 /*ratio*/,
+        false /*register_for_restart*/);
```
drwells, line 211: I don't see anything in this class that supports restarts - don't we need that for computing long-time flow statistics?

tests/IBTK/snapshot_cache_01.cpp
```diff
@@ -0,0 +1,251 @@
...
+    // Create the index that we fill with data
+    Pointer<VariableType> var = new VariableType(test_name);
+    auto var_db = VariableDatabase<NDIM>::getDatabase();
+    int var_idx = var_db->registerVariableAndContext(var, var_db->getContext("ctx"), 1 /*ghosts*/);
+    int err_idx = var_db->registerVariableAndContext(var, var_db->getContext("err"));
+
+    // Create a hierarchy interpolator.
+    SnapshotCache<VariableType> snapshot_cache(test_name + "::SnapshotCache");
+    auto hier_math_ops = HierarchyDataOpsManager<NDIM>::getManager();
+    Pointer<HierarchyDataOpsReal<NDIM, double> > hier_data_ops = hier_math_ops->getOperationsDouble(var, new_hierarchy);
```
drwells, line 202:
```suggestion
    Pointer<HierarchyDataOpsReal<NDIM, double> > hier_data_ops = hier_math_ops->getOperationsDouble(var, new_hierarchy, true);
```
We always want unique math ops to avoid leaks.

ibtk/src/utilities/snapshot_utilities.cpp
```diff
@@ -0,0 +1,132 @@
...
+        t_low = snapshot_time_low;
+    }
+    // Create a valid hier_data_ops object
+    if (!hier_data_ops)
+    {
+        auto hier_math_ops = HierarchyDataOpsManager<NDIM>::getManager();
+        auto var_db = VariableDatabase<NDIM>::getDatabase();
+        Pointer<Variable<NDIM> > var;
+        var_db->mapIndexToVariable(u_idx, var);
+        Pointer<HierarchyDataOpsReal<NDIM, double> > hier_data_ops = hier_math_ops->getOperationsDouble(var, hierarchy);
```
drwells, line 75:
```suggestion
        Pointer<HierarchyDataOpsReal<NDIM, double> > hier_data_ops = hier_math_ops->getOperationsDouble(var, hierarchy, true);
```

drwells (changes requested):
I have a few comments.

One more thing - this doesn't work with autotools yet.

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,247 @@
...
+                                         Pointer<PatchHierarchy<NDIM> > current_hierarchy)
+{
+    ++d_num_snapshots_stored;
+    // Push back a new variable
+    Pointer<VariableType> snapshot_var =
+        new VariableType(d_object_name + "::SnapshotVar_" + std::to_string(d_num_snapshots_stored), d_depth);
+    auto var_db = VariableDatabase<NDIM>::getDatabase();
+    int snapshot_idx = var_db->registerVariableAndContext(snapshot_var, d_ctx, d_gcw);
+    d_snapshot_idxs.push_back(snapshot_idx);
+    d_snapshot_vars[snapshot_idx] = snapshot_var;
```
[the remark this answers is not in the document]
abarret: We can use the same variable from the provided index, but we do need a new context. We need to specify a potentially different ghost cell width.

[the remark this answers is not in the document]
drwells: Got it. If we have a separate hierarchy per snapshot then would just reusing the same patch data index and variable on  each hierarchy work?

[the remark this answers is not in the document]
abarret: Because we have our own copy of the patch hierarchy, I suppose we could avoid the `VariableDatabase` altogether and manage the `PatchDescriptor` and `PatchDataFactory` objects directly.

ibtk/src/utilities/SnapshotCache.cpp
abarret: That would probably be useful

ibtk/src/utilities/SnapshotCache.cpp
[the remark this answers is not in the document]
boyceg: Avoiding `VariableDatabase` sounds like a good thing to do, particularly for "internal variables".

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,202 @@
...
+/*!
+ * \brief Class SnapshotCache provides a method of storing snapshots of patch data on a snapshot of the patch hierarchy.
+ *
+ * This class is templated over VariableType, which should be one of the SAMRAI variable types. This currently supports
+ * double data at face centerings cell, face, side, edge, and node.
+ *
+ * When retrieving or storing data with this class, the same patch data layout must be used, e.g. you must consistently
+ * store/retrieve CellData<NDIM, double> with depth of 2.
+ */
+template <class VariableType>
```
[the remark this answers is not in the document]
abarret: If we clone the patch data factory for the prescribed index, I think we can get rid of templates all-together. Then this class will be able to take snapshots with different data centerings, which seems both problematic and useful.

ibtk/src/utilities/snapshot_utilities.cpp
boyceg: Should we make our own function to get these things that always generates unique objects?

abarret:
> @abarret Lets get this done. Can you address all the remaining issues in time for the next release?

I should be able to do that.

ibtk/src/utilities/SnapshotCache.cpp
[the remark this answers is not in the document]
abarret: My understanding is that `PatchData` is allocated by a `PatchDataFactory` object, which is in turn handled by the `PatchDescriptor` object. This object is passed from the patch hierarchy to the patch level to the individual patches so that each patch has the same `PatchDescriptor` object. On construction, the `PatchHierarchy` object uses the `PatchDescriptor` created by the `VariableDatabase`, and there's no method to change which `PatchDescriptor` a `PatchHierarchy` uses.

I see two paths forward:

1. Avoid `VariableDatabase` and `PatchHierarchy` and manage the snapshot `PatchLevel`'s directly. This class would have to carry around the different `PatchLevel` objects and a `PatchDescriptor` object, but we wouldn't have to use `VariableDatabase`. Writing restart files might also be a bit tricky.
2. Use `VariableDatabase` to create one variable and context pairs per instance of `SnapshotCache`.

@boyceg and @drwells, do you have an opinion?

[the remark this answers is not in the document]
boyceg: Option 2 seems like the simplest.

However, from the standpoint of having an accurate representation of the data, would it make sense to actually store the `PatchHierarchy` for the snapshotted-configuration, so that the snapshot is not slowly degraded by dynamic mesh refinement/coarsening? To do this, it would be useful to have an internal `PatchHierarchy`/`PatchLevel` that _only_ has data allocated for the captured data.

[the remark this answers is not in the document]
abarret: That's what this is currently set up to do. We store a snapshot of the patch hierarchy and only allocate data for the captured patch index of that snapshot.

We were trying to avoid having to use the `VariableDatabase` to get a patch index to allocate data on the snapshotted patch hierarchy. The issue is an instance of `PatchHierarchy` can only use the `PatchDescriptor` object managed by `VariableDatabase`. So the options are we store a `PatchHierarchy` object (which contains a snapshot of the patch configuration) and use `VariableDatabase` to get a patch index to store data on. Or we store the `PatchLevel`s ourselves and manage our own `PatchDescriptor` object to give us a patch index.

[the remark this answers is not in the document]
drwells: I think the best way forward is to just create one `PatchHierarchy` per snapshot. I think we all agree on this.

In general its just not possible to avoid the variable database - at some point we need to map between indices and `Variable` objects and that is what it does.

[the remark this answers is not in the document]
abarret: If we want to use a `PatchHierarchy` object, then we have to use the `VariableDatabase`. We could add our own index to the `PatchDescriptor` object, but that's functionally the same as using `VariableDatabase`.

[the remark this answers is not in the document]
boyceg: OK --- I think it seems easiest to manage `PatchHierarchy` and use `VariableDatabase`.

abarret:
I've removed the templates. The data layout is now specified in the constructor by a `Variable` which is also used to construct an internal patch index. This is also set up for restarts.

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,226 @@
...
+
+    /*!
+     * Clear the snapshots stored in this object. This deallocates all patch indices stored by this object.
+     */
+    void clearSnapshots();
+
+    /*!
+     * Create a snapshot. This creates a snapshot of the current hierarchy, and copies that data from this patch index
+     * into a patch index maintained by this object. If you want to overwrite a snapshot, you should call
+     * updateSnapshot() instead. Data is not allocated in this class until this function is called.
```
boyceg, line 89: If you call "set" twice (instead of "set" and then "update"), is there an error?

boyceg: Related questions:

1. If you want to capture several fields at the same time and on the same patch hierarchy, do you call "set" once and then "update" for the rest? What happens if you call "set" twice?
2. What happens if you want to capture fields at the same time but on different patch hierarchies? Does the implementation assume a single patch hierarchy configuration is used for each time point?

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,226 @@
...
+    /*!
+     * Transfer the snapshot at the specified time value within a given tolerance to the patch index u_idx on the
+     * supplied patch hierarchy. The patch index u_idx must contain sufficient ghost cell width to perform the
+     * operations used by the refinement operator.
+     *
+     * Note this can cause a significant amount of communication if the supplied patch hierarchy has a different
+     * configuration of patches than the snapshot patch hierarchy.
+     *
+     * This function does not synchronize the data on the current hierarchy (i.e. no coarsening is performed). It
+     * transfers the snapshot using the refinement operator.
```
boyceg, line 117: Using which refinement operator?

boyceg: A couple things to consider:
1. Different data centerings will need different refine operators
2. Depending on the patch configuration, you may need to coarsen the snapshot for the refine operator to work correctly --- this might be something that you state is required of the caller before saving the snapshot, or it could be specified in the method to store the snapshot

abarret: The default refinement operator is read in via the input database in the constructor. Otherwise, you can specify which refinement operator to use in the `getSnapshot` function.

I'll include a statement that the snapshot should be synchronized prior to being stored.

ibtk/include/ibtk/SnapshotCache.h
abarret: This is a little tricky, because we are storing times as doubles, there's always a question of machine precision. I'll outline the possibilities:

1. If the user calls `setSnapshot` and the time point is **exactly** equal to a previous time point, then there should be no error, and the snapshot should be updated to whatever argument was provided by the function.
2. If the user calls `setSnapshot` and the time point is **different** (including machine precision), then an additional snapshot is taken and stored at that new time point.
3. If the user calls `updateSnapshot` and the time point is **within the specified tolerance**, then the snapshot at the found time point is updated.
4. If the user calls `updateSnapshot` and the time point is **not within the specified tolerance**, an error will occur.

abarret: This class is only designed to store one snapshot of one variable at various points in time. It should be relatively straightforward to store one snapshot of multiple variables at various points in time. All that's required is wrapping `vector`'s around the objects and adding some logic to determine which variable to update.

boyceg: > This is a little tricky, because we are storing times as doubles, there's always a question of machine precision. I'll outline the possibilities:
> 
> 1. If the user calls `setSnapshot` and the time point is **exactly** equal to a previous time point, then there should be no error, and the snapshot should be updated to whatever argument was provided by the function.
> 2. If the user calls `setSnapshot` and the time point is **different** (including machine precision), then an additional snapshot is taken and stored at that new time point.
> 3. If the user calls `updateSnapshot` and the time point is **within the specified tolerance**, then the snapshot at the found time point is updated.
> 4. If the user calls `updateSnapshot` and the time point is **not within the specified tolerance**, an error will occur.

I think it makes sense to have both `setSnapshot` and `updateSnapshot` take an optional tolerance with a reasonable default (sqrt of machine precision?).

boyceg: > This class is only designed to store one snapshot of one variable at various points in time. It should be relatively straightforward to store one snapshot of multiple variables at various points in time. All that's required is wrapping `vector`'s around the objects and adding some logic to determine which variable to update.

I think it is fine to restrict to a single field.

One other thing --- does the interface allow us to look up the patch hierarchy that the snapshot was stored on?

ibtk/include/ibtk/SnapshotCache.h
boyceg: It might be good to provide a convenience function that will do the synchronization if needed --- I think we are just talking about coarsening from finest to coarsest levels.

abarret: Sure. I can include that.

ibtk/include/ibtk/SnapshotCache.h
abarret: > I think it makes sense to have both setSnapshot and updateSnapshot take an optional tolerance with a reasonable default (sqrt of machine precision?).

Assuming the `PatchHierarchy` automatically deletes the data allocated when it's destroyed (which should happen), there will be functionally no difference between these calls. We can just have a single function `storeSnapshot` which has the behavior

1. If the user calls `storeSnapshot` and a stored time point is within the specified tolerance, then the snapshot at the found time point is updated.
2. If the user calls `storeSnapshot` and a stored time point is not within the specified tolerance, then an additional snapshot is taken and stored at that new time point.

> One other thing --- does the interface allow us to look up the patch hierarchy that the snapshot was stored on?

Not currently, but that's an easy function to add.

abarret:
Another question, how should this class handle errors? Currently, if a user attempts to look up a snapshot that isn't stored, an error will occur. Should we start throwing exceptions instead of terminating the program?

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,226 @@
...
+class SnapshotCache : SAMRAI::tbox::Serializable
+{
+public:
+    /*!
+     * The constructor for class SnapshotCache sets some default values, reads in configuration information from input
+     * databases, and registers a variable/context pair with the VariableDatabase. Can optionally set up this class for
+     * restarts.
+     *
+     * The optinal input database is searched for the following keys
+     * - 'refine_type' : String that is the default refine operation for transfering data between hierarchies.
```
drwells, line 64: Could you provide some advice here on good values for this parameter?

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,226 @@
...
+     * - 'gcw' : Integer array that is the ghost cell width with which the internal variable is set up.
+     *
+     * \note If this object is being created from a restart, the grid geometry used the patch hierarchy must be
+     * supplied. An error will be thrown if the grid geometry is an invalid pointer.
```
drwells, line 68: Lets just make this unconditionally required. We should try to avoid the possibility of different code paths for restart and not restart runs.

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,226 @@
...
+     */
+    SnapshotCache(std::string object_name,
+                  SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > var,
+                  SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db = nullptr,
```
drwells, line 72: Similarly, unless there are reasonable defaults we can use 100% of the time, this argument should be mandatory.

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,226 @@
...
+     * Note this can cause a significant amount of communication if the supplied patch hierarchy has a different
+     * configuration of patches than the snapshot patch hierarchy.
+     *
+     * This function does not synchronize the data on the current hierarchy (i.e. no coarsening is performed). It
+     * transfers the snapshot using the refinement operator.
+     */
+    //@{
+    void getSnapshot(int u_idx,
+                     double time,
+                     SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > current_hierarchy,
```
drwells, line 122: These functions should just return a `tbox::Pointer<hier::PatchHierarchy>` by value.

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,226 @@
...
+    std::set<double> d_snapshot_times;
+
+    /*!
+     * We store a copy of the patch hierarchy at which the snapshot is stored. As needed, we will transfer data
+     * from the snapshot hierarchy to the provided hierarchy.
+     *
+     * We use a map with a double as a key. This means only values stored in d_snapshot_times should be used to access
+     * an element in this map, otherwise we may not find an element because of finite precision.
+     */
+    std::map<double, SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > > d_snapshot_hierarchies;
```
drwells, line 218: I don't think its a good idea to store the keys in two places at once - we might update one and not the other.

A better alternative would be to store
```cpp
std::vector<std::pair<double, tbox::Pointer<hier::PatchHierarchy>>
```
instead and keep that sorted ourselves.

drwells (changes requested):
How we store the snapshots (and the question of numerical tolerances) depends on how we want to use snapshots. I think it would be better to store the data contiguously and return 'closest' snapshots by using `std::lower_bound()` than providing a tolerance - the caller can then decide from there if they want to use a different snapshot. We could also provide `begin()` and `end()` and let the user write this stuff themselves too.

ibtk/include/ibtk/SnapshotCache.h
abarret: The hierarchy is not a return here. We interpolate the snapshot from the snapshotted hierarchy onto `current_hierarchy` and store it in `u_idx`.

abarret:
I suppose one option is to turn this class into a strict cache that only stores snapshots and does no interpolation. We can write convenience functions that, when given an iterator and patch hierarchy, interpolates the data onto the provided patch hierarchy.

drwells:
Part of my confusion comes from just not understanding some of the code.

I think it would be better to keep this as a pure snapshot cache and move the interpolation functions somewhere else.

boyceg:
I think that most use cases would involve evaluating the data on a different hierarchy than they were stored on originally, and so combining lookup with interpolation makes sense to me, particularly if there is a mechanism to obtain the data on the original hierarchy as well.

On the other hand, for interpolation, could something like `HierarchyGhostCellInterpolation` (or a similar class) do the job?

To @drwells 's point --- it would be more flexible to have the snapshot class "just take snapshots" (e.g. copy data onto a stand-alone patch hierarchy) and not do anything w.r.t. coarsening or refining the captured data. That gives more control down the line for any decisions that need to be made about how to evaluate the field on a different grid.

If this class doesn't provide a data transfer capability itself, then it really should be essentially a "one liner" to do it outside of this class.

abarret:
> On the other hand, for interpolation, could something like `HierarchyGhostCellInterpolation` (or a similar class) do the job?

This class won't work because it's specialized to only use a single `PatchHierarchy`. In general, we will have two patch hierarchies and we want to transfer from one to the other. But if this is something that might be useful to do generally, we can introduce another class that can interpolate between hierarchies.

> If this class doesn't provide a data transfer capability itself, then it really should be essentially a "one liner" to do it outside of this class.

I think we can just pull out the functions that interpolate between hierarchies and make them standalone functions.

abarret:
@drwells take a look and let me know what you think. `SnapshotCache` now only stores snapshots and retrieval of snapshots occurs by returning iterators. There are a couple of functions that can retrieve data from the `SnapshotCache` and interpolate it onto a different `PatchHierachy` object.

If the design looks ok, I'll start updating the tests.

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,285 @@
...
+    SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > d_snapshot_var;
+    SAMRAI::tbox::Pointer<SAMRAI::hier::VariableContext> d_ctx;
+    int d_snapshot_idx = IBTK::invalid_index;
+    SAMRAI::hier::IntVector<NDIM> d_gcw = 1;
+
+    /*!
+     * Vector of snapshots consisting of time points and hierarchies. Each hierarchy allocates data only with the patch
+     * data index stored by this class.
+     */
+    std::vector<SnapshotType> d_snapshots;
```
drwells, line 221: If we are returning iterators anywhere then this is not going to work since when `std::vector` grows it must allocate a completely new array. If we can guarantee that we never delete snapshots in the middle of the range then either `std::deque<T>` or `std::vector<std::unique_ptr<T>>` is the right choice for avoiding this problem.

ibtk/include/ibtk/SnapshotCache.h
```diff
@@ -0,0 +1,285 @@
...
+ *
+ * This function does not synchronize the data on the current hierarchy (i.e. no coarsening is performed). It
+ * transfers the snapshot using the refinement operator.
+ */
+void fillSnapshotOnHierarchy(const SnapshotCache& cache,
+                             int u_idx,
+                             double time,
+                             SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > current_hierarchy,
+                             const std::string& snapshot_refine_type,
+                             double tol = 1.0e-8);
```
drwells, line 256: Shouldn't this now be a utility function?

ibtk/include/ibtk/SnapshotCache.h
abarret: The storage type is a double and a pointer. Resizing a vector will be very cheap already. If you're concerned about iterator invalidation, `std::deque` will have also have iterators invalidated when an element is inserted. I'm not sure I understand the benefit of changing to `deque`.

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,359 @@
...
+    var_db->removePatchDataIndex(d_snapshot_idx);
+    d_snapshot_var = nullptr;
+    d_snapshot_idx = IBTK::invalid_index;
+}
+
+SnapshotCache::const_iterator
+SnapshotCache::getSnapshot(double time, double tol) const
+{
+    return std::find_if(d_snapshots.begin(), d_snapshots.end(), [time, tol](const SnapshotType& t) -> bool {
+        return IBTK::abs_equal_eps(t.first, time, tol);
```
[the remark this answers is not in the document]
abarret: `std::find_if` does return an iterator.

ibtk/src/utilities/SnapshotCache.cpp
```diff
@@ -0,0 +1,359 @@
...
+        restart_manager->registerRestartItem(d_object_name, this);
+        var_db->registerPatchDataForRestart(d_snapshot_idx);
+        d_registered_for_restart = true;
+    }
+    bool from_restart = restart_manager->isFromRestart();
+    if (from_restart)
+    {
+        if (!grid_geom) TBOX_ERROR("SnapshotCache needs the grid geometry for restarts!\n");
+        getFromRestart(grid_geom);
+    }
```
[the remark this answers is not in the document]
abarret: `grid_geom` is a required argument, but we only check that it's valid before it's used. Would you rather always check that it's valid, even if we don't use it?

ibtk/include/ibtk/SnapshotCache.h
abarret: I wasn't sure if we had a set style for placing stand-alone functions. I can pull these into a separate file. Although I thought we might want to let the compiler do it's job best and put all functions based on `SnapshotCache` into the same translation unit.

drwells: we typically put these things in files like `ibmesh_utilities.h`.

ibtk/include/ibtk/SnapshotCache.h
drwells: Resizing on its own is definitely not a problem, just the resulting iterator invalidation.

The point of changing to `std::deque` is that if we never need to insert a snapshot into the middle (i.e., we only a ever append or prepend) then these iterators are always valid. I imagine that, in practice, this is always what we will do.

If we want to support that case then we really do need to not return iterators and use `std::vector<std::unique_ptr<T>>` since the iterators will be invalidated every time we add enough snapshots. That combination (no iterators and unique pointers) will always work.

ibtk/src/utilities/SnapshotCache.cpp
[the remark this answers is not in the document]
drwells: yup

ibtk/include/ibtk/SnapshotCache.h
drwells: that's a good idea - lets put that in `ibtk_utilities.h`.

We could also require C++14, but sometimes people still want to build IBAMR with GCC 4.x, so lets do C++11 for at least one more release.

abarret:
@drwells `getSnapshot` now returns a pointer to the element. We still have iterators that can be invalidated via the `begin()` and `end()` methods. The tests still aren't updated. Let me know what you think.

CMakeLists.txt
```diff
@@ -147,6 +147,20 @@ IF("${IBAMR_FORCE_BUNDLED_MUPARSER}" STREQUAL "")
...
+CHECK_CXX_SOURCE_COMPILES(
+	"
+	#include <memory>
+	int main()
+	{
+	  auto ptr = std::make_unique<int>(42);
+	  return 0;
+	}
+	"
+	IBAMR_HAVE_MAKE_UNIQUE)
```
drwells, line 162: Lets not do this - we should just use our own version of this function everywhere.

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -119,6 +119,18 @@ static const bool ENABLE_TIMERS = true;
...
+#define ibtk_make_unique std::make_unique
+#else
+template <typename T, typename... Args>
+std::unique_ptr<T>
+make_unique(Args&&... args)
+{
+    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+#define ibtk_make_unique IBTK::make_unique
+#endif
```
drwells, line 132: Same here - we should just have `IBTK::make_unique()` and we can deprecate that at some point when we move to C++14.

drwells (changes requested):
Looks good - we should avoid adding macros and extra version checks if we can, though.

abarret:
> This is fine as-is, but I thought about it some more and there is one further change we should consider - we should just return the pairs by value (i.e., never return references - its just a pointer and a double precision number, after all) and then get rid of all the `unique_ptr` code. That would make this a lot easier to work with and avoid the problem with iterator invalidation much more cleanly. What do you think?

That's fine with me. There's an issue of returning an invalid snapshot via `getSnapshot`, but we can just return something like `numeric_limits<double>::max` and `nullptr`.

abarret:
@drwells I've removed `unique_ptr` from the class. Let me know what you think, and then I can squash this before we merge.

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -119,6 +120,13 @@ static const bool ENABLE_TIMERS = true;
 
 namespace IBTK
 {
+template <typename T, typename... Args>
+std::unique_ptr<T>
+make_unique(Args&&... args)
+{
+    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
+
```
drwells, line 129: We can get rid of this now

drwells (changes requested):
We need to clean up the free functions but once that's done this should be good to go.

abarret:
I've removed `make_unique` and fixed the documentation. I've also squashed the commits. This is ready for a final review.

ibtk/include/ibtk/snapshot_utilities.h
```diff
@@ -0,0 +1,94 @@
...
+/*!
+ * Update a snapshot at the given time. This function copies the data stored in u_idx into the snapshot index in the
+ * cache object. The snapshotted hierarchy is also updated to the current hierarchy.
+ *
+ * This outputs an error if a snapshot with that time point within the specified tolerance can not be found.
+ *
+ * Calls to this function may use a different patch index than the one used in setSnapshot(), but the underlying
+ * data layout must be consistent. This data layout is specified by the variable used to construct the cache.
+ */
+void update_snapshot(const SnapshotCache& cache,
```
drwells, line 45: Now that we updated the documentation I actually understand what is going on.

This function modifies the input `SnapshotCache`, right? If so, even though this works (since we modify things it owns via pointers) we should not mark it as `const` since the object is being modified. The same change should be made every time we are modifying the snapshot cache.

abarret: What about the functions in the class itself? The iterators return a pointer to the patch hierarchy, which could in theory change. Should those not be marked as `const`? Also the getter function that returns the patch index used by the class. Should that be marked as `const`?

drwells: I suspect the answer is "don't use const with SAMRAI" since doing things by pointers evades, for language reasons, constness checks. Honestly I'd probably just remove const everywhere when dealing with `tbox::Pointer`s.

abarret: In that case, should we change all these functions to take `shared_ptr`s instead (or maybe `tbox::Pointer` and switch to shared whenever we go to SAMRAI 4)?

drwells: It probably isn't worth it. As soon as pointers get involved its just too hard to maintain const-correctness - that's a problem with any PatchHierarchy, regardless of how we get it (const or not).

abarret: I don't think I was clear. I mean instead of references to `SnapshotCache`, should these external functions take `std::shared_ptr<SnapshotCache>` as arguments?

abarret: I can also remove `const` from the class definitions.

drwells: No - I don't see why we should limit ourselves to shared pointers managing `SnapshotCache`s. Is there some advantage to this?

abarret: No. It's just what SAMRAI 4 does.

abarret: Fixed

drwells (approved):

PR #1432, opened 2022-02-23
Fix sccache population

drwells:
If you look at the log files from the last PR you can see that sccache never gets used. I'm trying to fix that now.

### DIFF
--- a/.github/workflows/push-pull.yml
+++ b/.github/workflows/push-pull.yml
@@ -23,6 +23,7 @@ on:
       - master
 
 jobs:
+  # build IBAMR with CMake and run tests
   build_ibamr:
     runs-on: ubuntu-latest
     name: Build IBAMR and run tests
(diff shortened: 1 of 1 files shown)


.github/workflows/push-pull.yml
```diff
@@ -43,52 +44,58 @@ jobs:
...
+          path: ~/.cache/sccache
+          # A fundamental problem with the cache is that it is identified by a
+          # single key and, should a cache with that key already exist, the
+          # cache will not be updated. This makes sense for things like NPM but
+          # not so much for sccache, since we get a new and unique cache each
+          # time we rebuild the library with a new branch.
+          #
+          # To get around this, always make the key unique and rely on
+          # restore-keys to find the highest-indexed key available. Explicitly
+          # try to restore things from the most recent builds.
```
boyceg, line 75: This is trying to avoid recompiling stuff when possible?

How do you know the key corresponds to the current build?

Any reason for 5 keys?

boyceg (approved):

.github/workflows/push-pull.yml
drwells: 1. Yes.

2. It won't. The key I pick depends on the build number in GitHub's CI, which should be new. 

3. Its a wild guess. The point here is that I try to find the most recent of the last five builds to look for cached compiler results. I wanted more than one so that if we start, say, 4 builds at once, then the last one we start will still try to find something reasonable.

This is a really hacky part of the CI (which is probably why it didn't work before). The intention of this GitHub action is for storing dependencies, not build artifacts, so we have to provide some extra information to make it do what we want (find a recently compiled copy of IBAMR and try to reuse as many object files as possible).

PR #1433, opened 2022-02-23
CI: error out on warnings.

drwells:
Another CI patch: error out if there are warnings. Lets see what happens.


drwells:
I had to fix some interesting warnings but we now compile cleanly (and fail to compile when there are warnings).

drwells:
There are a few other small CI patches I'd like to get in. Can someone take a look and approve/disapprove?

abarret (approved):

Issue #1355, opened 2021-09-07
Put together another docker container that uses all bundled dependencies


drwells:
It looks like I broke `master` with the bundled copy of Eigen in #1344 (I am working on a patch). We should prevent this in the future by putting together a second docker container that uses bundled dependencies (and the oldest compiler toolchain we can find).

drwells:
Patch is in #1356.

drwells:
This is technically fixed by https://github.com/IBAMR/docker-files/blob/master/dependencies-0.10.1/Dockerfile, which uses all bundled dependencies. I'll consider this closed once we switch over to the new image here:

https://hub.docker.com/r/wellsd2/ibamr/tags

drwells:
Fixed by #1435.

PR #1435, opened 2022-02-25
Switch to using a new docker image.

drwells:
This is likely to explode so let me run it here first.


drwells:
It took some iterations but this is now much improved. A few advantages of this new setup:
1. The new docker image is much smaller (circa 250 MB instead of 1.3 G)
2. we use all bundled dependencies (fixing #1355)
3. the cache numbering scheme has been further improved - we won't accidentally use autotools artifacts when trying to compile with CMake

drwells:
getting the caching working correctly saves us a ton of time: the autotools build finished in 16 minutes because it didn't need to recompile anything. We only spend about 5 minutes actually executing tests with the CMake builds, so those take about 85 minutes to compile the library, tests, and examples: most of that work is repeated from previous runs if the cache is not populated correctly.

.github/workflows/push-pull.yml
```diff
@@ -73,26 +77,40 @@ jobs:
...
-            ${{ runner.os }}-build-${{ steps.keys.outputs.key3 }}
-            ${{ runner.os }}-build-${{ steps.keys.outputs.key4 }}
-            ${{ runner.os }}-build-${{ steps.keys.outputs.key5 }}
-            ${{ runner.os }}-build-
+            ${{ runner.os }}-build-cmake-${{ steps.keys.outputs.key1 }}
+            ${{ runner.os }}-build-cmake-${{ steps.keys.outputs.key2 }}
+            ${{ runner.os }}-build-cmake-${{ steps.keys.outputs.key3 }}
+            ${{ runner.os }}-build-cmake-${{ steps.keys.outputs.key4 }}
+            ${{ runner.os }}-build-cmake-${{ steps.keys.outputs.key5 }}
+            ${{ runner.os }}-build-cmake-${{ steps.keys.outputs.key10 }}
```
boyceg, line 87: Why does this jump from 5 to 10?

boyceg (approved):

.github/workflows/push-pull.yml
drwells: Originally I thought it might help to provide a specific key to look for in case the last five don't yet exist (like if we do some combination of creating PRs and merging PRs that starts five or more jobs at once) but in that case GitHub should default to just picking the next newest key (though this isn't documented; I'm guessing how they decide which key is best). I don't have any evidence that picking 'ten keys ago' will work any better than leaving GitHub to its own choices so I'll delete that line.

Issue #1301, opened 2021-07-16
Augment an example to track a few IB point coordinates


drwells:
This came up in the group meeting this morning - we should add an example for this that incorporates Aaron's old notes and shows how to do this.

Cole - I don't think you have a GitHub account, but if you want to give this a try let me know!

drwells:
See also 
[ibamr_lag_data.pdf](https://github.com/IBAMR/IBAMR/files/6831436/ibamr_lag_data.pdf)

PR #1437, opened 2022-02-28
Added new example which demonstrates how to track Eulerian IB coordin…

colegruninger97:
This example seeks to satisfy issue #1301 .

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


examples/navier_stokes/ex7/Makefile
```diff
@@ -0,0 +1,33 @@
...
+OBJS = $(SRC:%.cpp=%.o) $(IBAMR_LIB_3D) $(IBTK_LIB_3D)
+main3d: $(OBJS)
+	$(CXX) $(CXXFLAGS) $(LDFLAGS) $(OBJS) $(LDFLAGS) $(LIBS) -DNDIM=$(PDIM) -o main3d
+
+clean:
+	$(RM) main3d
+	$(RM) *.o *.lo *.objs *.ii *.int.c
+	$(RM) -r .libs
+
+-include $(SRC:%.cpp=%.d)
```
drwells, line 33: The examples are built as part of IBAMR itself, not as independent programs. The build system as a whole needs to be aware of what's going on. Unfortunately this is not trivial to do unless you have prior experience doing it.

If you're willing, I'd still like you to work out how to do this instead of relying on one of the principal developers to write the correct code to connect this to the rest of IBAMR. Could you please look at how each `Makefile.am` is set up in 

https://github.com/IBAMR/IBAMR/commit/c0123bf893e91efe03494f3ed349932df87ff16e

and make the corresponding changes here? Once we add the automake files (the files ending in Makefile.am) and update the one in `navier_stokes` we should instruct the build system to use them by adding the correct lines to `configure.ac`. Once those files are in we can use the standard autotools rigamarole to add the correct `Makefile.in`s and appropriate changes to `configure` (those files are always automatically generated for us). I'll show you how to do that once we set up the non-generated files first.

Once that is done we get to add this example with CMake, which is a bit easier.

drwells (changes requested):
I have a few comments for getting us started. The primary problem to work on now is hooking this into the build system.

examples/navier_stokes/ex7/Makefile
abarret: Also, should this be a `navier_stokes` example or an `IB` example?

colegruninger97: Hmmmm... I think it might make a bit more sense for this to be an `IB` example. I'll implement that change as well.

drwells:
This is a good start. Here's what we need to do now to get this working with the build system:

### autotools
- We need to generate a `configure` script that is aware of the new automake file. To do this, list the new automake file in `configure.ac` (should be line 190).
- Regenerate `configure` with the correct version of autotools. If you execute the `bootstrap` script this should be done for you automatically (that script downloads and compiles autotools and then uses that to regenerate `configure`)
- Add any new `Makefile.in`s to the branch as well as changes to other files.

### cmake
- List the new example in `examples/IB/explicit/CMakeLists.txt`
- Add a new `CMakeLists.txt` script for `ex7`. This is pretty similar to `ex4`: start by copying and pasting `examples/IB/explicit/ex4/CMakeLists.txt` and updating the example number and list of input files.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,441 @@
...
+ *                                                                             *
+ * For restarted run, command line is:                                         *
+ *                                                                             *
+ *    executable <input file name> <restart directory> <restart number>        *
+ *                                                                             *
+ *******************************************************************************/
+
+void print_eul_data(Pointer<PatchHierarchy<NDIM> > hierarchy, LDataManager* l_data_manager);
+
+// NOTE: These values are read in below from the input file. The values listed here are the default values.
```
drwells, line 61: Could you add the `.vertex` or whatever files to the PR? We typically include those as well as the scripts that generate them.

drwells (changes requested):
More minor comments.

drwells:
@colegruninger97 Nice progress. There are two things we need to address before we proceed:
1. We do not do merge commits (outside of putting branches into master). Merging to get updates in a feature branch is always the wrong choice since it intercalates your changes with updates to the library itself, but we always want commits for a certain feature to be contiguous in history. You should rebase instead.
2. We still need to fix the merge conflict. I recommend rebasing and then resolving conflicts by running the bootstrap script to keep regenerating new `configure` scripts. Do you know how to do it?

This should only take a few minutes - let me know if you have questions.

drwells:
This looks good. There are a few minor things we need to fix.

1. The input file doesn't work for me - I get some warnings and a missing variable error:
```
[drwells@archway ex7]$ ./main3d ./input3d  
Warning in ./input3d at line 233 column 92 : Double truncated to integer
   domain_boxes = [ (0,0,0),(ASPECT_RATIO_X*N - 1,ASPECT_RATIO_Y*N - 1,ASPECT_RATIO_Z*N - 1) ]
                                                                                            ^
Warning in ./input3d at line 233 column 92 : Double truncated to integer
   domain_boxes = [ (0,0,0),(ASPECT_RATIO_X*N - 1,ASPECT_RATIO_Y*N - 1,ASPECT_RATIO_Z*N - 1) ]
                                                                                            ^
Warning in ./input3d at line 233 column 92 : Double truncated to integer
   domain_boxes = [ (0,0,0),(ASPECT_RATIO_X*N - 1,ASPECT_RATIO_Y*N - 1,ASPECT_RATIO_Z*N - 1) ]
                                                                                            ^
IBStandardInitializer:  Reading from input files.
  base filename: posts_128
  assigned to level 2 of the Cartesian grid patch hierarchy
  NOTE: UNIFORM target point stiffnesses are being employed for the structure named posts_128

MemoryDatabase: Key ``POST_FORCE'' does not exist in the database...
```
2.  Please add a 'major' changelog - the format is pretty self-descriptive. Just make a new one based on what we have in `doc/news/changes/minor` (but add it to the major folder - tutorials are always major improvements)

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,435 @@
...
+#include <ibtk/AppInitializer.h>
+#include <ibtk/IBTKInit.h>
+#include <ibtk/IBTK_MPI.h>
+#include <ibtk/LData.h>
+#include <ibtk/LDataManager.h>
+#include <ibtk/muParserCartGridFunction.h>
+#include <ibtk/muParserRobinBcCoefs.h>
+
+#include <fstream>
+#include <iostream>
```
drwells, line 41:
```suggestion
```

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+        // Cleanup Eulerian boundary condition specification objects (when
+        // necessary).
+        for (unsigned int d = 0; d < NDIM; ++d) delete u_bc_coefs[d];
+
+    } // cleanup dynamically allocated objects prior to shutdown
+} // main
+
+void
+print_eul_data(Pointer<PatchHierarchy<NDIM> > hierarchy, LDataManager* l_data_manager)
```
drwells, line 416: These are already printed to a different log file - we don't need to print them again. We should remove this function.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,435 @@
...
+                pout << "\nWriting restart files...\n\n";
+                RestartManager::getManager()->writeRestartFile(restart_dump_dirname, iteration_num);
+            }
+            if (dump_timer_data && (iteration_num % timer_dump_interval == 0 || last_step))
+            {
+                pout << "\nWriting timer data...\n\n";
+                TimerManager::getManager()->print(plog);
+            }
+            if (dump_postproc_data && (iteration_num % postproc_data_dump_interval == 0 || last_step))
+            {
```
drwells, line 404: ?

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,435 @@
...
+            loop_time += dt;
+
+            pout << "\n";
+            pout << "At end       of timestep # " << iteration_num << "\n";
+            pout << "Simulation time is " << loop_time << "\n";
+            pout << "+++++++++++++++++++++++++++++++++++++++++++++++++++\n";
+            pout << "\n";
+
+            std::ofstream fout("Coordinates.txt", std::ios::app);
+
```
drwells, line 361: We already created the output stream above:
```suggestion
```
In addition, this isn't going to work in parallel - each process will attempt to create the same text file and the program will crash. You will need to decide if we want to use one coordinate file for each process or one giant coordinate file that saves things from each process.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,435 @@
...
+        {
+            pout << "\n\nWriting visualization files...\n\n";
+            time_integrator->setupPlotData();
+            visit_data_writer->writePlotData(patch_hierarchy, iteration_num, loop_time);
+            silo_data_writer->writePlotData(iteration_num, loop_time);
+        }
+        
+        // Create Text File which Stores the locations of each IB point
+
+        ofstream fout("Coordinates.txt", ios::out);
```
drwells, line 330:
```suggestion
        std::ofstream fout("Coordinates.txt", ios::out);
```

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,435 @@
...
+            const std::vector<LNode*>& local_nodes = l_mesh->getLocalNodes();
+            for (const auto& node : local_nodes)
+            {
+                const int lag_idx = node->getLagrangianIndex();
+                const int petsc_idx = node->getLocalPETScIndex();
+                Eigen::Map<Vector3d> X(&x_vals[petsc_idx * NDIM]);
+                fout.unsetf(ios_base::showpos);
+                fout.setf(ios_base::scientific);
+                fout.precision(5);
+                fout << loop_time << "," << X(0) << "," << X(1) << "," << X(2) << "\n";
```
drwells, line 378:
```suggestion
                fout << loop_time << "," << X << "\n";
```

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,435 @@
...
+            }
+
```
drwells, line 380: We need a matching call to `VecRestoreArray()` here.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,435 @@
...
+            Pointer<LData> X_data = ib_method_ops->getLDataManager()->getLData("X", finest_ln);
+            Vec X_vec = X_data->getVec();
+            double* x_vals;
+            int ierr = VecGetArray(X_vec, &x_vals);
+            IBTK_CHKERRQ(ierr);
+            Pointer<LMesh> l_mesh = ib_method_ops->getLDataManager()->getLMesh(finest_ln);
+            const std::vector<LNode*>& local_nodes = l_mesh->getLocalNodes();
+            for (const auto& node : local_nodes)
+            {
+                const int lag_idx = node->getLagrangianIndex();
```
drwells, line 372:
```suggestion
```

examples/IB/explicit/ex7/example.cpp
drwells: For consistency we should check PETSc's return code.

History lesson: PETSc functions always return an integer indicating whether or not they succeeded (success is `0`, anything else is some sort of failure). Traditionally this is named `ierr` because in the old days of FORTRAN-77, variables starting with `i`,`j`, ... `n` were implicitly integers and those starting with everything else were implicitly floats. This function isn't going to fail but for the sake of good style we should never ignore the return code - we can use `IBTK_CHKERRQ` again down here to match the first PETSc call.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -373,6 +373,7 @@ main(int argc, char* argv[])
                 fout << loop_time << "," << X << "\n";
             }
             VecRestoreArray(X_vec,&x_vals);
```
drwells, line 375:
```suggestion
     ierr = VecRestoreArray(X_vec,&x_vals);
```

boyceg:
@colegruninger97, need any help in finishing this up?

colegruninger97:
> @colegruninger97, need any help in finishing this up?

Yeah I may need some help finishing this up. I'd rather not use MPI I/O here, but I can't seem to figure out how to open the .txt file where the coordinates are printed only on one MPI process.

drwells:
What have you tried to do? Remember, you can guard blocks by checking the rank of the current process.

colegruninger97:
This last push is my initial attempt to make the printing of the Eulerian IB coordinates compatible with multiple MPI processes. I'm afraid my code might be embarrassingly inefficient. When you get a chance, let me know what you think. In the meantime, I'll try to optimize the code a bit more.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -439,21 +439,24 @@ main(int argc, char* argv[])
...
+                    count[j] = NDIM*nodes_per_proc[j];
                     displacements[j] = sum_elem;
 
                 }
 
                 //Gather the data
 
                 MPI_Gatherv(Local_Coords.data(),Local_Coords.size(),MPI_DOUBLE,RECV_BUF.data(),count,displacements,MPI_DOUBLE,0,IBTK_MPI::getCommunicator());
+                delete [] displacements;
+                delete [] count;
```
drwells, line 459:
```suggestion
```

drwells (changes requested):
I wouldn't worry about efficiency - its just a tutorial.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,515 @@
...
+            {
+                for(int r = 1; r<n_proc; ++r)
+                {
+                    MPI_Recv(&nodes_per_proc[r],1,MPI_INT,r,0,IBTK_MPI::getCommunicator(),MPI_STATUS_IGNORE);
+                }
+            }
+
+            IBTK_MPI::barrier();
+            //initialize vector to store the nodal coordinates on each processor
+            std::vector<double> Local_Coords(num_nodes*NDIM);
```
drwells, line 408:
```suggestion
            std::vector<double> local_coordinates(num_nodes*NDIM);
```
we typically don't use capital letters for variables.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,515 @@
...
+            std::vector<double> RECV_BUF;
+            //Specify what's done on the root process
+            int total_nodes = 0;
+            if(rank == 0)
+            {
+
+                for(std::vector<int>::iterator it = nodes_per_proc.begin(); it != nodes_per_proc.end(); ++it)
+                     {
+                        total_nodes += *it;
+                     }
```
drwells, line 438:
```suggestion
                const auto total_nodes = std::accumulate(nodes_per_proc.begin(), nodes_per_proc.end(), 0);
```
same thing but shorter. Like elsewhere, this should be 'global' if it refers to everything (not total).

drwells (changes requested):
This looks good. One thing to improve is consistently naming things as local or global. Can you also run clang-format on this code?

drwells:
It looks like you need to rebase and then rerun the bootstrap script - this happens when someone else adds a new file to the (autotools) build system.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -0,0 +1,520 @@
...
+#include <ibtk/muParserCartGridFunction.h>
+#include <ibtk/muParserRobinBcCoefs.h>
+
+#include <BergerRigoutsos.h>
+#include <CartesianGridGeometry.h>
+#include <LoadBalancer.h>
+#include <StandardTagAndInitialize.h>
+#include <mpi.h>
+#include <stdio.h>
+#include <stdlib.h>
```
drwells, line 41:
```suggestion
```

drwells (approved):
Looks good - we get the same output file with different processor counts 👍🏻

PR #1439, opened 2022-03-02
Test oldest configuration

drwells:
Now that the CI is set up its relatively easy to copy and paste what we have to add more testers.

We can't catch bugs like #1425 with our present configurations since we only use relatively recent compilers. This PR adds another tester based on CENTOS 7 (first released in 2014, formally supported through 2024) which uses gcc 4.8, libmesh 1.1 and PETSc 3.7.7.

This might take a few rounds to get working correctly.

### DIFF
--- a/.github/cmake/configure_centos7.cmake
+++ b/.github/cmake/configure_centos7.cmake
@@ -1,0 +1,36 @@
+## ---------------------------------------------------------------------
+##
+## Copyright (c) 2022 - 2022 by the IBAMR developers
+## All rights reserved.
+##
+## This file is part of IBAMR.
+##
+## IBAMR is free software and is distributed under the 3-clause BSD
+## license. The full text of the license can be found in the file
+## COPYRIGHT at the top level directory of IBAMR.
+##
+## ---------------------------------------------------------------------
+
+SET(CTEST_USE_LAUNCHERS "ON" CACHE STRING "")
+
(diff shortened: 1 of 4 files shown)


drwells:
Isn't that interesting - this version of boost doesn't even work with this version of GCC. I'll put together a patch.

drwells:
This is good to go, if someone wants to take a look. Some comments:
1. The boost fixes are a bit arcane. IIRC boost will inspect compiler versions to determine which features are available and which are not - hence a lot of C++11 stuff is implemented as `BOOST_MATH_CONSTEXPR` (which expands to `constexpr` or nothing) instead of just `constexpr` since not all compilers supported `constexpr` math functions. Boost is pretty conservative in labeling these things - hence, when someone added some forward declarations in another header, they forgot to use this macro and things happened to work, but its inconsistent with the declarations elsewhere and thats not allowed and doesn't work in general (and for us).
2. CENTOS 7 is AFAICT the oldest presently supported distro release which uses gcc 4.8. I think we are compatible with 4.7 but we don't have a good way to test it. These compilers are pretty old by this point (its hard to believe the massive GCC 5 rebuild was 7 years ago!) but I don't see any reason that we shouldn't support them.
3. All we use this container for is downloading and running our own code. We can keep is around without any security concerns indefinitely.
4. libMesh now requires C++17. I don't see any reason for us to move past C++11 right now (though of course people can use whatever they want in their own projects).

boyceg (approved):

PR #1443, opened 2022-03-17
Libmesh dev compat

drwells:
We presently don't compile with the master branch of libMesh and deprecated features turned off because we rely on their old const-incorrect iterator handling (e.g., you could get an `Elem *` to a `const Mesh`, which doesn't make sense and is now fixed). I did have to change the API of `FESurfaceDistanceEvaluator`: @amneetb no examples or tests use this function: can we remove it from the public interface?

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/examples/IBFE/explicit/ex0/example.cpp
+++ b/examples/IBFE/explicit/ex0/example.cpp
@@ -548,11 +548,11 @@ main(int argc, char* argv[])
             const std::vector<std::vector<VectorValue<double> > >& dphi = fe->get_dphi();
             TensorValue<double> FF;
             boost::multi_array<double, 2> X_node;
-            const MeshBase::const_element_iterator el_begin = mesh.active_local_elements_begin();
-            const MeshBase::const_element_iterator el_end = mesh.active_local_elements_end();
-            for (MeshBase::const_element_iterator el_it = el_begin; el_it != el_end; ++el_it)
+            const auto el_begin = mesh.active_local_elements_begin();
+            const auto el_end = mesh.active_local_elements_end();
+            for (auto el_it = el_begin; el_it != el_end; ++el_it)
             {
-                Elem* const elem = *el_it;
+                const auto elem = *el_it;
                 fe->reinit(elem);
                 for (unsigned int d = 0; d < NDIM; ++d)
                 {
--- a/ibtk/src/lagrangian/FEDataManager.cpp
+++ b/ibtk/src/lagrangian/FEDataManager.cpp
@@ -3086,7 +3086,11 @@ FEDataManager::collectActivePatchElements(std::vector<std::vector<Elem*> >& acti
                                           const int finest_elem_ln)
 {
     // Get the necessary FE data.
-    const MeshBase& mesh = d_fe_data->d_es->get_mesh();
+    //
+    // We once relied on libMesh not correctly implementing iterator constness
+    // correctly. Here, we don't need to change the mesh, but we use Elem *s
+    // (not const Elem *s): hence we cannot use a const MeshBase& any more.
+    MeshBase& mesh = d_fe_data->d_es->get_mesh();
     System& X_system = d_fe_data->d_es->get_system(COORDINATES_SYSTEM_NAME);
 
     // Setup data structures used to assign elements to patches.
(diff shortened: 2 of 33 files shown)


boyceg (commented):
Could we also go ahead and switch to range-based for loops while we are at it?

amneetb:
> `FESurfaceDistanceEvaluator`: @amneetb no examples or tests use this function: can we remove it from the public interface?

We have both 

https://github.com/IBAMR/IBAMR/tree/master/tests/level_set
https://github.com/IBAMR/IBAMR/blob/master/examples/IBLevelSet/ex0/example.cpp#L844-L861

drwells:
@boyceg we could but that would require bumping the minimum required version of libMesh to 1.3.1 (released in October 2018). I'd prefer to handle that in a followup (the CI will also need to be updated).

@amneetb I should have been clearer: `FESurfaceDistanceEvaluator` is good to go, but we need to change the return type of `FESurfaceDistanceEvaluator::getNeighborIntersectionsMap()` which is used in no examples or tests.

examples/IBFE/explicit/ex0/example.cpp
```diff
@@ -548,11 +548,11 @@ main(int argc, char* argv[])
             const std::vector<std::vector<VectorValue<double> > >& dphi = fe->get_dphi();
             TensorValue<double> FF;
             boost::multi_array<double, 2> X_node;
-            const MeshBase::const_element_iterator el_begin = mesh.active_local_elements_begin();
```
boyceg, line 551: 👍

ibtk/src/lagrangian/FEDataManager.cpp
```diff
@@ -3086,7 +3086,11 @@ FEDataManager::collectActivePatchElements(std::vector<std::vector<Elem*> >& acti
                                           const int finest_elem_ln)
 {
     // Get the necessary FE data.
-    const MeshBase& mesh = d_fe_data->d_es->get_mesh();
+    //
+    // We once relied on libMesh not correctly implementing iterator constness
+    // correctly. Here, we don't need to change the mesh, but we use Elem *s
```
boyceg, line 3091: Could we use `const Elem*`s here?

src/level_set/FESurfaceDistanceEvaluator.cpp
```diff
@@ -563,7 +559,7 @@ FESurfaceDistanceEvaluator::calculateSurfaceNormals()
         for (std::size_t node_n = 0; node_n < n_nodes; ++node_n) d_node_to_elem[elem->node_ptr(node_n)].insert(elem);
 
         // for each edge, store the current element in the edge to element map.
-        std::vector<Node*> edge_nodes(2);
+        std::array<const Node*, 2> edge_nodes;
```
boyceg, line 562: 👍

boyceg (approved):
Looks good to me!

ibtk/src/lagrangian/FEDataManager.cpp
drwells: Yes, but it would require that we change some more APIs - we would have to change the type of `getActivePatchElementMap()` and any data structure that presently stores `Elem`s for partitioning purposes.

drwells:
@amneetb, are you OK with this change?

amneetb:
Does this affect any of my stuff? Is that why you are asking me?

include/ibamr/FESurfaceDistanceEvaluator.h
```diff
@@ -77,7 +77,7 @@ class FESurfaceDistanceEvaluator
     /*!
      * \brief Get the map maintaining triangle-cell intersection and neighbors.
      */
-    const std::map<SAMRAI::pdat::CellIndex<NDIM>, std::set<libMesh::Elem*>, IBTK::CellIndexFortranOrder>&
+    const std::map<SAMRAI::pdat::CellIndex<NDIM>, std::set<const libMesh::Elem*>, IBTK::CellIndexFortranOrder>&
     getNeighborIntersectionsMap();
```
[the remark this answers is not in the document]
amneetb: Sure, that's fine. I think @abarret is making use of this class in his work, in which case I am pinging him to seek his advice.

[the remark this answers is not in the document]
abarret: This is fine with me.

PR #1446, opened 2022-04-11
Check element jacobians

### LINKED ISSUE #1444: Let IBFEMethod check which elements are inverted
We should have a better error message that tries to sort things out as soon as we load the mesh. Right now we error out in 

https://github.com/IBAMR/IBAMR/blob/50d5c37d8f2952abc21f05ab224f22003d23d6e7/ibtk/src/lagrangian/FEMapping.cpp#L230-L249

which is pretty far along the pipeline and doesn't print any useful messages.

This was discussed in group meeting and will fix a problem @labdala ran into.

drwells:
Fixes #1444.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/minor/20220411DavidWells
+++ b/doc/news/changes/minor/20220411DavidWells
@@ -1,0 +1,4 @@
+Improved: IBAMR::FEMechanicsBase and IBAMR::IBFEMethod now print an error message if
+any of the elements in the provided mesh are inverted.
+<br>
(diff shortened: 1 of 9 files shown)


boyceg (approved):

boyceg:
Looks like this also has a performance improvement for the FE cache?

PR #1449, opened 2022-04-27
Add capture variables to function pointers in IBRedundantInitializer.

abarret:
IBRedundantInitializer uses function pointers with no way to forward object states. This adds in `void *` to all the arguments with defaults of `nullptr` to bring the class in line with what happens in the rest of the library.

This change can potentially break user codes, but only requires adding the parameter `void *` to any functions that are passed to IBRedundantInitializer.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg (approved):
Looks good to me.

drwells:
Making things more uniform (and adding a spot for callbacks) is good. This only impacts two examples so the breakage isn't that bad.

drwells (approved):

PR #1450, opened 2022-04-27
Attest return code

drwells:
More CI updates.

A minor bug with the CI is that attest can fail (i.e., tests can fail or some other bad thing can happen) and we still get a CI success since attest does not communicate the failure. Fix this by returning 1 if attest failed in any way and 0 if we succeed.

Presently, two CIB tests always time out, so this should fail - I want to check that and then I will disable them (i.e., I want to verify that we 'fail successfully' on GitHub before finishing this PR).


drwells:
This failed correctly so I added the extra commit to disable (in CI) the tests which consistently time out.

boyceg (approved):

PR #1459, opened 2022-05-20
Allow multiple AdvDiffHierarchyIntegrator objects to be registered with the INS solvers.

abarret:
It's useful to be able to evolve different fields with different kinds of advection diffusion integrators. This replaces the single `AdvDiffHierarchyIntegrator` object in `INSHierarchyIntegrator` with a vector of integrators.

Note that we can not use a similar strategy as `IBStrategySet` for this as the INS solver needs to directly interact with the advection diffusion solver.

Also, for the VC INS solvers, the correct index of the advection diffusion integrator that owns the density or viscosity variable needs to be set.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/minor/20220526AaronBarrett
+++ b/doc/news/changes/minor/20220526AaronBarrett
@@ -1,0 +1,4 @@
+Improved: Multiple advection diffusion integrators can be registered with the
+navier stokes integrator.
+<br>
+(Aaron Barrett, 2022/05/26)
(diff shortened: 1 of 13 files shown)


amneetb:
@abarret : Is this for accommodating turbulence models in INSVC solvers?

abarret:
> Is this for accommodating turbulence models in INSVC solvers?

I don't think there's a need for two types of advection diffusion solvers, although potentially it could be used.

This is for a project in which some fields need the standard advection diffusion solver and some need a separate semi-Lagrangian solver.

drwells (approved):

PR #1463, opened 2022-06-08
CI: pick a better key for the no-libMesh build.

drwells:
For various reasons, picking the compilation cache from the libMesh build in the non-libMesh build just doesn't work that well. Lets just save the cache we generate here instead.


drwells:
Note that the configurations in which we only compile the library complete in 3 minutes, while this one takes 77 minutes - most of that time is spent redoing known work.

PR #1464, opened 2022-06-13
Fix harmonic averaging for the corner case of arguments with zero value

Raamkrishnan:
Hi @drwells, 

I have fixed the harmonic averaging for the corner case of arguments with zero value. This avoids the division by zero. I have also moved the arithmetic and harmonic averaging routines into one Fortran module to do function overloading. Please take a look.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/math/fortran/vclaplace2d.f.m4
+++ b/ibtk/src/math/fortran/vclaplace2d.f.m4
@@ -18,34 +18,6 @@ include(SAMRAI_FORTDIR/pdat_m4arrdim2d.i)dnl
 c
 ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
 c
-c     Compute the arithmetic average of four inputs
-c
-ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
-c
-      REAL function a_avg(a0,a1,a2,a3)
-      implicit none
-      REAL a0,a1,a2,a3
-      a_avg = 0.25d0*(a0+a1+a2+a3)
-      return
-      end
-c
-ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
-c
-c     Compute the harmonic average of four inputs
-c
-ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
-c
-      REAL function h_avg(a0,a1,a2,a3)
-      implicit none
-      REAL a0,a1,a2,a3
-      REAL dmr
-      dmr = 1.d0/a0+1.d0/a1+1.d0/a2+1.d0/a3
-      h_avg = 4.d0/dmr
-      return
-      end
-c
-ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
-c
 c     Computes (f0,f1) = alpha div mu (grad (u0,u1) + grad(u0,u1)^T) + beta (u0,u1) +
 c     gamma (v0,v1).
 c
@@ -70,11 +42,10 @@ c
      &     use_harmonic_interp)
 c
       implicit none
-
 c
 c     Functions.
 c
-      REAL a_avg, h_avg
+      REAL a_avg4, h_avg4
 c
 c     Input.
 c
@@ -123,14 +94,14 @@ c
             endif
 
             if (use_harmonic_interp .eq. 1) then
-                mu_upper = h_avg(mu(i0,i1),mu(i0+1,i1),
+                mu_upper = h_avg4(mu(i0,i1),mu(i0+1,i1),
      &                          mu(i0,i1+1),mu(i0+1,i1+1))
-                mu_lower = h_avg(mu(i0,i1),mu(i0-1,i1),
+                mu_lower = h_avg4(mu(i0,i1),mu(i0-1,i1),
      &                       mu(i0,i1+1),mu(i0-1,i1+1))
             else
-                mu_upper = a_avg(mu(i0,i1),mu(i0+1,i1),
+                mu_upper = a_avg4(mu(i0,i1),mu(i0+1,i1),
      &                          mu(i0,i1+1),mu(i0+1,i1+1))
-                mu_lower = a_avg(mu(i0,i1),mu(i0-1,i1),
+                mu_lower = a_avg4(mu(i0,i1),mu(i0-1,i1),
      &                       mu(i0,i1+1),mu(i0-1,i1+1))
             endif
 
@@ -157,16 +128,16 @@ c
             endif
 
             if (use_harmonic_interp .eq. 1) then
-               mu_upper = h_avg(mu(i0,i1),mu(i0+1,i1),
+               mu_upper = h_avg4(mu(i0,i1),mu(i0+1,i1),
      &                          mu(i0,i1+1),mu(i0+1,i1+1))
 
-               mu_lower = h_avg(mu(i0,i1),mu(i0+1,i1),
+               mu_lower = h_avg4(mu(i0,i1),mu(i0+1,i1),
      &                          mu(i0,i1-1),mu(i0+1,i1-1))
             else
-               mu_upper = a_avg(mu(i0,i1),mu(i0+1,i1),
+               mu_upper = a_avg4(mu(i0,i1),mu(i0+1,i1),
      &                          mu(i0,i1+1),mu(i0+1,i1+1))
 
-               mu_lower = a_avg(mu(i0,i1),mu(i0+1,i1),
+               mu_lower = a_avg4(mu(i0,i1),mu(i0+1,i1),
      &                          mu(i0,i1-1),mu(i0+1,i1-1))
             endif
 
--- a/ibtk/src/solvers/impls/fortran/patchsmoothers2d.f.m4
+++ b/ibtk/src/solvers/impls/fortran/patchsmoothers2d.f.m4
@@ -744,8 +744,7 @@ c
 c
 c     Functions.
 c
-      REAL a_avg, h_avg
-
+      REAL a_avg4, h_avg4
 c
 c     Input.
 c
@@ -793,14 +792,14 @@ c
          endif
 
          if (use_harmonic_interp .eq. 1) then
-            mu_upper = h_avg(mu(i0,i1),mu(i0+1,i1),
+            mu_upper = h_avg4(mu(i0,i1),mu(i0+1,i1),
      &                       mu(i0,i1+1),mu(i0+1,i1+1))
-            mu_lower = h_avg(mu(i0,i1),mu(i0-1,i1),
+            mu_lower = h_avg4(mu(i0,i1),mu(i0-1,i1),
      &                       mu(i0,i1+1),mu(i0-1,i1+1))
          else
-            mu_upper = a_avg(mu(i0,i1),mu(i0+1,i1),
+            mu_upper = a_avg4(mu(i0,i1),mu(i0+1,i1),
      &                       mu(i0,i1+1),mu(i0+1,i1+1))
-            mu_lower = a_avg(mu(i0,i1),mu(i0-1,i1),
+            mu_lower = a_avg4(mu(i0,i1),mu(i0-1,i1),
      &                       mu(i0,i1+1),mu(i0-1,i1+1))
          endif
 
@@ -827,14 +826,14 @@ c
          endif
 
          if (use_harmonic_interp .eq. 1) then
-            mu_upper = h_avg(mu(i0,i1),mu(i0+1,i1),
+            mu_upper = h_avg4(mu(i0,i1),mu(i0+1,i1),
      &                       mu(i0,i1+1),mu(i0+1,i1+1))
-            mu_lower = h_avg(mu(i0,i1),mu(i0+1,i1),
+            mu_lower = h_avg4(mu(i0,i1),mu(i0+1,i1),
      &                       mu(i0,i1-1),mu(i0+1,i1-1))
          else
-            mu_upper = a_avg(mu(i0,i1),mu(i0+1,i1),
+            mu_upper = a_avg4(mu(i0,i1),mu(i0+1,i1),
      &                       mu(i0,i1+1),mu(i0+1,i1+1))
-            mu_lower = a_avg(mu(i0,i1),mu(i0+1,i1),
+            mu_lower = a_avg4(mu(i0,i1),mu(i0+1,i1),
      &                       mu(i0,i1-1),mu(i0+1,i1-1))
          endif
 
@@ -882,7 +881,7 @@ c
 c
 c     Functions.
 c
-      REAL a_avg, h_avg
+      REAL a_avg4, h_avg4
 c
 c     Input.
 c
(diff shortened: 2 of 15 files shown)


ibtk/src/math/fortran/vclaplace2d.f.m4
```diff
@@ -18,31 +18,100 @@ include(SAMRAI_FORTDIR/pdat_m4arrdim2d.i)dnl
...
-c
-      REAL function h_avg(a0,a1,a2,a3)
-      implicit none
-      REAL a0,a1,a2,a3
-      REAL dmr
-      dmr = 1.d0/a0+1.d0/a1+1.d0/a2+1.d0/a3
-      h_avg = 4.d0/dmr
-      return
-      end
+      module average
```
drwells, line 24: Since we use de-facto fortran 77 everywhere else I would prefer to use the old-fashioned version of function declarations for style uniformity.

In general, modules are much better, but unless we want to convert everything I don't think it makes sense to mix and match right now.

ibtk/src/solvers/impls/fortran/patchsmoothers2d.f.m4
```diff
@@ -878,6 +879,7 @@ c
      &     use_harmonic_interp,
      &     red_or_black)
 c
+      use average
       implicit none
 c
 c     Functions.
```
drwells, line 885: in particular, on new line 887 we are still declaring h_avg and a_avg in the old way.

src/navier_stokes/INSVCStaggeredPressureBcCoef.cpp
```diff
@@ -327,9 +327,14 @@ INSVCStaggeredPressureBcCoef::setBcCoefs(Pointer<ArrayData<NDIM, double> >& acoe
...
-                                 2.0 * ((*mu_data)(i_g) * (*mu_data)(i_i)) / ((*mu_data)(i_g) + (*mu_data)(i_i)) :
-                                 -1.0e305;
+                        double mu_harmonic;
+                        if (IBTK::rel_equal_eps((*mu_data)(i_g), 0.0) || IBTK::rel_equal_eps((*mu_data)(i_i), 0.0))
+                            mu_harmonic = 0.0;
+                        else
+                            mu_harmonic =
+                                2.0 * ((*mu_data)(i_g) * (*mu_data)(i_i)) / ((*mu_data)(i_g) + (*mu_data)(i_i));
+
+                        mu = mu_contains_cells ? mu_harmonic : -1.0e305;
```
drwells, line 337: What's going on here? Should this throw an error instead?

If we need to assign a value that should never be read again then we should use a signaling nan.

drwells (changes requested):
Nice work.

Do we have any example code that previously crashed? That would be good to add as a test.

src/navier_stokes/INSVCStaggeredPressureBcCoef.cpp
```diff
@@ -327,9 +327,14 @@ INSVCStaggeredPressureBcCoef::setBcCoefs(Pointer<ArrayData<NDIM, double> >& acoe
                     }
                     else if (d_mu_interp_type == VC_HARMONIC_INTERP)
                     {
-                        mu = mu_contains_cells ?
-                                 2.0 * ((*mu_data)(i_g) * (*mu_data)(i_i)) / ((*mu_data)(i_g) + (*mu_data)(i_i)) :
-                                 -1.0e305;
+                        double mu_harmonic;
+                        if (IBTK::rel_equal_eps((*mu_data)(i_g), 0.0) || IBTK::rel_equal_eps((*mu_data)(i_i), 0.0))
```
abarret, line 331: You should be very careful about using `rel_equal_eps` to compare numbers that are close to zero. I think you want to use `abs_equal_eps`. See these tests
https://github.com/IBAMR/IBAMR/blob/master/tests/IBTK/equal_eps.output

ibtk/src/math/fortran/vclaplace2d.f.m4
Raamkrishnan: There is no other way to do function overloading in Fortran. We searched and found only this. We would like to keep this.

drwells: This is true - one workaround would be to not do function overloading and instead either define several independent functions or one function which takes an array and array length as argument.

Independently of the style considerations, CMake requires that Fortran module dependencies be handled in a somewhat explicit way. Most of the CI configurations cannot find the `.mod` files right now (GCC 12 appears to work by coincidence only). Similarly, we cannot compile this with autotools for similar reasons.

Since we don't really compile Fortran but instead preprocess m4 files into Fortran, specifying these dependencies is somewhat complex and is another reason that we should use the old-fashioned functions declarations. No matter what solution we work with it needs to work with the configurations we test in the CI.

boyceg:
I cannot look at this in any detail now, but can you accomplish this with m4 macros instead of explicit functions?

ibtk/src/math/fortran/vclaplace2d.f.m4
amneetb: I see. Let's then use several independent functions. Modules don't seem to be worth the trouble, especially if their compilation `.mod` is handled differently by different compilers.

src/navier_stokes/INSVCStaggeredPressureBcCoef.cpp
amneetb: Thanks @abarret. We switched to `abs_equal_eps`. Any idea what would be its equivalent function in fortran? We are using the `EPSILON()` function, i.e., `EPSILON(0.d0)`.

amneetb: > https://github.com/IBAMR/IBAMR/blob/master/tests/IBTK/equal_eps.output

https://github.com/IBAMR/IBAMR/blob/master/tests/IBTK/equal_eps.output#L65-L67

@abarret  Why are a = 1e-8 and b = 0 equal in the absolute test?

amneetb: similarly I don't understand this equality either
https://github.com/IBAMR/IBAMR/blob/master/tests/IBTK/equal_eps.output#L77-L79

abarret: The equivalent check in fortran to `IBTK::abs_equal_eps` with the default tolerance would be
```
eps = EPSILON(0.d0)
abs_equal_eps = ABS(a - b) .LE. SQRT(eps)
```

amneetb: > The equivalent check in fortran to `IBTK::abs_equal_eps` with the default tolerance would be
> 
> ```
> eps = EPSILON(0.d0)
> abs_equal_eps = ABS(a - b) .LE. SQRT(eps)
> ```

Thanks!

Raamkrishnan:
@drwells I have removed the module and implemented it as separate functions. Since these are independent of dimensions, I have added these functions in a new fortran class named "Averaging.f.m4" located at ibtk/src/utilities/fortran. I verified the code by running a few multiphase tests. Please take a look and let me know your suggestions.

ibtk/src/utilities/fortran/Makefile.am
```diff
@@ -0,0 +1,17 @@
...
+## IBAMR is free software and is distributed under the 3-clause BSD
+## license. The full text of the license can be found in the file
+## COPYRIGHT at the top level directory of IBAMR.
+##
+## ---------------------------------------------------------------------
+
+## Process this file with automake to produce Makefile.in
+include $(top_srcdir)/config/Make-rules
+
+EXTRA_DIST = averaging.f.m4
```
drwells, line 17: We also need `BUILT_SOURCES` and `CLEANFILES` here - take a look at one of the other fortran automake files (like math/fortran/Makefile.am)

drwells (changes requested):
We also need to update the CMake build system. Take a look at `ibtk/src/CMakeLists.txt`. Since these files don't depend on dimension we will need a little bit of extra logic to handle their preprocessing.

ibtk/src/utilities/fortran/averaging.f.m4
```diff
@@ -0,0 +1,115 @@
...
+     &    abs(a2) .le. sqrt(eps) .or. abs(a3) .le. sqrt(eps) .or.
+     &    abs(a4) .le. sqrt(eps) .or. abs(a5) .le. sqrt(eps) .or.
+     &    abs(a6) .le. sqrt(eps) .or. abs(a7) .le. sqrt(eps) .or.
+     &    abs(a8) .le. sqrt(eps) .or. abs(a9) .le. sqrt(eps) .or.
+     &    abs(a10) .le. sqrt(eps) .or. abs(a11) .le. sqrt(eps)) then
+          h_avg12 = 0.d0
+      else
+          dmr = 1.d0/a0+1.d0/a1+1.d0/a2+1.d0/a3+1.d0/a4+1.d0/a5
+     &             +1.d0/a6+1.d0/a7+1.d0/a8+1.d0/a9+1.d0/a10+1.d0/a11
+          h_avg12 = nmr/dmr
```
drwells, line 112: @boyceg I don't think its feasible to add something like this as a macro.

If we are worried about performance then we might want to investigate turning on link-time optimization. That would be a more permanent (and performant) fix.

Raamkrishnan:
> Running tests manually is good but it doesn't ensure that this bug is permanently fixed. Can you add a test which previously failed or generated wrong output to this PR?

None of the existing cases take the parameter to be zero. I encountered this bug when I was simulating Stefan's problem with the outflow boundary condition. I will commit the phase-change code in the coming weeks. I can add the test once the phase-change code is reviewed.

Raamkrishnan:
> We also need to update the CMake build system. Take a look at `ibtk/src/CMakeLists.txt`. Since these files don't depend on dimension we will need a little bit of extra logic to handle their preprocessing.

@drwells  Can you take care of this? I do not know how to work with CMake.

drwells:
One answer is that autoibamr would not be possible to write with autotools since there is no `make install` :)

Otherwise: some of the bugs I have seen on the mailing list about corrupted archives or incorrect dependency handling (e.g., changing something in IBAMR and then recompiling a user project does not propagate changes correctly) don't happen with CMake. This PR itself is a good example of another advantage of CMake: I count eight changed files in autotools compared to one with CMake.

drwells:
to respond to your edit: autoibamr is

https://github.com/ibamr/autoibamr

it automatically downloads, compiles, and installs IBAMR's dependencies (excluding MPI) and IBAMR.

amneetb:
Also there is a pedagogical value in building third party libraries for students who are extending IBAMR in terms of physics. For many of them it is the first gateway to linux and scientific computing libraries. Automating everything is a good way to increase the user base but necessarily the developer base. My 2 cents.

boyceg:
> > One answer is that autoibamr would not be possible to write with autotools since there is no `make install` :)
> > Otherwise: some of the bugs I have seen on the mailing list about corrupted archives or incorrect dependency handling (e.g., changing something in IBAMR and then recompiling a user project does not propagate changes correctly) don't happen with CMake. This PR itself is a good example of another advantage of CMake: I count eight changed files in autotools compared to one with CMake.
> 
> But these are all internal stuff (more files to change or more scripts to write) that an external user is not bothered about. I am thinking more in terms of comfort level of a user. If they prefer autotools to work with, that option should be there. It is not that we have to build autotools from scratch, we just need to maintain it. I am pretty sure if you ask PETSc team about CMake they will give you 100 reasons where CMake fails, so I don't think there is a simple reason to choose one build system over the other.

We (including @knepley) discussed build systems and agreed on moving to CMake. The alternative was to move to the PETSc BuildSystem. (There are some old PRs that attempted to do this, but I never quite got it to work.) Ultimately we decided on CMake because more it has much more documentation. Notice also that PETSc does not use autotools (and has not for some time). 😁

I am not 💯 on every aspect of CMake, but for what we are doing, it is mostly easier to work with than autotools. It also will be necessary to use with SAMRAI 4, and so it is coming one way or the other. 😁

boyceg:
> Also there is a pedagogical value in building third party libraries for students who are extending IBAMR in terms of physics. For many of them it is the first gateway to linux and scientific computing libraries. Automating everything is a good way to increase the user base but necessarily the developer base. My 2 cents.

Use of autoibamr certainly is not required.

amneetb:
Uh. Difficult to wrap my head around how building one library affects the build system of the other library. At the end of the day we just need directory locations.

boyceg:
> Uh. Difficult to wrap my head around how building one library affects the build system of the other library. At the end of the day we just need directory locations.

Here is one answer. This is the extent of the CMake logic required to link to SAMRI 4:
```
FIND_PACKAGE(SAMRAI REQUIRED)
```
Please compare to [configure_samrai.m4](https://github.com/IBAMR/IBAMR/blob/9ffc33ab8495ebf667e431f0486b64d9b54b3f0e/m4/configure_samrai.m4).

Which one seems easier to debug and maintain? 😁

boyceg:
At any rate, I'd suggest that we move build system discussions elsewhere (an issue or discussion on GitHub, or to the ibamr-dev mailing list).

drwells:
Yes (sorry @Raamkrishnan for taking over your PR). Lets move this to https://github.com/IBAMR/IBAMR/discussions/1467 - that way we can edit an FAQ document answering all these questions in one place.

Raamkrishnan:
I have modified CMakeLists.txt. Please take a look. The code compiles fine with CMake. The tests are also working fine.

Raamkrishnan:
@drwells If you are ok with the code, can you merge this code? I am refactoring my phase change code. I want to rebase with the master (along with these changes). I will add a test once my phase change code is reviewed.

drwells:
As long as we add the test eventually I'm OK with merging this - I'll hold you to it!

drwells:
I did one last check and it looks like this causes the bp_adv_diff test (which uses 8 processors and is not run in the CI) to fail - can you take a look before we merge? The CMake looks good.

drwells:
Nevermind - that test also fails on this machine on `master`.

drwells (approved):

amneetb:
> I did one last check and it looks like this causes the bp_adv_diff test (which uses 8 processors and is not run in the CI) to fail - can you take a look before we merge? The CMake looks good.

IIRC, that was fixed in these commits:

https://github.com/IBAMR/IBAMR/commit/7619ed4d432353e3c02346f0ad54629f600a8c7a
https://github.com/IBAMR/IBAMR/commit/12195b4930d6584ed9f15e31f7d2fe813069a4c4

and then @abarret reported that the Brinkman test passed, and you had approved the change here:
https://github.com/IBAMR/IBAMR/pull/1268

abarret:
> I did one last check and it looks like this causes the bp_adv_diff test (which uses 8 processors and is not run in the CI) to fail - can you take a look before we merge? The CMake looks good.

I have never had good experiences with this test. We certainly made it more robust in #1268, but it still will sometimes fail in ways I don't understand (seemingly at random). Curiously, I can't get it to fail now...

drwells:
I thought so too but it looks like it still exhibits some form of machine-dependency. This might be a problem with SAMRAI rather than anything we can directly control.

Raamkrishnan:
> Nice work.
> 
> Do we have any example code that previously crashed? That would be good to add as a test.

With the fix, the code works fine. Should I add that example code as a test?

drwells:
That would be great!

PR #1466, opened 2022-06-14
Fix running tests on CI

drwells:
There is some bug present in the lines

```cmake
# Not all CI configurations actually run tests
IF(TARGET tests)
  CTEST_BUILD(
    NUMBER_WARNINGS num_warnings
    RETURN_VALUE build_result
    TARGET tests
    )
ELSE()
  CTEST_BUILD(
    NUMBER_WARNINGS num_warnings
    RETURN_VALUE build_result
    )
ENDIF()
```
where, even though we need this in case `tests` is not defined, the CI always takes the non-test branch and as a result we never actually run tests. The first commit makes `attest` fail if it does not detect any test (independent of regular expression limiting) - that check should make the CI fail correctly.


drwells:
The issue is that even though we run that script in CMake, it isn't really part of the build system - it does not have a way to determine whether or not `tests` is a target except by trying and failing.

I have never used the cdash dashboard. Right now we require everything to pass at 100% in order to merge a PR so its not that useful to record our results (they are always the same). Hence, I removed the build test logging under ctest and replaced it with `make -j2 tests`. This fixes another bug where ctest does not propagate the set number of processors correctly.

PR #1468, opened 2022-06-15
Clean up some CMake comments.

drwells:
One comment is out of date - the other two could use grammar improvements.

Since this doesn't modify the library itself the usual checklist doesn't apply.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1249,13 +1249,6 @@ IF(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND
   STRING(APPEND CMAKE_CXX_FLAGS " -Wno-unneeded-internal-declaration")
 ENDIF()
 
-#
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -854,7 +854,7 @@ ENDIF()
 # ---------------------------------------------------------------------------- #
 
 #
-# Helper function that checks that we can compile and link _src with the given
+# Helper function which checks that we can compile and link _src with the given
```
boyceg, line 857:
```suggestion
# Helper function that checks that we can compile and link _src with the given
```
"that" is correct usage here. 😁

boyceg (approved):

CMakeLists.txt
drwells: I shouldn't leave UNC until I get this right!

PR #1469, opened 2022-06-16
Skip finding mpifort.

drwells:
Related to #1467: as of CMake 3.10 we can skip all the MPI Fortran stuff. Since we never call MPI routines from Fortran this should make things a little simpler.

Lets see what the CI thinks!


amneetb:
@drwells: Perhaps in the future or for a specific application we need `mpif90`. It would be good to fix the check instead of removing it?

boyceg:
I’d be happy to omit checking for MPI fortran since it can be hard to set up correctly on MacOS.

boyceg:
@amneetb I think that the change here just makes it so that only the C and C++ MPI compiler wrappers are required. I don't think it prevents discovery of the MPI Fortran compiler wrapper. @drwells is that right?

drwells:
This is a really good question - I checked and the following works:
```cmake
PROJECT(inline LANGUAGES C CXX Fortran)
CMAKE_MINIMUM_REQUIRED(VERSION 3.15.0)

ADD_LIBRARY(test test.f)
FIND_PACKAGE(MPI REQUIRED)

FIND_PACKAGE(IBAMR 0.8.0 REQUIRED)
TARGET_LINK_LIBRARIES(test IBAMR::IBAMR2d)
TARGET_LINK_LIBRARIES(test MPI::MPI_Fortran)
```

i.e., just because we do not find `MPI::MPI_Fortran` ourselves doesn't mean downstream users cannot do it if they would like.

boyceg:
Could we make Fortran an optional component, something like:
```
FIND_PACKAGE(MPI REQUIRED COMPONENTS C CXX OPTIONAL_COMPONENTS Fortran)
```

drwells:
In principle, yes, but there isn't much advantage to doing so - IBAMR is a C++ library and doesn't make much sense to use our stuff directly from Fortran. We never do MPI calls inside Fortran so a user should just set that up if they have the very odd situation of wanting to use IBAMR from an exclusively Fortran project.

amneetb:
I'm thinking of a situation where a user wants to use a third party MPI fortran library to do some extraneous calculations, e.g., doing loosely coupled physics with IBAMR objects.

drwells:
I had to think about it but ultimately I don't think making a change here will impact that scenario. Call that external Fortran library (which uses MPI) X. If X exports its link dependencies then we are set. If it does not (e.g., it is underlinked or staticly linked):
1. If we call it from C or C++: this will fail either way since `mpic++` will not add MPI fortran link flags.
2. If we call it from Fortran: a user would still need to set up their Fortran stub to depend on MPI. They can either use the MPI fortran wrappers or `FIND_PACKAGE(MPI REQUIRED)` themselves

Ultimately this change makes the build system somewhat more robust and it might require users in an obscure circumstance to write one more line of CMake - I think its okay. We should help accommodate our dependencies but I don't think it is a good use of our time to try to work around possible flaws in our users' dependencies.

boyceg:
If we made it optional would it at least simplify the user's `CMakeLists.txt` setup?

drwells:
Technically yes, but it seems odd to me to add a configuration feature which does not impact IBAMR at all.

The root cause of the original problem was not with mpifort - if you two still think we should find and set up MPI Fortran under some circumstances then we should skip this PR. Compromise solutions, like conditionally setting up `MPI::MPI_Fortran` are too user-specific for us to do (and likely to break if no one ever uses them).

boyceg:
I think we should skip searching for it.

A user can always pass in the MPI compiler wrappers as “the compilers” anyway, can’t they?

drwells:
Users can always provide their own compilers (though CMake prefers to set up MPI as a library dependency instead of using compiler wrappers).

I don't think using `mpifort` will fix the underlinkage problem in this hypothetical situation. In particular, unless we compile a Fortran executable we will not get the Fortran link line (just compiling an object file is not enough), even with the compiler wrappers. A user, in the case where `X` is underlinked, really will need to provide the link flags

boyceg:
OK, it sounds like this change removes the illusion that one can use MPI in Fortran without actually changing anything that users can actually do — is that right?

drwells:
I had to try this out to answer the questions - this might be more accurate than some of my previous posts.

1. We need an explicit
```cpp
TARGET_LINK_LIBRARIES(test MPI::MPI_Fortran)
```
or set `CMAKE_Fortran_COMPILER` to `mpifort` to use Fortran MPI regardless of whether or not we use C++ or C. CMake is smart enough to propagate the correct flags to link against the Fortran MPI library in the C++ application.

2. We could do
```cmake
FIND_PACKAGE(MPI REQUIRED COMPONENTS C CXX)
FIND_PACKAGE(MPI COMPONENTS Fortran)
```
which accomplishes what you want - it will error out if we cannot use MPI from C (or the C++ bindings) but will just not set up `MPI::MPI_Fortran` and proceed if the second line fails.

3. The C and C++ components refer to parts of the MPI standard. The C MPI is what we always use - its the standard `MPI_Send()` etc. We can rely on `MPI::MPI_C` in a C++ project just like we can rely on any other C library. The  [C++ interface to MPI](https://web.archive.org/web/20170421220544/http://blogs.cisco.com/performance/the-mpi-c-bindings-what-happened-and-why/) has been deprecated for about 10 years and never really got a lot of traction. I only have it in now since some projects (like libMesh) are underlinked and need the C++-MPI library on the link line to be happy.

boyceg:
Is there any harm in setting `CMAKE_Fortran_COMPILER` to `mpifort` if there is a working MPI Fortran wrapper?

drwells:
I don't think so. CMake prefers that we set up MPI like other dependencies (e.g., provide a root directory for it) and we have some additional code in place to set `MPI_ROOT` based on an MPI C++ compiler should it be necessary.

boyceg:
@drwells what should happen here?

drwells:
I cannot think of a situation in which someone would want to mix MPI in Fortran code and IBAMR. Even if they did they can always set the compiler by writing maybe one or two lines of code to load the Fortran MPI interface. I'd still like to make this change but we need to double-check it again.

boyceg:
@drwells shall we merge this?

PR #1475, opened 2022-06-29
Use consistent Laplacian discretization

abarret:
This fixes a bug in the staggered Stokes operator where an inconsistent discretization of the Laplacian operator is used at transverse boundaries. Originally, linear interpolation was used to fill ghost cells at physical boundaries. This leads to a zeroth order error in the Laplacian stencil. This allows options on how to fill ghost cells to be read in via an optional database argument in the constructor of the operator.

For reasons that I do not understand, linear interpolation does not affect the overall order of accuracy of the fluid velocity, only the order of the gradient of the fluid velocity. Therefore, this does not change the default behavior of the solver.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


abarret:
@colegruninger97, can you edit the `navier_stokes/stokes_operator` test to check that we get the right solution for pipe flow? You'll need to add an option in the input file to specify `bdry_interp_type = "QUADRATIC"` and then add the database as an argument to the `StaggeredStokesOperator`.

boyceg:
Is this interpolation or extrapolation at the boundary?

colegruninger97:
> @colegruninger97, can you edit the `navier_stokes/stokes_operator` test to check that we get the right solution for pipe flow? You'll need to add an option in the input file to specify `bdry_interp_type = "QUADRATIC"` and then add the database as an argument to the `StaggeredStokesOperator`.

bash-3.2$ ./main2d input2d 
|e|_oo = 0
|e|_2  = 0
|e|_1  = 0

Screen Shot 2022-06-30 at 2 18 51 PM



Looks like we are good to go!

abarret:
> 
> bash-3.2$ ./main2d input2d |e|_oo = 0 |e|_2 = 0 |e|_1 = 0
> 
> Looks like we are good to go!

Can you add the test to this branch too?

abarret (commented):
There's no need to make a second test executable. You can change the original test to read in boundary conditions and give the boundary conditions to the Stokes operator if there are physical boundaries. Everything else can stay the same.

boyceg:
> > Is this interpolation or extrapolation at the boundary?
> 
> Interpolation

Isn't the flag determining how ghost cell values are determined --- which would be extrapolation using BC and solution data?

abarret:
> Isn't the flag determining how ghost cell values are determined --- which would be extrapolation using BC and solution data?

In `HierarchyGhostCellInterpolation`, "extrapolation" refers to filling ghost cells using only interior data and "interpolation" fills ghost cells using interior data and BC's. There should be no "extrapolation" occurring in the Stokes operator.

boyceg:
> > Isn't the flag determining how ghost cell values are determined --- which would be extrapolation using BC and solution data?
> 
> In `HierarchyGhostCellInterpolation`, "extrapolation" refers to filling ghost cells using only interior data and "interpolation" fills ghost cells using interior data and BC's. There should be no "extrapolation" occurring in the Stokes operator.

Did I come up with that convention? Ugh. It seems like both are extrapolation...

abarret:
> Did I come up with that convention? Ugh. It seems like both are extrapolation...

I don't know who came up with the convention, but #665 introduced the option to use quadratic "interpolation" at the boundaries. Prior to that, only the "extrapolation" type could be set.

tests/navier_stokes/stokes_operator.cpp
```diff
@@ -54,45 +58,106 @@ main(int argc, char* argv[])
...
+        Pointer<CellVariable<NDIM, double>> f_cc_var = new CellVariable<NDIM, double>("f_cc");
 
         // Error terms.
-        Pointer<SideVariable<NDIM, double> > e_sc_var = new SideVariable<NDIM, double>("e_sc");
-        Pointer<CellVariable<NDIM, double> > e_cc_var = new CellVariable<NDIM, double>("e_cc");
+        Pointer<SideVariable<NDIM, double>> e_sc_var = new SideVariable<NDIM, double>("e_sc");
+        Pointer<CellVariable<NDIM, double>> e_cc_var = new CellVariable<NDIM, double>("e_cc");
 
         // Register patch data indices...
+        // Need to figure out what the registerVariableAndContext function does ...
```
abarret, line 119:
```suggestion
```

The `registerVariableAndContext` function sets up a patch data index that corresponds to the `Variable` and `VariableContext` pair. This allows data to be allocated on the patches in the patch hierarchy.

tests/navier_stokes/stokes_operator.cpp
```diff
@@ -158,34 +226,67 @@ main(int argc, char* argv[])
...
+        
+
+        // Interpolate the side-centered data to cell centers for output.
+        static const bool synch_cf_interface = true;
+        hier_math_ops.interp(draw_u_idx, draw_u_var, u_sc_idx, u_sc_var, nullptr, 0.0, synch_cf_interface);
+        hier_math_ops.interp(draw_f_idx, draw_f_var, f_sc_idx, f_sc_var, nullptr, 0.0, synch_cf_interface);
+        hier_math_ops.interp(draw_e_idx, draw_e_var, e_sc_idx, e_sc_var, nullptr, 0.0, synch_cf_interface);
+
+        // Output data for plotting.
+        visit_data_writer->writePlotData(patch_hierarchy, 0, 0.0);
```
abarret, line 272:
```suggestion
```

Same as above

tests/navier_stokes/stokes_operator.cpp
```diff
@@ -158,34 +226,67 @@ main(int argc, char* argv[])
...
 
         // Deallocate level data
         // Allocate data on each level of the patch hierarchy.
+        
+        //Question-> Why does data need to be deallocated?
```
abarret, line 277:
```suggestion
```
We deallocate data to prevent memory leaks. While it may not be a big deal in this test (since data will be deallocated  in a few lines automatically), it's a good habit to get into.

abarret (commented):
This looks good. A few comments. Can you run clang-format on the test?

abarret:
The tests pass and the examples that I've checked all run, so I'll go ahead and merge.

PR #1476, opened 2022-06-30
fix harmonic avg in PoissonUtilities

Raamkrishnan:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

I missed this class while fixing harmonic averaging to avoid division by zero. I also checked through git grep to make sure no other files require a harmonic average fix. The test suite runs fine.

### DIFF
--- a/ibtk/src/math/PoissonUtilities.cpp
+++ b/ibtk/src/math/PoissonUtilities.cpp
@@ -96,7 +96,12 @@ compute_mu_harmonic_avg(const hier::Index<NDIM>& i, const NodeData<NDIM, double>
     int total_nodes = 0;
     for (NodeIterator<NDIM> n(node_box); n; n++, total_nodes++)
     {
-        avg_mu += 1.0 / mu_data(n(), /*depth*/ 0);
+        const double mu = mu_data(n(), /*depth*/ 0);
+        if (IBTK::abs_equal_eps(mu, 0.0))
+        {
+            return 0.0;
+        }
+        avg_mu += 1.0 / mu;
     }
 
 #if !defined(NDEBUG)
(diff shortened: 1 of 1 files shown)


ibtk/src/math/PoissonUtilities.cpp
```diff
@@ -96,6 +96,12 @@ compute_mu_harmonic_avg(const hier::Index<NDIM>& i, const NodeData<NDIM, double>
     int total_nodes = 0;
     for (NodeIterator<NDIM> n(node_box); n; n++, total_nodes++)
     {
+        // if mu is zero on any node, then return zero.
+        if (IBTK::abs_equal_eps(mu_data(n(), /*depth*/ 0), 0.0))
+        {
+            const double mu_harmonic = 0.0;
+            return mu_harmonic;
```
amneetb, line 103: The two lines can be compactly written as 

```
return 0.0
```

ibtk/src/math/PoissonUtilities.cpp
```diff
@@ -141,6 +147,12 @@ compute_mu_harmonic_avg(const hier::Index<NDIM>& i, const EdgeData<NDIM, double>
     {
         for (EdgeIterator<NDIM> e(edge_box, axis); e; e++, total_edges++)
         {
+            // if mu is zero on any edge, then return zero.
+            if (IBTK::abs_equal_eps(mu_data(e(), /*depth*/ 0), 0.0))
+            {
+                const double mu_harmonic = 0.0;
+                return mu_harmonic;
+            }
```
amneetb, line 155: same here.

ibtk/src/math/PoissonUtilities.cpp
amneetb: You can also access the `mu` data once and use it at two spots of the code. 

```
const double mu =  mu_data(n(), /*depth*/ 0);
if (IBTK::abs_equal_eps(mu, 0.0))
{
     return 0.0;
}
avg_mu += 1.0/mu
```

amneetb (approved):

drwells (approved):

PR #1484, opened 2022-08-16
Averaged data manager

abarret:
This is the second part of #1428. This introduces the `HierarchyAveragedDataManager` which can detect and store (quasi) periodic states using the `SnapshotCache` class. This class can also write visualization files for the mean (and deviations from the mean) field. The time points at which to keep track of the field are specified either in the input file or the constructor.

I've added a test to check that we can compute the mean of a "noisy" field.

This class is currently incomplete. It is currently not set up for restarts and needs more testing.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


abarret:
@drwells or @boyceg If you have suggestions on the way I've set this class up, I'd be happy to hear them.

drwells:
This class also needs to be added to autotools.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,251 @@
...
+{
+/*!
+ * \brief Class HierarchyAveragedDataManager provides a method of determining and storing average fields over periodic
+ * intervals.
+ */
+class HierarchyAveragedDataManager
+{
+public:
+    /*!
+     * The constructor for class HierarchyAveragedDataManager sets some default values and determines data centering.
```
drwells, line 47: We should list these out, and also explain what should be in the input database.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,251 @@
...
+                                           double tol = 1.0e-8)
+    {
+        return updateTimeAveragedSnapshot(u_idx, time, hierarchy, d_mean_refine_type, wgt_idx, tol);
+    }
+    bool updateTimeAveragedSnapshot(int u_idx,
+                                    double time,
+                                    SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy,
+                                    const std::string& mean_refine_type,
+                                    const int wgt_idx = IBTK::invalid_index,
+                                    double tol = 1.0e-8);
```
drwells, line 107:
```suggestion

```

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,251 @@
...
+    /*!
+     * Returns the exact data point at which points are being stored in the internal map. Note that if a time point is
+     * not found within the specified tolerance, an unrecoverable error occurs.
+     */
+    double getTimePt(double time, double tol);
+
+    /*!
+     * Get the SnapshotCache object
+     */
+    const std::unique_ptr<SnapshotCache>& getSnapshotCache()
```
drwells, line 148:
```suggestion
    const SnapshotCache& getSnapshotCache()
```
There's no reason to return a unique pointer by const reference - we should just return the object directly (by reference).

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,251 @@
...
+    {
+        return d_snapshot_cache;
+    }
+
+    /*!
+     * Set the threshold for achieving a steady state.
```
drwells, line 154: However this is defined should be described in the class documentation - I don't know how this is supposed to work so a longer description would help a lot.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,251 @@
...
+     * \brief Registers a scratch variable with the variable database. This data is allocated and deallocated as needed.
+     */
+    void commonConstructor(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db,
+                           SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy);
+    /*!
+     * \brief Default constructor.
+     *
+     * \note This constructor is not implemented and should not be used.
+     */
+    HierarchyAveragedDataManager() = delete;
```
drwells, line 172: We can do this, but
1. Since we have a non-default constructor the default one is implicitly deleted
2. Since we have non-copyable members (the unique pointer) the coppy and assignment constructors are implicitly deleted

so we don't need these.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,251 @@
...
+#include <string>
+#include <vector>
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBTK
+{
+/*!
+ * \brief Class HierarchyAveragedDataManager provides a method of determining and storing average fields over periodic
+ * intervals.
```
drwells, line 41: Could you provide some example code in the class documentation explaining how this is supposed to work? After reading through this I'm still missing the big picture.

drwells (changes requested):
Basic stuff.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
abarret: I've added more documentation. Does this help explain the class?

drwells:
I finally got back to this - yes the documentation is fine.

I see you modified some of the test build scripts, but there are no tests. Could you add them? I think seeing this class in action would resolve most of my questions

abarret:
> I finally got back to this - yes the documentation is fine.
> 
> I see you modified some of the test build scripts, but there are no tests. Could you add them? I think seeing this class in action would resolve most of my questions

There is a test added:
`tests/IBTK/hierarchy_avg_data_01.cpp`

boyceg:
@drwells, @abarret, what should we do here? I think we want to have good support for computing mean flows and higher-order expressions like TKE and Reynolds stresses, which I think is the target for this functionality.

abarret:
This was superseded by #1786.

PR #1489, opened 2022-08-25
Lumped mass matrix as a preconditioner

AminKolah:
In this PR a condition is added so that we can compute the L2 projection of some jump quantities using consistent mass matrix for first order MONOMIAL basis functions. Previously, lumped matrix was used as a preconditioner for all cases. We would like to not use that for the case of first order MONOMIAL (avoiding a tbox assertion error in assert_qrule_is_nodal function that only allows for CONSTANT order with MONOMIAL).

Instead of singling out the case first order DG MONOMIAL mentioned about, we could also try to add a separate flag that decides whether we want to use lumped mass matrix as a precondition for all cases or not.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/lagrangian/FEProjector.cpp
+++ b/ibtk/src/lagrangian/FEProjector.cpp
@@ -115,36 +115,28 @@ build_nodal_qrule(const Order order, const unsigned int dim)
     return {};
 }
 
-inline void
-assert_qrule_is_nodal(const FEType& fe_type, const QBase* const qrule, const Elem* const elem)
+inline bool
+qrule_is_nodal(const FEType& fe_type, const QBase* const qrule)
 {
     auto fe_order = fe_type.order;
-    auto elem_type = elem->type();
-    std::vector<FEFamily> permitted_fe_families{ LAGRANGE, L2_LAGRANGE, MONOMIAL };
-    TBOX_ASSERT(std::find(permitted_fe_families.begin(), permitted_fe_families.end(), fe_type.family) !=
-                permitted_fe_families.end());
-    if (fe_type.family == MONOMIAL)
+    auto elem_type = qrule->get_elem_type();
+    bool nodal_cond = false;
+    if (fe_type.family == LAGRANGE || fe_type.family == L2_LAGRANGE || fe_type.family == MONOMIAL)
     {
-        // Monomials can only be considered nodal for piecewise constants
-        TBOX_ASSERT(fe_order == CONSTANT);
-    }
-    switch (fe_order)
-    {
-    case CONSTANT:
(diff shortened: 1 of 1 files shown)


drwells:
Thanks for fixing this. One further improvement we can make is by putting all the "we can use a nodal quadrature rule" detection logic in one place - i.e., redo `assert_qrule_is_nodal` to instead just check if a quadrature rule is nodal and then return a boolean. We can then check that where you presently check for first-order monomials - this way all the checks are in one place. Can you make that change?

AminKolah:
@drwells whenever you get a chance, would you mind checking if my last commit does what you recommended? Thanks.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -115,36 +115,31 @@ build_nodal_qrule(const Order order, const unsigned int dim)
...
-    case SECOND:
-        TBOX_ASSERT(qrule->type() == QSIMPSON);
-        TBOX_ASSERT((elem_type == EDGE3) || (elem_type == TRI6 || elem_type == QUAD9) ||
-                    (elem_type == TET10 || elem_type == HEX27));
-        break;
-    default:
-        TBOX_ERROR("FEProjector::assert_qrule_is_nodal(): unsupported element order "
-                   << Utility::enum_to_string<Order>(fe_order) << "\n");
-    }
+        if (fe_type.family == LAGRANGE || fe_type.family == L2_LAGRANGE || fe_type.family == MONOMIAL)
```
drwells, line 126: This returns true for all monomial types - I don't think that's what you intended.

AminKolah: Ah you're right! how about now?

drwells: I think you forgot to push an update.

drwells:
A couple of the higher-order tests now fail to run. A few other fail because we set up the lumped mass matrix at a different point (so the logs are different). Can you take a look?

AminKolah:
> A couple of the higher-order tests now fail to run. A few other fail because we set up the lumped mass matrix at a different point (so the logs are different). Can you take a look?

I believe all issues are fixed and also tested with a few examples to make sure it's doing what we expect. Do you mind taking another look? Thanks.

AminKolah:
@drwells Also let me know if you want to squash commits before the merge

drwells:
Yup, please squash! Can you also run clang-format?

AminKolah:
Closed this in favor of #1491

PR #1493, opened 2022-10-04
Use nodal values for staggered velocities.

drwells:
This looks a lot better. Since forces aren't defined outside the domain (and typically don't have ghost values) there isn't a good way to plot them as nodal values - hence, for now, this is just for the velocity field.

Here's an x-velocity with cell data:
visit0000
and node:
visit0001

Since force values aren't defined outside the domain I don't think there's a good way to plot them - hence lets just use nodal velocities for now.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
OOoh.

Note that the ghost filling routines do know how to do different kinds of extrapolation at physical boundaries; that can be used to interpolate at the boundaries for fields without given boundary conditions.

include/ibamr/INSStaggeredHierarchyIntegrator.h
```diff
@@ -423,6 +422,11 @@ class INSStaggeredHierarchyIntegrator : public INSHierarchyIntegrator
 
     std::string d_U_P_bdry_interp_type = "LINEAR";
 
+    /*!
+     * Variables for graphical output.
+     */
+    SAMRAI::tbox::Pointer<SAMRAI::pdat::NodeVariable<NDIM, double> > d_U_nc_var;
```
boyceg, line 428: 1. Could we make the plotting variables all of type `Variable` and then decide data centerings at run time? This would mean there would be things like `U_plot_var` and `P_plot_var`, which could be the same as the true state variables if the data centerings match.
2. It would be good to have the option to output at least force and vorticity as nodal as well.
3. Pressure is naturally cell-centered, but for visualization it might be useful to have a nodal representation that is consistent with the coarse-fine interface treatment. E.g. if one wants to do isocontours/surfaces of pressure, those will be much cleaner as nodal until we get ghost cells into the VisIt output itself.

boyceg: There could / should be an operator to evaluate the node-centered curl of a staggered grid vector field; it probably is preferable to use that over evaluating the vorticity on another representation and then averaging to the nodes.

boyceg:
One other thing --- I don't know if it is easy to do this, but it would be great if there was a test to verify that hanging nodes are set to give a continuous field.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -584,7 +584,7 @@ INSStaggeredHierarchyIntegrator::INSStaggeredHierarchyIntegrator(std::string obj
     d_Q_var = INSHierarchyIntegrator::d_Q_var;
     d_N_old_var = new SideVariable<NDIM, double>(d_object_name + "::N_old");
 
-    d_U_cc_var = new CellVariable<NDIM, double>(d_object_name + "::U_cc", NDIM);
+    d_U_nc_var = new NodeVariable<NDIM, double>(d_object_name + "::U_nc", NDIM);
```
boyceg, line 587: I think that you need to do:
```suggestion
    d_U_nc_var = new NodeVariable<NDIM, double>(d_object_name + "::U_nc", NDIM, /*fine_boundary_represents_var*/ false);
```
and then _refine_ from coarse to fine levels to get continuity at coarse-fine interfaces.

boyceg: The syncronization would be:
1. _coarsen_ data from fine levels to coarse levels (using `OuternodeVariable`/`OuternodeData` for efficiency).
2. _refine_ data from coarse levels to fine levels (using `LINEAR_REFINE`) to ensure everything matches along the coarse-fine interface.

boyceg: (All of the synchronization probably belongs in `HierarchyMathOps::interp()`. It would be great if the implementation was flexible enough to handle both `fine_boundary_represents_var==true` and `fine_boundary_represents_var==false`.)

include/ibamr/INSStaggeredHierarchyIntegrator.h
drwells: Yes, definitely - lets handle the vorticity in a follow-up PR.

drwells:
> One other thing --- I don't know if it is easy to do this, but it would be great if there was a test to verify that hanging nodes are set to give a continuous field.

I don't think that we can easily access these internal plotting variables but we could check this outside of the INS code by using the same procedure and checking that we get zero error with a bi/trilinear field.

boyceg:
> > One other thing --- I don't know if it is easy to do this, but it would be great if there was a test to verify that hanging nodes are set to give a continuous field.
> 
> I don't think that we can easily access these internal plotting variables but we could check this outside of the INS code by using the same procedure and checking that we get zero error with a bi/trilinear field.

It might be better to use something nonlinear to confirm that the coarse data are correctly interpolated to the fine patches. If you set everything to linear and then "synchronize" then nothing should change.

boyceg:
> > > One other thing --- I don't know if it is easy to do this, but it would be great if there was a test to verify that hanging nodes are set to give a continuous field.
> > 
> > 
> > I don't think that we can easily access these internal plotting variables but we could check this outside of the INS code by using the same procedure and checking that we get zero error with a bi/trilinear field.
> 
> It might be better to use something nonlinear to confirm that the coarse data are correctly interpolated to the fine patches. If you set everything to linear and then "synchronize" then nothing should change.

This sounds like a kind of painful test to write, but one way to do this would be to confirm that the hanging node values are linear interpolants of the fine nodes that exactly overlap the coarse nodes.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
boyceg: > (All of the synchronization probably belongs in `HierarchyMathOps::interp()`. It would be great if the implementation was flexible enough to handle both `fine_boundary_represents_var==true` and `fine_boundary_represents_var==false`.)

Basically, if `fine_boundary_represents_var==true`, then synching the values just requires step 1 (coarsening) and not step 2 (refining) --- step 2 will not change any grid values.

boyceg: One of the points of writing nodal data is to get a continuous representation into VisIt, so that things like isocontours are continuous.

boyceg: Notice also that the synchronization is almost the same as what is already done for side-centered data in `HierarchyMathOps` --- it uses `OutersideData` to synch up values on coarse-fine interfaces.

boyceg: I think that `CONSTANT_COARSEN` already provides the correct injection-based coarsen operator (see: https://ibamr.github.io/IBAMR-docs/samrai/2.4.4/html/package_patchdata_operators.html).

drwells: Oh yeah, that works!

visit0004

I think the very small jumps are due to VisIt itself. I'll write some test code to see if I can verify that.

boyceg: Despite my claims to the contrary elsewhere, it is probably OK just to coarsen and refine a linear nodal field and verify that nothing changes. You just want to make sure that `fine_boundary_represents_var==false` in the `NodeVariable` constructor.

boyceg: (Note that if you coarsen directly from the `NodeData` it may not do the right thing at coarse-fine interfaces if `fine_boundary_represents_var==false` -- although I am not sure!)

drwells: I spent a few hours this afternoon looking at this with a reduced test case and I think what we need (computing boundary values so that the solution is H1-conforming) is missing from SAMRAI - it only seems to schedule the priority of operators with this information but I cannot find any case in which it actually uses this information to go back and 'fix' the nodes on patch boundaries. I'll put something together that does this.

boyceg:
You have to use a refine schedule to communicate from coarse to fine levels.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
boyceg: > In particular, `CONSTANT_COARSEN` makes fine-grid nodes match the coarse-grid ones, but it does not go on to fill in the hanging nodes (which it really should!)

Wait a minute, `CONSTANT_COARSEN` changes the _fine_ data?

drwells: I was kinda tired when I did this work on Friday - it at least synchronized them. I'll check!

boyceg:
@drwells, for the custom (multi-)linear refine operator, is the plan to refine from _node_ data on the coarse level to _outer-node_ data on the fine level?

drwells:
I think we should be able to duplicate the side/edge coarse-fine logic in reverse to copy coarse node data onto fine outernode data, refine that, and then copy back. The present implementation doesn't do that (I copied and pasted coarsening code, but it should be refining code).

drwells:
Lets hold off on the H1-conforming stuff for now - we should just get what we have (nodal) working and make it an option for the caller.

drwells:
We now synch values via outernode data - I added tests and things appear to work. I also verified some graphical output from the NSE examples.

Since collocated nodal values are now equal we could go back and use the fine-grid values and linear/bilinear interpolation to correct the hanging nodes. We would be setting values in the 'wrong' direction but things would still be H1-conforming without needing additional ghost synchronization - do you want to do that now?

boyceg:
Walking the “fine side” of the coarse-fine interface to fix up the hanging node values sounds good to me.

drwells:
I'll write a utility function to do that but it probably won't be done until tomorrow.

ibtk/include/ibtk/HierarchyMathOps.h
```diff
@@ -840,11 +841,17 @@ class HierarchyMathOps : public SAMRAI::tbox::DescribedClass
      * on each coarse-fine interface are synchronized prior to performing the
      * interpolation.
      *
+     * @note The option @p dst_cf_bdry_synch presently synchronizes values by
```
boyceg, line 844: I think this probably should never change ---  with the idea being that the fine value should be "more accurate" than the coarse one, and so we want to make the coarse value agree with the fine value, and then make the overall field continuous by correcting "hanging node" values along the coarse-fine interface.

drwells: I agree and this should be updated. There's no compelling reason to set fine values from coarse at present, especially if we are only doing this for visualization, so the present version should be fine (albeit I need to finish the "enforce hanging node constraints" code).

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -2322,18 +2333,57 @@ INSStaggeredHierarchyIntegrator::applyGradientDetectorSpecialized(const Pointer<
...
+            hier_bdry_fill->initializeOperatorState(comp, d_hierarchy);
+            hier_bdry_fill->fillData(0.0);
+        }
+
         d_hier_math_ops->interp(
-            U_cc_idx, d_U_cc_var, U_sc_idx, d_U_var, d_no_fill_op, d_integrator_time, synch_cf_interface);
+            U_nc_idx, d_U_nc_var, synch_cf_interface, d_U_scratch_idx, d_U_var, d_no_fill_op, d_integrator_time, synch_cf_interface);
+
+        // ensure that things are continuous at coarse-fine interfaces by
+        // refining but not coarsening (since fine_boundary_represents_var ==
```
boyceg, line 2377: Currently `fine_boundary_represents_var` is `true`?

drwells:
Well, that took longer than I thought it would, but that should finally be done.

I think its easier to keep the manual calculations for some of the box shifts since we need to do them anyway - we don't want an adjacent ghost box, we want a codim 1 box exactly on the boundary of the patch.

ibtk/src/math/HierarchyMathOps.cpp
```diff
@@ -2053,20 +2062,40 @@ HierarchyMathOps::interp(const int dst_idx,
...
         }
+        if (ln + 1 <= d_finest_ln && dst_cf_bdry_synch)
+        {
+            xeqScheduleOuternodeRestriction(dst_idx, d_on_idx, ln);
+            d_hierarchy->getPatchLevel(ln + 1)->deallocatePatchData(d_on_idx);
+        }
     }
+
+    // fixup data at hanging nodes.
+    if (dst_cf_bdry_synch)
```
boyceg, line 2094: Should this also be conditional on whether the `Variable` is configured for the coarse value to represent the solution on coarse-fine interfaces?

ibtk/src/math/HierarchyMathOps.cpp
```diff
@@ -2053,20 +2062,40 @@ HierarchyMathOps::interp(const int dst_idx,
...
-                Pointer<OutersideData<NDIM, double> > os_data = patch->getPatchData(d_os_idx);
-                os_data->copy(*src_data);
+                if (src_cf_bdry_synch)
+                {
+                    Pointer<OutersideData<NDIM, double> > os_data = patch->getPatchData(d_os_idx);
+                    os_data->copy(*src_data);
+                }
+                if (dst_cf_bdry_synch)
+                {
+                    Pointer<OuternodeData<NDIM, double> > on_data = patch->getPatchData(d_on_idx);
```
boyceg, line 2074: We might need to be careful here --- the code that I wrote to set up `d_on_idx` made it scalar in 2D and vector-valued in 3D (since I was using it to synch up the vorticity).

tests/IBTK/nodal_interpolation_01.cpp
```diff
@@ -32,6 +32,8 @@
 // Set up application namespace declarations
 #include <ibtk/app_namespaces.h>
 
+#include "../tests.h"
```
boyceg, line 35: Would you mind adding a case where you do not synch up the coarse-fine interface in `HierarchyMathOps::interp()` but instead use `HierarchyGhostCellInterpolation` to do the synch? If `"LINEAR_REFINE"` does the right thing (which it sounds like it doesn't), it ***should*** enforce the hanging node constraints for us so long as the `Variable` is configured so that the coarse values represent the solution. We probably should eventually fix whatever bad thing is happening in the linear refine operator for node data... 😬

ibtk/src/math/HierarchyMathOps.cpp
boyceg: We also should go back and use this with the nodal vorticity operator.

boyceg (approved):
👍

ibtk/src/math/HierarchyMathOps.cpp
boyceg: It might simplify thing to use helper functions like `getOuternodePatchDataIndex(depth)` that will manage setting up / tracking the patch data indices and variables for the different depths instead of doing this "by hand".

drwells: Yup - that's where the failing test fails. It also fails because SAMRAI can generate patches which whose sizes are not integer multiples of the refinement ratio - good thing we added all these tests.

boyceg:
> Yup - that's where the failing test fails. It also fails because SAMRAI can generate patches which whose sizes are not integer multiples of the refinement ratio - good thing we added all these tests.

Ack. That means this fix up probably needs to happen with or following the application of a refine schedule.

drwells:
I need to double-check that point - I was using `getRatio()` instead of `getRatioToCoarserLevel()`. I think the gridding algorithm forbids the situation I described above.

boyceg:
Yes you need to use:
> getRatioToCoarserLevel()

ibtk/src/math/HierarchyMathOps.cpp
drwells: Hm. We don't really support the "coarse values represent the solution case" because IIRC those coarsening operators haven't been written yet, so I'd prefer to ignore that check.

boyceg:
> On Nov 10, 2022, at 12:06 PM, David Wells ***@***.***> wrote:
> 
> ﻿
> @drwells commented on this pull request.
> 
> In ibtk/src/math/HierarchyMathOps.cpp:
> 
> >      }
> +
> +    // fixup data at hanging nodes.
> +    if (dst_cf_bdry_synch)
> Hm. We don't really support the "coarse values represent the solution case" because IIRC those coarsening operators haven't been written yet, so I'd prefer to ignore that check.
> 
I think that this is exactly that case — we are determining the fine values from the overlying coarse ones on coarse-fine interfaces during refinement. I think that is all that flag is meant to imply.

tests/IBTK/nodal_interpolation_01.cpp
drwells: It does enforce hanging node constraints but the coarse and fine shared nodal values don't match - I added a test. I think this is what we expected to see. At some point I got really bad results with one of these coarsening or refinement operators but I can't remember where I described the issue.

drwells:
The present implementation sets coarse values from fine ones right now (that's how we set up the outernode sync) - I don't think the other one has been implemented. Should we do something else here?

boyceg:
I am not certain, but I think that outer node coarsen doesn’t care about the value of this flag.

drwells:
That's definitely true!

drwells:
I'll put this down here so it doesn't get lost in the thread.

I can't locate my original notes but we cannot use `LINEAR_REFINE` with nodal data - the boxes SAMRAI sets up are wrong so some data that is output is uninitialized. For example, here is a run of the present test:

```
|e|_oo = 5.671686642740574e+52
level = 0
0, 0.5: 0.4335454509658621
0.25, 0.5: -1.811039348550182e+50
0.5, 0.5: -0.3189087212405671
0.75, 0.5: 1.772402075856429e+51
1, 0.5: -0.2802923477026677
level = 1
0.25, 0.5: -0.04890488997854661
0.2578125, 0.5: -0.1428353079333652
0.265625, 0.5: -0.198031695789549
0.2734375, 0.5: -0.2522494483956232
0.28125, 0.5: -0.3052206318550948
0.2890625, 0.5: -0.3566834725808279
0.296875, 0.5: -0.4063836509306854
0.3046875, 0.5: -0.4540755580071827
0.3125, 0.5: -0.4995235094102851
0.3203125, 0.5: -0.542502909945213
0.328125, 0.5: -0.5828013635294893
0.3359375, 0.5: -0.6202197228142756
0.34375, 0.5: -0.6545730733329689
0.3515625, 0.5: -0.6856916473135825
0.359375, 0.5: -0.7134216626390206
0.3671875, 0.5: -0.7376260828092658
0.375, 0.5: -0.7581852941498877
0.3828125, 0.5: -0.7749976969202378
0.390625, 0.5: -0.7879802074001848
0.3984375, 0.5: -0.7970686684741723
0.40625, 0.5: -0.8022181666835707
0.4140625, 0.5: -0.8034032541805086
0.421875, 0.5: -0.8006180744863334
0.4296875, 0.5: -0.7938763914332223
0.4375, 0.5: -0.7832115211459224
0.4453125, 0.5: -0.7686761673997533
0.453125, 0.5: -0.7503421611684995
0.4609375, 0.5: -0.7283001056493013
0.46875, 0.5: -0.702658928518761
0.4765625, 0.5: -0.6735453436329319
0.484375, 0.5: -0.6411032248313663
0.4921875, 0.5: -0.6054928949397658
0.5, 0.5: -0.5784813866033905
0.25, 0.5: -0.04890488997854661
0.2578125, 0.5: -0.1428353079333652
0.265625, 0.5: -0.198031695789549
0.2734375, 0.5: -0.2522494483956232
0.28125, 0.5: -0.3052206318550948
0.2890625, 0.5: -0.3566834725808279
0.296875, 0.5: -0.4063836509306854
0.3046875, 0.5: -0.4540755580071827
0.3125, 0.5: -0.4995235094102851
0.3203125, 0.5: -0.542502909945213
0.328125, 0.5: -0.5828013635294893
0.3359375, 0.5: -0.6202197228142756
0.34375, 0.5: -0.6545730733329689
0.3515625, 0.5: -0.6856916473135825
0.359375, 0.5: -0.7134216626390206
0.3671875, 0.5: -0.7376260828092658
0.375, 0.5: -0.7581852941498877
0.3828125, 0.5: -0.7749976969202378
0.390625, 0.5: -0.7879802074001848
0.3984375, 0.5: -0.7970686684741723
0.40625, 0.5: -0.8022181666835707
0.4140625, 0.5: -0.8034032541805086
0.421875, 0.5: -0.8006180744863334
0.4296875, 0.5: -0.7938763914332223
0.4375, 0.5: -0.7832115211459224
0.4453125, 0.5: -0.7686761673997533
0.453125, 0.5: -0.7503421611684995
0.4609375, 0.5: -0.7283001056493013
0.46875, 0.5: -0.702658928518761
0.4765625, 0.5: -0.6735453436329319
0.484375, 0.5: -0.6411032248313663
0.4921875, 0.5: -0.6054928949397658
0.5, 0.5: -0.5784813866033905
0.5, 0.5: -0.5784813866033905
0.5078125, 0.5: -0.5581984547727112
0.515625, 0.5: -0.5366433806059768
0.5234375, 0.5: -0.5139226852098928
0.53125, 0.5: -0.4901486499711988
0.5390625, 0.5: -0.4654387616832545
0.546875, 0.5: -0.4399151319484903
0.5546875, 0.5: -0.4137038937259236
0.5625, 0.5: -0.3869345780058909
0.5703125, 0.5: -0.3597394736923502
0.578125, 0.5: -0.332252973856095
0.5859375, 0.5: -0.3046109115895738
0.59375, 0.5: -0.2769498887453934
0.6015625, 0.5: -0.2494066008757557
0.609375, 0.5: -0.2221171617088478
0.6171875, 0.5: -0.1952164305004972
0.625, 0.5: -0.1688373455851926
0.6328125, 0.5: -0.1431102674199346
0.640625, 0.5: -0.1181623343674711
0.6484375, 0.5: -0.09411683440251314
0.65625, 0.5: -0.07109259584584243
0.6640625, 0.5: -0.04920340013718583
0.671875, 0.5: -0.02855741954882346
0.6796875, 0.5: -0.009256682618643097
0.6875, 0.5: 0.008603430055629913
0.6953125, 0.5: 0.0249346571693379
0.703125, 0.5: 0.03965629288159356
0.7109375, 0.5: 0.05269558564830779
0.71875, 0.5: 0.06398809774658404
0.7265625, 0.5: 0.07347802371377865
0.734375, 0.5: 0.0811184661275603
0.7421875, 0.5: 0.08687166736411656
0.75, 0.5: 0.508865076532288
```
some level 0 data is clearly wrong. To emphasize, this is from using `LINEAR_REFINE` from inside `HierarchyGhostCellInterpolation`: the hanging node code implemented with coarsening outernode data + our fixup routine is fine.

boyceg:
So I think that we should always coarsen if the synch argument is set to true, but only constrain the fine values to be the interpolation of the coarse values of the variable is configured so that the fine boundary represents the variable. I think that this is what happens (or is supposed to happen) with the side and face variables, for instance.

drwells:
I don't understand the sequencing here: if we coarsen then all the coarse nodes on the interface match the corresponding fine node values. The process of constraining hanging nodes is independent of whether we do fine to coarse or coarse to fine in the previous step - its just there for conformity reasons. What's the motivation behind only fixing up the hanging nodes when the fine boundary represents the variable?

boyceg:
> I don't understand the sequencing here: if we coarsen then all the coarse nodes on the interface match the corresponding fine node values. The process of constraining hanging nodes is independent of whether we do fine to coarse or coarse to fine in the previous step - its just there for conformity reasons. What's the motivation behind only fixing up the hanging nodes when the fine boundary represents the variable?

Sorry for being confusing --- Daniel is home sick and I have been replying on my phone. 👎 

I just took a look at some of the side- and face-centered implementations, and they don't behave exactly like I was suggestion.

Here is what I think makes sense here:
1. If `dst_cf_bdry_synch == true`, coarsen from fine to coarse (via injection).
2. If `dst_cf_bdry_synch == true && dst_var->fineBoundaryRepresentsVariable() == false` then apply the constraint operation on the fine nodes along the coarse-fine interface; otherwise, skip it.
The original fine interpolation ***should*** be more accurate than the linear/bilinear interpolation of the coarse values, and so we should allow callers to keep those values if they want them.

The idea is that if `dst_var->fineBoundaryRepresentsVariable() == false`, then the caller is saying that they want to use the coarse values along the coarse-fine interface to determine the fine solution along the interface; otherwise, the caller is saying that they want to fine values along the coarse-fine interface to determine the fine solution. (The implication in the latter is that the solution might not be continuous along the coarse-fine interface. For a finite difference scheme, maybe that is 💯 OK.)

Does that make sense? I will annotate the changes accordingly...

ibtk/src/math/HierarchyMathOps.cpp
```diff
@@ -2053,20 +2096,40 @@ HierarchyMathOps::interp(const int dst_idx,
...
         }
+        if (ln + 1 <= d_finest_ln && dst_cf_bdry_synch)
+        {
+            xeqScheduleOuternodeRestriction(dst_idx, on_idx, ln);
+            d_hierarchy->getPatchLevel(ln + 1)->deallocatePatchData(on_idx);
+        }
+    }
+
+    // fixup data at hanging nodes.
+    if (dst_cf_bdry_synch)
```
boyceg, line 2128:
```suggestion
    // If the fine boundary represents the variable value at the coarse fine interface, then
    // do not overwrite values along the "fine side" of the coarse-fine interface.
    if (dst_cf_bdry_synch && !dst_var->fineBoundaryRepresentsVariable())
```
If the fine boundary represents the variable, we don't want to overwrite the fine

drwells:
Another option would be to let callers decide whether or not they want to use the fixup function themselves instead of checking the value of `fineBoundaryRepresentsVariable()` (in fact that's what we should do for vorticity) - would you be OK with just calling the fixup function that way instead? I think thats clearer.

boyceg:
> Another option would be to let callers decide whether or not they want to use the fixup function themselves instead of checking the value of `fineBoundaryRepresentsVariable()` (in fact that's what we should do for vorticity) - would you be OK with just calling the fixup function that way instead? I think thats clearer.

That would be more flexible. At least in my way of thinking, configuring a `Variable` so that `fineBoundaryRepresentsVariable() == false` means that synching the coarse-fine interface should involve setting the fine values on the coarse-fine interface to interpolate the coarse values, whereas configuring a `Variable` so that `fineBoundaryRepresentsVariable() == true` means that you want to keep the "higher resolution" values. Callers that want a different behavior can always go back and do a custom synch themselves.

boyceg:
At least at this point, we want the nodal velocity, vorticity, and (eventually) pressure to be conformal at the coarse-fine interface solely for _visualization_ purposes --- e.g. so that iso-contours that cut the coarse-fine interface do not have jump discontinuities.

boyceg:
It also seems fine only to call the fixup routine outside of `interp()` --- after all, we do not currently apply similar fixups to other data centerings (side and face) that have overlapping coarse and fine representations on coarse-fine interfaces. (This is probably the first time we have ever had `fineBoundaryRepresentsVariable() == false`!)

drwells:
I went that way - I think it is clearer. We can now easily set up conforming vorticity.

ibtk/src/math/HierarchyMathOps.cpp
```diff
@@ -135,10 +135,12 @@ HierarchyMathOps::HierarchyMathOps(std::string name,
...
       d_finest_ln(finest_ln),
       d_fc_var(new FaceVariable<NDIM, double>(d_object_name + "::scratch_fc")),
-      d_nc_var(new NodeVariable<NDIM, double>(d_object_name + "::scratch_nc", NDIM)),
+      d_nc_s_var(new NodeVariable<NDIM, double>(d_object_name + "::scratch_nc_s", 1)),
+      d_nc_v_var(new NodeVariable<NDIM, double>(d_object_name + "::scratch_nc_v", NDIM)),
       d_sc_var(new SideVariable<NDIM, double>(d_object_name + "::scratch_sc")),
       d_of_var(new OuterfaceVariable<NDIM, double>(d_object_name + "::scratch_of")),
-      d_on_var(new OuternodeVariable<NDIM, double>(d_object_name + "::scratch_on", NDIM == 2 ? 1 : NDIM)),
+      d_on_s_var(new OuternodeVariable<NDIM, double>(d_object_name + "::scratch_on_s", 1)),
+      d_on_v_var(new OuternodeVariable<NDIM, double>(d_object_name + "::scratch_on_v", NDIM)),
```
drwells, line 143: Its ugly, but this should make everything work for both scalars and vectors in 2D and 3D.

ibtk/src/math/HierarchyMathOps.cpp
```diff
@@ -2033,14 +2064,32 @@ HierarchyMathOps::interp(const int dst_idx,
 {
     if (src_ghost_fill) src_ghost_fill->fillData(src_ghost_fill_time);
 
+    if (dst_cf_bdry_synch && dst_var->fineBoundaryRepresentsVariable() == false)
+    {
+        TBOX_ERROR("The combination of a destination boundary sync and coarse values representing the values on "
+                   "coarse-fine interfaces is not implemented.");
+    }
```
drwells, line 2071: There is no operator for outernode data corresponding to this setup so lets just abort.

boyceg: Are you sure that this doesn't work? I think that this is what is done in the side-to-node curl implementation.

My preference would be always to coarsen if `dst_cf_bdry_synch == true`.

boyceg: I just double-checked: the side-to-node curl implementation seems to work fine with either value of `fine_boundary_represents_variable`. I looked at the visual output and also ran the test suite.

drwells: It does but that's because SAMRAI (and by extension us) always ignore `fine_boundary_represents_variable`. My reasoning here is that if that is `true` then synchronization should be coarsening (set coarse values with fine), whereas if it is `false` then we should use refinement (set fine with coarse). In both `curl` and `interp` we use restriction (`xeqScheduleOuternodeRestriction`), which works since SAMRAI provides the class `OuternodeDoubleConstantCoarsen`. However: there is no `OuternodeDoubleConstantRefine` so the other way isn't implemented.

drwells: We can change the behavior for the sake of convenience but I think this is the correct setup in terms of "this means coarsening" and which outernode operators are implemented right now.

boyceg: I do not agree. We should only use the coarse grid interpolation to set the coarse grid values along the coarse-fine interface if the user says not to synchronize (`dst_cf_bdry_synch == false`). The reason is that the coarse grid interpolation will be using "invalid" values, whereas the fine grid interpolation will be using "valid" values.

If we want to synchronize coarse-fine interface values, we should always determine the overlying coarse grid values from the underlying fine grid values.

If a caller wants to do something different, they are able to do so by setting `dst_cf_bdry_synch == false` and then doing what they want along the coarse-fine interface.

Similarly, if a caller indicates that they want to use the coarse grid values in the `NodeVariable` to determine the value of the variable, then after setting the coarse grid values, we should go back and force the underling fine grid values in the `NodeVariable` to correspond to the coarse grid values. As above, if the caller wants to do something different (although I am having trouble seeing a use case), they are able to do so by setting `dst_cf_bdry_synch == false` and then doing what they want along the coarse-fine interface.

drwells: I'm missing something important here - the coarse-grid side-to-node interpolation is not done with the outernode synchronization step, correct? Where does that part happen/where should that part happen?

boyceg: One way to think about this is that the side-to-node interpolation first happens patch-by-patch on each level of the patch hierarchy, independent of values on other patch levels. We then have to decide what (if anything) to do at coarse-fine interfaces and in the "invalid" regions of coarser levels.

My thinking in these operators is that if we view a particular variable as a "composite grid variable", we want to use the "valid" values to determine coarse-grid values. So generally, at overlaps between levels, we want to define coarse values in terms of overlying fine values.

If `dst_cf_bdry_synch == true`, then I think that the coarse values along the coarse-fine interface should be determined from the overlying fine values. That "synchronizes" the coarse values along the coarse-fine interface.

In addition, if `dst_cf_bdry_synch == true` and `fineBoundaryRepresentsVariable() == false`, then that says that ***in the composite grid variable***, the coarse values should determine the values of the overlying fine values along the coarse-fine interface. Ensuring that requires applying the "H1 fixup" along the coarse-fine interface.

Does this make sense?

boyceg: One other thing: here, I don't think that `dst_var->fineBoundaryRepresentsVariable() == false` means that the coarse-grid node data should be determined from the coarse-grid side data. Instead, I think that `dst_var->fineBoundaryRepresentsVariable() == false` simply implies that along the coarse-fine interface, the fine values should be determined from the overlying coarse values, e.g., through the H1 fixup routine.

boyceg: > You're probably right but I want to make sure I understand what's going on.
> 
> > If we want to synchronize coarse-fine interface values, we should always determine the overlying coarse grid values from the underlying fine grid values.
> 
> I'm stuck at this step - isn't this the outernode restriction which we presently use? If so then doesn't this step always imply that `fine_boundary_represents_variable` must be true?

As far as I can tell, this flag is only used in SAMRAI for refining and not for coarsening.

boyceg: To summarize:

If `dst_cf_bdry_synch == false`, then we simply apply the patch-wise interpolation routine on every patch in the hierarchy, with no interlevel communication.

If `dst_cf_bdry_synch == true`, then we always restrict the results from finer levels onto coarser levels along the coarse-fine interface (via `xeqScheduleOuternodeRestriction`), independent of the value of `dst_var->fineBoundaryRepresentsVariable()`.

If `dst_cf_bdry_synch == true && dst_var->fineBoundaryRepresentsVariable() == false`, then we also apply the H1 fixup routine, to ensure that the fine coarse-fine interface values interpolate the overlying coarse-grid values. We consider this part of the "synchronization process" along coarse-fine interfaces.

boyceg: Specifically, if `dst_var` is configured so that `fineBoundaryRepresentsVariable() == true`, we interpret that as an indication that the caller ***does not*** want to constrain the fine-grid values along the coarse-fine interface to interpolate the overlying coarse-grid values.

boyceg: Finally, notice that the behavior here can be customized by setting `dst_cf_bdry_synch = false` and then applying a custom synchronization routine after the call to `HierarchyMathOps::interp()`.

boyceg: (One reason why it is actually ***good*** that SAMRAI ignores the result of `fineBoundaryRepresentsVariable()` in coarsening is that the only way to perform composite-grid calculations in SAMRAI that use both fine and coarse DOFs is on the ***fine patches*** along the coarse-fine interface; the result can then be communicated to the coarse patch through a coarsening operator.)

drwells: We chatted about this over zoom and the conclusion is:

1. if "synchronize dst" = false then we don't do anything with outernode or the H1 fixup routine
2. if "synchronize dst" = true and "fine side represents the solution" = false then we do outernode coarsening and the H1 fixup routine
3. if "synchronize dst" = true and "fine side represents the solution" = false then we do outernode coarsening (but no H1 fixup)

drwells:
That should be it!

ibtk/include/ibtk/HierarchyMathOps.h
```diff
@@ -81,6 +81,11 @@ namespace IBTK
  * "composite-grid" mathematical operations on a range of levels in a
  * SAMRAI::hier::PatchHierarchy object.
  *
+ * For nodal destination data, if the provided variable has
```
boyceg, line 84: 👍

ibtk/include/ibtk/HierarchyMathOps.h
```diff
@@ -81,6 +81,11 @@ namespace IBTK
...
+ * fineBoundaryRepresentsVariable() == false and synch_dst_cf_interface == true
+ * then each routine will call HierarchyMathOps::enforceHangingNodeConstraints()
+ * to achieve a conforming bi/trilinear discretization.
```
drwells, line 87: we don't actually do this yet but lets fix that in a followup.

boyceg:
Let's merge once the checks pass.

PR #1494, opened 2022-10-07
Modified the ex7 printing IB coordinates tutorial to keep track of the lagrangian indices

colegruninger97:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


examples/IB/explicit/ex7/example.cpp
```diff
@@ -330,9 +343,24 @@ main(int argc, char* argv[])
...
+        offsets[2] = offsetof(Lag_Coords, Y);
+        offsets[3] = offsetof(Lag_Coords, Z);
+
+        MPI_Type_create_struct(items, blocklength, offsets, types, &mpi_lag_coords_type);
+        MPI_Type_commit(&mpi_lag_coords_type);
+
         if (rank == 0)
         {
-            fout.open("Coordinates.txt", std::ios::out);
+            fout.open("IBCoordinates.txt", std::ios::out);
```
drwells, line 363:
```suggestion
            fout.open("IBCoordinates.txt");
```

examples/IB/explicit/ex7/example.cpp
```diff
@@ -330,9 +343,24 @@ main(int argc, char* argv[])
...
+        int blocklength[4] = { 1, 1, 1, 1 };
+        MPI_Datatype types[4] = { MPI_INT, MPI_DOUBLE, MPI_DOUBLE, MPI_DOUBLE };
+        MPI_Datatype mpi_lag_coords_type;
+        MPI_Aint offsets[4];
+        offsets[0] = offsetof(Lag_Coords, lag_idx);
+        offsets[1] = offsetof(Lag_Coords, X);
+        offsets[2] = offsetof(Lag_Coords, Y);
+        offsets[3] = offsetof(Lag_Coords, Z);
+
+        MPI_Type_create_struct(items, blocklength, offsets, types, &mpi_lag_coords_type);
```
drwells, line 358: This needs a corresponding `MPI_Type_free()` call once we are done with it to prevent leaking memory.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -391,95 +419,81 @@ main(int argc, char* argv[])
...
+                // put the lag idx as the first entry of the local_coordinates vector since the petsc indices may change
+                // throughout a computation
                 Eigen::Map<Vector3d> X(&x_vals[petsc_idx * NDIM]); // X is a vector containing the local coordinates
-                for (int m = 0; m < NDIM; m++)
-                {
-                    local_coordinates[NDIM * k + m] = X[m];
-                }
 
-                k = k + 1;
+                local_coordinates.push_back({ lag_idx, X[0], X[1], X[2] });
```
drwells, line 431: I'm surprised that this works - I don't think its valid in plain C++11. Could you add a constructor to `Lag_Coords` and call that instead?

examples/IB/explicit/ex7/example.cpp
```diff
@@ -54,6 +53,20 @@
...
+    double Y;
+    double Z;
+};
+
+bool
+operator<(const Lag_Coords& lhs, const Lag_Coords& rhs)
+{
+    return lhs.lag_idx < rhs.lag_idx;
+}
+
```
drwells, line 69:
```suggestion
```
(we can avoid defining this here)

examples/IB/explicit/ex7/example.cpp
```diff
@@ -391,95 +419,81 @@ main(int argc, char* argv[])
...
+            // Now if we are on the root process, let's go ahead and sort the global_coordinates vector according to the
+            // lag index and print out the coordinates to "IBCoordinates.txt"
             if (rank == 0)
             {
                 fout.unsetf(ios_base::showpos);
                 fout.setf(ios_base::scientific);
                 fout.precision(5);
 
-                for (int l = 0; l < total_nodes; l++)
+                std::stable_sort(global_coordinates.begin(), global_coordinates.end());
```
drwells, line 491:
```suggestion
                std::stable_sort(global_coordinates.begin(), global_coordinates.end(),
                [](const Lag_Coords &a, const Lag_Coords &b) {return a.lag_idx < b.lag_idx;});
```
Do we need a stable sort? Could we ever have two structs with equal Lagrangian indices?

drwells (changes requested):

drwells:
@colegruninger97 will you have time to address these by the end of the month? I'd like to get this in the next release.

colegruninger97:
> @colegruninger97 will you have time to address these by the end of the month? I'd like to get this in the next release.

Hi David, Yes! I should be able to get to this by the end of the month. I will try and have it finished before holidays. Thanks for the reminder.

examples/IB/explicit/ex7/example.cpp
```diff
@@ -54,6 +54,28 @@
...
 
+struct Lag_Coords
+{
+    int lag_idx; // The global lag index
+    double X;
+    double Y;
+    double Z;
+
+    // Default constructor
+    Lag_Coords(){};
```
drwells, line 65:
```suggestion
    Lag_Coords()
    : lag_idx(std::numeric_limits<int>::max()
    , X(std::numeric_limits<double>::quiet_NaN()
    , Y(std::numeric_limits<double>::quiet_NaN()
    , Z(std::numeric_limits<double>::quiet_NaN()
    {}
```
we should put in invalid values to prevent accidentally reading an uninitialized structure.

drwells (changes requested):
Thanks for getting back to this - lets fix one more minor thing.

drwells (approved):

drwells:
Thanks!

PR #1238, opened 2021-04-05
add support for uniform body forces for IB structures

boyceg:
### IBAMR Pull Request Checklist
- [x] Does the test suite pass?
- [X] Was clang-format run?
- [X] Were relevant issues cited? Please remember to add `Fixes #12345` to close
      the issue automatically if we are fixing the problem.
- [x] Is this a change others will want to know about? If so, then has a
      changelog entry been added?
- [X] If new data structures have been added to a class then have they been set
      up appropriately for restarts? If so, ensure that the restart version
      number is incremented.
- [x] Does this change include a bug fix or new functionality? If so, a new test
      or tests should be added. New tests should run quickly (less than a minute
      in release mode). If possible, an older test should gain a new option so
      that we do not need to compile more test executables.
- [X] Did you (if your account has permission to do so) set relevant labels on
      GitHub for the pull request?

PR #1498, opened 2022-10-20
Fix initialization of director vectors for gIB.

abarret:
From #1238, the retrieval of director vectors was changed in a way that's incompatible with `IBStandardInitializer`. @boyceg, do you remember the reason this was changed? If directors aren't initialized, should we return an arbitrary set of vectors, or should we throw an error?

I'll work on a test for `IBStandardInitializer`.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
I cannot remember. I guess it was to fix some bug? Looks like we may need more gIB tests. 😬

abarret:
I've changed the director vector retrieval function to output a warning if vectors are not set. This should be a loud warning in the log file that something is probably incorrect. I can't imagine a scenario when anyone would want to use gIB without first setting director vectors. I've also updated the test to supply vectors.

src/IB/IBRedundantInitializer.cpp
```diff
@@ -1496,8 +1496,11 @@ IBRedundantInitializer::getVertexBdryMassSpec(const std::pair<int, int>& point_i
 const std::vector<double>&
 IBRedundantInitializer::getVertexDirectors(const std::pair<int, int>& point_index, const int level_number) const
 {
-    if (!d_init_director_and_rod_on_level_fcn)
+    if (static_cast<size_t>(point_index.second) >= d_directors[level_number][point_index.first].size())
     {
+        TBOX_WARNING("Could not find enough director vectors on level " << level_number << " for structure "
+                                                                        << point_index.first << "\n"
+                                                                        << "Using the standard basis.\n");
```
boyceg, line 1503: This seems a little opaque; maybe something like:
```suggestion
                                                                        << "Initializing director vectors to the standard basis.\n");
```
Also, is it all directors for the structure?
```suggestion
                                                                        << "Initializing all director vectors for structure " << point_index.first  << " on level " << level_number << " to the standard basis.\n");
```

src/IB/IBRedundantInitializer.cpp
```diff
@@ -1496,8 +1496,11 @@ IBRedundantInitializer::getVertexBdryMassSpec(const std::pair<int, int>& point_i
 const std::vector<double>&
 IBRedundantInitializer::getVertexDirectors(const std::pair<int, int>& point_index, const int level_number) const
 {
-    if (!d_init_director_and_rod_on_level_fcn)
+    if (static_cast<size_t>(point_index.second) >= d_directors[level_number][point_index.first].size())
     {
+        TBOX_WARNING("Could not find enough director vectors on level " << level_number << " for structure "
```
boyceg, line 1501: Instead of emitting a warning, how about just emitting an error?

It seems like we should require that the correct number of director vectors is always provided, or perhaps allow for a special case in which _no_ directors are provided --- and in that case alone default to the standard basis vectors.

abarret: We do a lot of error checking when we read the input files (or run the function) to make sure everything is consistent sizes, e.g. `num_vertex == num_directors`. This warning will only be emitted if no directors are registered (either through input file or function), and then it will be emitted as many times as there are vertices. The idea was to make the log file clear that something is almost definitely wrong.

I think emitting an error is the best thing to do. The only situation in which initializing directors to the standard basis is valid is if the structure's tangential vector is in the z-direction.

boyceg: I guess the other option would be to initialize the directors so that the structure is in equilibrium...?

boyceg:
@abarret are there more fixes needed?

abarret:
I need to put together a test for `IBStandardInitializer`, but this is the only fix needed to get gIB working. I need to change the warning to an error.

abarret:
This is ready. I've added another test for the `ib_body_force_kirchhoff` test that uses `IBStandardInitializer`.

src/IB/IBRedundantInitializer.cpp
```diff
@@ -1496,11 +1496,9 @@ IBRedundantInitializer::getVertexBdryMassSpec(const std::pair<int, int>& point_i
 const std::vector<double>&
 IBRedundantInitializer::getVertexDirectors(const std::pair<int, int>& point_index, const int level_number) const
 {
-    if (!d_init_director_and_rod_on_level_fcn)
-    {
-        static std::vector<double> II = { 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0 };
-        return II;
-    }
+    if (static_cast<size_t>(point_index.second) >= d_directors[level_number][point_index.first].size())
```
boyceg, line 1499: 👍

boyceg (approved):
👍

PR #1499, opened 2022-10-23
Implement composite, analytically divergence-free B-spline kernels

boyceg:
This is intended to provide the noninterpolatory divergence-free kernels from https://doi.org/10.1016/j.jcp.2022.111500.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
This is intended to provide the noninterpolatory divergence-free kernels from https://doi.org/10.1016/j.jcp.2022.111500.

boyceg:
@colegruninger97 could you take a look and see if it corresponds to your MATLAB implementation?

drwells:
Adding tests should be pretty easy - just duplicate input files for `interpolate_velocity_01_2d` (and make sure we get good max norm errors). We don't really have 'manufactured solutions' tests with spreading - there we just print values.

ibtk/src/lagrangian/LEInteractor.cpp
```diff
@@ -1367,6 +1513,18 @@ string_to_kernel(const std::string& kernel_fcn)
...
+    {
+        switch (kernel_fcn.back())
+        {
+        case '2':
+            return COMPOSITE_BSPLINE_32;
+        case '3':
+            return COMPOSITE_BSPLINE_43;
+        }
+        break;
+    }
```
drwells, line 1527: ✔️

ibtk/src/lagrangian/fortran/lagrangian_interaction2d.f.m4
```diff
@@ -3187,3 +3187,486 @@ c
...
+     &                 (X(d,s)+Xshift(d,l)-X_cell(d))/dx(d))
+               endif
+            enddo
+         enddo
+c
+c     Interpolate u onto V.
+c
+         INTERPOLATE_2D_SPECIALIZE_FIXED_WIDTH(ic_lower(1), ic_upper(1),
+                                               ic_lower(0), ic_upper(0),
+                                               3)
```
drwells, line 3298: if we are reading a 3x3 stencil here then we need to populate the extra rows (or columns) of `w` with zeros.

boyceg: That happens automatically because the basic kernel functions all return 0 outside of their radius of support.

boyceg: However, per #1501, we could probably improve performance by making specialized helper macros that only use the nonzero values. For staggered data, we also probably also could relax the ghost cell width requirements for these kernels.

colegruninger97:
Okay, after a bit of foiling, I can confirm that you are using the same Bspline combinations as https://www.sciencedirect.com/science/article/pii/S0021999122005629?via%3Dihub

boyceg:
Great. Do you have code for the interpolatory versions of those kernels that could be added for comparison?

colegruninger97:
> Great. Do you have code for the interpolatory versions of those kernels that could be added for comparison?

I haven't yet tried implementing the interpolatory verions of the kernels; however, I'd be happy to give it a shot.

boyceg:
I think this is ready to merge.

boyceg:
@drwells I think this is good to merge. We still need to investigate the performance of these kernels, but that seems outside the scope of this PR. 😁

boyceg:
Whoops, just found a Fortran interface error.

boyceg:
OK, fixed now!

tests/spread/spread_01_2d.composite_bspline_43.mpirun=4.output
```diff
@@ -0,0 +1,35 @@
...
+[(8,4),(11,7)]
+[(12,0),(15,3)]
+[(12,4),(15,7)]
+rank: 3 level: 1 boxes:
+[(8,8),(11,11)]
+[(8,12),(11,15)]
+[(12,8),(15,11)]
+[(12,12),(15,15)]
+
+rank: 0
```
drwells, line 29: I think N needs to be bigger to get some values printed here.

boyceg: Hrumm.

boyceg: Damn Fortran. I lopped off the last argument of the composite B-spline 4/3 spreading subroutine.

boyceg: Take another look please. 😁

boyceg:
@drwells, I fixed the error that was resulting in empty output for the `COMPOSITE_BSPLINE_43` spreading test.

Issue #1477, opened 2022-07-04
harmonic avg changelog and tests


drwells:
Discussed in #1464  - we also need a changelog.

drwells:
@Raamkrishnan can we get this done in time for the release in December?

Raamkrishnan:
@drwells Yes. I will be able to do this before December.

Issue #1299, opened 2021-07-15
Start using clang-format


drwells:
It shouldn't be that hard to add another GitHub action that runs clang format and somehow communicates the result to the PR author. Some options:
- we could rig things up to set up a PR onto the callers PR that fixes indentation
- minimally, we can generate a patch for them to apply

I think the second one is probably okay - we've gotten away with less than that in the deal.II CI. In particular, we have an `indent` target in CMake that runs clang-format which makes it very easy for users to just fix the indentation locally on their own.

Hence, lets do the following:
- [ ] borrow `download_clang_format` and `make indent` from deal.II
- [ ] set up a CI test that makes sure PRs conform to clang-format
- [ ] have the CI indenter generate a patch

drwells:
If we pick a specific version we should also pick a new language standard so that lambdas are formatted correctly.

boyceg:
Sounds good.

PR #1505, opened 2022-10-27
amneetb/phase change code review

amneetb:
- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
For starters we need to fix the problems with the CI and add CMake support - do you see how to do that?

amneetb:
@Raamkrishnan : Do we not have CMake with this branch yet?

amneetb:
> For starters we need to fix the problems with the CI and add CMake support - do you see how to do that?

We will add the missing CMake support. That will fix most of the issues with CI (I think). Do you see any other issues apart from the missing CMake support?

drwells:
the autotools CI isn't happy either - I didn't investigate but it looks like something is missing there too.

This is a big PR! New features are good.

amneetb:
> This is a big PR! New features are good.

This has a lot of new physics as well, including a low Mach solver and the phase field method.

Raamkrishnan:
> @Raamkrishnan : Do we not have CMake with this branch yet?

No. This branch does not have CMake yet. But I can add it.

include/ibamr/AdvDiffConservativeCUIConvectiveOperator.h
```diff
@@ -0,0 +1,221 @@
...
+class RefineSchedule;
+} // namespace xfer
+} // namespace SAMRAI
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * \brief Class AdvDiffConservativeCUIConvectiveOperator is a concrete ConvectiveOperator
```
boyceg, line 64: How does this class differ from `AdvDiffCUIConvectiveOperator`, which also is supposed to have a "convective" mode (determined by the `ConvectiveDifferencingType` specified in the class constructor)?

The name `AdvDiffConservativeCUIConvectiveOperator` makes it sound like this is a specialization of `AdvDiffCUIConvectiveOperator`. Is that right?

include/ibamr/AdvDiffConservativeCUIConvectiveOperator.h
```diff
@@ -0,0 +1,221 @@
...
+{
+public:
+    /*!
+     * \brief Class constructor.
+     */
+    AdvDiffConservativeCUIConvectiveOperator(std::string object_name,
+                                             SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > Q1_var,
+                                             SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > Q2_var,
+                                             SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db,
+                                             ConvectiveDifferencingType difference_form,
```
boyceg, line 92: Should this class always use "conservative" differencing?

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,304 @@
...
+class BasePatchHierarchy;
+} // namespace hier
+} // namespace SAMRAI
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * \brief Class MassIntegrator is an abstract class which integrates
```
boyceg, line 65: Could this be a specialization of `HierarchyIntegrator`? There seems to be a good amount of duplicated functionality here. I would prefer to avoid introducing a new interface unless the existing one is not suitable and cannot made to be suitable.

include/ibamr/PhaseChangeHierarchyIntegrator.h
```diff
@@ -0,0 +1,511 @@
...
+     * \brief Get the operator that is used to evaluate the right-hand side of energy equation.
+     */
+    SAMRAI::tbox::Pointer<IBTK::LaplaceOperator>
+    getHelmholtzRHSOperatorForEnergyEquation(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > var);
+
+    /*!
+     * \brief Register an IBTK::CartGridFunction object to specify the value of the
+     * source term for energy equation.
+     */
+    void setEnergyEquationSourceTermFunction(SAMRAI::tbox::Pointer<IBTK::CartGridFunction> F_fcn);
```
boyceg, line 265: Should we standardize on whether or not to abbreviate "function"? Both are used here.

include/ibamr/PhaseChangeHierarchyIntegrator.h
```diff
@@ -0,0 +1,511 @@
...
+     * \brief Get the solver of the energy equation.
+     */
+    SAMRAI::tbox::Pointer<IBTK::PoissonSolver>
+    getHelmholtzSolverForEnergyEquation(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > var);
+
+    /*!
+     * \brief Get the operator that is used to evaluate the right-hand side of energy equation.
+     */
+    SAMRAI::tbox::Pointer<IBTK::LaplaceOperator>
+    getHelmholtzRHSOperatorForEnergyEquation(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > var);
```
boyceg, line 259: As above:
```suggestion
    getEnergyEquationHelmholtzRHSOperator(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > var);
```

include/ibamr/AllenCahnHierarchyIntegrator.h
```diff
@@ -0,0 +1,352 @@
...
+    SAMRAI::solv::RobinBcCoefStrategy<NDIM>* getLiquidFractionPhysicalBcCoef();
+
+    /*!
+     * Get the convective operator being used by this solver class for Allen-Cahn equation.
+     *
+     * If the convective operator has not already been constructed, then this
+     * function will initialize a default convective operator.
+     */
+    SAMRAI::tbox::Pointer<ConvectiveOperator>
+    getConvectiveOperatorForAllenCahnEquation(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > lf_var,
```
boyceg, line 182: Similar to suggestions in a few other places:
```suggestion
    getAllenCahnEquationConvectiveOperator(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > lf_var,
```

include/ibamr/BrinkmanPenalizationStrategy.h
```diff
@@ -78,11 +78,6 @@ class BrinkmanPenalizationStrategy : public SAMRAI::tbox::Serializable
      */
     virtual void postprocessComputeBrinkmanPenalization(double current_time, double new_time, int num_cycles);
 
-    /*!
-     * \brief Set Brinkman penalization penalty factor.
-     */
-    virtual void setBrinkmanCoefficient(double chi);
```
boyceg, line 84: Since this is in the public interface for this strategy class, I'd suggest that we indicate that the interface is deprecated rather than removing it entirely.

include/ibamr/BrinkmanPenalizationStrategy.h
```diff
@@ -98,14 +93,6 @@ class BrinkmanPenalizationStrategy : public SAMRAI::tbox::Serializable
         return d_object_name;
     } // getName
 
-    /*
-     * \brief Get the Brinkman coefficient.
-     */
-    double getBrinkmanCoefficient() const
```
boyceg, line 104: Same as above: since this is in the public interface for this strategy class, I'd suggest that we indicate that the interface is deprecated rather than removing it entirely.

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
```diff
@@ -282,6 +282,11 @@ class INSVCStaggeredHierarchyIntegrator : public INSHierarchyIntegrator
     virtual void
     registerBrinkmanPenalizationStrategy(SAMRAI::tbox::Pointer<IBAMR::BrinkmanPenalizationStrategy> brinkman_force);
 
+    /*!
+     * \brief Register function to compute the source term of the div U equation.
```
boyceg, line 286: Is this an optional function?

boyceg: (I mean the function to compute the source term, not the class member. 😁 )

scripts/formatting/setup_clang_format_precommit_hook.sh
```diff
@@ -12,5 +12,5 @@
 ##
 ## ---------------------------------------------------------------------
 
-cp scripts/check_formatting.sh .git/hooks/pre-commit
+cp scripts/formatting/check_formatting.sh .git/hooks/pre-commit
```
boyceg, line 15: 👍

boyceg (commented):
I did a quick review of the public interface and have a few comments and suggestions.

include/ibamr/AdvDiffConservativeCUIConvectiveOperator.h
Raamkrishnan: Yes. `AdvDiffConservativeCUIConvectiveOperator` class is a specialization of `AdvDiffCUIConvectiveOperator` class. Because this class also uses `CUI` as a convective operator type.  But the difference is `AdvDiffConservativeCUIConvectiveOperator` class computes \div{u Q1 Q2} whereas `AdvDiffCUIConvectiveOperator` class computes \div{u Q} or u.\grad{Q} based on the ConvectiveDifferencingType.

boyceg: I think this class may need a new name.

We do not have a 100% consistent naming system, but `AdvDiffConservativeCUIConvectiveOperator` suggests that it is a specialization (in the C++/object oriented programming sense) of `AdvDiffCUIConvectiveOperator`, but the header indicates that it is actually a specialization of class `ConvectiveOperator`. 

In fact, it seems like its main use case (to compute terms of the form `N = div(Q1 Q2 u)` doesn't really fit with the `ConvectiveOperator` interface. (`applyConvectiveOperator(int Q1_idx, int Q2_idx, int N_idx)` is only provided by this implementation.)

I haven't looked into the implementation, but does this basically just apply the standard CUI scheme with `Q = Q1*Q2`? If so, would suggest:

1. Provide an implementation of `applyConvectiveOperator(int Q1_idx, int Q2_idx, int N_idx)` in the base class in terms of the virtual method `applyConvectiveOperator(int Q_idx, int N_idx)`.
2. Provide an adaptor class that can use any `ConvectiveOperator` implementation to evaluate `applyConvectiveOperator(int Q1_idx, int Q2_idx, int N_idx)`.

The current design will require duplicating every `ConvectiveOperator` to also provide this other form, which I would prefer to avoid if possible.

include/ibamr/AdvDiffConservativeCUIConvectiveOperator.h
boyceg: Following up from above --- if the class always uses conservative differencing, then it should not take in `difference_form` as an argument here.

Raamkrishnan:
> For starters we need to fix the problems with the CI and add CMake support - do you see how to do that?



> For starters we need to fix the problems with the CI and add CMake support - do you see how to do that?

The CI is happy now. I have also added the CMake support. I am adding a test that solves Stefan's problem using zero viscosity, and harmonic interpolation. This will complete #1477 also.

include/ibamr/AdvDiffConservativeCUIConvectiveOperator.h
Raamkrishnan: @boyceg The class name is `AdvDiffCUIConservativeConvectiveOperator`. Since this class is a specialization of `AdvDiffCUIConvectiveOperator` it is derived from the `AdvDiffCUIConvectiveOperator` class. This class's main use is to compute `N = div(Q P u)`.

I have added a virtual function `applyConvectiveOperator(int Q_idx, int P_idx, int N_idx)` in the base class `ConvectiveOperator`. But the implementation is provided only in `AdvDiffCUIConservativeConvectiveOperator`.

Please take a look and let me know your suggestions.

include/ibamr/AdvDiffConservativeCUIConvectiveOperator.h
Raamkrishnan: Yes. Now the class `AdvDiffCUIConservativeConvectiveOperator` do not take `difference_form` as an argument.

include/ibamr/MassIntegrator.h
Raamkrishnan: @boyceg `MassIntegrator` is a lightweight specialized class used to integrate the mass conservation equation at both the cell center and side center within each time step. This class does not maintain any state. It is a temporary integrator.

include/ibamr/PhaseChangeHierarchyIntegrator.h
```diff
@@ -0,0 +1,511 @@
...
+    /*!
+     *  \brief Compute source terms of energy equation.
+     */
+    virtual void computeEnergyEquationSourceTerm(int F_scratch_idx, const double dt) = 0;
+
+    /*!
+     * \brief Get the solver of the energy equation.
+     */
+    SAMRAI::tbox::Pointer<IBTK::PoissonSolver>
+    getHelmholtzSolverForEnergyEquation(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > var);
```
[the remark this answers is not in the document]
Raamkrishnan: 👍

include/ibamr/PhaseChangeHierarchyIntegrator.h
Raamkrishnan: 👍

include/ibamr/AllenCahnHierarchyIntegrator.h
Raamkrishnan: 👍

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
Raamkrishnan: Yes. It is optional. By default \div u = 0 is used in the code. Only when `d_Div_U_F_fcn` is registered, we solve \div u \ne 0.

scripts/formatting/check_formatting.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3.10
```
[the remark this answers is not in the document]
Raamkrishnan: 👍

include/ibamr/BrinkmanPenalizationStrategy.h
Raamkrishnan: Instead of `chi`, we now use `penalty_factor` and the function name is  `setBrinkmanPenaltyFactor(double penalty_factor)`.

include/ibamr/BrinkmanPenalizationStrategy.h
Raamkrishnan: Changed the function name from `getBrinkmanCoefficient()` to ` getBrinkmanPenaltyFactor()`.

amneetb:
@boyceg and @drwells : Can we get this merged soon? We are building applications on top of this branch which ideally should be based on master. 

@Raamkrishnan : Can you commit the `T_ref` code and this branch up to speed with Master?

boyceg:
Sorry this dropped off my radar. I’ll take another look this week.

Raamkrishnan:
@amneetb I have added T_ref changes in the enthalpy code along with a few minor changes. I have also rebased against the master branch. Please take a look.

amneetb:
One problem I see with breaking this PR into smaller pieces is that each individual piece is not useful and it will make all examples defunct. Consider ex0, which is a simple standard Stefan problem. It requires almost all pieces of this PR to work. Many things are needed to even start doing a simple problem. 

Here is a roadmap that I think could help in understanding the algorithms of this PR:

- Section 4.2 of this preprint explain the time stepping scheme (https://arxiv.org/pdf/2301.06256.pdf)
- The starting point should be `EnthalpyHierarchyIntegrator` class. That class is derived from `PhaseChangeHierarchyIntegrator` class. 
- Currently we have two ways of doing phase change: Allen Cahn based version (phase field method) and enthalpy method.
- We have not written any paper on Allen Cahn approach yet, but it will happen sometime next fall. We are working on pushing enthalpy approach to the finish line first. 
- This mean you can first review EnthalpyHierarchyIntegrator first.


For `EnthalpyHierarchyIntegrator` here are few main things to keep in mind

1. It works with three phases. Gas, liquid, and solid. Gas is taken to be of distinct chemical phase than liquid and solid, which belong to the same material. The combined liquid and solid phase is termed the "phase changing material" (PCM).
2. Level set function $\phi$ is used to distinguish the gas-PCM region.
3. Liquid and solid regions are distinguished via the liquid fraction variable $\varphi$,  $\varphi = 1$ in liquid 0 in solid and between 0 and 1 in the so-called "mushy region"
4. \varphi is solved at the same time with the enthalpy equation. This is what makes the enthalpy equation nonlinear. All material properties like conductivity etc. that appear in the enthalpy equation, depend upon the unknown varphi variable.
5. The mushy region is "compressible" but in the Low Mach sense. 
6. The RHS of div.u \ne 0 because of the density jump between liquid and solid phases.

include/ibamr/AllenCahnHierarchyIntegrator.h
```diff
@@ -0,0 +1,337 @@
...
+{
+/*!
+ * \brief Class AllenCahnHierarchyIntegrator is a concrete class that manages the
+ * time integration of the Allen-Cahn and energy equation.
+ *
+ * In this class, implementations are provided for phase change of a pure material
+ * using the phase-field method.
+ *
+ *
+ *
```
drwells, line 67:
```suggestion
```

include/ibamr/AllenCahnHierarchyIntegrator.h
```diff
@@ -0,0 +1,337 @@
...
+    int d_H_sc_idx = IBTK::invalid_index;
+    int d_T_lf_N_scratch_idx = IBTK::invalid_index;
+
+    /*!
+     * Allen-Cahn equation parameters.
+     * M_lf  - Mobility
+     * lambda_lf - Mixing energy density
+     * eps_lf - Thickness of the liquid-solid interface
+     */
+    double d_M_lf, d_lambda_lf, d_eps_lf;
```
drwells, line 321: These should have default values, even if it's just `std::numeric_limits<double>::signaling_NaN()`.

include/ibamr/BrinkmanPenalizationStrategy.h
drwells: Boyce's comment still holds - other people are using this code so lets not change this without deprecating it first.

src/utilities/ConvectiveOperator.cpp
```diff
@@ -80,6 +80,15 @@ ConvectiveOperator::apply(SAMRAIVectorReal<NDIM, double>& x, SAMRAIVectorReal<ND
...
 } // apply
 
+void
+ConvectiveOperator::applyConvectiveOperator(int /*Q_idx*/, int /*P_idx*/, int /*N_idx*/)
+{
+    TBOX_ERROR(
+        "ConvectiveOperator::applyConvectiveOperator is a base class and the implementation is provided in "
+        "AdvDiffCUIConservativeConvectiveOperator class.\n");
+    return;
+} // applyConvectiveOperator
```
drwells, line 90: Is there a plan to implement this function in all of the derived classes? If there isn't a way to do so then this function should just be in the derived class.

drwells (changes requested):
The main thing to do next is to move some things into the library which are present in the majority of the examples. For instance, `SetFluidProperties.cpp` is essentially identical in all but one (ex5) example (and the headers always essentially identical) - we should move this into the library and perhaps add a common implementation instead of copying and pasting. The same is true for several other files:

- TemperatureInitialCondition.(h,cpp) is the same except for a sign flip in ex6
- SetLSProperties.(h,cpp) is always the same
- GravityForcing.(h,cpp) is always the same
- TagInterfaceRefinementCells.(h,cpp) is always the same

It would be great to take this a step further and abstract out the computations done in `LSLocateInterface`, `LevelSetInitialCondition`, and `LiquidFractionInitialCondition` since those are also always identical except for an inner loop. ex0 and ex5 have notably different `SetFluidProperties` - the other examples are the same AFAICT. One possibility would be to make those template classes which take a Functor as an argument (which can be called without any overhead in the inner loop).

drwells:
There's something wrong with how the examples are getting linked:
```
/usr/bin/ld: CMakeFiles/phase_change-ex5-2d.dir/example.cpp.o: in function `main':
example.cpp:(.text+0x1f72): undefined reference to `TagInterfaceRefinementCells::TagInterfaceRefinementCells(SAMRAI::tbox::Pointer<IBAMR::AdvDiffHierarchyIntegrator>, SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<2, double> >, double, double)'
/usr/bin/ld: example.cpp:(.text+0x1fba): undefined reference to `callTagInterfaceRefinementCellsCallbackFunction(SAMRAI::tbox::Pointer<SAMRAI::hier::BasePatchHierarchy<2> >, int, double, int, bool, bool, void*)'
collect2: error: ld returned 1 exit status
```

src/utilities/ConvectiveOperator.cpp
amneetb: I think this is implemented in a derived class. @Raamkrishnan can point out to the exact case.

examples/phase_change/ex0/SetFluidProperties.cpp
```diff
@@ -0,0 +1,345 @@
...
+                                               Pointer<Variable<NDIM> > Cp_var,
+                                               Pointer<IBTK::HierarchyMathOps> hier_math_ops,
+                                               const int cycle_num,
+                                               const double time,
+                                               const double current_time,
+                                               const double new_time,
+                                               void* ctx)
+{
+    // Set the density from the level set information
+    static SetFluidProperties* ptr_SetFluidProperties = static_cast<SetFluidProperties*>(ctx);
```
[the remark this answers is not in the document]
amneetb: The main idea behind this design was that we would like to set density/viscosity using any sort of interface capturing method (level set, VoF, phase field), and the user can define it based upon their choice during the application setup. We done this using level set and phase field as of now, but in the future we will also like to set it using VoF.

include/ibamr/AdvDiffCUIConservativeConvectiveOperator.h
boyceg, line 1: It seems like the main feature is the ability to evaluate `div(P Q u)`, not the ability to do "conservative CUI differencing". Does the `div(P Q u)` operator have a standard name in the literature where these kinds of things show up?

amneetb: No. The operator is basically advection of enthalpy $\nabla \cdot (\rho C_p T \bf{u})$. Here `P` is specific heat $C_p$ and $Q$ is temperature $T$.

include/ibamr/ConvectiveOperator.h
```diff
@@ -84,6 +84,11 @@ class ConvectiveOperator : public IBTK::GeneralOperator
      */
     virtual void applyConvectiveOperator(int Q_idx, int N_idx) = 0;
 
+    /*!
+     * \brief Compute N = div(Q1 Q2 u).
+     */
```
boyceg, line 89: Please make the variable name consistent between the comment and the function specification.

examples/phase_change/ex0/SetFluidProperties.cpp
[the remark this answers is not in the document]
amneetb: Perhaps `SetLSProperties.(h,cpp)` can be a helper function in the library. `GravityForcing.(cpp, h)` can also be, but the idea of Gravity forcing is that a user should be able to define it in any suitable manner (depending upon the interface tracking method). `TagInterfaceRefinementCells.(h,cpp)` cannot be because it is application dependent (what to tag for refinement).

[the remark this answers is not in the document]
boyceg: I'd like to keep application-specific stuff out of the library itself.

It might make sense to have a helper library for the examples to avoid code duplication.

On the other hand, I have often thought of the examples as providing self-contained starter code for a "real" application -- in which case it is helpful to have everything in place in the example directory.

include/ibamr/ConvectiveOperator.h
```diff
@@ -84,6 +84,11 @@ class ConvectiveOperator : public IBTK::GeneralOperator
...
+    virtual void applyConvectiveOperator(int Q_idx, int P_idx, int N_idx);
```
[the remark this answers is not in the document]
amneetb: For smooth things maybe yes, but it is not a good idea for sharp contrast to multiply Q and P. Physically speaking, the right thing to do is (which is what we do) is to have a mass flux $\rho \bf{u}$ computed from phase advection, apply limiter on $T$. For $C_p$ it is not clear whether it should be limited or not, because the quantity being advected is $T$ and not $C_p$. I think we tried both ways---limiting $C_p$ and just interpolation---and we have settled on one versus the other, don't recall which one. @Raamkrishnan ?

examples/phase_change/ex0/SetFluidProperties.cpp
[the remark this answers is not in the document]
amneetb: > 
> On the other hand, I have often thought of the examples as providing self-contained starter code for a "real" application -- in which case it is helpful to have everything in place in the example directory.

Yes, that is the idea behind the examples. It is just that we have many (useful) examples that are benchmarking cases from the literature.

src/utilities/ConvectiveOperator.cpp
amneetb: I think Boyce suggested to have it in the base `ConvectiveOperator` class. We didn't have it before. It was done (with some effort) after the suggestion.

examples/phase_change/ex2/SetFluidProperties.cpp
```diff
@@ -0,0 +1,525 @@
...
+                const Pointer<CellData<NDIM, double> > rho_data = patch->getPatchData(rho_idx);
+                // Li et al, 2015
+                for (Box<NDIM>::Iterator it(patch_box); it; it++)
+                {
+                    CellIndex<NDIM> ci(it());
+                    const double liquid_fraction = (*lf_data)(ci);
+                    const double heaviside = (*H_data)(ci);
+
+                    (*rho_data)(ci) = d_rho_gas + (d_rho_solid - d_rho_gas) * heaviside +
+                                      (d_rho_liquid - d_rho_solid) * liquid_fraction * heaviside;
```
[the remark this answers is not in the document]
amneetb: This is a mixture formula, which is not unique. It is up to the user to define a density mixture model --- this mostly depends upon a particular interface tracking technique. In phase field or conservative level set method 

$\rho = \rho_s \phi_s + \rho_s \phi_l + \rho_g  \phi_g$

where $\phi_i$ is the volume fraction of phase $I$.

include/ibamr/AdvDiffConservativeCUIConvectiveOperator.h
boyceg: I think we need to come up with a more precise name for the concept that is being provided here. Calling it a "conservative" specialization of the parent operator seems misleading, since the original CUI operator implementation itself also can be set up to perform conservative differencing.

From an implementation standpoint, it might be better to split out the cell-to-face flux computation and synchronization from the conservative differencing operation (which itself could be done in the base class) --- that way any convective operator implementation could be used to perform these operations without requiring a custom specialization with lots of duplicated code.

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,301 @@
...
+     */
+    virtual ~MassIntegrator() = default;
+
+    /*!
+     * \brief Integrate density and compute the convective operator of the momentum and/or energy.
+     */
+    virtual void integrate(double dt) = 0;
+
+    /*!
+     * \name General operator functionality.
```
boyceg, line 91: Should this be "integrator"?

include/ibamr/MassIntegrator.h
boyceg: I don't follow why this is not a `HierarchyIntegrator`.

Can you add some comments to the header explaining why this doesn't implement the `HierarchyIntegrator` interface?

Alternatively: It might make more sense to create a "lightweight" integrator base class that is specialized by `HierarchyIntegrator`.

include/ibamr/AdvDiffConservativeCUIConvectiveOperator.h
boyceg: > From an implementation standpoint, it might be better to split out the cell-to-face flux computation and synchronization from the conservative differencing operation (which itself could be done in the base class) --- that way any convective operator implementation could be used to perform these operations without requiring a custom specialization with lots of duplicated code.

I think this is really the right thing to do --- i.e. splitting out the functionality into separate functions. I think it will reduce the amount of duplicated code between the different operator implementations. I will try to put up an implementation in the next day or two...

include/ibamr/MassIntegrator.h
amneetb: This class is integrating the mass balance PDE over a single time step and passing the mass flux to the main HieracyIntegrator.

HierarchyIntegrator's philosophy is very different. It is derived from TagAndInitializeStrategy and is responsible for data management and tagging and regridding.

OTOH, mass integrator is basically solving the PDE over one time step and giving the flux to main HierachyIntegrator (e.g., PhaseChangeHI or INSVC). It is not doing any stuff related to hierarchy management.

We can call it MassPDEIntegrator?

boyceg:
@amneetb, @Raamkrishnan, would you all be able to split all of the examples into a separate PR?

amneetb:
> @amneetb, @Raamkrishnan, would you all be able to split all of the examples into a separate PR?

You mean `examples/phase_change` ?

drwells:
Yes. We should figure out the changes to the advection integrators first and then do the examples second since the discussions for each are unrelated.

amneetb (commented):
@drwells: It appears that the new clang-format (v 16.0) is doing formatting very differently. @Raamkrishnan called ./scripts/formatting/reformat script on this branch.

drwells:
I don't know what's up with that. Please undo that commit - we should not mix in changes to other parts of the library here.

amneetb:
> I don't know what's up with that. Please undo that commit - we should not mix in changes to other parts of the library here.

Yes we will do that. But how to do the formatting in future? We have two branches one with old formatting (this current PR) and a new temporary one with new format. We can't rebase them because of the formatting issues.

amneetb:
Should IBAMR ship with its own clang-format script so that we all are using the same thing?

drwells:
The status quo we've had for a couple years of asking people to run clang-format themselves and running it over everything for the release has been OK - lets discuss changes to that in #1299.

We're already at 77 conversation items and multiple unresolved discussion threads - we need to split this up into multiple PRs. How soon can you put up separate changesets for the new features and for the examples?

amneetb:
I don't know how to do it. If we break it into multiple commits, does that mean that all commits need to compile and pass the testsuite? This will mean that intermediate Makefiles need to be generated for these commits.

amneetb:
By multiple commits I meant multiple PRs. This will make it a mess. I think you were mainly concerned about code duplicity in the examples. This is unavoidable if we want to keep library general or we want to have examples.

drwells:
> I don't know how to do it. If we break it into multiple commits, does that mean that all commits need to compile and pass the testsuite? This will mean that intermediate Makefiles need to be generated for these commits.

Yup - this shouldn't be too bad, we just need to regenerate `configure` a couple of times.

We need to go through the same review process (e.g., dealing with the multilevel bugs) regardless of how many PRs we open. Yes, it will be messy, but splitting it up will be less so since we can tackle one thing at a time.

If we merge this PR as-is we will end up with 20 different copies of `GravityForcing.h` - splitting things up will give us an opportunity to examine how we can write a common interface for these functions so that we don't need to do that. I'm in favor of adding whatever we need to add to IBAMR to make that easier. I'm generally in favor of adding these kinds of common forces to the library. The same goes for the other mostly-duplicated files.

amneetb:
That's just too much work with little gains. We can have this as a separate branch that we can keep rebasing against master.

drwells:
We all have limited time to work on these things; merging small pieces quickly is the fastest route forward for everyone.

It's relatively straightforward to chop this up into two commits - I'm able to compile all of the library and tests with one branch containing commit A and all of the examples (except for the pre-existing link error in phase_change/ex5) on a second branch containing commit A and commit B. If that works for you I can set those up as two branches and we can proceed from that point.

amneetb:
Two PRs is what we had discussed earlier. When you mentioned multiple, implying greater than 2,it seemed that you wanted a different workflow.

Sure, please chop as you see fit. We really are not that great with Git and merging, and afraid that our workflow will break the branch and make it useless for us to carry out the day-to-day research. 

Yesterday I tried that myself but there is a clang-formatting issue with my workflow. It made my branch unsuitable to merge with rest of the IBAMR. 

Breaking the branch into PRs is really not that easy for some of us :-(

drwells:
We might need more than two, but we should start with two (library and examples). Further splits might be more obvious afterwards. This is one place where git is somewhat bimodal - this either takes 3 minutes or 3 hours depending on how good you are at these manipulations.

> afraid that our workflow will break the branch and make it useless for us to carry out the day-to-day research.

all the more reason to work efficiently. I'll set up the branches in a minute.

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
amneetb: @boyceg `registerFluidSourceFunction()` is defined in `INSHierarchyIntegrator`. We don't know what it is suppose to be doing. It is defined in `INSStaggeredHierarchyIntegrator` that cannot handle variable density, so it cannot be $\nabla \cdot \bf{u} = -1/\rho \frac{D \rho}{D t}$. We thought it purpose is different than the low Mach model, hence we defined a new function. 

If  `registerFluidSourceFunction` is supposed to be low Mach model, we can use this function instead of defining `registerDivergenceVelocitySourceFunction` in INSVC.

boyceg: `registerFluidSourceFunction` gives a way to set $\nabla \cdot {\bf u} = Q` for the supplied `Q` `CartGridFunction`. It seems like that is the same thing that `registerDivergenceVelocitySourceFunction` is doing?

amneetb: Yes. We will then use the existing function.

Alternatively, we can rename registerfluidSourceFunction to registerDivergenceVelocityFunction to make it more clear?

boyceg: That would be fine, although we should deprecate the old function and have it call the new one. Suggest `registerVelocityDivergenceFunction`.

PR #1506, opened 2022-10-28
Provide side-to-node curl in 3D

boyceg:
This is intended to provide robust support for side-to-node curl operators, and to use them to output the vorticity for `INSStaggeredHierarchyIntegrator`.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
The problem is with setting nodal ghost values, right? I am pretty sure this is totally broken. Hopefully `Outernode` will work.

drwells:
Another possibility would be to dig deep into SAMRAI and find the box they forgot to convert from cell to nodal, but that is very much a needle in a haystack problem.

boyceg:
No, the problem actually is that side ghost values are not being set (for some reason that I don't understand yet).

boyceg:
Ah, it is because the default "BC fill op" does not do any refining! Whoops...

drwells:
Good thing this works!

boyceg:
> Good thing this works!

Well, almost --- I also think there is an error in one or more of the finite difference formulas.

boyceg:
OK, I think I sorted it out. Will try to push something later tonight.

BTW, this will benefit from the same coarse-fine fix-up as velocity interpolation. Maybe that could all go in a separate PR?

boyceg:
@drwells, @abarret, this is ready for review.

tests/IBTK/curl_01_3d.output
```diff
@@ -0,0 +1 @@
+|e|_oo = 2.14735
```
boyceg, line 1: The large error is because of the coarse-fine interface treatment, which is using `CONSERVATIVE_LINEAR_REFINE` to fill ghost values --- so we only expect (and realize) first order accuracy.

tests/IBTK/curl_01_2d.output
```diff
@@ -0,0 +1 @@
+|e|_oo = 0.39974
```
boyceg, line 1: The large error is because of the coarse-fine interface treatment, which is using `CONSERVATIVE_LINEAR_REFINE` to fill ghost values --- so we only expect (and realize) first order accuracy.

drwells: I checked and we do get first-order convergence in the max norm - for some reason the visualization of the error doesn't include this jump and looks smooth. Must be a visit thing.

drwells (approved):

PR #1509, opened 2022-11-01
add a central pressure tap to IBFE instrument panel

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBFEInstrumentPanel.cpp
+++ b/src/IB/IBFEInstrumentPanel.cpp
@@ -29,7 +29,6 @@
 
 namespace IBAMR
 {
-
 /////////////////////////////// STATIC ///////////////////////////////////////
 
 namespace
@@ -360,7 +359,6 @@ IBFEInstrumentPanel::readInstrumentData(const int U_data_idx,
     const double* const dx_coarsest = grid_geom->getDx();
     TBOX_ASSERT(grid_geom->getDomainIsSingleBox());
     const Box<NDIM> domain_box = grid_geom->getPhysicalDomain()[0];
-    int bcast_root = IBTK::invalid_index;
     for (unsigned int meter_idx = 0; meter_idx < d_num_meters; ++meter_idx)
     {
         const libMesh::Point x_centroid = put_point_in_domain(
@@ -398,10 +396,12 @@ IBFEInstrumentPanel::readInstrumentData(const int U_data_idx,
                 x_centroid_local_ln = ln;
             }
         }
-        x_centroid_global_ln = IBTK_MPI::maxReduction(x_centroid_local_ln);
+        int bcast_root = -1;
+        x_centroid_global_ln = IBTK_MPI::maxReduction(x_centroid_local_ln, &bcast_root);
         if (x_centroid_global_ln == x_centroid_local_ln)
         {
-            bcast_root = IBTK_MPI::getRank();
+            // The centroid should be on a patch owned by exactly one processor
+            TBOX_ASSERT(bcast_root == IBTK_MPI::getRank());
             Pointer<PatchLevel<NDIM> > level = hierarchy->getPatchLevel(x_centroid_local_ln);
             if (x_centroid_local_patch_idx != IBTK::invalid_index)
             {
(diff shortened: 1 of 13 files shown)


drwells:
This logic for looping across the patch levels should work, but could we just do the following instead?
1. set `my_intersecting_patch_level = -1`
2. loop across all locally owned patches and levels, update `my_intersecting_patch_level` if it contains the point we want
3. do a max reduction over `my_intersecting_patch_level`
4. if `my_intersecting_patch_level == max(my_intersecting_patch_level)` then use `LEInteractor` and broadcast that result

boyceg:
Sure.

drwells:
Do we want to get this done ASAP? If so I can put together a test.

boyceg:
A test would be great!

boyceg:
(We also should make tests for the case where the nodeset “misses” some points along the perimeter for Charles’ PR.)

drwells:
```

        IBFE/instrument_panel_01_3d.input --------------------------- RUN FAILED
test failed with output:
    P=00000:Program abort called in file ``/__w/IBAMR/IBAMR/ibtk/src/lagrangian/LEInteractor.cpp'' at line 2281
    P=00000:ERROR MESSAGE: 
    P=00000:Failed assertion: Q_size / Q_depth == X_size / X_depth
```
looks like we have a few other things to fix too.

boyceg:
@drwells: I have made the requested changes. I am also updating the way that quadrature point assignments are made for the meter meshes, to align with the way that it is done for the central point. I will make that a separate commit --- I am not sure it is clearly giving a reduction in complexity. 😁

boyceg:
I guess that one advantage of this approach is that it could work with fully distributed Cartesian grid data structures (which would be available in SAMRAI 3/4).

boyceg:
@drwells: Looks like all tests are passing. Let me know what you think about the second commit.

src/IB/IBFEInstrumentPanel.cpp
```diff
@@ -330,9 +351,85 @@ IBFEInstrumentPanel::readInstrumentData(const int U_data_idx,
         }
     }
 
+    // Interpolate the pressure at the centroid of each meter mesh.
+    //
+    // TODO: Factor out common code for finding assignments of points to patches.
```
drwells, line 356: it would be nice to re-use FEDataManager here but I don't think its flexible enough - see also https://github.com/drwells/fiddle/pull/84

src/IB/IBFEInstrumentPanel.cpp
```diff
@@ -330,9 +351,85 @@ IBFEInstrumentPanel::readInstrumentData(const int U_data_idx,
...
+                    TBOX_ERROR("no pressure data!\n");
+                }
+            }
+        }
+    }
+
     // Synchronize the values across all processes.
     IBTK_MPI::sumReduction(d_flow_rate_values.data(), d_num_meters);
     IBTK_MPI::sumReduction(d_mean_pressure_values.data(), d_num_meters);
+    IBTK_MPI::sumReduction(d_centroid_pressure_values.data(), d_num_meters);
```
drwells, line 432: While this is correct I would still prefer to broadcast unique values instead of doing a reduction. Perhaps we can improve this later.

drwells (approved):
The extra changes to the quadrature point indexing should be OK since the tests pass - did this absorb @cpuelz's patch?

Lets add a test for the pressure before we merge.

src/IB/IBFEInstrumentPanel.cpp
boyceg: NP. I just made this change.

boyceg:
> The extra changes to the quadrature point indexing should be OK since the tests pass - did this absorb @cpuelz's patch?
> 
> Lets add a test for the pressure before we merge.

No, this does not include @cpuelz's patch (#1508). I am going to take a look at that code separately.

drwells:
Do you want to write the test or should I handle it?

boyceg:
If you could write or update a test, that would be wonderful. I don't think I have enough time today both to write a test for this and also to review #1508.

drwells:
I can wrap this up this afternoon (I need to go teach right now). The CI isn't happy about some signed vs. unsigned comparisons - I can fix those too.

boyceg:
@drwells I will force push a fix for the comparison warning.

boyceg:
> @drwells I will force push a fix for the comparison warning.

Comparison warning is fixed now. (I need to figure out why I'm not seeing these things in my editor...)

drwells:
My best guess is that that particular CI image is using a newer and pickier version of GCC.

boyceg:
> My best guess is that that particular CI image is using a newer and pickier version of GCC.

It turns out that I need to add `-Wextra` to get those warnings to show up in my UI.

drwells:
Sometimes we don't calculate the meter value correctly in parallel - I'm investigating.

drwells:
I fixed the broadcast bug and added three tests - this should be good to go once the CI check is complete.

PR #1510, opened 2022-11-01
CI: Don't use set-output.

drwells:
This is deprecated - see

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


drwells:
This is enough of a fix to work around our uses of deprecated features - we need to upgrade from cache v2 to cache v3 to fix the remainder of the warnings.

boyceg (approved):

PR #1513, opened 2022-11-03
update .gitignore to include .clang-tidy files

boyceg:
The usual checklist does not apply.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ compile_flags.txt
 *.sublime-workspace
 .cache
 .clangd
+.clang-tidy
 
 # Created by https://www.gitignore.io/api/linux,osx,c,c++,fortran
 


drwells (approved):

boyceg:
@drwells should we pick some `clang-tidy` checks that we think everyone should use? 😁

drwells:
We could definitely set up the CI to run clang-tidy and fail if there are any warnings - see also https://github.com/dealii/dealii/blob/981214b60b491f465dd8cf04e24989a2d806229c/contrib/ci/Jenkinsfile.tidy

drwells:
How do we do these days with clang-tidy? Are there thousands of warnings? We do have `-Werror` on for part of the CI which helps keep things clean.

boyceg:
It depends on which checks are on. With:
```
Checks:		'modernize-*,mpi-*,performance-*,redability-*,
		 -modernize-use-trailing-return-type,
		 -performance-unnecessary-value-param'
```
it is not too bad.

`bugprone-*` and/or `cppcoreguidelines-*` generate a ton of warnings.

PR #1514, opened 2022-11-03
remove static system names in the FEMechanicsBase, IBFEMethod, and FEDataManager interfaces

boyceg:
This will make it easier to have multiple instances of FEMechanicsBase in a single simulation code.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
Somehow this seems to have messed up restart... 😕

boyceg:
Hmmm, it looks like there are complex fluids examples that are not in the autotools build system (@abarret 😁).

abarret:
> Hmmm, it looks like there are complex fluids examples that are not in the autotools build system (@abarret 😁).

Which examples? They should all be there. I think the complex fluid code predates the CMake build system.

examples/IBFE/explicit/ex0/example.cpp
```diff
@@ -293,7 +293,7 @@ main(int argc, char* argv[])
         {
             for (unsigned int d = 0; d < NDIM; ++d)
             {
-                u_bc_coefs[d] = NULL;
+                u_bc_coefs[d] = nullptr;
```
drwells, line 296: Could we move this modernization stuff to another PR?

boyceg: Sure. I got carried away. 😁

boyceg: I am going to split this into 3 commits --- one for the core library, one for the tests, one for the examples.

drwells:
As far as changing examples goes - is this going to break people's applications? Ideally everything in the examples would continue to work.

boyceg:
This is necessarily a breaking change.

boyceg:
The intention here is to make it possible for there to be, e.g., independent `IBFEMethod` and `FEMechanicsExplicitIntegrator` objects running in the same application / working with the same mesh. I was thinking that we need to have different variable names for the different instances, but maybe that is not necessary...

boyceg:
@drwells, how does `attest` run the restart tests? I am getting a failure in the IBFE test `explicit_ex4_2d.amr.restart=50.mpirun=4` but cannot seem to run it "by hand".

BTW --- if you think this is not the right thing to do, I can just abandon the branch.

boyceg:
I'm currently running the `IBFE` tests, but it looks like all of the other sets of tests that involve `FEMechanicsBase`, `IBFEMethod`, and/or `FDEDataManager` are working.

abarret:
> This is necessarily a breaking change.

If we're going to be breaking interfaces anyway, can we go ahead and replace the static names in `IIMethod` too?

boyceg:
> > This is necessarily a breaking change.
> 
> If we're going to be breaking interfaces anyway, can we go ahead and replace the static names in `IIMethod` too?

I was going to do that once `IBFEMethod` works. But I am not going to do any more unless @drwells gives this his blessing. 😁

drwells:
The primary benefit of this change is we can start doing DLM and some other interesting things. The downside is that every single user of IBFE is going to have to make significant changes to their own code to keep up to date (e.g., we need to pass a lot more system information in ex5). In practice that just means most users will keep using the older IBAMR release and never migrate. Hence I don't think its worth doing.

For the purposes of DLM or other interesting coupling strategies we should just move to fiddle/deal.II which don't have these issues.

boyceg:
> The primary benefit of this change is we can start doing DLM and some other interesting things. The downside is that every single user of IBFE is going to have to make significant changes to their own code to keep up to date (e.g., we need to pass a lot more system information in ex5). In practice that just means most users will keep using the older IBAMR release and never migrate. Hence I don't think its worth doing.
> 
> For the purposes of DLM or other interesting coupling strategies we should just move to fiddle/deal.II which don't have these issues.

Is this ready to go now with complex models? If not, we need to be able to do this in `IBFEMethod` --- even if it is something that is a configure-time option.

boyceg:
BTW, all tests work with this branch when I run in opt mode.

boyceg:
Alternatively, perhaps @jabrown893 can just use this branch for the TAVR model, and we occasionally synch it up with `master`.

boyceg:
(BTW, I am not sure that IBFE ex5 is a representative example, since it has the peculiar feature of selecting between a surface and volume structure representation at runtime.)

drwells:
We have a solid mechanics integrator written by @marshallrdavey which can do general implicit elasticity (these repos are not public so not everyone can read the source code)

https://github.com/IBAMR/project-dlm/blob/master/include/dlm/mechanics/ImplicitMechanics.h

we also have an implementation of approximate DLM with spring forces:

https://github.com/IBAMR/project-dlm/blob/master/include/dlm/interaction/DLMMethod.h

It runs but I wasn't convinced that it works - I wasn't able to get something to converge to a manufactured solution but that problem may be conceptual rather than with the implementation. Hence the state is pretty similar to what we have now in IBAMR in the sense that we have equivalents to IBFEMethod and FEMechanicsBase written but the coupling is not finished.

drwells:
Its not ideal but I would strongly prefer using this branch with TAVR (and maintaining this branch) than making downstream projects change things.

boyceg:
> Its not ideal but I would strongly prefer using this branch with TAVR (and maintaining this branch) than making downstream projects change things.

OK with me! I will still need to finish the examples then. 😀

drwells:
It will be much faster to support Jordan than it will be to answer all the angry user emails too!

boyceg:
> > Hmmm, it looks like there are complex fluids examples that are not in the autotools build system (@abarret 😁).
> 
> Which examples? They should all be there. I think the complex fluid code predates the CMake build system.

See [48aa518](https://github.com/IBAMR/IBAMR/pull/1514/commits/48aa5182305b514de6086cf017f72f57da667970).

abarret:
> > > Hmmm, it looks like there are complex fluids examples that are not in the autotools build system (@abarret grin).
> > 
> > 
> > Which examples? They should all be there. I think the complex fluid code predates the CMake build system.
> 
> See [48aa518](https://github.com/IBAMR/IBAMR/pull/1514/commits/48aa5182305b514de6086cf017f72f57da667970).

That is bizarre. These examples have never been buildable with autotools, which means there was a good 9 months where these examples could not be built at all.

boyceg:
> > > > Hmmm, it looks like there are complex fluids examples that are not in the autotools build system (@abarret grin).
> > > 
> > > 
> > > Which examples? They should all be there. I think the complex fluid code predates the CMake build system.
> > 
> > 
> > See [48aa518](https://github.com/IBAMR/IBAMR/pull/1514/commits/48aa5182305b514de6086cf017f72f57da667970).
> 
> That is bizarre. These examples have never been buildable with autotools, which means there was a good 9 months where these examples could not be built at all.

Looks like my initial fix was incomplete; see [fa51556](https://github.com/IBAMR/IBAMR/pull/1514/commits/fa51556cc22a71aa623552faba41b0f7943b8492).

examples/Makefile.am
```diff
@@ -13,14 +13,14 @@
 
 ## Process this file with automake to produce Makefile.in
 include $(top_srcdir)/config/Make-rules
-SUBDIRS = CIB ConstraintIB IB IBFE IBLevelSet IIM IMP adv_diff advect fe_mechanics navier_stokes level_set multiphase_flow vc_navier_stokes wave_tank
+SUBDIRS = CIB ConstraintIB IB IBFE IBLevelSet IIM IMP adv_diff advect complex_fluids fe_mechanics navier_stokes level_set multiphase_flow vc_navier_stokes wave_tank
```
drwells, line 16: Wild! Can we move this to a new PR?

boyceg: I will leave it to @abarret. 😀

boyceg:
What about making the symbol visibility something that can be set at `configure` time, defaulting to the old API?

drwells:
That's probably too confusing for users - people will have mutually incompatible copies of IBAMR.

boyceg:
Since we can accomplish the goals of this without changing the API (see #1517), I will close this PR. (Although I think we should consider making this change when* we update to SAMRAI 3/4!!)

drwells:
One way to proceed would be to add `[[deprecated("use accessor functions instead")]]` to all of the global constants and add the accessor functions (but make them return the same values). That would give people an upgrade path.

boyceg:
@drwells I am happy to do that if it is OK with you.

drwells:
There may be some additional subtleties which I am missing but I think that should be OK - I am fine with breaking people's code as long as they are informed ahead of time of the upgrade path.

PR #1517, opened 2022-11-04
add DLM FSI example

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
@jabrown893: this is an example of how to set up a model that uses a separate (explicit) mechanics solver with subcycling in the mechanics solver. We should try to do the same thing for your TAVR models --- at least for the stents, which will have a mass density larger than the background fluid.

boyceg:
@amneetb or @Raamkrishnan, would one of you all be able to make a version of this new example that uses the multiphase flow solver instead of the constant coefficient solver?

This would be variable density and (for now) constant viscosity, with the density field _prescribed_ by the IB model. I.e., the density field should not be advected by the fluid solver.

examples/DLM/ex0/example.cpp
```diff
@@ -0,0 +1,789 @@
...
+            }
+            for (int k = 0; k < num_fe_substeps; ++k)
+            {
+                const double sub_dt = 0.5 * dt / num_fe_substeps;
+                const double start_time = loop_time + 0.5 * dt + k * sub_dt;
+                const double end_time = loop_time + 0.5 * dt + (k + 1) * sub_dt;
+                fem_solver->preprocessIntegrateData(start_time,
+                                                    end_time,
+                                                    /*num_cycles*/ 1);
+                fem_solver->modifiedTrapezoidalStep(start_time, end_time);
```
drwells, line 619: Interesting approach. A major source of complication in `DLMMethod` (from the DLM project) was that we interpolated positions in time - it looks like that is skipped here.

boyceg: You don't need to do that if you couple via fractional stepping.

boyceg: I think fractional stepping (e.g. using Strang) makes a lot of sense here because the idea (at least for stiff elasticity) is that you want to be taking smaller structural time steps _anyway_.

drwells: Is there any loss of accuracy from holding the IBFE position constant over the substeps?

boyceg: Surely it impacts accuracy --- but I think it does not impact order of accuracy. Also, I don't think this is the major determinant of accuracy here.

(BTW, we do the same thing here: https://arxiv.org/abs/2204.12907.)

boyceg: BTW, presumably the better way to do this is to use a more modern time stepping scheme that properly handles multiple time scales.

drwells: I might give this a try with the deal.II code later :)

boyceg: I just added some input files demonstrating how to increase structural stiffness without reducing the overall time step size (although, of course, it is necessary to increase the number of structural sub-steps).

drwells:
We may be missing some input files. The CMake stuff should be OK once things compile cleanly - we can see the example being compiled in the log.

boyceg:
I want to double check one of the numerical parameters before merging...

amneetb:
> @amneetb or @Raamkrishnan, would one of you all be able to make a version of this new example that uses the multiphase flow solver instead of the constant coefficient solver?
> 
> This would be variable density and (for now) constant viscosity, with the density field _prescribed_ by the IB model. I.e., the density field should not be advected by the fluid solver.

Done ;-)

https://github.com/IBAMR/IBAMR/tree/amneetb/add-multiphase-elasticity-example/examples/multiphase_flow/ex14

boyceg:
I think that this is ready to merge.

boyceg:
@drwells, does this look OK?

drwells (approved):
I have some minor suggestions. I see some nice vortex shedding with `input2d`.

boyceg:
Thanks for the updates. Will merge once the CI is happy.

PR #1522, opened 2022-11-15
INSStaggeredHierarchyIntegrator: nodal vorticity.

drwells:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

This should be good to go - I'm doing pressure now (that requires writing a new 3D cell-to-nodal function)


tests/IBTK/curl_01_2d.conforming.output
```diff
@@ -0,0 +1 @@
+|e|_oo = 0.195336
```
drwells, line 1: Interestingly, this choice halves the error.

boyceg (approved):

tests/IBTK/curl_01_2d.conforming.output
boyceg: Hmmm.

PR #1523, opened 2022-11-15
Nodal pressure

drwells:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

This is probably OK but I want to do some more checks first since we really need to change the API for the cell-to-node interpolation function to make this work correctly. The present version lets us populate ghost values with interpolation, which we don't do for other data centerings.


src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -571,6 +571,7 @@ INSStaggeredHierarchyIntegrator::INSStaggeredHierarchyIntegrator(std::string obj
     // use a conforming discretization (the solution, with bilinear/trilinear
     // nodal data, will be continuous)
     d_U_nc_var = new NodeVariable<NDIM, double>(d_object_name + "::U_nc", NDIM, /*fine_boundary_represents_var*/ false);
+    d_P_nc_var = new NodeVariable<NDIM, double>(d_object_name + "::P_nc", 1, /*fine_boundary_represents_var*/ false);
     d_F_cc_var = new CellVariable<NDIM, double>(d_object_name + "::F_cc", NDIM);
```
boyceg, line 575: `F_cc_var` is also just for output --- probably should switch that one too!

boyceg:
I agree that we should have a consistent API for ghost cell filling. Source data `HierarchyGhostCellInterpolation` objects probably should be included in all or none.

ibtk/src/math/HierarchyMathOps.cpp
```diff
@@ -2003,30 +2003,64 @@ HierarchyMathOps::interp(const int dst_idx,
...
         for (PatchLevel<NDIM>::Iterator p(level); p; p++)
         {
             Pointer<Patch<NDIM> > patch = level->getPatch(p());
 
             Pointer<NodeData<NDIM, double> > dst_data = patch->getPatchData(dst_idx);
             Pointer<CellData<NDIM, double> > src_data = patch->getPatchData(src_idx);
 
-            d_patch_math_ops.interp(dst_data, src_data, patch, dst_ghost_interp);
+            // We don't want to fill ghosts here - that will be done in another way
+            d_patch_math_ops.interp(dst_data, src_data, patch, false);
```
boyceg, line 2042: I don't know what it means to fill ghost cells in a patch operator. This probably needs to be changed in `PatchMathOps`.

boyceg: It looks like this flag extends the averaging region into the ghost cell region. This is confusing because it is not how most other patch or hierarchy math operators are set up. It looks like the variable-coefficient solver relies on this.

I can see the argument for evaluating on a larger box than the patch box --- it requires a larger volume of communication but only one ghost filling step. Perhaps there could be a configuration flag in the constructors for both `HierarchyMathOps` and `PatchMathOps` that control whether operations are evaluated on the patch box, or in the largest box for which ghost cells are available.

boyceg: #1524 provides a possible way to go here.

drwells:
I fixed most of the failing tests - this works for IBFE but introduces some bugs for multiphase.

@marshallrdavey for the sake of the heart model you should be able to use this branch to get nodal output for everything we would want to plot (including forces).

drwells:
Perhaps not yet - there are still some failing tests.

drwells:
I redid this with the changes discussed in the other PR: i.e., the functions which interpolate into ghost regions have been renamed. All the tests pass.

boyceg (approved):

PR #1524, opened 2022-11-16
HierarchyMathOps and PatchMathOps can use a fill box that is different from the patch box

boyceg:
- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/HierarchyMathOps.h
+++ b/ibtk/include/ibtk/HierarchyMathOps.h
@@ -102,7 +102,8 @@ public:
                      SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy,
                      int coarsest_ln = invalid_level_number,
                      int finest_ln = invalid_level_number,
-                     std::string coarsen_op_name = "CONSERVATIVE_COARSEN");
+                     std::string coarsen_op_name = "CONSERVATIVE_COARSEN",
+                     int dst_fill_box_gcw = 0);
 
     /*!
      * \brief Destructor.
@@ -812,7 +813,6 @@ public:
      */
     void interp(int dst_idx,
                 SAMRAI::tbox::Pointer<SAMRAI::pdat::NodeVariable<NDIM, double> > dst_var,
-                bool dst_ghost_interp,
                 int src_idx,
                 SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > src_var,
                 SAMRAI::tbox::Pointer<HierarchyGhostCellInterpolation> src_ghost_fill,
@@ -874,7 +874,6 @@ public:
      */
     void interp(int dst_idx,
                 SAMRAI::tbox::Pointer<SAMRAI::pdat::EdgeVariable<NDIM, double> > dst_var,
-                bool dst_ghost_interp,
                 int src_idx,
                 SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > src_var,
                 SAMRAI::tbox::Pointer<HierarchyGhostCellInterpolation> src_ghost_fill,
@@ -918,7 +917,6 @@ public:
      */
     void harmonic_interp(int dst_idx,
                          SAMRAI::tbox::Pointer<SAMRAI::pdat::NodeVariable<NDIM, double> > dst_var,
-                         bool dst_ghost_interp,
                          int src_idx,
                          SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > src_var,
                          SAMRAI::tbox::Pointer<HierarchyGhostCellInterpolation> src_ghost_fill,
@@ -939,7 +937,6 @@ public:
      */
     void harmonic_interp(int dst_idx,
                          SAMRAI::tbox::Pointer<SAMRAI::pdat::EdgeVariable<NDIM, double> > dst_var,
-                         bool dst_ghost_interp,
                          int src_idx,
                          SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > src_var,
                          SAMRAI::tbox::Pointer<HierarchyGhostCellInterpolation> src_ghost_fill,
@@ -1492,6 +1489,12 @@ private:
     // Housekeeping.
     std::string d_object_name;
 
+    // Destination fill box ghost cell width.
+    const int d_dst_fill_box_gcw;
+
+    // Patch math operations.
+    PatchMathOps d_patch_math_ops;
+
     // Patch hierarchy information.
     SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > d_hierarchy;
     SAMRAI::tbox::Pointer<SAMRAI::geom::CartesianGridGeometry<NDIM> > d_grid_geom;
@@ -1532,9 +1535,6 @@ private:
     SAMRAI::tbox::Pointer<SAMRAI::math::HierarchyCellDataOpsReal<NDIM, double> > d_hier_cc_data_ops;
     SAMRAI::tbox::Pointer<SAMRAI::math::HierarchyFaceDataOpsReal<NDIM, double> > d_hier_fc_data_ops;
     SAMRAI::tbox::Pointer<SAMRAI::math::HierarchySideDataOpsReal<NDIM, double> > d_hier_sc_data_ops;
-
-    // Patch math operations.
-    PatchMathOps d_patch_math_ops;
 
     // The cell weights are used to compute norms of data defined on the patch
     // hierarchy.
--- a/ibtk/include/ibtk/PatchMathOps.h
+++ b/ibtk/include/ibtk/PatchMathOps.h
@@ -65,7 +65,7 @@ public:
     /*!
      * \brief Default constructor.
      */
-    PatchMathOps() = default;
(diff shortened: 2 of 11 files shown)


boyceg:
This would remove the non-standard API for certain interpolation routines while at the same time making the entire math operations framework more flexible.

drwells:
Now would be a good time to add some more VC tests to make sure we don't break anything inside the library - I'll take a look at that (and this PR) once I'm done with teaching today.

boyceg:
@amneetb, are these changes OK with you in principle?

Another alternative would be to have each operator have an optional argument that specifies the fill box ghost cell width. @drwells, do you think that would be better?

boyceg:
I may make a mockup of that other option today if I get a few spare minutes.

amneetb:
> @amneetb, are these changes OK with you in principle?

I'll take a close look today. We have a quite a few tests for VC solvers. If they are all passing then it is a good sign.

@Raamkrishnan can you also take a look into this PR and see if it requires us to change any phase change code? I hope it is all internal changes,and things on higher level remain as it is.

> 
> 
> 
> Another alternative would be to have each operator have an optional argument that specifies the fill box ghost cell width. @drwells, do you think that would be better?

boyceg:
@amneetb, the only changes required are:
1. `HierarchyMathOps` and `PatchMathOps` take an optional argument in their constructors to set the fill box ghost cell width.
2. Interfaces to the interpolation functions that took a boolean to determined whether the fill box should be grown by 1 have been changed to remove that boolean.

I updated the VC solvers to account for these changes by adding an extra hierarchy math ops object to the parent VC solver class that uses a larger fill box. That math ops object is only used for the interpolation routines that asked for the larger fill box.

boyceg:
One thing we are trying to sort out is that there are many operator implementations should (but do not currently) synchronize values along coarse-fine interfaces in the output variable. We will not get all of the cases any time soon, but we are making progress in, e.g., #1506, #1522, and #1523.

The usual pattern for the API used elsewhere in `HierarchyMathOps` for this is to have a boolean after the specification of the output patch data index and variable that indicates if the caller wants to perform this synchronization. This breaks with the cell-to-node interpolation routine since that interface has a boolean to indicate if the fill box should be extended.

This PR standardizes the interface (by removing those booleans) and then generalizes the support for larger fill boxes (by allowing callers to specify the fill box GCW when setting up the HierarchyMathOps / PatchMathOps objects).

boyceg:
> Another alternative would be to have each operator have an optional argument that specifies the fill box ghost cell width. @drwells, do you think that would be better?

I started doing this, but it seems like a much more invasive set of changes that seems to be more error prone.

drwells (approved):
This seems like the right approach, for now - ideally we would let other functions handle ghost updates but it is important to not break the multiphase flow code.

boyceg:
@amneetb, @Raamkrishnan, are these changes OK with you all?

drwells:
There's a lot of duplicated code in the second commit. Is there a straightforward way to generalize the error checks to a single place?

boyceg:
> There's a lot of duplicated code in the second commit. Is there a straightforward way to generalize the error checks to a single place?

Yes, most of the checks are similar for different functions. Is there an easy way to refactor this? If not, I am not sure that I will be able to do it "by hand".

boyceg:
I mean, I can factor out the new test pretty easily, but there will remain a ton of similar/duplicated checks (which are all my fault also).

drwells:
I don't think its possible to automatically update this but I'd prefer to at least not add more duplicated code and instead move new things into a separate checking routine.

boyceg:
> I don't think its possible to automatically update this but I'd prefer to at least not add more duplicated code and instead move new things into a separate checking routine.

Will do.

Raamkrishnan:
> > @amneetb, are these changes OK with you in principle?
> 
> I'll take a close look today. We have a quite a few tests for VC solvers. If they are all passing then it is a good sign.
> 
> @Raamkrishnan can you also take a look into this PR and see if it requires us to change any phase change code? I hope it is all internal changes,and things on higher level remain as it is.
> 
> > Another alternative would be to have each operator have an optional argument that specifies the fill box ghost cell width. @drwells, do you think that would be better?

No. Phase change code does not require any change. Because phase change code does not involve any interpolation from cell to node (or edge). Only INSVCStaggeredHierarchyIntegrator requires a cell to node (or edge) for the viscosity. @Boyce already did those changes in INSVCStaggeredHierarchyIntegrator class and its child integrators.

Raamkrishnan (approved):

ibtk/include/ibtk/HierarchyMathOps.h
```diff
@@ -1492,6 +1489,12 @@ class HierarchyMathOps : public SAMRAI::tbox::DescribedClass
     // Housekeeping.
     std::string d_object_name;
 
+    // Destination fill box ghost cell width.
+    const int d_dst_fill_box_gcw;
```
drwells, line 1493:
```suggestion
    int d_dst_fill_box_gcw;
```
I know we don't permit copying these things but making them `const` is still weird.

ibtk/include/ibtk/PatchMathOps.h
```diff
@@ -65,7 +65,7 @@ class PatchMathOps : public SAMRAI::tbox::DescribedClass
     /*!
      * \brief Default constructor.
      */
-    PatchMathOps() = default;
+    PatchMathOps(const int dst_fill_box_gcw = 0) : d_dst_fill_box_gcw(dst_fill_box_gcw){};
```
drwells, line 68: Could we document this ctor and move the definition to the source file?

ibtk/include/ibtk/HierarchyMathOps.h
```diff
@@ -102,7 +102,8 @@ class HierarchyMathOps : public SAMRAI::tbox::DescribedClass
                      SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy,
                      int coarsest_ln = invalid_level_number,
                      int finest_ln = invalid_level_number,
-                     std::string coarsen_op_name = "CONSERVATIVE_COARSEN");
+                     std::string coarsen_op_name = "CONSERVATIVE_COARSEN",
+                     int dst_fill_box_gcw = 0);
```
drwells, line 106: We should document what this parameter means.

ibtk/include/ibtk/PatchMathOps.h
```diff
@@ -764,6 +760,9 @@ class PatchMathOps : public SAMRAI::tbox::DescribedClass
      * \return A reference to this object.
      */
     PatchMathOps& operator=(const PatchMathOps& that) = delete;
+
+    /// Ghost cell width to extend the fill box.
+    const int d_dst_fill_box_gcw = 0;
```
drwells, line 765:
```suggestion
    int d_dst_fill_box_gcw = 0;
```

drwells (commented):
A few minor comments.

Do you have time to work on this this week? If not I can take it over.

boyceg:
I will try to finish this up. I got close to factoring out almost all of the error checks in `PatchMathOps` before Thanksgiving!

drwells:
@boyceg do you have time to clean this up or should I do it? Some of our projects (Jordan's and Marshall's) need nodal pressure, and we need this before we do that.

drwells:
I want to wrap this up soon so we can get the nodal pressure branch merged in.

One issue with this approach is that it computes values in ghost regions in a different manner than the rest of the library. Here, we coarsen/refine and then apply a stencil, whereas everywhere else in the library we apply a stencil and then coarsen/refine. Is there a mathematical reason to do it this way? If not we should stick to the standard approach to ghost value updates (i.e., the cell that owns those values is responsible for computing things). Writing into ghost regions is usually a bad idea.

boyceg:
@drwells, the alternative would be to change the way that the harmonic interpolation operator works (to not take in a boolean to determine if to compute over the ghost box). That would be OK with me. In fact, if it is not too much work, it seems preferable.

drwells:
I read through the code and noticed that the other harmonic interpolation is
https://github.com/IBAMR/IBAMR/blob/b7ae5f1a8accc41b662fac57223355406a6b9f7a/ibtk/src/math/HierarchyMathOps.cpp#L2192-L2198
i.e., code does a `dst_cf_bdry_synch` - i.e., that boolean is something completely different based on the type of variable. We should make these functions consistent and I am in favor of not interpolating into ghost regions.

boyceg:
Sounds good to me. Should we scrap this PR?

drwells:
I spent a few more hours digging through this and it looks like we only use this functionality in an intermediate step to interpolate viscosity to either nodal or edge values. The correct fix here is to completely eliminate this temporary variable and just pass along the original one to the variable-coefficient Laplace solver (which can then do the interpolation on-the-fly while differencing). That would resolve questions about 'coarsen/refine then interpolate' vs. the other way around.

I don't want to rewrite the VC Laplace solver. It already works. Hence - lets rename the existing functions which interpolate in the ghost regions, add some caveats to the documentation, and then add back in functions with 'sync_cf_bdry` as the boolean argument. That way everything still works and we can restore a consistent interface to this function.

boyceg:
I generally agree. However, before we jettison this, one concrete reason to add this flexibility would be to provide matrix-free support for some overlapping smoothers / preconditioners.

drwells:
That could be useful - do we have a writeup on what we might want to accomplish there? We can always update this branch later.

boyceg:
@drwells I think this one should go away?

PR #1525, opened 2022-11-18
add guard to target point logging to skip non-target points

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBStandardForceGen.cpp
+++ b/src/IB/IBStandardForceGen.cpp
@@ -1248,7 +1248,7 @@ IBStandardForceGen::computeLagrangianTargetPointForce(Pointer<LData> F_data,
 #if (NDIM == 3)
             F_node[idx + 2] += K * (X_target[2] - X_node[idx + 2]) - E * U_node[idx + 2];
 #endif
-            if (d_log_target_point_displacements)
+            if (d_log_target_point_displacements && (K > 0.0 || E > 0.0))
             {
                 dX = 0.0;
                 dX += (X_target[0] - X_node[idx + 0]) * (X_target[0] - X_node[idx + 0]);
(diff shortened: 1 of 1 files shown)


src/IB/IBStandardForceGen.cpp
```diff
@@ -1248,7 +1248,7 @@ IBStandardForceGen::computeLagrangianTargetPointForce(Pointer<LData> F_data,
 #if (NDIM == 3)
             F_node[idx + 2] += K * (X_target[2] - X_node[idx + 2]) - E * U_node[idx + 2];
 #endif
-            if (d_log_target_point_displacements)
+            if (d_log_target_point_displacements && (K > 0.0 || E > 0.0))
```
drwells, line 1251: I don't know this part of IBAMR that well but what's going on here? Are these not considered target points if they don't have springs attached?

boyceg:
> 
> In src/IB/IBStandardForceGen.cpp:
> 
> > @@ -1248,7 +1248,7 @@ IBStandardForceGen::computeLagrangianTargetPointForce(Pointer<LData> F_data,
>  #if (NDIM == 3)
>              F_node[idx + 2] += K * (X_target[2] - X_node[idx + 2]) - E * U_node[idx + 2];
>  #endif
> -            if (d_log_target_point_displacements)
> +            if (d_log_target_point_displacements && (K > 0.0 || E > 0.0))
> I don't know this part of IBAMR that well but what's going on here? Are these not considered target points if they don't have springs attached?
> 
If those are both zero, then the point is not targeted to anything.

boyceg:
Let's just close this for now --- there was a reason that I did this, but I cannot remember the particular case that motivated.

PR #1533, opened 2022-12-27
0.12 patches

drwells:
I collated the remaining necessary changes (and two minor build system fixes) into one set of patches. Once this is in we can create the release.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


ibtk/src/lagrangian/LEInteractor.cpp
```diff
@@ -1573,8 +1573,7 @@ string_to_kernel(const std::string& kernel_fcn)
 double (*LEInteractor::s_kernel_fcn)(double r) = &ib4_kernel_fcn;
 int LEInteractor::s_kernel_fcn_stencil_size = 4;
 
-void
-LEInteractor::setFromDatabase(Pointer<Database> /*db*/)
+void LEInteractor::setFromDatabase(Pointer<Database> /*db*/)
```
abarret, line 1576: Don't we usually put return types on their own line? What did clang-format do here?

src/IB/IMPInitializer.cpp
```diff
@@ -616,8 +616,7 @@ IMPInitializer::getVertexPosn(const std::pair<int, int>& point_index, const int
     return d_vertex_posn[level_number][point_index.first][point_index.second];
 } // getVertexPosn
 
-void
-IMPInitializer::getFromInput(Pointer<Database> /*db*/)
+void IMPInitializer::getFromInput(Pointer<Database> /*db*/)
```
abarret, line 619: Same here

abarret (approved):

ibtk/src/lagrangian/LEInteractor.cpp
drwells: We do. I had to check and the current `.clang-format` file is set up so that is true for all returned objects - however it is not for the 'typeless type' of void. In other files `void` is not on its own line.

PR #1536, opened 2023-01-09
FEMechanicsBase: fix the component coupling.

drwells:
We should set these values immediately before 'assemble'. I don't think this will make a significant difference since we do MAT_IGNORE_ZERO_ENTRIES elsewhere but not allocating them at all is an improvement.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/FEMechanicsBase.cpp
+++ b/src/IB/FEMechanicsBase.cpp
@@ -525,7 +525,6 @@ FEMechanicsBase::doInitializeFEEquationSystems()
             {
                 X_system.add_variable("X_" + std::to_string(d), d_fe_order_position[part], d_fe_family_position[part]);
             }
-            X_system.get_dof_map()._dof_coupling = &d_diagonal_system_coupling;
 
             auto& dX_system = equation_systems.add_system<ExplicitSystem>(getDisplacementSystemName());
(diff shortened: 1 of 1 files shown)


src/IB/FEMechanicsBase.cpp
```diff
@@ -525,29 +525,25 @@ FEMechanicsBase::doInitializeFEEquationSystems()
...
                 U_system.add_variable("U_" + std::to_string(d), d_fe_order_position[part], d_fe_family_position[part]);
             }
-            U_system.get_dof_map()._dof_coupling = &d_diagonal_system_coupling;
 
             auto& F_system = equation_systems.add_system<ExplicitSystem>(getForceSystemName());
             for (unsigned int d = 0; d < NDIM; ++d)
             {
                 F_system.add_variable("F_" + std::to_string(d), d_fe_order_force[part], d_fe_family_force[part]);
             }
-            F_system.get_dof_map()._dof_coupling = &d_diagonal_system_coupling;
```
drwells, line 550: These don't get set if we are using a restart file.

boyceg (approved):

Issue #1535, opened 2023-01-09
double check that libmesh restart files have unique names


drwells:
We have the following code right now (Jordan is about to put up a PR to fix it) in FEMechanicsBase:

```cpp
std::string
FEMechanicsBase::libmesh_restart_file_name(const std::string& restart_dump_dirname,
                                           unsigned int time_step_number,
                                           unsigned int part,
                                           const std::string& extension)
{
    std::ostringstream file_name_prefix;
    file_name_prefix << restart_dump_dirname << "/libmesh_data_" << d_object_name << "_part_" << part << "." << std::setw(6)
                     << std::setfill('0') << std::right << time_step_number << "." << extension;
    return file_name_prefix.str();
}
```
There's a similar thing going on in IIMethod. This is not going to work with `IBStrategySet` etc. since the file name is not dependent on the object name.

Lets double-check that all of these save to unique restart files and increment the restart file format number accordingly.

abarret:
I talked with @jabrown893 about this. You can have multiple libMesh based objects (`IIMethod`, `IBFEMethod`, etc.) write restart files to _different folders_ by modifying constructor arguments and restart writing function arguments, but the folder must already exist on the filesystem. SAMRAI ensures that the restart folder that SAMRAI is using exists. I don't think that's true with libMesh.

One fix is to make sure the folder exists in the `writeFEDataToRestartFile` before trying to write to it.

If we want to write multiple types of restart files to the _same folder_, then we have to do something like above. We could also add an argument to `libmesh_restart_file_name` that replaces `"libmesh_data_"`.

jabrown893:
It seems like the options are between giving the user tools to put the restart files in different directories (easiest to implement with what's currently there? -- puts the responsibility/knowledge requirement on the reader), giving the user tools to give the restart files different names (also puts the responsibility/knowledge requirement on the reader), or us giving the restart files different names on behalf of the user (which probably makes the user experience easier long-term but will increment the restart file format).

jabrown893:
(In any case, we should probably set up the FE restart data writer to create the directory it needs, since we allow the user to specify the directory -- especially since this is how the SAMRAI part works?)

drwells:
We do have `mkDirRecursive` or some other similarly named utility function to make directories - that can be done automatically. However - we specify restart directories on the command line. It is inconsistent to have multiple restart directories.

[...]

PR #1537, opened 2023-01-10
Use class-specific names for libMesh restart files.

drwells:
This fixes the "FEMechanicsExplicitIntegrator and IBFEMethod try to write the same restart file" bug in 99% of cases. We should still add some extra error checking and perhaps optional file prefixes.

Part of #1535. The downside to this approach is we need to increment the restart file number, which we haven't in awhile.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/include/ibamr/FEMechanicsBase.h
+++ b/include/ibamr/FEMechanicsBase.h
@@ -543,10 +543,10 @@ protected:
     /*!
      * Get the libMesh restart file name.
      */
-    static std::string libmesh_restart_file_name(const std::string& restart_dump_dirname,
-                                                 unsigned int time_step_number,
-                                                 unsigned int part,
-                                                 const std::string& extension);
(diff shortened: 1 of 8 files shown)


boyceg:
@drwells, could we instead just have a filename prefix variable that gets different default values for the different solvers (but could be overridden for cases with multiple instantiations of the same solver class)?

drwells:
Yes - we would just need to add a database argument for that.

boyceg:
@drwells, I know that you want to move away from `IBFEMethod`, but since folks are still using it, should we fix this?

drwells:
Do you want to go ahead and add database arguments?

drwells:
Done!

It looks like I forgot to add a test in 2023 - I added one which uses IBStrategySet and confirmed that we create independent XDR files.

boyceg (approved):
Looks good to me.

PR #1538, opened 2023-01-10
Fix restarts with the DLM example.

drwells:
This, combined with the class-specific restart file names patch, fixes this example.


boyceg (approved):
If we are going to merge in #1537, then we might as well merge in this one as well.

PR #1539, opened 2023-01-10
CI: upgrade to cache@v3.

drwells:
cache@v2 uses some deprecated stuff so we need to upgrade.

Lets see if the rest of our CI still works with this.


drwells:
Some of the caches didn't work but I think that's due to something else - this should be good to go.

PR #1547, opened 2023-02-14
Cmake fixes + bump silo version

drwells:
Two small build system fixes:
1. Fix some issues with boost that only occur when the user installs boost on their own in a non-system directory
2. Enforce using SILO 4.11 or newer. I saw someone use SILO 4.10 last week, which is not compatible with IBAMR - it is not const-correct so we get compilation errors.


drwells:
Looks like the containers failed correctly 😅 - I'll update those and redo the check.

boyceg (approved):

drwells:
The only reason we didn't merge this already is that it conflicted with the oldest CI runner - now that that's retired we should just merge it.

PR #1549, opened 2023-02-16
CI: Update the first Fedora image.

drwells:
We need to update the CI to not use Silo 4.10 to get the other PR to work. This PR updates the first CI image (the basic Fedora one we use for half of our configurations). The new image is here:

https://hub.docker.com/r/wellsd2/ibamr/tags

Additional subchanges:
1. Don't hard-code fedora33 everywhere (we now use Fedora 36)
2. Get rid of sccache (ccache is the standard)


boyceg (approved):

drwells:
I need to finish this - it isn't hard but it is somewhat tedious to squelch all of the samrai warnings.

ccache is the standard tool for compilation caching - all the major package managers provide it. sccache is a specialized tool that KitWare uses which is less available (e.g., it isn't provided by Fedora) - it does approximately the same thing but can also work over a network. Since we don't store a compilation cache on a network (in the way sccache expects) we should just use the standard tool.

drwells:
Another thing we should do here is make sure we are consistently using `ld.gold`.

drwells:
Superceded by #1693.

PR #1555, opened 2023-03-09
Adding a simple example to fix the regridding issues

amneetb:
- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


amneetb:
@boyceg : Please check examples/adv_diff/ex8 for a simple test case. There are changes in `AdvDiffSemiImplicitHierarchyIntegrator` and `HierarchyIntegrator` that are needed for this example to work.

The example is solving 

$\frac{\partial Q}{\partial t} = f(x,t)$

where $Q = \frac{1}{\sigma\sqrt{2\pi}} \  \exp(-(x-x_0)/2\sigma^2) \cos(\omega t)$ and $f(x,t)$ is manufactured from it. If $Q > 0.1$ tagging happens. In this case $Q$ goes above and below the 0.1 value, and therefore, refining and coarsening is expected.

amneetb:
@boyceg  and @drwells: I was able to do this fixes without making any change to `HierarchyIntegrator.cpp`, which is a good thing. Please take a look. I just forced push the changes to this PR.

src/adv_diff/AdvDiffSemiImplicitHierarchyIntegrator.cpp
```diff
@@ -908,23 +908,37 @@ AdvDiffSemiImplicitHierarchyIntegrator::postprocessIntegrateHierarchy(const doub
 /////////////////////////////// PROTECTED ////////////////////////////////////
 
 void
-AdvDiffSemiImplicitHierarchyIntegrator::resetHierarchyConfigurationSpecialized(
-    const Pointer<BasePatchHierarchy<NDIM> > base_hierarchy,
-    const int coarsest_level,
-    const int finest_level)
+AdvDiffSemiImplicitHierarchyIntegrator::regridHierarchyBeginSpecialized()
```
boyceg, line 911: Can you remove this specialization since it only calls the parent class implementation?

src/adv_diff/AdvDiffSemiImplicitHierarchyIntegrator.cpp
```diff
@@ -908,23 +908,37 @@ AdvDiffSemiImplicitHierarchyIntegrator::postprocessIntegrateHierarchy(const doub
...
     for (const auto& Q_var : d_Q_var)
     {
         d_Q_convective_op_needs_init[Q_var] = true;
     }
+    AdvDiffHierarchyIntegrator::regridHierarchyBeginSpecialized();
+    return;
+} // regridHierarchyBeginSpecialized
+
+void
+AdvDiffSemiImplicitHierarchyIntegrator::resetHierarchyConfigurationSpecialized(
```
boyceg, line 922: Same as above: Can you remove this specialization since it only calls the parent class implementation?

src/adv_diff/AdvDiffSemiImplicitHierarchyIntegrator.cpp
boyceg: Whoops, I think I misread this diff --- it is doing more than calling the parent class implementation.

src/adv_diff/AdvDiffSemiImplicitHierarchyIntegrator.cpp
boyceg: This one looks like it could go away though?

boyceg:
@drwells, do you know what is going on with the Cmake CI builds?

drwells:
I see
```
CMake Error: File /__w/IBAMR/IBAMR/examples/adv_diff/ex8/input3d does not exist.
```
looks like we are missing an input file, or an input file is listed which doesn't exist.

drwells:
Overall this looks good. Could you convert the example into a test? Since we aren't demonstrating any new physics I don't think we need to add a new example.

drwells:
We also need a changelog entry!

amneetb:
@boyceg and @drwells : We also need to fix similar issues in INS classes. Should that be a different PR? I was thinking doing it in here.

drwells:
I don't think it matters much how we split it up. I'm happy to merge this PR by itself once we convert the example into a test and get the build working.

amneetb:
@drwells : Converted the example to a test. If this looks good to you, please merge with master. I will rebase the phase change branch on top of master, and then test it with PhaseChangeHierarchyIntegrator and also INS solvers. We have a test problem with INS solvers in the phase change branch that will help to further smoke out regrinding issues with INS integrators.

src/adv_diff/AdvDiffSemiImplicitHierarchyIntegrator.cpp
Raamkrishnan: > Same as above: Can you remove this specialization since it only calls the parent class implementation?

Done 👍

Raamkrishnan:
@drwells If everything looks good, can you please merge this PR? My phase change code needs this change.

drwells:
This looks good. Two small things:
1. Can you take a look at the failing CI? It looks like there are still some warnings being printed.
2. Can you add the test to the CMake build system?

Raamkrishnan:
> > This looks good. Two small things:
> > 
> > 1. Can you take a look at the failing CI? It looks like there are still some warnings being printed.
> > 2. Can you add the test to the CMake build system?
> 
> 👍



> This looks good. Two small things:
> 
> 1. Can you take a look at the failing CI? It looks like there are still some warnings being printed.
> 2. Can you add the test to the CMake build system?

The CI is happy now. I have also added the test to the CMake build system.

drwells (approved):

PR #1556, opened 2023-03-13
Discontinuous (in local coordinates) FE families for IIM quantities

AminKolah:
###  A changelog to clarify some of the added functionalities: 
- [ ] Do all the combinations of pressure jump, viscous jump, traction representations with MONOMIAL FIRST and CONSTANT or L2_LAGRANGE FIRST work as expected?
- [ ] Do all rigid and flexible examples using discontinuous representation of jumps/traction quantities work as expected? Rigid/stationary examples include ex1 flow past a square block and ex3 flow past a thin swimming airfoil. Flexible examples include the thin beam in the Turek-Hron benchmark (ex6) and the flexible plate inside the flow phantom (ex9).
- [ ] These functionalities need to be also separately tested for an internal flow simulation with a step (similar to backward facing step in 2D or the stationary housing in the pulse duplicator experiment in 3D. We can add the 2D case of the backward facing step to our example list if necessary).

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


AminKolah:
It seems like it's a non-IIM related test that is generating an error?

boyceg:
@AminKolah, it looks like you changed the order in which certain operations are performed, which results in changed output in some tests.

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -742,6 +741,7 @@ FEProjector::computeL2Projection(PetscVector<double>& U_vec,
         if (qrule_is_nodal(fe_type, qrule.get()))
         {
             // use the lumped matrix as the preconditioner:
+            PetscMatrix<double>& lumped_mass = *buildLumpedL2ProjectionSolver(system_name).second;
```
boyceg, line 744: @AminKolah, I think this is the change that is causing tests to fail.

AminKolah: The change that is made here is basically saying that only build the lumped L2 projection solver if the qrule is nodal. Is that not correct? There is a `qrule_is_nodal`  assertion in `buildLumpedL2ProjectionSolver` itself that triggers an error if qrule is not nodal.

drwells: I suspect the reason is more banal - we rely on log files to verify that results haven't changed, so moving this down changes the order in which lines are printed (and causes a test failure). Is there a reason this was moved?

AminKolah: So currently the 3 choices with USE_DISCON_ELEMS = TRUE are 1) CONSTANT MONOMIAL, 2) FIRST order L2_LAGRANGE and 3) FIRST order MONOMIAL. That line was moved down inside the if condition because currently the 3rd option (FIRST order MONOMIAL) fails the assertion for `qrule_is_nodal`. Practically speaking, we may never even need to use FIRST order MONOMIAL because the second option (FIRST order L2_LAGRANGE) would give the same result. So I can put that line back if we all agree there is no need to use FIRST order MONOMIAL for the jump discontinuous systems.

AminKolah: To be more precise, this is the line that causes an error when the first order monomial is used. @drwells is it necessary to keep this check?
https://github.com/IBAMR/IBAMR/blob/833bf42f40123190d5fe6afe62f5b26f1f9117b2/ibtk/src/lagrangian/FEProjector.cpp#L127

boyceg:
@AminKolah you don't need to revert the change to `FEProjector` if the change is correct. 😁 We just will need to update the test output.

AminKolah:
> @AminKolah you don't need to revert the change to `FEProjector` if the change is correct. grin We just will need to update the test output.

It's already looking like I'll be squashing some commits later, but do you think it's necessary to test FIRST order MONOMIAL in addition to L2_LAGRANGE?  So long as we stick to either constant MONOMIAL or L2_LAGRANGE things are working as expected. In any case, I'm still unclear why is necessary for the MONOMIAL to only be constant order if nodal qrule is used? That check was already in the library so I didn't touch that part.

boyceg:
I think it is OK to specialize to `L2_LAGRANGE` elements in cases that they are available.

AminKolah:
Okay I further clarified in the input file on what should be used when changing the FE family for the jumps/traction:

https://github.com/IBAMR/IBAMR/blob/af298486fe0827d5698a1579eb9d972c2b05a0a2/examples/IIM/ex1/input2d#L53-L55

boyceg:
If there are restrictions on the allowable choices of `FEFamily` and `Order`, we should include a check in the code that reads in the options, and/or in the code that sets up the FE fields.

ibtk/src/lagrangian/FEProjector.cpp
drwells: If I recall correctly, the monomial space is `{1, x, y}` etc: past using just constants, there is no corresponding nodal quadrature rule, so the check is correct. What do you expect to happen here? I think `L2_Lagrange` is going to be the right choice.

examples/IIM/ex1/input2d
```diff
@@ -121,10 +121,15 @@ IIMethod {
...
-   tau_fe_family = "L2_LAGRANGE"
+  velocity_jump_fe_family = "L2_LAGRANGE" //"MONOMIAL" 
+    pressure_jump_fe_family = "L2_LAGRANGE" //"MONOMIAL" 
+	  wss_fe_family = "L2_LAGRANGE" //"MONOMIAL"
+	tau_fe_family = "L2_LAGRANGE" //"MONOMIAL"
+   
+   	pressure_jump_fe_order 		= "FIRST" //"CONSTANT" 
+	velocity_jump_fe_order 		= "FIRST" //"CONSTANT" 
+	wss_fe_order 				= "FIRST" //"CONSTANT" 
+	tau_fe_order 				= "FIRST" //"CONSTANT"
```
drwells, line 132: I think your editor is set up with a different tab width than the rest of this file - could you just use spaces (and indent with three spaces) here?

src/IB/IIMethod.cpp
```diff
@@ -1003,27 +1053,31 @@ IIMethod::interpolateVelocity(const int u_data_idx,
...
+							for (unsigned int d = 0; d < NDIM; ++d)
+							{
+								for (unsigned int k = 0; k < n_nodes_jump; ++k)
+								{
+									const double& p2 = phi_DU_jump[k][qp];
+									DU_jump_qp[axis][NDIM * (qp_offset + qp) + d] += DU_jump_node[axis][k][d] * p2;
+								}
+							}
+						}
+					}
```
drwells, line 1080: The same is true here - use four spaces or just let clang-format handle it.

drwells:
Overall this looks good - can you add a changelog entry?

src/IB/IIMethod.cpp
```diff
@@ -762,6 +810,7 @@ IIMethod::interpolateVelocity(const int u_data_idx,
             U_t_vec = d_U_t_new_vecs[part];
             X_vec = d_X_new_vecs[part];
         }
+        
         copy_and_synch(*X_vec, *X_ghost_vec);
```
drwells, line 814: We already do the time-based selection of the correct vectors here. Could you use these in `batch_vec_ghost_update()` instead?

If vectors are used with patch data then we need to ensure that they are IB-ghosted: normally `FEDataManager` handles that but its all done manually here. Which of these need the IB ghosting? I'm surprised that only the position and velocity jump are explicitly labeled as such.

src/IB/IIMethod.cpp
AminKolah: Seems like the address given in this file needs to be modified (the file must have moved from the original directory?) https://github.com/IBAMR/IBAMR/blob/af298486fe0827d5698a1579eb9d972c2b05a0a2/scripts/formatting/setup_clang_format_precommit_hook.sh#L15

examples/IIM/ex1/input2d
```diff
@@ -121,10 +121,15 @@ IIMethod {
    p_calc_width				   = P_CALC_WIDTH
    wss_calc_width			  = WSS_CALC_WIDTH
    min_ghost_cell_width = 6
-   pressure_jump_fe_family = "L2_LAGRANGE"
-   velocity_jump_fe_family = "L2_LAGRANGE"
-   wss_fe_family = "L2_LAGRANGE"
-   tau_fe_family = "L2_LAGRANGE"
+  velocity_jump_fe_family = "L2_LAGRANGE" //"MONOMIAL"
```
boyceg, line 124: @AminKolah is the intention for the comments to suggest other options? If so, I'd suggest making this a bit more explicit. Otherwise, I'd suggest removing the comments.

Also, would you mind aligning the option settings in the input file?

examples/IIM/ex1/input3d
```diff
@@ -156,10 +156,15 @@ IIMethod {
    p_calc_width				   = P_CALC_WIDTH
    wss_calc_width			  = WSS_CALC_WIDTH
    min_ghost_cell_width = 6
-   pressure_jump_fe_family = "L2_LAGRANGE"
-   velocity_jump_fe_family = "L2_LAGRANGE"
-   wss_fe_family = "L2_LAGRANGE"
-   tau_fe_family = "L2_LAGRANGE"
+  velocity_jump_fe_family = "MONOMIAL"
```
boyceg, line 159: Same as above --- would you mind aligning these settings?

boyceg: (It looks like you might be using tabs instead of spaces.)

src/IB/IIMethod.cpp
```diff
@@ -239,6 +239,16 @@ IIMethod::registerDisconElemFamilyForJumps(const unsigned int part)
 {
     TBOX_ASSERT(!d_fe_equation_systems_initialized);
     TBOX_ASSERT(part < d_num_parts);
+    if (d_velocity_jump_fe_family == LAGRANGE || d_pressure_jump_fe_family == LAGRANGE
```
boyceg, line 242: If you need to set this in the input file, what is the job of the function `registerDisconElemFamilyForJumps`?

src/IB/IIMethod.cpp
```diff
@@ -1732,14 +1788,17 @@ IIMethod::computeFluidTraction(const double data_time, unsigned int part)
                     {
                         const double& p_X = phi_X[k][qp];
                         x_qp[NDIM * (qp_offset + qp) + i] += x_node[k][i] * p_X;
+					}
```
boyceg, line 1791: This is another place where it looks like there are tabs instead of spaces.

src/IB/IIMethod.cpp
```diff
@@ -4281,7 +4355,13 @@ IIMethod::getFromInput(Pointer<Database> db, bool /*is_from_restart*/)
     if (d_use_pressure_jump_conditions)
     {
         if (db->isString("pressure_jump_fe_family"))
+        {
             d_pressure_jump_fe_family = Utility::string_to_enum<FEFamily>(db->getString("pressure_jump_fe_family"));
+		}
+		if (db->isString("pressure_jump_fe_order"))
```
boyceg, line 4361: As above --- looks like this is using tabs.

boyceg:
@AminKolah are you able to finish this up so it can be merged?

ibtk/src/lagrangian/FEProjector.cpp
```diff
@@ -124,7 +124,7 @@ qrule_is_nodal(const FEType& fe_type, const QBase* const qrule)
     if (fe_type.family == LAGRANGE || fe_type.family == L2_LAGRANGE || fe_type.family == MONOMIAL)
     {
         nodal_cond = true;
-        if (fe_type.family == MONOMIAL && fe_order != CONSTANT) return false;
+        if (fe_type.family == MONOMIAL && !(fe_order == CONSTANT || fe_order == FIRST)) return false;
```
drwells, line 127: I thought the first-degree monomials were `1, x, y, z`: that doesn't correspond to a nodal quadrature rule. What's the explanation for this change?

src/IB/IIMethod.cpp
```diff
@@ -235,13 +235,84 @@ IIMethod::getFEDataManager(const unsigned int part) const
...
+IIMethod::registerDisconElemFamilyForTraction(const unsigned int part, 
+                                          libMesh::FEFamily fe_family,
+                                          libMesh::Order fe_order)
 {
     TBOX_ASSERT(!d_fe_equation_systems_initialized);
     TBOX_ASSERT(part < d_num_parts);
-    d_use_discon_elem_for_jumps[part] = true;
+    // The acceptable options are CONSTANT MONOMIAL,  FIRST order MONOMIAL, or FIRST order L2_LAGRANGE.
+    // NOTE: For now if separate orders are used for viscous vs. pressure jumps, the traction order 
+    // need to be the the lowest of the two.
```
drwells, line 246: Can you move this documentation to the header file?

src/IB/IIMethod.cpp
```diff
@@ -235,13 +235,84 @@ IIMethod::getFEDataManager(const unsigned int part) const
...
+    if ((fe_family == L2_LAGRANGE && fe_order == FIRST) || (fe_family == MONOMIAL && fe_order == CONSTANT)
+          || (fe_family == MONOMIAL && fe_order == FIRST))
+    {
+		//if (fe_order == FIRST && (d_viscous_jump_fe_order[part] == CONSTANT || d_pressure_jump_fe_order[part] == CONSTANT))
```
drwells, line 250:
```suggestion
```

src/IB/IIMethod.cpp
```diff
@@ -2839,7 +2965,7 @@ IIMethod::spreadForce(const int f_data_idx,
                 DU_jump_vec[d]->localize(*DU_jump_ghost_vec[d]);
             }
         }
-
+
```
drwells, line 2968:
```suggestion

```

drwells (changes requested):
We need some tests here too but overall this looks fine.

AminKolah:
> We need some tests here too but overall this looks fine.

I agree. Should we generate a simple test based on ex1? A few of other existing examples also could use this option but I think flow past a square block /cube (ex1) could be a good choice as a test.

ibtk/src/lagrangian/FEProjector.cpp
AminKolah: This check was throwing a `TBOX_ERROR` when building the L2 projection matrix. I guess part of my confusion was that if you choose `FIRST` order L`2_LAGRANGE` then this check returned true for nodal quadrature rule, but `FIRST `order `MONOMIAL` was returning false. Is that what expect? Maybe I'm not understanding why the two situations would be different in terms of using nodal quadrature use.

src/IB/IIMethod.cpp
AminKolah: Done.

src/IB/IIMethod.cpp
AminKolah: oops, sorry seems like I already cleaned this up in my new commit.

drwells:
I think pretty much anything that uses the new code is fine - the more discontinuous things we turn on the better.

AminKolah:
Yes, good idea. We can add that to the current flow past a cylinder and sphere tests. Though we wouldn't normally want to use these for a smooth surface, it won't hurt to add the flags. Should I generate separate input file that uses the same application code for a test?

AminKolah:
I also created a changelog. I'll add more items if I think of any.

drwells:
@AminKolah we're wrapping up the IBAMR release this week and I'd like to get this in - can you add the tests today or tomorrow?

drwells:
It looks like you still need to update the build system to compile the tests.

Can you add the changelog?

AminKolah:
> It looks like you still need to update the build system to compile the tests.
> 
> Can you add the changelog?

To clarify a bit more on this: I created two new input files, namely  `flow_past_cylinder_DG.mpirun=2.input` and  `flow_past_sphere_DG.mpirun=4.input `(and their corresponding output files) for the already existing tests of `flow_past_cylinder.cpp` and `flow_past_sphere.cpp` respectively. The new generated input files can be run along with the previous input files (`flow_past_cylinder.mpirun=2.input` and `flow_past_sphere.mpirun=4.input`) for the same examples. Do I still need to update the build systems given what I just mentioned? If yes, would you mind providing a bit more instructions? I don't see anywhere in the build system files where the name of input files are specifically mentioned?

AminKolah:
For the changelog I added the added capabilities at the top in this PR. Should I also add a file to the folder at https://github.com/IBAMR/IBAMR/tree/master/doc/news/changes/minor ?

drwells:
I misread the changelog - the build system is fine but we need to rename those input files to associate them with the existing tests.

`flow_past_sphere_DG.mpirun=4.input` is going to be paired with `flow_past_sphere_DG` which does not exist - the way input/output files are associated with tests is that we look for an executable with the same name up to the first `.`. Hence this should work if you rename the input and output files as `flow_past_sphere.dg.mpirun=4.input` etc. You can always run attest yourself to verify that things get picked up correctly.

drwells:
Yes, the summary at the top of this page looks good, but I'd call this a major change so put it in the other folder please.

AminKolah:
Okay I was just in the process of changing _DG to .DG :) ...That sounds good. I'll make the other change now too. Thanks!

drwells (approved):

Issue #1554, opened 2023-03-09
Extract the 'compute pressure at the centroid' meter mesh code


drwells:
This is something @jabrown893 needs which wouldn't be hard to do. Some applications want something simpler than IBFEInstrumentPanel - they just need to sample the Eulerian fields at known points. We do exactly this for computing the pressure at the meter mesh centroid.

Lets write a new function which samples the background grid at some given point. We can basically just copy-and-paste lines 400-475 of IBFEInstrumentPanel to do this.

boyceg:
It would be useful to be able to treat AMR fields as continuous functions that can be evaluated at “arbitrary" positions. Is that what you have in mind?

drwells:
Yup! I think we can write a function
```cpp
double
interpolate_scalar(const int data_idx, const PatchHierarchy<NDIM> patch_hierarchy, const Point p);
```
That's basically what we use in `IBFEInstrumentPanel` to compute the pressure at the mesh centroid.

abarret:
There's some code that does something like this here
https://github.com/IBAMR/IBAMR/blob/master/examples/complex_fluids/ex2/InterpolationUtilities.h

drwells:
Nice - we should move that into the library itself. We can also make the kernel a parameter.

abarret:
I was assigned this three years ago, #776. I guess it's time to actually do it.

drwells:
I forgot about that one.

I see a few things that `IBFEInstrumentPanel` does that these functions don't:
1. Assumes ghosts are present
2. Works with multiple patch levels

jabrown893:
That would be great!

Issue #776, opened 2019-12-05
Move InterpolationUtilities into IBTK


drwells:
From a comment in #725: we should move `examples/complex_fluids/ex2/InterpolationUtilities.{h,cpp}` into IBTK.

PR #1560, opened 2023-03-20
Add interpolation utilities.

abarret:
This is WIP to satisfy #1554 and #776. This adds two ways to interpolate a cell centered quantity to an arbitrary point in the computational domain. One uses a bilinear interpolant, and the other uses a least squares fit.

There's some things that still need updating (comments, autotools, etc.), but I want to know if this satisfies what @jabrown893 needs.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


jabrown893:
Did you say this is only for cell-centered quantities? The main thing I need is actually velocity. But I think the plan was first to get set up for pressure, and then adapt/expand for velocity.

ibtk/include/ibtk/interpolation_utilities.h
```diff
@@ -0,0 +1,88 @@
...
+#include <PatchLevel.h>
+
+#include <vector>
+
+namespace IBTK
+{
+namespace Interpolation
+{
+/*
+ * Interpolations the cell centered data stored in patch index data_idx to the point X. Uses bilinear interpolation.
```
boyceg, line 41: Should we use `LEInteractor` under the hood and allow for user-specified interpolation kernels?

drwells: We should definitely use `LEInteractor` - that would also correctly support multiple data centerings.

boyceg:
@drwells, @abarret, what should we do here? It seems like @drwells and I both think that this should be done using `LEInteractor`.

ibtk/include/ibtk/interpolation_utilities.h
```diff
@@ -0,0 +1,62 @@
...
+
+namespace IBTK
+{
+/*
+ * Interpolates the data stored in data_idx to the location X using the provided kernel function.
+ *
+ * Q_var must correspond to Cell, Side, Node, or Edge centered double values, or an unrecoverable error will occur.
+ * Ghost cells must be filled in for the data stored in data_idx prior to this function call.
+ *
+ * The returned values are synchronized across all processors.
```
boyceg, line 38: Does the input position need to be the same across all processors?

ibtk/include/ibtk/interpolation_utilities.h
```diff
@@ -0,0 +1,62 @@
...
+                                SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > patch_hierarchy,
+                                std::string kernel_fcn = "IB_4");
+
+/*
+ * Interpolates the data stored in data_idx to the locations provided in X using the provided kernel function.
+ *
+ * Q_var must correspond to Cell, Side, Node, or Edge centered double values, or an unrecoverable error will occur.
+ * Ghost cells must be filled in for the data stored in data_idx prior to this function call.
+ *
+ * The returned values are synchronized across all processors.
```
boyceg, line 54: Same question as above.

abarret:
> I think that what it does is fine. It might be useful to have a "fast" version of this that only operates on local patches and doesn't require parallel synchronization.

I've updated the documentation to clarify that inputs need to be synced across ranks. The wrapper "fast" version will take more time. If you want this version in the release, then I can set up a new branch. Otherwise, I'll add the wrapped version to this branch later. I don't have a strong opinion either way.

boyceg:
I'll leave it up to @drwells about whether to merge or not. The fast version seems outside of the scope of this PR in any case, however.

boyceg (approved):

PR #1569, opened 2023-03-29
phase change: library and tests

drwells:
First half of #1505 - just changes to the library. Some immediate TODOs:

- [ ] we need to not incompatibly change the Brinkman penalization code (old tests should work without modifications). We can deprecate the old stuff if we need to, but we need to mark it as deprecated before we delete it.
- [ ] Boyce said we can generalize some of the advection code which will vastly simplify the implementation - we need a roadmap for that.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


amneetb:
@drwells Thanks! 

1. I'll take care of registerFluidSourceFunction and registerDivergenceVelocitySourceFunction issue, once I know what the former function is doing. 
2. I'll also mark some functions deprecated in the Brinkman class.
3. Boyce will take care of advective operators. 

Are there any other outstanding issues with the library code of this PR?

include/ibamr/BrinkmanPenalizationStrategy.h
```diff
@@ -18,7 +18,7 @@
 
 /////////////////////////////// INCLUDES /////////////////////////////////////
 
-#include <ibamr/config.h>
+#include "ibtk/ibtk_utilities.h"
```
drwells, line 21: We need to update the CI to check this, but every one of our headers needs to include either `ibtk/config.h` or `ibamr/config.h`. Can you add it back in?

boyceg:
I want to update the basic implementations of the convective operators so that it is no longer necessary to create the specialized `AdvDiffCUIConservativeConvectiveOperator` implementation. I will try to get this done by some time next week.

I also want to think about whether it would make sense to unify the time integrator classes. I would prefer not to have separate time integrator concepts in the library. One thought is if it might make sense to make the time integrator class introduced in this PR the base class for `HierarchyIntegrator`.

boyceg:
@abarret, @amneetb, what do you think about trying to make the time integrator class introduced here the base class for `HierarchyIntegrator`?

boyceg:
We would need to deprecate & rename `HierarchyIntegrator::advanceHierarchy()`, but otherwise it seems like it should be natural to do this?

abarret:
> @abarret, @amneetb, what do you think about trying to make the time integrator class introduced here the base class for `HierarchyIntegrator`?

It is unclear to me what creating another base class class for `HierarchyIntegrator` will do. We already have a few classes that integrate equations that do not inherit from `HierarchyIntegrator`, e.g. `FEMechanicsExplicitIntegrator` and `LSInitStrategy`. What will having an interface to these kinds of objects actually do for us?

`HierarchyIntegrator` is a large class that does a lot of things. It would be nice if we could pair down the number of things that the class does, e.g. separate "integrating equations" from "regridding and managing states". But I don't see a good way of doing that without completely rewriting the class and it's derivatives.

amneetb:
> > @abarret, @amneetb, what do you think about trying to make the time integrator class introduced here the base class for `HierarchyIntegrator`?
> 
> It is unclear to me what creating another base class class for `HierarchyIntegrator` will do. We already have a few classes that integrate equations that do not inherit from `HierarchyIntegrator`, e.g. `FEMechanicsExplicitIntegrator` and `LSInitStrategy`. What will having an interface to these kinds of objects actually do for us?
> 
> `HierarchyIntegrator` is a large class that does a lot of things. It would be nice if we could pair down the number of things that the class does, e.g. separate "integrating equations" from "regridding and managing states". But I don't see a good way of doing that without completely rewriting the class and it's derivatives.

Exactly! It is hard to have a single interface for integrating specialized PDEs. These specialized integrators should not be mixed with HierarchyIntegrator class that does a lot of things. In the case of MassIntegrator, the job is very specific: compute the temporary mass fluxes for variable density solvers. I don't think it has any other use to the rest of the library. 

I was thinking that HierarchyIntegrator should be renamed as HierarchyManager or something like that. It acts more like a manager of data structures and hierarchy than an integrator.

amneetb:
Another thought is we can have `HierarchyManager` and `TimeIntegrator` as base classes. `HierarchyIntegrator` is derived from `HierarchyManager` and `TimeIntegrator`. The idea is that `TimeIntegrator` is a very lightweight class that has just two or three virtual functions, and a very few data structures. But again this is just an added layering. It does not do anything good.

boyceg:
> But again this is just an added layering. It does not do anything good.

I disagree that it would not do anything good. I think that it is good to have as small a number of concepts in the API as possible, and to avoid duplicating interfaces.

amneetb:
It's fine with me. We can adapt to the new API. I would suggest to keep the `TimeIntegrator` API to be as light as possible.

abarret:
> I disagree that it would not do anything good. I think that it is good to have as small a number of concepts in the API as possible, and to avoid duplicating interfaces.

It's not clear to me when we would need a `TimeIntegrator` over the specialized version of it. The phase change code needs a specialized mass integrator. The IBFE code needs a specialized finite element integrator. The level set code needs a specialized re-initialization integrator. These integrators are not interchangeable, and so it seems like adding a base class is adding complexity for the sake of complexity.

I admit that may be short sited: just because I don't see a use for it now, doesn't mean that a use won't crop up in the future (reduced order modeling??). As long as it's a very bare-bones interface, I think the added complexity will be minimal.

boyceg:
> > I disagree that it would not do anything good. I think that it is good to have as small a number of concepts in the API as possible, and to avoid duplicating interfaces.
> 
> It's not clear to me when we would need a `TimeIntegrator` over the specialized version of it. The phase change code needs a specialized mass integrator. The IBFE code needs a specialized finite element integrator. The level set code needs a specialized re-initialization integrator. These integrators are not interchangeable, and so it seems like adding a base class is adding complexity for the sake of complexity.
> 
> I admit that may be short sited: just because I don't see a use for it now, doesn't mean that a use won't crop up in the future (reduced order modeling??). As long as it's a very bare-bones interface, I think the added complexity will be minimal.

Well, `TimeIntegrator` would be an ABC, so you would always need a specialization. The point is to avoid having slightly different software interfaces for what amounts to the same concept.

drwells:
There's about 15,000 lines of code now in this PR - by comparison there are 12,000 lines in the deal.II reimplementation of `IBFEMethod/IBFESurfaceMethod`. I'm going to need a few more days to work through this.

The conversation on abstract base classes is good.

abarret:
> Well, `TimeIntegrator` would be an ABC, so you would always need a specialization. The point is to avoid having slightly different software interfaces for what amounts to the same concept.

Yes I understand that. I am wondering is why we should or need to enforce an interface. Specializations will not be interchangeable, so polymorphism isn't needed. Is the point of the interface to provide users with a basic understanding of how to use specializations of `TimeIntegrator`?

My understanding of this code is that even with a `TimeIntegrator` interface, there would still be a need for the `MassIntegrator` class (to set up specializations for integrating the density), so adding in another base class `TimeIntegrator` does not decrease the code complexity.

@boyceg  are you envisioning something more than this?
```
class TimeIntegrator
{
TimeIntegrator();
virtual ~TimeIntegrator();
void initializeTimeIntegrator() = 0;
void integrate() = 0;
void deinitializeTimeIntegrator() = 0;
};
```

boyceg:
> @boyceg are you envisioning something more than this?
> 
> ```
> class TimeIntegrator
> {
> TimeIntegrator();
> virtual ~TimeIntegrator();
> void initializeTimeIntegrator() = 0;
> void integrate() = 0;
> void deinitializeTimeIntegrator() = 0;
> };
> ```

Yes, exactly.

include/ibamr/BrinkmanPenalizationStrategy.h
amneetb: Added that.

amneetb:
@Raamkrishnan : The regridding issue is fixed in #1570. Can you rebase this branch with #1573 and update the convective operators? Once that is done, rebase #1570 against this branch (#1569). Sorry for the convoluted re-basing, but I think finally #1570 will be the one that will get merged with master.

amneetb:
All of the outstanding issues listed here are done in #1570. These are:

1. Deprecating some of the Brinkman functions.

2. Regridding issues with changing number of levels.

3. Convective operators.

This branch has been merged with #1570.

boyceg:
@amneetb, note that #1570 seems a little wonky. E.g. `ScopedVectorCopy` is showing up as a new class. Could we either get this branch sorted out first, or break out #1570 into a collection of smaller PRs?

amneetb:
> @amneetb, note that #1570 seems a little wonky. E.g. `ScopedVectorCopy` is showing up as a new class. Could we either get this branch sorted out first, or break out #1570 into a collection of smaller PRs?

I think it will be better to work directly with #1570. If you can break and stick them together soon and give us an intact branch that would be great. The reason is we want to list the URL of examples in the paper (one reviewer has specifically asked for it). This also means that we want to ensure reproducibility of the results reported in the paper.

amneetb:
> If you prefer to break up #1570, that is fine, but you will need to do it, and then I can review the individual pieces. Ideally different functionality would be in different PRs (e.g. changes to the way that the convective operator works, new time integrator classes, etc.) --- and you all are the ones that know how everything fits together.

Yes, definitely I can do that. Let us first work with #1570 to complete the paper revision and ensure results reproducibility. Once that is done, I will break the working branch into smaller PRs for you to review. Hopefully we all can review them quickly and merge them to master. 
 
> 
> Personally I think it is silly to put URLs to code repositories in archival publications. Like even if this was in master already, it could conceivably move around between now and the time of publication, or by the time someone tries to reproduce several years down the line. (@knepley do you have any thoughts about managing this?)

I was thinking along the lines of your recent JCP paper, where you write something like 

```
This benchmark test is provided in IBAMR version within the directory examples/IIM/ex7.
```

> 
> I don't know if this is the best strategy, but I think it could make sense to list the git hash that was used for the tests along with the names of the examples. That would allow someone to pick out exactly the case that you ran without needing to know anything about which branch was used, etc.

Perhaps I can list #1570 PR hash, but that will go away. So I am not sure what would be a good way to list a hash for branches that are not merged with master.

boyceg:
[In most cases, branches will stick around forever (even if deleted).](https://github.com/orgs/community/discussions/23262) So I think listing a hash to a branch (even if it is subsequently deleted) should be reliable. I think it could even make sense to make a temporary branch for running tests that is subsequently deleted.

amneetb:
> [In most cases, branches will stick around forever (even if deleted).](https://github.com/orgs/community/discussions/23262) So I think listing a hash to a branch (even if it is subsequently deleted) should be reliable. I think it could even make sense to make a temporary branch for running tests that is subsequently deleted.

Great! This solves the problem of having an intact branch (temporary branch #1570 will serve the purpose) and or merging it with master before the re-submission of paper.

boyceg:
I must admit that I am not sure what happens when you rebase --- I think that those hashes eventually do go away. I would suggest possibly making a branch named something like "JCP-phase-change-test-branch" that has no PR associated with it, and using it for the tests.

amneetb:
Sounds good. We will create this type of branch out of #1570 to list it in the paper.

drwells:
I'm a bit late to the party this morning!

> I think it will be better to work directly with https://github.com/IBAMR/IBAMR/pull/1570. If you can break and stitch them together soon and give us an intact branch that would be great. The reason is we want to list the URL of examples in the paper (one reviewer has specifically asked for it). This also means that we want to ensure reproducibility of the results reported in the paper by rerunning all/most of the cases done in the paper.

We could use #1570, but the branch corresponding to the original version (#1505) is still around. If you just need something to point reviewers to, a good alternative would be to set up the phase change stuff as a project dependent on IBAMR. Everything should work as expected with the last IBAMR release. This has the advantage that you can set things up on Zenodo which will make even the pickiest reviewer happy.

amneetb:
That's a great idea!! Let me and Rama make sure #1570 is correct in terms of reproducibility of results. Then we can create a new repository out of it.

boyceg:
Can you point us to any documents that describe the workflow for setting this up / best practices? We should be doing this at least for all of our methods-focused papers, and maybe for our applications-focused ones as well.

amneetb:
In the past when David taught us Git rebasing, we noted down everything. That document serves us well till today. I intend to document these things as well, when David helps us further.

PR #1577, opened 2023-04-12
CI: update to checkout v3.

drwells:
This version of `checkout` also fixed some silly bugs with paths - let's keep the workarounds for now since they are more robust.


drwells:
I reran a job to verify that the cache is still populated correctly. This is good to go.

boyceg (approved):

PR #1580, opened 2023-04-12
Fix marker points

drwells:
Most of what we need for particles.

As discussed on slack, it looks like marker points which move into ghost regions go missing, so this isn't done yet - I'm working on it.

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
I posted a long description of some problems this ran into on slack - in summary, its rather difficult to consistently move points from coarse to fine patches since a point near the interface may have its velocity computed with coarse data on the coarse side and fine data on the fine side. It's probably fixable but it's going to be easier to remove the dependency on SAMRAI's coarsening and refinement operators.

In case we want it in the future the present state of that attempt is on

https://github.com/IBAMR/IBAMR/commits/drwells/samrai-lmarker-partial-fixes

PR #1583, opened 2023-04-21
Markers

drwells:
Reimplementation of marker points, advected with the midpoint method.

This first PR just adds a class for handling marker points. I'll add it to `IBExplicitHierarchyIntegrator` (and, relatedly, test restarts) in the next PR.

- [x] Was clang-format run?
      changelog entry been added?
- [x] ~If new data structures have been added to a class then have they been set
      number is incremented.~ I'm skipping this for now since this doesn't yet interface with the rest of the library
      GitHub for the pull request?


tests/IBTK/marker_points_02.cpp
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        // sync ghost data:
+        using ITC = IBTK::HierarchyGhostCellInterpolation::InterpolationTransactionComponent;
+        std::vector<ITC> ghost_cell_components(1);
+        ghost_cell_components[0] = ITC(
+            u_idx, "CONSERVATIVE_LINEAR_REFINE", true, "CONSERVATIVE_COARSEN", "LINEAR", true, u_bc_coef_ptrs, nullptr);
+        IBTK::HierarchyGhostCellInterpolation ghost_fill_op;
+        ghost_fill_op.initializeOperatorState(ghost_cell_components, patch_hierarchy);
+        ghost_fill_op.fillData(/*time*/ 0.0);
```
drwells, line 155: There's something slightly wrong with this ghostfilling approach: some corners are not filled correctly. I rigged up the tests to avoid that problem but it would still be nice to fix this.

boyceg:
@drwells: note that if the same time stepping and kernel functions are used for the IB structures and the markers, then co-located structure and marker points should always move together. This could also make for a good test. It should work for both `IBMethod` and _nodal_ (but not _elemental_) `IBFEMethod`.

drwells:
Yes - things should be set up so that we do exactly the same thing with the midpoint rule in both cases. I only implemented the midpoint rule since that's our default with IBFE.

amneetb:
I'm not sure if the new initialization PR would affect this PR or not, but if it be good to keep that in mind while implementing the new marker code.

drwells:
Yes, it's amusing timing - isn't it funny how we sometimes all encounter the same issue at once? This one in particular is a priority since Jordan needs markers ASAP to finish his dissertation.

I think this won't interact much with our regridding redo patches since markers are now stored in an independent data structure - if `IBFEMethod` works then this will be fine as well during regrids.

This PR doesn't actually interface with the rest of the library so that's a problem for #1584.

drwells:
> Should we try to issue a warning if the time stepping schemes are different?

We should probably just implement the rest of the timestepping schemes. How this interacts with the rest of the library should be discussed in #1584 - this PR is the independent implementation.

boyceg:
I think you have to ask the BC filling routine to do the corners. I think this also came up with the nodal vorticity patch — check out that code to see how to get the corners filled. If you cannot find it, I’ll try to dig it up when I get to the office.

drwells:
Sounds good. I think I figured out how to do this, but the way we fill corners for IBFEMethod is incomplete - I'll tidy up the marker point code and then we can address that problem next.

ibtk/include/ibtk/MarkerPoints.h
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Advect the markers by some given velocity field at half and new times
+     * and the interaction kernel @p kernel.
+     *
+     * The default method is the 'explicit trapezoid' method, which uses the
+     * present velocity to predict a new position and then uses a combination
+     * of a new velocity (computed at the predicted position) and current
+     * velocity to compute a new solution.
+     */
```
boyceg, line 165: Should we add a comment about whether ghost values need to be set? (Presumably the do.)

ibtk/include/ibtk/MarkerPoints.h
```diff
@@ -0,0 +1,190 @@
...
+    // class for timestepping.
+    friend class MarkerPoints;
+};
+
+/*!
+ * Implementation of marker points, also sometimes called particles. These
+ * points are akin to IB points except they are simply advected by the fluid
+ * and do not apply any force upon it.
+ */
+class MarkerPoints : public SAMRAI::tbox::DescribedClass
```
boyceg, line 121: `MarkerPatchHierarchy` ?

boyceg: (I guess I don't like the asymmetry between `MarkerPatch` and `MarkerPoints`.)

ibtk/include/ibtk/MarkerPoints.h
```diff
@@ -0,0 +1,190 @@
...
+     * Set marker velocities with some given velocity field @p u_idx and
+     * interaction kernel @p kernel.
+     */
+    void setVelocities(const int u_idx, const std::string& kernel);
+
+    /**
+     * Advect the markers by some given velocity field at half and new times
+     * and the interaction kernel @p kernel.
+     *
+     * The default method is the 'explicit trapezoid' method, which uses the
```
boyceg, line 161:
```suggestion
     * The default method is the 'explicit midpoint' method, which uses the
```

ibtk/include/ibtk/MarkerPoints.h
```diff
@@ -0,0 +1,190 @@
...
+#if 0
+    /**
+     * Save the present state of the object to a SAMRAI database.
+     */
+    void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db) override;
+#endif
+
+    /**
+     * Set marker velocities with some given velocity field @p u_idx and
+     * interaction kernel @p kernel.
```
boyceg, line 153: Do we list the available kernels somewhere? Or maybe point to `LEInteractor` for kernel names?

ibtk/include/ibtk/MarkerPoints.h
```diff
@@ -0,0 +1,190 @@
...
+ */
+class MarkerPoints : public SAMRAI::tbox::DescribedClass
+{
+public:
+    /**
+     * Constructor.
+     */
+    MarkerPoints(const std::string& name,
+                 SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > patch_hierarchy,
+                 const EigenAlignedVector<IBTK::Point>& positions,
```
boyceg, line 129: Can we indicate how the position and velocity data are expected to be arranged?

drwells:
I fixed all the review comments.

drwells:
@boyceg want to merge this one? :)

doc/news/changes/major/20230421DavidWells
```diff
@@ -0,0 +1,3 @@
+New: Added a class MarkerPoints for tracking marker points.
```
boyceg, line 1:
```suggestion
New: Added a class MarkerPatchHierarchy for tracking marker points.
```

boyceg (approved):
Looks good.

Should we deprecate (or just remove if it doesn't work?) the existing marker point code?

drwells:
> Looks good.
> 
> Should we deprecate (or just remove if it doesn't work?) the existing marker point code?

Yes - I removed it from `IBHierarchyIntegrator in 7f3a2ca1ce0290af159dd10c8bd28062d72361ba. Since the present version doesn't work with coarse-fine interfaces (or markers outside the physical domain) we should remove the whole thing later - I'll do that after the 'add markers' PRs are in.

PR #1584, opened 2023-04-24
Markers 2

drwells:
Followup to #1583 - this isn't done yet (and half the commits in this branch are in that one) but it shows where I'm going with all this.

Remaining items:
- [x] unconditionally do ghost updates for the particle velocity fields
- [x] serialization/restart
- [x] HDF5 output
- [x] moving particles to a temporary structure during regridding
- [x] add a method to set up particles from input vectors - users should only specify their positions and we should interpolate their velocities as needed
- [x] implement missing timestepping schemes

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


src/IB/IBExplicitHierarchyIntegrator.cpp
```diff
@@ -381,7 +387,14 @@ IBExplicitHierarchyIntegrator::postprocessIntegrateHierarchy(const double curren
                                          getGhostfillRefineSchedules(d_object_name + "::u"),
                                          new_time);
 
-    // This class is not responsible for postprocessing any other objects so proceed to the base class:
+    // Update the marker points, should they exist:
+    if (d_marker_points)
+    {
+        TBOX_ASSERT(d_time_stepping_type == MIDPOINT_RULE);
+        const std::string d_marker_kernel = "PIECEWISE_LINEAR";
+        d_marker_points->midpointStep(new_time - current_time, d_u_half_idx, u_new_idx, d_marker_kernel);
```
drwells, line 395: This is set up to save the IB-ghosted data (we'll have to redo the ghost updates in the case of the scratch hierarchy ourselves here)  used to advance the structures with the midpoint rule so that we do exactly the same timestepping on the marker points.

drwells:
Things appear to all work, though I need to write some tests. Doing unconditional ghost updates outside of the usual mechanism (i.e., the refinement and coarsening algorithms stored by HierarchyIntegrator and instead just using HierarchyGhostCellInterpolation directly) works.

drwells:
This PR is getting pretty long so I took out the IBExplicitHierarchyIntegrator patches - this PR just implements everything else we need for markers.

ibtk/include/ibtk/MarkerPatchHierarchy.h
```diff
@@ -146,33 +176,58 @@ class MarkerPatchHierarchy : public SAMRAI::tbox::DescribedClass
...
+     * and the interaction kernel @p kernel.
+     *
+     * This time integrator is the 'explicit trapezoid' method, which uses the
+     * present velocity to predict a new position and then uses the
+     * trapezoid rule to correct that position.
+     *
+     * @note This function assumes that @p u_new_idx has up-to-date ghost values
+     * with sufficient width for @p kernel. See LEInteractor for kernel names.
+     */
+    void trapezoidalStep(const double dt, const int u_new_idx, const std::string& kernel);
```
boyceg, line 229: Don't you need two velocities (u_current and u_new) for trapezoidal step?

Using `J` for the interpolation operator, the explicit trapezoidal rule should be:
```
   X(n+1,~) := X(n) + dt J[X(n)] u(n)
   X(n+1) := 0.5*dt*(J[X(n)] u(n) + J[X(n+1,~)] u(n+1))
```

ibtk/include/ibtk/MarkerPatchHierarchy.h
```diff
@@ -146,33 +176,58 @@ class MarkerPatchHierarchy : public SAMRAI::tbox::DescribedClass
...
+     * IBStrategy objects this uses the name `forwardEulerStep` to describe this
+     * timestepping method. After the markers are moved their velocities are
+     * reset to values interpolated from @p u_new_idx via the IB kernel @p
+     * kernel.
+     *
+     * @note This function assumes that @p u_new_idx has up-to-date ghost
+     * values with sufficient width for @p kernel. See LEInteractor for kernel
+     * names.
+     */
+    void forwardEulerStep(const double dt, const int u_new_idx, const std::string& kernel);
```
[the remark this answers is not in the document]
drwells: This should be `u_new_idx`. The way this class works is it assumes the velocities are always set to the correct current values, so for this function we move the markers and then interpolate the new velocity to keep that invariant.

ibtk/include/ibtk/MarkerPatchHierarchy.h
drwells: Yes, but like the other one - the present velocity is already known (markers store position and velocity).

ibtk/src/utilities/MarkerPatchHierarchy.cpp
```diff
@@ -343,6 +428,242 @@ MarkerPatchHierarchy::reinit(const EigenAlignedVector<IBTK::Point>& positions,
...
+    const auto& velocities = pair.second;
+
+    const std::array<std::string, NDIM> position_datasets{ { "x",
+                                                             "y",
+#if NDIM == 3
+                                                             "z"
+#endif
+    } };
+
+    // technically momentum, but marker points don't have mass so this is the
```
boyceg, line 506: Why are you using the momentum nomenclature --- is it required by VisIt?

drwells: Yes - H5Part stores momenta, not velocity. I'll add a comment.

ibtk/include/ibtk/MarkerPatchHierarchy.h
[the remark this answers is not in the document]
drwells: I'll add a comment explaining that for all the timestepping schemes.

ibtk/src/utilities/MarkerPatchHierarchy.cpp
```diff
@@ -380,6 +753,49 @@ MarkerPatchHierarchy::setVelocities(const int u_idx, const std::string& kernel)
...
+                for (unsigned int i = 0; i < marker_patch.d_positions.size(); ++i)
+                {
+                    new_positions[i] = marker_patch.d_positions[i] + dt * marker_patch.d_velocities[i];
+                }
+
+                // 2. Interpolate the velocity at the new time:
+                std::vector<double> new_velocities(marker_patch.d_velocities);
+                do_interpolation(u_new_idx, new_positions, patch, kernel, new_velocities);
+
+                // 3. Do a backward Euler step:
```
boyceg, line 782: I had to look up that this is what `FORWARD_EULER` time stepping does for particles with `IBExplicitHierarchyIntegrator`. Seems like I might have missed adding a `break` statement in one of those switch statements. 🤷‍♂️

Speaking of --- is there a reason not to follow the pattern in `IBExplicitHierarchyIntegrator` and pass in the time stepping type and have a common time stepping routine? I guess one reason is that it makes it easier to see through the interface what time stepping is provided...

drwells: > I had to look up that this is what FORWARD_EULER time stepping does for particles with IBExplicitHierarchyIntegrator. Seems like I might have missed adding a break statement in one of those switch statements. :man_shrugging:

Yeah, I wasn't sure what was going on with that one but I think it's still a convergent time integrator? Should we fix that? I have no idea if anyone is actually relying on that time integrator right now.

> s there a reason not to follow the pattern in IBExplicitHierarchyIntegrator and pass in the time stepping type and have a common time stepping routine?

We could, but it is slightly simpler to defer the choice of timestepping to the caller. I'd rather not carry the extra state around.

Interestingly, since markers have a one-way coupling with the velocity field, all the time integrators are actually equivalent - the velocity field is linear in time so midpoint/trapezoid/whatever forward Euler turns out to be are all equal. If we use the same time integrator we get exactly the same coordinates for the markers as IB points. If we use a different one then it is off by roundoff.

boyceg (approved):

PR #1585, opened 2023-04-27
ensure that regridHierarchyConfiguration is called when expected

boyceg:
- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
Note that this drops some commits from #1555 to restore the original behavior of `AdvDiffHierarchyIntegrator` / `AdvDiffSemiImplicitHierarchyIntegrator` --- and so the history is screwy.

boyceg:
Implementations of `HierarchyIntegrator::resetHierarchyConfiguration` often assume that function is called any time there is a change to the patch hierarchy, but SAMRAI appears not call that function if the only change to the hierarchy is that the finest level was dropped. In this PR, there is extra logic to see if the number of levels changed but `HierarchyIntegrator::resetHierarchyConfiguration` was not called.

@amneetb, I think this is a simpler fix than what is in #1555 and #1567.

The other issue is that many of the solver objects' deallocation routines break if the number of levels changes. I think that the right fix is to make sure that the deallocation routines work if the number of levels changes. This mostly entails extra calls to `SAMRAIVectorReal::resetLevels`. I started doing that here.

boyceg:
The relevant changes (so far) are:
* https://github.com/IBAMR/IBAMR/pull/1585/commits/73c2247b0d044db3cb37e90c7bcb8a6eb0465e93
* https://github.com/IBAMR/IBAMR/pull/1585/commits/daf1b390cf056862b31fe7b78be77ef06e593015

drwells:
Can you explicitly use `git revert` so that the history is cleaner? That would help me understand where the new stuff is.

PR #1587, opened 2023-05-01
Markers 3

drwells:
~Only the last five commits are relevant - the rest are more of the marker implementation and are in the other PR.~

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/tests/IBFE/explicit_ex4.cpp
+++ b/tests/IBFE/explicit_ex4.cpp
@@ -63,6 +63,7 @@
  * 3. IBFE part inactivation
  * 4. IBFECentroidPostProcessor
  * 5. IBTK::MergingLoadBalancer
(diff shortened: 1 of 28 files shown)


tests/IBFE/explicit_ex4.cpp
```diff
@@ -530,6 +531,31 @@ main(int argc, char** argv)
...
+                X[1] += 0.5;
+#if NDIM == 3
+                X[2] += 0.5;
+#endif
+
+                positions.push_back(X);
+            }
+
+            time_integrator->setMarkers(positions);
+        }
```
drwells, line 557: @jabrown893 You'll be the first user - does something like this make sense for how you plan on using this new feature?

boyceg (approved):

PR #1589, opened 2023-05-04
Vector RAII

drwells:
Spin-off from #1586.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


doc/news/changes/minor/20230504DavidWells
```diff
@@ -0,0 +1,4 @@
+New: added two new classes ScopedVectorCopy and ScopedVectorDuplicate to enable
+usage of SAMRAI vectors with RAII semantics.
+<br>
+(David Wells, 2023/05/04)
```
drwells, line 4: We'll write a major changelog entry for the three of us once this is all done.

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -324,8 +326,101 @@ using FRDV = FreeRigidDOFVector;
...
+    x.resetLevels(coarsest_ln_in, finest_ln_in);
+    return;
+}
+
+/*!
+ * Utility function which asserts that the SAMRAI pointer is not null.
+ *
+ * This is useful for writing generic code in which we might want to assert that
+ * a pointer is not null in the initialization list where we cannot use the
+ * normal assertion macro. assertion.
```
boyceg, line 397:
```suggestion
 * normal assertion macro.
```
?

ibtk/include/ibtk/private/ScopedVectorDuplicate-inl.h
```diff
@@ -0,0 +1,97 @@
...
+    : ScopedVectorDuplicate(checked_dereference(vector), name)
+{
+}
+
+template <typename TYPE>
+ScopedVectorDuplicate<TYPE>::ScopedVectorDuplicate(const SAMRAI::solv::SAMRAIVectorReal<NDIM, TYPE>& vector,
+                                                   const std::string& name)
+{
+    d_vector = vector.cloneVector(name);
+    d_vector->allocateVectorData();
```
boyceg, line 47: Do we want to allow for users to pass in a value for the `timestamp` argument to `allocateVectorData()`?

boyceg:
Looks good. Are there more places in the library where we should switch to using this class before merging?

boyceg:
I wonder if we should just make our own `AMRVector`, `HierarchyVector`, and/or `LevelVector` class(es) and stop using `SAMRAIVectoReal`.

ibtk/include/ibtk/private/ScopedVectorDuplicate-inl.h
drwells: We could, but I don't think its necessary - these vectors should not participate in time interpolation.

drwells:
I switched over one more class- that's two fewer persistent vectors. In general I think we should always copy these in `solve()` instead of setting up scratch vectors.

boyceg:
I am not sure, but I think it would be good to avoid generating unneeded copies of the vector _data_.

ibtk/src/solvers/impls/FACPreconditioner.cpp
```diff
@@ -94,48 +96,43 @@ FACPreconditioner::setTimeInterval(const double current_time, const double new_t
...
+    if (deallocate_after_solve) initializeSolverState(x, b);
 
     // Set the initial guess to equal zero.
-    u.setToScalar(0.0, /*interior_only*/ false);
+    x.setToScalar(0.0, /*interior_only*/ false);
+
+    // Clone the right-hand-side vector to avoid modifying it during the
+    // preconditioning operation.
+    ScopedVectorCopy<double> f(b);
+    ScopedVectorDuplicate<double> r(b);
```
drwells, line 111: I agree that we should avoid extra data copies. What's going on here is that we set up `f` as a deep copy of `b` and `r` as a vector of the same size with all zeros. This avoids one extra data copy (we don't need to set `r = f` - that's done in the residual computation)

drwells:
This is ready to go on my end.

drwells:
@boyceg if you're OK with this then lets merge this and rebase the other PRs on top of it.

boyceg:
Checking now.

boyceg:
Looks good. Comments about "duplicate" in the scoped copy class may need some clarification. One thought is --- there are multiple different vector concepts in the code. Would it be useful to include "SAMRAIVector" in the class name? Perhaps `SAMRAIVectorScopedCopy` and `SAMRAIVectorScopedDuplicate`?

boyceg (approved):

drwells:
I renamed them (it only takes a moment to update such things). I think its wise that we not have `IBTK::Vector` and `IBTK::ScopedVectorCopy` have such similar names.

doc/news/changes/minor/20230504DavidWells
```diff
@@ -0,0 +1,4 @@
+New: added two new classes ScopedVectorCopy and ScopedVectorDuplicate to enable
```
boyceg, line 1:
```suggestion
New: added two new classes SAMRAIScopedVectorCopy and SAMRAIScopedVectorDuplicate to enable
```

boyceg:
> I renamed them (it only takes a moment to update such things). I think its wise that we not have `IBTK::Vector` and `IBTK::ScopedVectorCopy` have such similar names.

Sounds good. I guess I like having `SAMRAIVector` in the name to make it clearer which vector class it is working with --- but do not have very strong feelings. (I guess it would be more clear as `SAMRAIVectorReal`, but that gets pretty verbose...)

drwells:
We don't do anything with complex numbers so that's just needlessly wordy.

abarret:
It's not a big deal, but this is potentially a breaking change for downstream projects that utilize the `protected` attributes in `FACPreconditioner` (like one I'm working on). It's probably worth noting this in the changelog.

PR #1593, opened 2023-05-08
Fix some more marker bugs.

drwells:
We can now unset markers and then set them up again later.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBExplicitHierarchyIntegrator.cpp
+++ b/src/IB/IBExplicitHierarchyIntegrator.cpp
@@ -511,7 +511,12 @@ IBExplicitHierarchyIntegrator::setMarkers(const EigenAlignedVector<IBTK::Point>&
     IBTK::Vector v;
     v.fill(0.0);
     std::fill(velocities.begin(), velocities.end(), v);
+    // explicitly destroy the present markers (should they exist) since the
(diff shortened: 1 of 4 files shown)


src/IB/IBExplicitHierarchyIntegrator.cpp
```diff
@@ -511,7 +511,10 @@ IBExplicitHierarchyIntegrator::setMarkers(const EigenAlignedVector<IBTK::Point>&
     IBTK::Vector v;
     v.fill(0.0);
     std::fill(velocities.begin(), velocities.end(), v);
+    // avoid having two markers at once
+    d_markers.setNull();
```
boyceg, line 515: If you reset the value of `d_markers` on the next line, do you need to call `setNull()` here?

drwells: Yes. We need to delete the current one to remove it from the restart database before adding a new one - otherwise we will have two objects with the same name.

boyceg: Ah --- can you add that to the comment?

drwells:
Fixed!

boyceg (approved):

PR #1599, opened 2023-05-11
look up the patch data using a patch data index

boyceg:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
Ooops. Yet case of "pointers can be converted to integers".

What's a bit surprising about this one is that I thought the new tests in that PR would execute that code path - I'll fix the test and append that to this PR.

ibtk/src/math/HierarchyMathOps.cpp
```diff
@@ -1987,7 +1987,7 @@ HierarchyMathOps::interp(const int dst_idx,
                          const bool src_cf_bdry_synch)
 {
     if (src_ghost_fill) src_ghost_fill->fillData(src_ghost_fill_time);
-    Pointer<NodeDataFactory<NDIM, double> > data_factory = dst_var->getPatchDataFactory();
+    Pointer<CellDataFactory<NDIM, double> > data_factory = dst_var->getPatchDataFactory();
```
drwells, line 1990: I fixed another copy-paste error.

drwells (approved):

drwells:
While I was at it - I noticed that the existing nodal interpolation tests weren't wrong, but they used periodic BCs and a nonperiodic manufactured solution which caused the infinity norm to be O(1). I fixed that and added a new test for node-to-cell.

ibtk/src/math/HierarchyMathOps.cpp
boyceg: Gotta get rid of the silently failing implicit conversion facilitated by `Pointer`.

PR #1600, opened 2023-05-11
mostly remove use of Pointer in HierarchyMathOps and PatchMathOps

boyceg:
- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/CartCellRobinPhysBdryOp.h
+++ b/ibtk/include/ibtk/CartCellRobinPhysBdryOp.h
@@ -232,7 +232,7 @@ public:
      * \param ghost_width_to_fill  Integer vector describing maximum ghost width to fill over
      * all registered scratch components.
      */
-    void accumulateFromPhysicalBoundaryData(SAMRAI::hier::Patch<NDIM>& patch,
+    void accumulateFromPhysicalBoundaryData(const SAMRAI::hier::Patch<NDIM>& patch,
                                             double fill_time,
                                             const SAMRAI::hier::IntVector<NDIM>& ghost_width_to_fill) override;
 
@@ -265,7 +265,7 @@ private:
                                    const SAMRAI::tbox::Array<SAMRAI::hier::BoundaryBox<NDIM> >& physical_codim1_boxes,
                                    double fill_time,
                                    const SAMRAI::hier::IntVector<NDIM>& ghost_width_to_fill,
-                                   SAMRAI::hier::Patch<NDIM>& patch,
+                                   const SAMRAI::hier::Patch<NDIM>& patch,
                                    bool adjoint_op);
 
     /*!
--- a/ibtk/include/ibtk/HierarchyMathOps.h
+++ b/ibtk/include/ibtk/HierarchyMathOps.h
@@ -224,10 +224,10 @@ public:
      * \see resetLevels
      */
     void curl(int dst_idx,
-              SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > dst_var,
+              const SAMRAI::pdat::CellVariable<NDIM, double>* dst_var,
               int src_idx,
-              SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > src_var,
-              SAMRAI::tbox::Pointer<HierarchyGhostCellInterpolation> src_ghost_fill,
+              const SAMRAI::pdat::CellVariable<NDIM, double>* src_var,
+              HierarchyGhostCellInterpolation* src_ghost_fill,
               double src_ghost_fill_time);
 
     /*!
@@ -242,10 +242,10 @@ public:
      * \see resetLevels
      */
     void curl(int dst_idx,
-              SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > dst_var,
+              const SAMRAI::pdat::CellVariable<NDIM, double>* dst_var,
               int src_idx,
-              SAMRAI::tbox::Pointer<SAMRAI::pdat::FaceVariable<NDIM, double> > src_var,
-              SAMRAI::tbox::Pointer<HierarchyGhostCellInterpolation> src_ghost_fill,
+              const SAMRAI::pdat::FaceVariable<NDIM, double>* src_var,
+              HierarchyGhostCellInterpolation* src_ghost_fill,
               double src_ghost_fill_time);
 
     /*!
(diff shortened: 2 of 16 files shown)


abarret:
Is there a reason you are using raw pointers instead of references in the interface?

boyceg:
Yes!
- In many functions, some of the arguments are optional (indicated by passing in a null pointer). We could make the non-optional arguments references and the optional ones pointers, but I am not crazy about this.
- This does not require any changes to user code.

ibtk/include/ibtk/CartCellRobinPhysBdryOp.h
```diff
@@ -232,7 +232,7 @@ class CartCellRobinPhysBdryOp : public RobinPhysBdryPatchStrategy
      * \param ghost_width_to_fill  Integer vector describing maximum ghost width to fill over
      * all registered scratch components.
      */
-    void accumulateFromPhysicalBoundaryData(SAMRAI::hier::Patch<NDIM>& patch,
+    void accumulateFromPhysicalBoundaryData(const SAMRAI::hier::Patch<NDIM>& patch,
                                             double fill_time,
                                             const SAMRAI::hier::IntVector<NDIM>& ghost_width_to_fill) override;
```
drwells, line 237: While making these things `const` is fine w.r.t. the type system, its a bit misleading because we modify the underlying `PatchData` object in a next step - could we leave these as-is?

ibtk/include/ibtk/HierarchyMathOps.h
```diff
@@ -142,7 +142,7 @@ class HierarchyMathOps : public SAMRAI::tbox::DescribedClass
      * If a cell is not refined in the hierarchy, its weight is set to its
      * volume.  If a cell is refined, its weight is set to zero.
      */
-    SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > getCellWeightVariable() const;
+    SAMRAI::pdat::CellVariable<NDIM, double>* getCellWeightVariable() const;
```
drwells, line 145: Like Aaron I don't see an advantage of this design: this is always going to return a non-null pointer so it should just be a reference.

ibtk/include/ibtk/HierarchyMathOps.h
```diff
@@ -224,10 +224,10 @@ class HierarchyMathOps : public SAMRAI::tbox::DescribedClass
      * \see resetLevels
      */
     void curl(int dst_idx,
-              SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > dst_var,
+              const SAMRAI::pdat::CellVariable<NDIM, double>* dst_var,
```
drwells, line 227: This is a good opportunity to switch these to references. Like I noted yesterday, one problem with this interface is that

```cpp
curl(0, 0, 0, 0, 0, 0);
```
compiles due to implicit conversions. We should leverage the type system here.

Is there a case in which we would know the type of variable but not have the `Pointer<Variable>` on-hand? I don't think I have ever seen that in IBAMR.

ibtk/include/ibtk/HierarchyMathOps.h
```diff
@@ -535,16 +535,16 @@ class HierarchyMathOps : public SAMRAI::tbox::DescribedClass
...
               int src1_idx,
-              SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > src1_var,
-              SAMRAI::tbox::Pointer<HierarchyGhostCellInterpolation> src1_ghost_fill,
+              const SAMRAI::pdat::CellVariable<NDIM, double>* src1_var,
+              HierarchyGhostCellInterpolation* src1_ghost_fill,
               double src1_ghost_fill_time,
               double beta = 0.0,
               int src2_idx = invalid_index,
-              SAMRAI::tbox::Pointer<SAMRAI::pdat::FaceVariable<NDIM, double> > src2_var = NULL,
+              const SAMRAI::pdat::FaceVariable<NDIM, double>* src2_var = NULL,
```
drwells, line 547: in contrast, since these are optional arguments, we should use pointers here. Can you use `nullptr`? How did `NULL` sneak back in?

drwells (changes requested):
Now might not be a great time to make this PR because of all the outstanding stuff for phase changes. Nevertheless we should do it.

boyceg:
I would argue that it actually is **not** a big change because `Pointer<T>` has implicit conversion to `T*`, so the only time that user (or library) code has to change is if some (but not all) of the arguments were being passed in with `Pointer<BaseT>` instead of `Pointer<T>`. Notice, in particular, that almost all of the changes are in `HierarchyMathOps` and `PatchMathOps`.

boyceg:
(Changing to references for some arguments would be a much more invasive change.)

ibtk/include/ibtk/HierarchyMathOps.h
boyceg: I needed to revert this change anyway --- we need to provide a `Pointer` here because ultimately these weights are used in `SAMRAIVectorReal`, which expects a `Pointer`.

ibtk/include/ibtk/HierarchyMathOps.h
boyceg: Changed to `nullptr`. I don't think that this PR introduced `NULL` here ...

boyceg:
@drwells, @abarret, tests are all now passing.

ibtk/include/ibtk/HierarchyMathOps.h
drwells: I'm not sure how NULL got back in - we need to avoid that (especially when we mix pointers and integers)

boyceg: Agh I must have dropped the commit where I fixed that.

boyceg: Ah, I did fix it -- I think you are looking at an outdated commit?

ibtk/include/ibtk/CartCellRobinPhysBdryOp.h
boyceg: So do you think we should make _all_ of these `Patch` arguments non-`const` references? These functions really should _not_ be modifying these `Patch`es.

drwells:
This looks good but I'm hesitant to make such a sweeping change right before the release. Are you OK with holding off on this until we create a release branch?

boyceg:
Fine with me.

I think we should make removing `Pointer<T>` wherever possible an explicit goal for the following release, along with eliminating use of implicit down-casting via the `Pointer<T>` interface.

boyceg:
@drwells @abarret : LMK what you think. I would like to limit the use of `SAMRAI::tbox::Pointer<T>`.

abarret:
> @drwells @abarret : LMK what you think. I would like to limit the use of `SAMRAI::tbox::Pointer<T>`.

I don't have any issues with this. If there are other sweeping or breaking changes we want to do, I think it would be good to do all of those in a single release.

boyceg:
> > @drwells @abarret : LMK what you think. I would like to limit the use of `SAMRAI::tbox::Pointer<T>`.
> 
> I don't have any issues with this. If there are other sweeping or breaking changes we want to do, I think it would be good to do all of those in a single release.

I think that there are basically no breaking changes here since the functions don't retain ownership of the objects passed in using `Pointer<T>`.

ibtk/include/ibtk/CartCellRobinPhysBdryOp.h
boyceg: I don't agree that it is misleading --- we are not changing the `Patch` object that is being passed in.

boyceg:
I was thinking some more about this, and if the goal is to make IBAMR work with both SAMRAI 2 and 3/4 (at least for some period of time), then I think it would be better not to do this, since `std::shared_ptr<T>` does not provide automatic conversion to `T*` --- which this is relying on.

drwells:
Yup (and it shouldn't).

It would be better to replace these pointers with references, but one thing at a time.

PR #1607, opened 2023-05-31
put mpi link in correct place in CMakeLists.txt

marshallrdavey:
The checklist is irrelevant because I just changed the build script.


drwells:
See also #1545.

PR #1610, opened 2023-06-06
PR to conserve mass with the level set method close to machine epsilon 

amneetb:
- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


amneetb:
@Kaustubh-Khedkar : Please check level_set/ex2 for a strenuous example to check mass loss issues for the level set method. You can use most of the code from here to test your new conservative level set method code.

src/level_set/LevelSetUtilities.cpp
```diff
@@ -307,6 +309,66 @@ LevelSetUtilities::computeIntegralHeavisideFcns(LevelSetContainer* lsc)
...
+                }
+            }
+        }
+    }
+
+    return IBTK_MPI::sumReduction(integral);
+
+} // computeNetInFlowPhysicalBoundary
+
 void
```
amneetb, line 372: @boyceg, @drwells and @abarret: does this implementation look correct for computing net inflow through a physical boundary?

drwells: Sorry for the slow response - I see that this is also here: https://github.com/IBAMR/IBAMR/discussions/1615.

This generally looks right and if we just need to do one side of the boundary then the problem is a lot easier than an arbitrary subregion. `IBFEInstrumentPanel` and `IBInstrumentPanel` technically have some code to compute these quantities but the implementations are either not great (`IBFEInstrumentPanel` uses a pizza mesh) or incomplete.

Can you add some tests? Its pretty easy to mess up indexing with side-centered data. That's the best way to know for sure.

amneetb: Thanks! We plan to add a test to check the implementation. 

BTW, should we move this function to some other utility class? As written, it does not need any level set information  (but this can be changed easily). We are assuming only a single phase is present that the boundary location, which is the most case even for the intended multiphase applications we have in mind.

boyceg: Could do something similar to `SpongeLayerForceFunction`, which lets you evaluate flow rates through different domain boundaries.

drwells (changes requested):
Thanks for doing this - we need to cut down on the amount of repeated stuff in the examples and this helps a lot.

Want to turn the new example into a test? Also we need a changelog entry

include/ibamr/LevelSetUtilities.h
```diff
@@ -0,0 +1,420 @@
...
+    static std::vector<double> computeHeavisideIntegrals3PhaseFlows(LevelSetContainer* lsc);
+
+    /*!
+     * \return Integral of inflow \f$ -\vec{u} \cdot \vec{n}\f$ at a physical boundary.
+     *
+     * \param location_idx of the boundary at which the integral is evaluated
+     */
+    static double computeNetInflowPhysicalBoundary(SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps> hier_math_ops,
+                                                   int u_idx,
+                                                   int bdry_loc_idx);
```
drwells, line 345: We also still need a test for this function.

amneetb:
I need to use functionality of this PR to remove repeated stuff in many other examples as well. I have done it for a handful of examples as of now. This PR will cut down a lot of repeated code from the library. I'm waiting for the phase change PRs 1-7 to get merged soon so that I can rebase this one against master and start cutting more.

src/level_set/RelaxationLSMethod.cpp
```diff
@@ -926,122 +901,110 @@ RelaxationLSMethod::applyMassConstraint(Pointer<CellData<NDIM, double> > dist_da
...
+                                              int dist_init_idx,
+                                              int ham_idx,
+                                              int dt_idx) const
 {
     Pointer<PatchHierarchy<NDIM> > hierarchy = hier_math_ops->getPatchHierarchy();
     const int coarsest_ln = 0;
     const int finest_ln = hierarchy->getFinestLevelNumber();
 
-    // First compute a Heaviside field from the level set field
+    auto signof = [](const double x) { return x > 0.0 ? 1.0 : (x < 0.0 ? -1.0 : 0.0); };
```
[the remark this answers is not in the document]
amneetb: @drwells: I checked (https://en.cppreference.com/w/cpp/numeric/math/signbit). It  returns bool: `false` for negative, `true` else. We need `-1.0', `0.0`, or `1.0` instead here. Do you mean a different function instead?

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -268,6 +268,20 @@ smooth_heaviside(const double& phi, const double& alpha)
     return Hphi;
 }
 
+/*!
+ * Smooth delta function.
```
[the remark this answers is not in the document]
amneetb: I think `smooth_delta` represents derivative of `smooth_heaviside` function that is also defined in this file. The name makes this connection explicit. `cosine_kernel` will hide.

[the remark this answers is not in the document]
boyceg: IDK, if you ask me, we should do everything we can to discourage people from using the cosine delta function. 😉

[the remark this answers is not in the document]
amneetb: Its purpose is different here. It is not used for interpolation or as such, but to define a smooth Heaviside function to mark domain of a phase. The integral of Heaviside function gives total volume of the phase while the derivative of Heaviside function gives the surface area of the phase. Not at all related to IB...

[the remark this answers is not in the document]
boyceg: It might be worth trying to come up with some way to make this clear in the function name --- maybe by attaching these functions to a different namespace or class.

include/ibamr/LevelSetUtilities.h
```diff
@@ -0,0 +1,420 @@
...
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * \brief Class LevelSetUtilities provides implementation of some helper functions
+ * to enforce mass/volume conservation of phases.
+ */
+
+class LevelSetUtilities
```
boyceg, line 69: @drwells do you have an opinion here about using inner classes vs. classes in a namespace? I feel like we do not do a lot of inner classes, although of course there are some.

include/ibamr/LevelSetUtilities.h
```diff
@@ -0,0 +1,420 @@
...
+     */
+    class LevelSetContainer : public virtual SAMRAI::tbox::DescribedClass
+    {
+    public:
+        /*!
+         * \brief Constructor of the class.
+         */
+        LevelSetContainer(SAMRAI::tbox::Pointer<AdvDiffHierarchyIntegrator> adv_diff_integrator,
+                          SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > ls_var)
+            : d_adv_diff_integrator(adv_diff_integrator)
```
boyceg, line 84: I think this 
```suggestion
            : d_adv_diff_integrator(adv_diff_integrator),
              d_ls_vars({ls_var})
```

include/ibamr/LevelSetUtilities.h
boyceg: Also, should the function implementations go in an `-inl.h` file in `include/ibamr/private`?

amneetb:
@boyceg : Can we first work on phase change PRs? I need to rebase this PR against #1628 and break it into code change + example/test changes. I'm doing this as I am typing this message ;-)

include/ibamr/LevelSetUtilities.h
```diff
@@ -0,0 +1,420 @@
...
+/*!
+ * \brief Class LevelSetUtilities provides implementation of some helper functions
+ * to enforce mass/volume conservation of phases.
+ */
+
+class LevelSetUtilities
+{
+public:
+    /*!
+     * \brief A lightweight class to hold the level set variable and the associated hierarchy
```
boyceg, line 73: Container has a fairly specific meaning in C++ / STL. Is there a reason that you want to call this a container?

Also, there are is at least one other data item stored (interface half width). Can the description be a little "higher level" --- maybe (just guessing) something like "encapsulate data needed to specify a level set formulation"?

include/ibamr/LevelSetUtilities.h
```diff
@@ -0,0 +1,420 @@
...
+        } // LevelSetContainer
+
+        LevelSetContainer(SAMRAI::tbox::Pointer<AdvDiffHierarchyIntegrator> adv_diff_integrator,
+                          std::vector<SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > > ls_vars)
+            : d_adv_diff_integrator(adv_diff_integrator), d_ls_vars(std::move(ls_vars))
+        {
+            // return;
+        } // LevelSetContainer
+
+        void setInterfaceHalfWidth(double ncells)
```
boyceg, line 97: Is there a reason not to include this in the constructor too? The other variables are essentially read only --- should this one be handled similarly?

include/ibamr/LevelSetUtilities.h
```diff
@@ -0,0 +1,420 @@
...
+         * Name of this object.
+         */
+        std::string d_object_name;
+
+        SAMRAI::tbox::Pointer<IBAMR::LSInitStrategy> d_ls_ops;
+
+    }; // SetLSProperties
+
+    /*!
+     * \brief A function that sets or resets the level set data on the patch hierarchy
```
boyceg, line 405: What does this mean?

amneetb:
#1630 is the first patch for this PR.

include/ibamr/LevelSetUtilities.h
amneetb: Yes. `ncells` is not a property of level set function, but rather the algorithm utilizing the level set variable. They can change them during the simulation.

boyceg: Also, is there source doe documentation somewhere for how this parameter is used?

include/ibamr/LevelSetUtilities.h
amneetb: This description is from Nishant. I think by sets he meant at t = 0, and by reset at other times. At initial time LS is set in a different manner than at other times.

include/ibamr/LevelSetUtilities.h
amneetb: Sounds good. Do you have a different name for container here?

include/ibamr/LevelSetUtilities.h
boyceg: What data are being set / reset?

include/ibamr/LevelSetUtilities.h
boyceg: I'm not sure --- what are you imagining that this class encapsulates?

include/ibamr/LevelSetUtilities.h
amneetb: Just the level set function (a scalar field).

include/ibamr/LevelSetUtilities.h
amneetb: Mainly the LS variable, its hierarchy integrator, and numerical interface width.

boyceg: I mean is it just a "level set function specification" ("level set specification")?

include/ibamr/LevelSetUtilities.h
boyceg: What is the algorithm that is used to set / reset?

Also, how does this relate to `SetLSProperties`, which at first glance appears to do the same thing (but with more flexibility since it takes in an `LSInitStrategy`)?

amneetb:
I will work on these suggestions for `LevelSetUtilities` class in #1631

include/ibamr/LevelSetUtilities.h
amneetb: I will add a comment in .h file, where `ncells` gets used.

include/ibamr/LevelSetUtilities.h
amneetb: Basically `setLSDataPatchHierarchy` is a callback function needed in `AdvDiffHI` to reset the level set function. This function (`setLSDataPatchHierarchy()`) takes a pointer to `SetLSProperties` to do the job. 

I think we had discussed many years ago to make `AdvDiffHI` agnostic of LS functions. This is done via callback functions to HI.

include/ibamr/LevelSetUtilities.h
amneetb: I reverted this back, as this way of constructing vector data members does not work (at least on my system). 

I tried a simple program

```
#include <iostream>
#include <vector>

class A
{
public:

A(double a = 1.5, double b= 2.7):d_vec({a,b})
{
return;
}

void print()
{
  for (int i = 0; i < d_vec.size(); ++i)
  {
     std::cout << d_vec[i];
  }
  return;
}
private:
std::vector<double> d_vec;
};

int main()
{
   A obj;
   obj.print();
   return 0;
}

$ clang++ vect.cpp -o ./exe
vect.cpp:8:40: error: expected expression
A(double a = 1.5, double b= 2.7):d_vec({a,b})
                                                                   ^
1 error generated.


```

include/ibamr/LevelSetUtilities.h
drwells: I think we talked about this in-person at some point: lets start transitioning things to being in namespaces.

drwells:
It looks like #1630 includes all the phase change patches now too. Was that intentional? I am generally fine with this PR as-is and we can work on it here if it doesn't rely on any of the new phase change code.

amneetb:
> It looks like #1630 includes all the phase change patches now too. Was that intentional? I am generally fine with this PR as-is and we can work on it here if it doesn't rely on any of the new phase change code.

Yes. We decided to break big PRs into small patches. #1630 and #1631 breaks this into two PRs related to code changes. There would be an additional PR that contains examples and tests from this PR, but I have not put that out yet.

amneetb:
BTW, I did not close this PR because it lists Boyce's suggestions. I think I have already addressed them in #1630 and #1631.

boyceg:
If these are superseded by the PRs that supersede #1630 and #1631, can we close this one?

boyceg:
The main library changes from this PR landed through #1687, which superseded #1630 and #1631, with related example cleanup in #1702.

Using Codex, I have checked this PR for remaining issues and opened four focused follow-ups:
- #1976 fixes the executable registration so the existing Poiseuille-flow test can run through `attest`.
- #1977 stabilizes the shared regularized Heaviside and delta evaluation near the smoothing cutoffs.
- #1978 safeguards global phase-volume correction and adds lifecycle, restart, and failure coverage.
- #1979 handles degenerate redistribution and adds strict regression checks for preserved resolved-case behavior.

The numerical changes form the PR stack #1977 → #1978 → #1979. #1976 is independent of that stack.  @amneetb, if you could take a look, that would be very helpful! I think these should be pretty quick to review. Maybe none of it is needed!

I am going to go ahead and close this older PR as superseded for now. (We can always reopen it later.)

PR #1622, opened 2023-07-03
Phase change patch 1: Changes to INSVCHI to get phase change code ready

amneetb:
Adding div.u \ne 0 and different types of mass-momentum integrators

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


amneetb:
@boyceg : This is the first patch that sets the stage. Patch 2 will be based on this PR (in a stacked sense).

drwells:
Thanks for splitting this up!

I apologize for not getting back to this earlier. I'm on vacation now (and will also be gone for the second half of next week) so I will get back to these patches either next Monday/Tuesday or the week of the 24th.

amneetb:
Thanks! No worries.

Raamkrishnan:
@drwells Can you please take a look at our patches when you have time?

drwells:
Yes - I'll start the review this week!

drwells (changes requested):
I read through the whole thing and I think its fine - I don't see anything wrong with moving some of the details pertaining to the conservation of mass to a new base class.

Can we skip making `INSVCStaggeredConservativeMassMomentumSSPRKIntegrator` a separate class? It just implements some more RK schemes - we should have that in `INSVCStaggeredConservativeMassMomentumRKIntegrator`.

Furthermore - all SSPRK stuff is new. Can we add some tests for it?

drwells:
@boyceg do you have strong feelings either way about `MassIntegrator`? I'd like to move this along and I don't think we need to make fundamental changes.

drwells:
> Furthermore - all SSPRK stuff is new. Can we add some tests for it?

To be clear, I don't see that class used in a test in #1628 either.

amneetb:
> I read through the whole thing and I think its fine - I don't see anything wrong with moving some of the details pertaining to the conservation of mass to a new base class.
> 
> Can we skip making `INSVCStaggeredConservativeMassMomentumSSPRKIntegrator` a separate class? It just implements some more RK schemes - we should have that in `INSVCStaggeredConservativeMassMomentumRKIntegrator`.
> 

Time stepping for SSPRK3 is more involved than RK2. Hence we separated them out. That will make them easier to debug and maintain in the long run. 

> Furthermore - all SSPRK stuff is new. Can we add some tests for it?

SSPRK has been there for a while now, since 2017 or 2018. That has been our default solver for multiphase flows.

amneetb:
This test (which is there is master) uses SSPRK3:

https://github.com/IBAMR/IBAMR/blob/master/tests/multiphase_flow/high_density_droplet.input#L181

include/ibamr/INSHierarchyIntegrator.h
```diff
@@ -203,9 +203,19 @@ class INSHierarchyIntegrator : public IBTK::HierarchyIntegrator
...
     /*!
      * Supply a fluid source/sink distribution.
+     *
+     * @deprecated Use registerVelocityDivergenceFunction() instead.
      */
     void registerFluidSourceFunction(SAMRAI::tbox::Pointer<IBTK::CartGridFunction> Q_fcn);
 
+    /*!
+     * Add a CartGridFunction that computes \f$ \nabla \cdot \mathbf{u} \f$.
+     * This is used to solve the low Mach equation \f$ \nabla \cdot \mathbf{u} = \mathrm{R.H.S} \f$.
```
boyceg, line 213: Let's avoid comments that are specific to subclasses in this base class.
```suggestion
```

include/ibamr/INSHierarchyIntegrator.h
```diff
@@ -223,9 +233,16 @@ class INSHierarchyIntegrator : public IBTK::HierarchyIntegrator
 
     /*!
      * Return a pointer to the source strength variable.
+     *
+     * @deprecated Use getVelocityDivergenceVariable() instead.
      */
     SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > getFluidSourceVariable() const;
 
+    /*!
+     * Return a pointer to the velocity divergence variable.
```
boyceg, line 242: With approximate projection methods, you don't actually get the value of the divergence that you specify --- so I think we should try to be careful not to imply that this is "the divergence of the velocity" (which is something you get from post-processing).

This seems too wordy, but maybe something like:
```suggestion
     * Return a pointer to the variable that specifies the divergence of the velocity.
```

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,301 @@
...
+
+namespace IBAMR
+{
+/*!
+ * \brief Class MassIntegrator is an abstract class which integrates
+ * the density field
+ *
+ *  \f$ \frac{\partial \rho}{\partial t} + \nabla \cdot (\rho u) = S(x,t) \f$
+ *
+ * and computes the conservative form of the convective operator
```
boyceg, line 67: I think one of the points is that you need to make these all consistent to get good results --- maybe could add that in the comments (otherwise it seems a little odd that these other terms are being computed in a solver for the mass).

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,301 @@
...
+class BasePatchHierarchy;
+} // namespace hier
+} // namespace SAMRAI
+
+/////////////////////////////// CLASS DEFINITION /////////////////////////////
+
+namespace IBAMR
+{
+/*!
+ * \brief Class MassIntegrator is an abstract class which integrates
```
boyceg, line 62: It seems like this duplicates functionality provided by the advection-diffusion solvers. Why is this additional class needed?

include/ibamr/INSVCStaggeredConservativeMassMomentumSSPRKIntegrator.h
```diff
@@ -0,0 +1,128 @@
...
+ * A generic framework for design of interface capturing schemes for multi-fluid flows</A>
+ *
+ * \note This class is specialized in that it computes a conservative discretization of the form
+ * \f$N = \nabla \cdot (u \rho u)\f$, where the density \f$\rho\f$ can vary in space and time.
+ * This operator is to be used in conjuction with the conservative form of the variable coefficient
+ * Navier-Stokes equations, which will produce better results for high density ratio flows.
+ *
+ * \see INSVCStaggeredHierarchyIntegrator
+ */
+class INSVCStaggeredConservativeMassMomentumSSPRKIntegrator : public INSVCStaggeredConservativeMassMomentumRKIntegrator
```
boyceg, line 73: It seems odd that this is a subclass of `INSVCStaggeredConservativeMassMomentumRKIntegrator` instead of `MassIntegrator`. Are the non-SSP time steppers actually useful?

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
```diff
@@ -570,7 +570,6 @@ class INSVCStaggeredHierarchyIntegrator : public INSHierarchyIntegrator
     SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > d_P_var;
     SAMRAI::tbox::Pointer<SAMRAI::pdat::SideVariable<NDIM, double> > d_F_var;
     SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > d_F_cc_var;
-    SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > d_Q_var;
```
boyceg, line 573: I would consider leaving `Q` here --- its role is very similar to `F`.

boyceg:
> @boyceg do you have strong feelings either way about `MassIntegrator`? I'd like to move this along and I don't think we need to make fundamental changes.

At this point I guess I am not sure I understand how the `MassIntegrator` differs from the existing advection-diffusion solvers.

boyceg:
> > I read through the whole thing and I think its fine - I don't see anything wrong with moving some of the details pertaining to the conservation of mass to a new base class.
> > Can we skip making `INSVCStaggeredConservativeMassMomentumSSPRKIntegrator` a separate class? It just implements some more RK schemes - we should have that in `INSVCStaggeredConservativeMassMomentumRKIntegrator`.
> 
> Time stepping for SSPRK3 is more involved than RK2. Hence we separated them out. That will make them easier to debug and maintain in the long run.
> 
> > Furthermore - all SSPRK stuff is new. Can we add some tests for it?
> 
> SSPRK has been there for a while now, since 2017 or 2018. That has been our default solver for multiphase flows.

If you don't need the non-SSP RK time steppers, could we remove them?

(BTW --- I think that FE is considered to be an SSP RK time stepping scheme.)

amneetb:
> > @boyceg do you have strong feelings either way about `MassIntegrator`? I'd like to move this along and I don't think we need to make fundamental changes.
> 
> At this point I guess I am not sure I understand how the `MassIntegrator` differs from the existing advection-diffusion solvers.

I think we had discussed this before. `MassIntegrator` is an intermediate or a lightweight integrator that is used within one time step to get a consistent mass flux $m_\rho = \rho \mathbf{u}$ for advecting either momentum $\mathbf{u}$ through $\nabla \cdot (\rho \mathbf{u} \otimes \mathbf{u})  = \nabla \cdot (m_\rho \otimes \mathbf{u} )$ or a scalar $\phi$ through  $\nabla \cdot (\rho \mathbf{u} \phi)  = \nabla \cdot (m_\rho  \phi)$. It does not maintain any hierarchy data, child or parent integrator, as done by advection-diffusion hierarchy integrator.

amneetb:
> > > I read through the whole thing and I think its fine - I don't see anything wrong with moving some of the details pertaining to the conservation of mass to a new base class.
> > > Can we skip making `INSVCStaggeredConservativeMassMomentumSSPRKIntegrator` a separate class? It just implements some more RK schemes - we should have that in `INSVCStaggeredConservativeMassMomentumRKIntegrator`.
> > 
> > 
> > Time stepping for SSPRK3 is more involved than RK2. Hence we separated them out. That will make them easier to debug and maintain in the long run.
> > > Furthermore - all SSPRK stuff is new. Can we add some tests for it?
> > 
> > 
> > SSPRK has been there for a while now, since 2017 or 2018. That has been our default solver for multiphase flows.
> 
> If you don't need the non-SSP RK time steppers, could we remove them?

We need non-SSP RK time steppers. We use them in #1626 and onwards. 

> 
> (BTW --- I think that FE is considered to be an SSP RK time stepping scheme.)

include/ibamr/INSVCStaggeredConservativeMassMomentumSSPRKIntegrator.h
amneetb: `INSVCStaggeredConservativeMassMomentumRKIntegrator` is derived from `MassIntegrator`. SSP-RK integrator is a bit more specialized than its parent RK integrator class.

boyceg:
> > > @boyceg do you have strong feelings either way about `MassIntegrator`? I'd like to move this along and I don't think we need to make fundamental changes.
> > 
> > 
> > At this point I guess I am not sure I understand how the `MassIntegrator` differs from the existing advection-diffusion solvers.
> 
> I think we had discussed this before. `MassIntegrator` is an intermediate or a lightweight integrator that is used within one time step to get a consistent mass flux mρ=ρu for advecting either momentum u through ∇⋅(ρu⊗u)=∇⋅(mρ⊗u) or a scalar ϕ through ∇⋅(ρuϕ)=∇⋅(mρϕ). It does not maintain any hierarchy data, child or parent integrator, as done by advection-diffusion hierarchy integrator.

I suggest that this be named something different than "integrator". And the discussion for how this differs from the advection-diffusion solvers should go into the class comments. Otherwise every time I see this class I am going to ask the same questions. 😁 

That said --- can you remind me why you cannot use (a subclass of) the advection-diffusion solver here? Ultimately are the implementations of this ABC basically providing a full (not light-weight) implementation?

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
amneetb: It is there. It is moved a bit down to line 605 with a comment, as to what `Q` represents. I think we were getting confused with `Q` of advection-diffusion integrator, so it separated it out from the block of variables.  

https://github.com/IBAMR/IBAMR/blob/phase_change_patch_1/include/ibamr/INSVCStaggeredHierarchyIntegrator.h#L605-L607

include/ibamr/INSVCStaggeredConservativeMassMomentumSSPRKIntegrator.h
boyceg: I may be missing something, but it looks like it just reimplements `integrate`. Do we really need to have all of these different classes just to provide these simple RK schemes?

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
boyceg: Yes, I saw it was moved --- I was just suggesting to keep it with F because it is the "same kind of variable" as F.

Raamkrishnan:
That said --- can you remind me why you cannot use (a subclass of) the advection-diffusion solver here? Ultimately are the implementations of this ABC basically providing a full (not light-weight) implementation?

The main reason is we want the convective flux in the "preprocessIntegrateHierarchy" and "integrateHierarchy" of the HierarchyIntegrator (It could be INSVC or PhaseChangeHI). We solve mass conservation equation using the current (nth level) density and velocity values in the "preprocessIntegrateHierarchy". Similarly, in the "integrateHierarchy", we solve the mass equation and get the convective flux using new (n+1 level) density and velocity values. 

We want to solve the mass equation within a timestep (within a "advanceHierarchy").

boyceg:
> That said --- can you remind me why you cannot use (a subclass of) the advection-diffusion solver here? Ultimately are the implementations of this ABC basically providing a full (not light-weight) implementation?
> 
> The main reason is we want the convective flux in the "preprocessIntegrateHierarchy" and "integrateHierarchy" of the HierarchyIntegrator (It could be INSVC or PhaseChangeHI). We solve mass conservation equation to get the convective flux using the current (nth level) density and velocity values in the "preprocessIntegrateHierarchy". Similarly, in the "integrateHierarchy", we solve the mass equation and get the convective flux using new (n+1 level) density and velocity values.
> 
> We want to solve the mass equation within a timestep (within a "advanceHierarchy").

Which of these things can you not do with the existing advection-diffusion solvers?

amneetb:
> > That said --- can you remind me why you cannot use (a subclass of) the advection-diffusion solver here? Ultimately are the implementations of this ABC basically providing a full (not light-weight) implementation?
> > The main reason is we want the convective flux in the "preprocessIntegrateHierarchy" and "integrateHierarchy" of the HierarchyIntegrator (It could be INSVC or PhaseChangeHI). We solve mass conservation equation to get the convective flux using the current (nth level) density and velocity values in the "preprocessIntegrateHierarchy". Similarly, in the "integrateHierarchy", we solve the mass equation and get the convective flux using new (n+1 level) density and velocity values.
> > We want to solve the mass equation within a timestep (within a "advanceHierarchy").
> 
> Which of these things can you not do with the existing advection-diffusion solvers?

We can't do that this with existing advection-diffusion (ADHI) solvers. There are two reasons: (i) For INS we need to integrator mass balance PDE on cell faces; ADHI does it on cell centers. (ii) ADHI aims to update Q. We do not care about updating Q, but rather the advective flux that is used in updating Q (= rho).

amneetb:
> > > That said --- can you remind me why you cannot use (a subclass of) the advection-diffusion solver here? Ultimately are the implementations of this ABC basically providing a full (not light-weight) implementation?
> > > The main reason is we want the convective flux in the "preprocessIntegrateHierarchy" and "integrateHierarchy" of the HierarchyIntegrator (It could be INSVC or PhaseChangeHI). We solve mass conservation equation to get the convective flux using the current (nth level) density and velocity values in the "preprocessIntegrateHierarchy". Similarly, in the "integrateHierarchy", we solve the mass equation and get the convective flux using new (n+1 level) density and velocity values.
> > > We want to solve the mass equation within a timestep (within a "advanceHierarchy").
> > 
> > 
> > Which of these things can you not do with the existing advection-diffusion solvers?
> 
> We can't do that this with existing advection-diffusion (ADHI) solvers. There are two reasons: (i) For INS we need to integrator mass balance PDE on cell faces; ADHI does it on cell centers. (ii) ADHI aims to update Q. We do not care about updating Q, but rather the advective flux that is used in updating Q (= rho).

And the (iii) is efficiency. ADHI has a massive memory footprint. It maintains so much related to hierarchy, temporary variables, etc.

amneetb:
BTW, there are many "pseudo time stepping PDEs" that are needed within a time step, that are not worth doing it via HI. Another example is propagating interface information to neighboring grid cells ([see](https://math.berkeley.edu/~sethian/2006/Papers/sethian.adalsteinsson.extension.pdf)), or level set reinitialization equation.

include/ibamr/INSVCStaggeredConservativeMassMomentumSSPRKIntegrator.h
amneetb: "simple" is an understatement ;-). To maintain stability there needs to be an orchestrated dance between time stepping of INSVCHI and these mass integrators. Even a slight mismatch at the time level of these two integrators leads to instability for high density ratio cases.

Synchronizing SSP-RK with INSVC is more involved than RK2 with INSVC.

boyceg: SSP-RK2 and SSP-RK3 are essentially linear combinations of forward Euler steps. (This is one reason why it seems like you should be able to put forward Euler in this class along with the higher-order time steppers.) If the time stepping is substantially different from that, I think it would be helpful to include a few comments describing the sense in which the time stepping is using SSP-RK.

Also, there are different SSP-RK schemes --- can we document which ones are implemented?

boyceg:
> > > @boyceg do you have strong feelings either way about `MassIntegrator`? I'd like to move this along and I don't think we need to make fundamental changes.
> > 
> > 
> > At this point I guess I am not sure I understand how the `MassIntegrator` differs from the existing advection-diffusion solvers.
> 
> I think we had discussed this before. `MassIntegrator` is an intermediate or a lightweight integrator that is used within one time step to get a consistent mass flux mρ=ρu for advecting either momentum u through ∇⋅(ρu⊗u)=∇⋅(mρ⊗u) or a scalar ϕ through ∇⋅(ρuϕ)=∇⋅(mρϕ). It does not maintain any hierarchy data, child or parent integrator, as done by advection-diffusion hierarchy integrator.

I would suggest putting some of this discussion in the class comments. What is the intended use / how does it differ from solvers in the library? The current class documentation does not make it clear (at least to me!) why this class exists along with the advection-diffusion solver class --- hence all of my evidently inane questions.

src/utilities/MassIntegrator.cpp
```diff
@@ -0,0 +1,309 @@
...
+            "MassIntegrator::"
+            "MassIntegrator():\n"
+            << "  unsupported density convective limiter: "
+            << IBAMR::enum_to_string<LimiterType>(d_density_convective_limiter) << " \n"
+            << "  valid choices are: PPM, CUI\n");
+    }
+
+    switch (d_density_time_stepping_type)
+    {
+    case FORWARD_EULER:
```
boyceg, line 146: Should these time stepping details go into the subclasses? It seems like the intent of the design is to let different implementations provide different time stepping schemes.

include/ibamr/INSVCStaggeredConservativeMassMomentumSSPRKIntegrator.h
boyceg: (BTW, documentation via pointing to papers seems fine!)

amneetb:
> A few final questions/thoughts:
> 
> 1. Is `MassIntegrator` intended only as an ABC for conservative VC INS integrators, or will you be using it for other kinds of solvers later on? I think that a lot of my confusion came from `MassIntegrator` sounding like it is providing a very general kind of interface. If it is intended to be used only as a base class for conservative and/or staggered-grid VC-INS solvers, it might be useful to encode that into the name --- maybe something like `INSVCStaggeredMassMomentumIntegrator` or `INSVCStaggeredConservativeMassMomentumIntegrator`. Or if it is to be used for different kinds of solvers, but always with staggered grid layouts, maybe it could be `StaggeredMassMomemtumIntegrator` or `StaggeredConservativeMassMomentumIntegrator` ?
> 2. It might make sense for most of the description about how `MassIntegrator` should be used to go in the class documentation for  `INSStaggeredConservativeHierarchyIntegrator` --- in which case, could point users to that class in Doxygen using `@see INSStaggeredConservativeHierarchyIntegrator`.
> 3. If the `MassIntegrator` implementation is specific for VC-INS solvers, I'd suggest putting the implementation into `src/navier_stokes` instead of `src/utilities`.

What is `ABC`? I'd meant to ask that earlier, but missed it.

amneetb:
> All good stuff to put somewhere in the code documentation too. 😁

Yep. We will document these points properly in the header file of `MassIntegrator`.

src/utilities/MassIntegrator.cpp
boyceg: FWIW, this is also the reason why I commented on the SSP implementation being derived from the non-SSP implementation instead of the ABC. I wonder if it would be better to have an abstract RK base class that only provide forward Euler, and then have two different concrete implementations (one that adds on the explicit midpoint rule, the other that adds on the SSP RK schemes).

boyceg:
Oh, yeah, one more thing --- I know that some of the algorithmic choices give bad results --- that is one of the points of your paper with Nishant on multiphase flow solvers. Are there any of the VC solvers that we are maintaining only to demonstrate that they don't work well? 😁

src/utilities/MassIntegrator.cpp
amneetb: One technicality is the velocity field used in the intermediate stages of RK and SSP-RK. While the RK is simple $0.5*(u^{n+1} + u^{n})$, SSP-RK employs a more complicated interpolation of $u^{n+1}, u^{n}, u^{n-1}$ to achieve full accuracy. Hence we just separated them out.

boyceg:
> > Are there any of the VC solvers that we are maintaining only to demonstrate that they don't work well? 😁
> 
> `INSVCNonConservativeHI` is what most of the literature had been doing. This HI is simple and works well for low density ratio flows. It does not solve any redundant mass equation. Most multiphase folks have started doing `INSVCConservativeHI` in one way or the other. Some have implemented exactly what we have done in the JCP 2019 paper.

Could we deprecate the non-conservative version, or do we want to continue to provide it?

amneetb:
> > > Are there any of the VC solvers that we are maintaining only to demonstrate that they don't work well? 😁
> > 
> > 
> > `INSVCNonConservativeHI` is what most of the literature had been doing. This HI is simple and works well for low density ratio flows. It does not solve any redundant mass equation. Most multiphase folks have started doing `INSVCConservativeHI` in one way or the other. Some have implemented exactly what we have done in the JCP 2019 paper.
> 
> Could we deprecate the non-conservative version, or do we want to continue to provide it?

I also debated about it. For efficiency reasons, one can use the non conservative integrator. It is similar to how we have kept the ins collocated HI. The two have their `historical` merits.

src/utilities/MassIntegrator.cpp
amneetb: The interpolation scheme for SSP-RK utilizing $u^{n-1}, u^{n}, u^{n+1}$ is explained in our JCP paper:
https://www.sciencedirect.com/science/article/pii/S0021999119302256

In particular see Eqs. (43)-(46) and 3rd order (a.k.a "full") accuracy in $\rho$ by retaining $u^{n-1}$ in Fig. 15.

Raamkrishnan:
@boyceg We have incorporated your suggestions. Please take a look.

src/navier_stokes/INSVCStaggeredConservativeHierarchyIntegrator.cpp
```diff
@@ -133,18 +133,17 @@ INSVCStaggeredConservativeHierarchyIntegrator::INSVCStaggeredConservativeHierarc
...
-            d_rho_p_integrator = new INSVCStaggeredConservativeMassMomentumRKIntegrator(
-                "INSVCStaggeredConservativeHierarchyIntegrator::MassMomentumRKIntegrator",
-                input_db->getDatabase("mass_momentum_integrator_db"));
-    }
+    Pointer<Database> mass_momentum_db = input_db->getDatabase("mass_momentum_integrator_db");
+    if (mass_momentum_db->getString("density_time_stepping_type") == "SSPRK3" ||
+        mass_momentum_db->getString("density_time_stepping_type") == "SSPRK2")
+        d_rho_p_integrator = new INSVCStaggeredConservativeMassMomentumSSPRKIntegrator(
+            "INSVCStaggeredConservativeHierarchyIntegrator::MassMomentumSSPRKIntegrator",
+            input_db->getDatabase("mass_momentum_integrator_db"));
```
amneetb, line 141: @Raamkrishnan : You can replace ` input_db->getDatabase("mass_momentum_integrator_db")` with `mass_momentum_db` here

src/navier_stokes/INSVCStaggeredConservativeHierarchyIntegrator.cpp
```diff
@@ -133,18 +133,17 @@ INSVCStaggeredConservativeHierarchyIntegrator::INSVCStaggeredConservativeHierarc
...
+    else
+        d_rho_p_integrator = new INSVCStaggeredConservativeMassMomentumRKIntegrator(
+            "INSVCStaggeredConservativeHierarchyIntegrator::MassMomentumRKIntegrator",
+            input_db->getDatabase("mass_momentum_integrator_db"));
```
amneetb, line 145: and here.

src/utilities/MassIntegrator.cpp
boyceg: Hrmmm. I don't remember this part of the paper. (Obviously. 😀) If the time stepping for $u$ is globally second-order, I am not sure if there is any benefit to overall order-of-accuracy to performing a local third-order reconstruction.

(The way I would think about it is that you cannot take the second-order accurate values $u^{n-1}$, $u^{n}$, and $u^{n+1}$ and perform a quadratic interpolation to recover a third-order accurate approximation to $u$ at $t \in [t^{n-1}, t^{n+1}]$. The reconstruction is still just second-order accurate.)

What happens if you omit using $u^{n-1}$ here? Is it important to use SSP-RK3 vs. SSP-RK2? Since everything else is solved using 2nd-order time stepping, I don't think using higher-order for the density equation should impact the order of accuracy, although these choices may impact the error constant. Omitting $u^{n-1}$ could also improve stability, and I think that the stability properties of SSP-RK2 and the 3-stage SSP-RK3 schemes are the same. (On the other hand, I think that 4-stage SSP-RK3 is more stable than 2-stage SSP-RK2.)

Typically, the main reason to use multi-step is to _avoid_ doing multiple stages (i.e., cycles). The usual tradeoff is typically in terms of stability. If I am doing multi-step for any part of the problem, I would try to do it for the whole thing, so that I only need to do a single linear solve per time step.

Note that there are SSP multistep methods --- see, e.g., https://arxiv.org/abs/1504.04107 and references therein. Maybe we should be using these instead of AB2 in the constant-coefficient solvers... 😁

amneetb: I think we were targeting accuracy of density, velocity, and pressure altogether. This particular interpolation scheme gave the maximum accuracy (and stability for a very high density ratio case) with ssp-rk3. Perhaps for practical problems, it does not matter. These days we are just using RK2 for density and finding it equally robust. Anyways, it is there as a valid option to try.

boyceg: SSK-RK2 is simply the explicit trapezoidal rule --- if it doesn't make any difference if you use SSP-RK2 or SSP-RK3, then do we need both the RK and SSP-RK classes?

boyceg: Note also that omitting the lagged value of the velocity will save some memory, which you have mentioned concerns about here. 😁

boyceg: BTW, to be clear, my intuition here is that SSP-RK2 without multistep extrapolation will be more stable than SSP-RK3 with multistep extrapolation.

amneetb: Here we are concerned with a different type of instability, which is mass-momentum inconsistency. The integration of rho with any sort of reasonable time stepping scheme is fine. The main point is using the same mass flux in the convective operator of the momentum equation. In this regard, rk, ssp-rk2/3 are all reasonable choices, because we utilize the same mass flux in the two advective operators.

I would like to keep ssp-rk2/3 because there was this paper that debated one needs to the same integrator in momentum eq. as in mass equation. Our implementation clearly shows that this is not true!

amneetb: IIRC ssp-rk3 was more stable than ssp-rk2 on some extremely high density ratio cases. @nishantnangia may recall better. I think this was the reason we stuck with ssp-rk3 in the paper (and even afterwards).

boyceg:
> Note that the concept of `MassIntegrator` can/is also be used for scalar variables. We have extended this to the enthalpy equation, which is solved with a cell center layout. The core concept behind `MassIntegrator` is to obtain a consistent mass flux mρ=ρu for stabilizing high density ratio flows. It also ensures a consistency condition that:
> 
> 1. When there are no viscous, pressure forces or any other body forces in the domain, the momentum equation becomes discretely equal to the mass equation.
> 2. Under isothermal/isenthalpic conditions, with no additional heat source, the energy/enthalpy equation reverts to mass balance equation.
> 
> To achieve these conditions, the time integrator scheme of `MassIntegrator` is tightly coupled to how INSVCHI or PhaseChangeHI integrate their respective variables. This is what the `MassIntegrator` class is doing. It is strongly coupled to the INSVC or PhaseChangeHIs. Note that when flow is incompressible, one does _not_ need to solve mass balance equation. We are solving a redundant equation in order to achieve consistency and stability.

Just to make sure I understand, in all of these cases, implementations of this class are doing mass density solves and then providing fluxes to the other dynamic equations? Or are subclasses eventually implementing time steppers for different dynamic equations?

If the class is intended to be tightly coupled to `INSVC`, I would like for `INSVC` to appear in the class name. (Since these names are getting really long, I think it could make sense to start partitioning functionality into different namespaces if you prefer. `INSVC::MassIntegrator` would make it clear that this thing goes with the INSVC solvers without requiring long names everywhere.)

I'd also like to think about whether there is a naming convention for the "lightweight" integrator concept that could being introduced here, to distinguish these from the "heavyweight" hierarchy integrators. (The name `MassIntegrator` sounds like something that might be a stand-alone solver, which I understand this is not.) Is it something like a "single time step integrator" --- possibly `STSIntegrator` --- or a "single time step flux integrator" --- `STSFluxIntegrator`? Other possibilities include `LightweightIntegrator` or `TemporaryIntegrator`, although I think it is less clear what that means. Perhaps `StatelessIntegrator`, `STSStrategy`, `SubIntegrator`, `IntegratorFluxEvaluator`, or `IntegratorFluxStrategy`?

(All of this single time step stuff is based on my possibly incorrect understanding of how this class is ultimately used --- I understand that it not keeping track of state but rather being used by `INSVCHI` to determine fluxes within a single time step.)

amneetb:
Correct these are tightly coupled to INSVCHI (later on with PhaseChangeHI) to provide mass fluxes. For incompressible flows, density is not a state variable but rather a phase property. So the rho that comes out of MassIntegrator is not a state variable but rather an intermediate quantity to stabilize the numerical scheme.

src/navier_stokes/INSVCStaggeredConservativeMassMomentumSSPRKIntegrator.cpp
```diff
@@ -0,0 +1,531 @@
...
+    const double new_time = d_current_time + dt;
+    std::vector<InterpolationTransactionComponent> new_transaction_comps(1);
+    new_transaction_comps[0] = InterpolationTransactionComponent(d_rho_scratch_idx,
+                                                                 d_rho_new_idx,
+                                                                 "CONSERVATIVE_LINEAR_REFINE",
+                                                                 false,
+                                                                 "CONSERVATIVE_COARSEN",
+                                                                 d_density_bdry_extrap_type,
+                                                                 false,
+                                                                 d_rho_bc_coefs);
```
boyceg, line 488: FWIW, notice that these can be initialized much more compactly with C++11 --- I think this should work:
```suggestion
    std::vector<InterpolationTransactionComponent> new_transaction_comps({{d_rho_scratch_idx,
                                                                                                                                      d_rho_new_idx,
                                                                                                                                      "CONSERVATIVE_LINEAR_REFINE",
                                                                                                                                      false,
                                                                                                                                      "CONSERVATIVE_COARSEN",
                                                                                                                                      d_density_bdry_extrap_type,
                                                                                                                                      false,
                                                                                                                                      d_rho_bc_coefs}});
```

boyceg: (Much might be a little bit of an exaggeration. 🤷‍♂️ )

amneetb:
I think I like STSMassFluxIntegrator. That can be derived from STSIntegrator (if we want that kind of class hierarchy)

src/navier_stokes/INSVCStaggeredConservativeMassMomentumSSPRKIntegrator.cpp
```diff
@@ -0,0 +1,531 @@
...
+            d_hier_sc_data_ops->axpy(d_V_composite_idx, w2, d_V_new_idx, d_V_composite_idx, /*interior_only*/ true);
+            std::vector<InterpolationTransactionComponent> v_update_transaction_comps(1);
+            v_update_transaction_comps[0] = InterpolationTransactionComponent(d_V_scratch_idx,
+                                                                              d_V_composite_idx,
+                                                                              "CONSERVATIVE_LINEAR_REFINE",
+                                                                              false,
+                                                                              "CONSERVATIVE_COARSEN",
+                                                                              d_velocity_bdry_extrap_type,
+                                                                              false,
+                                                                              d_u_bc_coefs);
```
boyceg, line 311:
```suggestion
            std::vector<InterpolationTransactionComponent> v_update_transaction_comps({{
                                                                              d_V_composite_idx,
                                                                              "CONSERVATIVE_LINEAR_REFINE",
                                                                              false,
                                                                              "CONSERVATIVE_COARSEN",
                                                                              d_velocity_bdry_extrap_type,
                                                                              false,
                                                                              d_u_bc_coefs}});
```
(This would need to be clang-formatted.)

src/navier_stokes/INSVCStaggeredConservativeMassMomentumSSPRKIntegrator.cpp
```diff
@@ -0,0 +1,531 @@
...
+                                 /*interior_only*/ true);
+    std::vector<InterpolationTransactionComponent> v_transaction_comps(1);
+    v_transaction_comps[0] = InterpolationTransactionComponent(d_V_scratch_idx,
+                                                               d_V_composite_idx,
+                                                               "CONSERVATIVE_LINEAR_REFINE",
+                                                               false,
+                                                               "CONSERVATIVE_COARSEN",
+                                                               d_velocity_bdry_extrap_type,
+                                                               false,
+                                                               d_u_bc_coefs);
```
boyceg, line 216:
```suggestion
    std::vector<InterpolationTransactionComponent> v_transaction_comps({{
                                                               d_V_scratch_idx,
                                                               d_V_composite_idx,
                                                               "CONSERVATIVE_LINEAR_REFINE",
                                                               false,
                                                               "CONSERVATIVE_COARSEN",
                                                               d_velocity_bdry_extrap_type,
                                                               false,
                                                               d_u_bc_coefs}});
```

src/navier_stokes/INSVCStaggeredConservativeMassMomentumSSPRKIntegrator.cpp
```diff
@@ -0,0 +1,531 @@
...
+    // Fill ghost cells for current density
+    std::vector<InterpolationTransactionComponent> rho_transaction_comps(1);
+    rho_transaction_comps[0] = InterpolationTransactionComponent(d_rho_scratch_idx,
+                                                                 d_rho_current_idx,
+                                                                 "CONSERVATIVE_LINEAR_REFINE",
+                                                                 false,
+                                                                 "CONSERVATIVE_COARSEN",
+                                                                 d_density_bdry_extrap_type,
+                                                                 false,
+                                                                 d_rho_bc_coefs);
```
boyceg, line 196:
```suggestion
    std::vector<InterpolationTransactionComponent> rho_transaction_comps({{
                                                                 d_rho_scratch_idx,
                                                                 d_rho_current_idx,
                                                                 "CONSERVATIVE_LINEAR_REFINE",
                                                                 false,
                                                                 "CONSERVATIVE_COARSEN",
                                                                 d_density_bdry_extrap_type,
                                                                 false,
                                                                 d_rho_bc_coefs}});
```

include/ibamr/INSHierarchyIntegrator.h
```diff
@@ -203,9 +203,18 @@ class INSHierarchyIntegrator : public IBTK::HierarchyIntegrator
 
     /*!
      * Supply a fluid source/sink distribution.
+     *
+     * @deprecated Use registerVelocityDivergenceFunction() instead.
      */
     void registerFluidSourceFunction(SAMRAI::tbox::Pointer<IBTK::CartGridFunction> Q_fcn);
 
+    /*!
+     * Add a CartGridFunction that computes \f$ \nabla \cdot \mathbf{u} \f$.
```
boyceg, line 212: For consistency with other function docs, suggest:
```suggestion
     * Supply a CartGridFunction that specifies \f$ \nabla \cdot \mathbf{u} = Q\f$.
```

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,326 @@
...
+
+    /*!
+     * \brief Get the current time interval.
+     */
+    std::pair<double, double> getTimeInterval() const;
+
+    /*!
+     * \brief Get the current time step size.
+     */
+    double getDt() const;
```
boyceg, line 197:
```suggestion
    double getTimeStepSize() const;
```
or...

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,326 @@
...
+
+    /*!
+     * \brief Set the previous time step value between times n - 1 (old) and n (current). This is used during velocity
+     * extrapolation.
+     */
+    void setPreviousTimeStepSize(double dt_prev);
```
boyceg, line 203: ...or
```suggestion
    void setPreviousDt(double dt_prev);
```

include/ibamr/INSVCStaggeredConservativeMassMomentumSSPRKIntegrator.h
boyceg: > "simple" is an understatement ;-). To maintain stability there needs to be an orchestrated dance between time stepping of INSVCHI and these mass integrators. Even a slight mismatch at the time level of these two integrators leads to instability for high density ratio cases.
> 
> Synchronizing SSP-RK with INSVC is more involved than RK2 with INSVC.

It looks like the implementations of `integrate` only differ in 2 or 3 places --- are you sure it wouldn't be less maintenance to have some switch statements that pick the SSP vs. non-SSP code paths where they differ?

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,326 @@
...
+ * and computes the conservative form of the convective operator
+ * \f$ \nabla \cdot (\rho \mathbf{u} \otimes \mathbf{u})\f$ for the momentum and
+ * \f$ \nabla \cdot (\rho \mathbf{u} Q)\f$ for the energy equation where \f$ Q = C_\textrm{p} T\f$
+ * or \f$ Q = h\f$ based on the form of the energy equation.
+ *
+ * \note MassIntegrator is an intermediate or a lightweight integrator that is used within one time step to
+ * get a consistent mass flux \f$ m_\rho = \rho \mathbf{u} \f$ for advecting either momentum \f$ \mathbf{u} \f$
+ * through \f$ \nabla \cdot (\rho \mathbf{u} \otimes \mathbf{u})  = \nabla \cdot (m_\rho \otimes \mathbf{u} ) \f$
+ * or a scalar \f$ Q \f$ through  \f$ \nabla \cdot (\rho \mathbf{u} Q)  = \nabla \cdot (m_\rho  Q) \f$. It does
+ * not maintain any hierarchy data, child or parent integrator, as done by advection-diffusion hierarchy integrator.
```
boyceg, line 76:
```suggestion
 * not maintain any hierarchy data, child or parent integrator, as done in implementations of HierarchyIntegrator.
```

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,326 @@
...
+ * Here, we are solving a redundant mass balance equation through MassIntegrator class in order to achieve consistency
+ * and stability.
+ *
+ * Reference
+ * Nangia et. al, <A HREF="https://www.sciencedirect.com/science/article/pii/S0021999119302256">
+ * A robust incompressible Navier-Stokes solver for high density ratio multiphase flows </A>
+ *
+ * Thirumalaisamy and Bhalla, <A HREF="https://arxiv.org/abs/2301.06256">
+ * A low Mach enthalpy method to model non-isothermal gas-liquid-solid flows with melting and solidification </A>
+ *
```
boyceg, line 98:
```suggestion
```

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,326 @@
...
+    /*!
+     * \name General operator functionality.
+     */
+    //\{
+
+    /*!
+     * \brief Compute hierarchy dependent data required for time integrating variables.
+     */
+    virtual void
+    initializeTimeIntegrator(SAMRAI::tbox::Pointer<SAMRAI::hier::BasePatchHierarchy<NDIM> > base_hierarchy) = 0;
```
boyceg, line 127: If we go with "single time step integrator", suggest naming these:
```suggestion
    initializeSTSIntegrator(SAMRAI::tbox::Pointer<SAMRAI::hier::BasePatchHierarchy<NDIM> > base_hierarchy) = 0;
```

boyceg: (I won't mark the others, but the same comment applies below.)

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,326 @@
...
+     */
+    virtual ~MassIntegrator() = default;
+
+    /*!
+     * \brief Integrate density and compute the convective operator of the momentum and/or energy.
+     */
+    virtual void integrate(double dt) = 0;
+
+    /*!
+     * \name General operator functionality.
```
boyceg, line 119: Should this be:
```suggestion
     * \name General time stepping functionality.
```
?

include/ibamr/MassIntegrator.h
```diff
@@ -0,0 +1,326 @@
...
+
+    /*!
+     * \brief Destructor.
+     */
+    virtual ~MassIntegrator() = default;
+
+    /*!
+     * \brief Integrate density and compute the convective operator of the momentum and/or energy.
+     */
+    virtual void integrate(double dt) = 0;
```
boyceg, line 116: Should this be part of the general time stepper functionality block?

src/navier_stokes/INSVCStaggeredConservativeMassMomentumRKIntegrator.cpp
```diff
@@ -1106,29 +1124,30 @@ INSVCStaggeredConservativeMassMomentumIntegrator::INSVCStaggeredConservativeMass
     TBOX_ASSERT(d_V_composite_idx >= 0);
 #endif
 
-    const std::string rho_sc_name = "INSVCStaggeredConservativeMassMomentumIntegrator::RHO_SIDE_CENTERED";
-    d_rho_sc_var = var_db->getVariable(rho_sc_name);
-    if (d_rho_sc_var)
+    const std::string rho_sc_name = "INSVCStaggeredConservativeMassMomentumRKIntegrator::RHO_SIDE_CENTERED";
```
boyceg, line 1127: Should these variables use the `d_object_name` to set up the variable name? Otherwise you can only allocate one of these objects at a time.

boyceg: (Same comment applies to the other variable names --- I will not mark those.)

src/navier_stokes/INSVCStaggeredHierarchyIntegrator.cpp
```diff
@@ -1563,6 +1556,8 @@ void
 INSVCStaggeredHierarchyIntegrator::initializeCompositeHierarchyDataSpecialized(const double /*init_data_time*/,
                                                                                const bool initial_time)
 {
+    if (initial_time && d_Q_fcn) return;
```
boyceg, line 1559: Why do you skip the projection in this case? I think the projection should work with div u = Q != 0.

src/navier_stokes/INSVCStaggeredNonConservativeHierarchyIntegrator.cpp
```diff
@@ -971,6 +971,17 @@ INSVCStaggeredNonConservativeHierarchyIntegrator::setupPlotDataSpecialized()
...
 {
+    // Here we want to impose the condition
+    // U := U* - 1/rho * Grad Phi
+    //
+    // Taking the divergence on both sides of the above equation, we get
+    // Div U = Div U* - Div (1/rho * Grad Phi)
+    //
+    //  ===>   - Div (1/rho * Grad Phi) =  Div U - Div U*
+    //
+    // Here, Div U is the externally supplied velocity divergence source
```
boyceg, line 982: I think that this comment would be clearer if you replace `Div U` by `Q`.

src/navier_stokes/INSVCStaggeredNonConservativeHierarchyIntegrator.cpp
```diff
@@ -971,6 +971,17 @@ INSVCStaggeredNonConservativeHierarchyIntegrator::setupPlotDataSpecialized()
...
+    // and U* is velocity after the regridding operation.
```
boyceg, line 983:
```suggestion
    // and U* is velocity after the regridding operation that is to be projected.
```

src/navier_stokes/INSVCStaggeredNonConservativeHierarchyIntegrator.cpp
```diff
@@ -1646,10 +1657,14 @@ INSVCStaggeredNonConservativeHierarchyIntegrator::setupSolverVectors(
             rhs_vec->getComponentDescriptorIndex(0), rhs_vec->getComponentDescriptorIndex(0), d_F_scratch_idx);
     }
 
-    // Account for internal source/sink distributions.
+    // Account for source term of a Div U equation.
```
boyceg, line 1660:
```suggestion
    // Account for a prescribed values of Q = Div U.
```

src/utilities/MassIntegrator.cpp
```diff
@@ -0,0 +1,274 @@
...
+            d_density_bdry_extrap_type = input_db->getString("bdry_extrap_type");
+        }
+        if (input_db->keyExists("density_bdry_extrap_type"))
+        {
+            d_density_bdry_extrap_type = input_db->getString("density_bdry_extrap_type");
+        }
+        if (input_db->keyExists("convective_limiter"))
+        {
+            d_density_convective_limiter =
+                IBAMR::string_to_enum<LimiterType>(input_db->getString("convective_limiter"));
```
boyceg, line 108: Do you need to include the `IBAMR` namespace here?
```suggestion
                string_to_enum<LimiterType>(input_db->getString("convective_limiter"));
```

boyceg (changes requested):
I've finished reviewing everything in this PR. Various requested changes are listed. (It looks like the code documentation has already been updated, which helps a lot in explaining what is going on with these classes --- thanks!!)

One final suggestion would be to consider adding functions to evaluate the interpolated / extrapolated velocities in `integrate()` to simplify their implementations a bit.

src/navier_stokes/INSVCStaggeredHierarchyIntegrator.cpp
amneetb: Yes, it can work in theory. However, what we learnt the hard way is that `Q` comes from from an underlying process that generally involves derivatives of quantities, which are not known at `t= 0`. More concretely, imagine we are doing phase change or a chemical reaction. The volume change `div u`  that needs to captured in `Q` would be some function of gradient of temperature or species concentration. These are not known at initial time. They gradually evolve during the simulation, which builds up `Q` over time. The best thing to do at `t = 0`, is basically not do anything!

boyceg: Sounds reasonable --- can you document this behavior somewhere in the class header, so that it shows up in the Doxygen documentation?

Raamkrishnan:
@boyceg We have made all the required changes. Please take a look.

include/ibamr/STSMassFluxIntegrator.h
```diff
@@ -69,73 +69,72 @@ namespace IBAMR
  * \f$ \nabla \cdot (\rho \mathbf{u} Q)\f$ for the energy equation where \f$ Q = C_\textrm{p} T\f$
  * or \f$ Q = h\f$ based on the form of the energy equation.
  *
- * \note MassIntegrator is an intermediate or a lightweight integrator that is used within one time step to
+ * \note STSMassFluxIntegrator is an intermediate or a lightweight integrator that is used within one time step to
  * get a consistent mass flux \f$ m_\rho = \rho \mathbf{u} \f$ for advecting either momentum \f$ \mathbf{u} \f$
```
amneetb, line 73: STSMassFluxIntegrator is an intermediate or a lightweight integrator that is used within one or a single time step (STS) to ...

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
```diff
@@ -463,6 +463,20 @@ class INSVCStaggeredHierarchyIntegrator : public INSHierarchyIntegrator
 
     /*!
      * Perform data initialization after the entire hierarchy has been constructed.
+     *
+     * \note If \f$ \nabla \cdot \mathbf{u} = Q \f$ is specified, then at the initial time
+     * the velocity projection step is omitted in this routine. This is because \f$ Q \f$ generally involves
+     * derivatives of transported variables, which are not known at the initial time. More concretely,
+     * imagine we are simulating a phase change or combustion process. the volume change effect
+     * \f$ \nabla \cdot \mathbf{u} \f$  that is captured by \f$ Q \f$ would be computed through
+     * gradient of temperature or species concentration. These gradients are not known at the
```
amneetb, line 472: through gradients of temperature or species concentration. 

or 

through the gradient of temperature or species concentration.

amneetb (commented):
A couple more...

include/ibamr/INSVCStaggeredHierarchyIntegrator.h
Raamkrishnan: I have done the changes and pushed it.

Raamkrishnan:
@boyceg If everything looks good, can we merge this?

boyceg:
Let me take a look.

amneetb:
Ping :-)

boyceg:
Sorry for the delay --- classes just started and childcare has been irregular (our preschool is currently closed).

drwells (approved):
I apologize again for letting this sit for so long. I think (with Boyce's suggestions) this is good-to-go.

It would be nice to unify the SSPRK and RK implementations (I commented on this somewhere else too - that should have been done here). However - the SSPRK does a bunch of stuff the normal one doesn't and it uses inheritance so I don't think its a big deal if we keep them separate (though the normal RK really should just be a special case).

amneetb:
@boyceg : If you can approve this then we will merge it and rebase patch 2 against the updated master (we can also rebase patch 2 against patch 1 if that is the preferred workflow)

include/ibamr/STSMassFluxIntegrator.h
```diff
@@ -0,0 +1,326 @@
...
+ * \f$ \nabla \cdot (\rho \mathbf{u} \otimes \mathbf{u})\f$ for the momentum and
+ * \f$ \nabla \cdot (\rho \mathbf{u} Q)\f$ for the energy equation where \f$ Q = C_\textrm{p} T\f$
+ * or \f$ Q = h\f$ based on the form of the energy equation.
+ *
+ * \note STSMassFluxIntegrator is an intermediate or a lightweight integrator that is used within one or a single time
+ * step (STS) to get a consistent mass flux \f$ m_\rho = \rho \mathbf{u} \f$ for advecting either momentum \f$
+ * \mathbf{u} \f$ through \f$ \nabla \cdot (\rho \mathbf{u} \otimes \mathbf{u})  = \nabla \cdot (m_\rho \otimes
+ * \mathbf{u} ) \f$ or a scalar \f$ Q \f$ through  \f$ \nabla \cdot (\rho \mathbf{u} Q)  = \nabla \cdot (m_\rho  Q) \f$.
+ * It does not maintain any hierarchy data, child or parent integrator, as done in implementations of
+ * HierarchyIntegrator..
```
boyceg, line 77:
```suggestion
 * HierarchyIntegrator.
```

boyceg (approved):
Looks great, thanks! Spotted a minor typo in a comment.

Raamkrishnan:
@drwells Once you merge this branch with the master, I will rebase ```phase_change_patch_2``` with the master and start working on it.

PR #1627, opened 2023-07-06
Add phase-change APIs with regression coverage

amneetb:
Add phase-change material-property and laser-source utilities with native regression coverage. Include the AMR, gradient-bound, and restart repairs needed by the examples in #1628.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
Could you rebase now that the actual implementation is merged?

The situation we have, right now, with examples is not great. We have 20 different copies of `GravityForcing.h` and 21 copies of `SetLSProperties.h`. This patch adds eight more copies of `SetLSProperties.h`. I appreciate that these are not all equal files but at the same time they are largely structurally equivalent. I found one which is copied six times and another which is copied seven times.

This patch adds about 11k lines of source code to the examples (there are 38k total new lines) - that brings us up to about 63k lines of source code for examples (and 5.5k of headers), whereas `src/` contains 92k source lines. That is 20% of the total example code we have ever added to IBAMR. This is going to make changing things in the future a lot harder for no appreciable gain in functionality.

I have a few ideas for how we can improve the situation:
1. Extract common cases and add them to the library instead of adding them to each example. We should not need to have any copies of `SetLSProperties.h` in `examples/`. I'm in favor of adding gravity forcing and other common things to the library itself. I like doing this a lot because it will make working with examples and writing new apps in the future easier.
2. Combine sufficiently similar examples with different input files setting up different physics.
3. Put some of these examples somewhere besides IBAMR.

This is a good opportunity to clean some of these things up and make writing examples easier.

All that being said, this is just my opinion on how we should handle examples. Perhaps the better option is to accept examples as-is and not worry about it. @IBAMR/ibamr-admin-team How about we vote on it?

:+1: take examples as-is
🚀 clean things up and add common functionality to the library

amneetb:
@drwells: I hear you, or actually have heard you. There is a PR #1631 that has `LevelSetUtilities` class which will remove these duplicated code files from examples. That PR was created on Aug 5, 2023. Now is Feb 7, 2024.

drwells:
I know, I know - I have a finite amount of time to review these things.

Where in that patch are we deleting things?

amneetb:
I have some examples in my local directory that are removing duplicated files. Have not committed that part. I am waiting for review to finish and interface to stabilize, before putting in the effort modifying examples.

drwells:
It sounds like we should merge that first. How about we close this PR, merge the tests independently, do that PR, and then revisit these?

amneetb:
Yes, sounds like a plan. #1631 is based on #1630. PRs #1630 - #1632 should be reviewed and merged next.

examples/phase_change/ex2/input2d
```diff
@@ -465,6 +468,14 @@ RelaxationLSMethod {
...
 }
 
+LevelSetMassFixer {
+   half_width = NUMBER_OF_INTERFACE_CELLS  // width of the interface in each phase to compute Heaviside integrals
+   max_its = 4                 // maximum no. of Newton's iterations
+   rel_tol = 1e-12             // error between current and target volume
+   correction_interval = 1     // frequency at which mass loss is corrected
+   enable_logging      = TRUE  // prints Netwon residuals
+}
+
```
amneetb, line 478: Are we using this functionality in phase change examples?

Raamkrishnan: I just added the mass loss fix to this example. This is a advection of bubble case. There is no phase change. So I added the mass loss fix to this example. But I am not using mass loss fix here.

Raamkrishnan: I am also adding three more examples: Bubble trap, Stefan melting (two-phase and three-phase). I am working on that. I will also test the restart and push it to this branch. Then this branch is ready for review.

amneetb:
@Raamkrishnan : If all examples are running fine and producing expected results, perhaps we can merge this PR.

@boyceg : do you agree? This PR is adding examples and making some changes to phase change code to support AMR and restart. The majority of the phase change code is already merged .

Raamkrishnan:
I will let you know after I verify the results from the examples. Once that is done, this branch is ready for review.

boyceg:
I haven't looked at the details, but is laser beam forcing a generic thing that could see reuse in multiple applications, or is it pretty particular to specific cases?

amneetb:
@Raamkrishnan : it should be named laser source rather than laser force. You may have changed the name already.

@boyceg : we plan on re-using and extending it (through derivation) in the future.

boyceg:
Another question is if the implementations of `LSLocateInterface` and similar classes are all the same or are they all different? I don't think it makes sense to put these into the main library, but if they are all the same / basically the same, I wonder if it makes sense to pull them into a phase change example / application support library.

boyceg:
> @Raamkrishnan : it should be named laser source rather than laser force. You may have changed the name already.
> 
> @boyceg : we plan on re-using and extending it (through derivation) in the future.

I guess my question isn't whether you all are planning on using it -- presumably you are! 😁  -- but if it is a generic thing that, for instance, is widely used in the literature.

If it is a generic thing that is widely used in the literature, then I think it makes sense to put into the library.

On the other hand, if it is pretty specific to specific applications, I wonder if it would make more sense to put it in a separate application support library.

amneetb:
Yeah, this model is quite common in the literature.

amneetb:
I believe LSLocate stuff was moved into the library to reduce code repeatability in the examples. This was done for multiphase flow examples first and now for the phase change examples. @Raamkrishnan is this right?

boyceg:
> I believe LSLocate stuff was moved into the library to reduce code repeatability in the examples. This was done for multiphase flow examples first and now for the phase change examples. @Raamkrishnan is this right?

There are many copies of this class and the LS initialization class in the examples in this PR.

Again, it doesn't necessarily need to be moved into the main library per se --- it could be part of an application support library that gets compiled with the phase change examples.

amneetb:
I believe in general specifying SDFs analytically is quite rare. So this is kept outside the library. My previous comment seems to be misinformed.

include/ibamr/EnthalpyHierarchyIntegrator.h
```diff
@@ -116,6 +117,13 @@ class EnthalpyHierarchyIntegrator : public PhaseChangeHierarchyIntegrator
      */
     void computeDivergenceVelocitySourceTerm(int Div_U_F_idx, const double new_time) override;
 
+    /*!
+     * \brief Register specific enthalpy variable \f$ h \f$.
+     */
+    virtual void
+    registerSpecificEnthalpyVariable(SAMRAI::tbox::Pointer<SAMRAI::pdat::CellVariable<NDIM, double> > h_var,
+                                     const bool output_h_var = true);
```
boyceg, line 125:
```suggestion
                                    bool output_h_var = true);
```

boyceg: Don't `const` qualify in the interface when passing by value. (This is an implementation detail.)

include/ibamr/EnthalpyHierarchyIntegrator.h
```diff
@@ -198,6 +225,30 @@ class EnthalpyHierarchyIntegrator : public PhaseChangeHierarchyIntegrator
...
+     * Euler time stepping scheme is used to perfom constant extrapolation with fixed 15 iterations and CFL = 0.3.
+     */
+    void extrapolateLiquidFractionToGasRegion(int lf_new_idx);
+
+    /*!
+     * \brief Compute advection velocity \f$ u_{\rm adv} = H(-\phi) n \f$.
+     */
+    void computeAdvectionVelocityForExtrapolation(int u_adv_fc_lf_extrap_current_idx,
+                                                  const double current_time,
+                                                  const double new_time);
```
boyceg, line 241:
```suggestion
                                                  double current_time,
                                                  double new_time);
```

include/ibamr/EnthalpyHierarchyIntegrator.h
```diff
@@ -242,6 +298,34 @@ class EnthalpyHierarchyIntegrator : public PhaseChangeHierarchyIntegrator
...
+    int d_u_adv_cc_lf_extrap_current_idx = IBTK::invalid_index;
+    int d_normal_lf_extrap_current_idx = IBTK::invalid_index;
+    int d_lf_extrap_rhs_scratch_idx = IBTK::invalid_index;
+
+    SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> d_lf_extrap_convective_op_input_db = d_default_convective_op_input_db;
+    SAMRAI::tbox::Pointer<CellConvectiveOperator> d_lf_extrap_convective_op = nullptr;
+    SAMRAI::tbox::Pointer<SAMRAI::solv::SAMRAIVectorReal<NDIM, double> > d_lf_extrap_sol, d_lf_extrap_rhs;
+    bool d_lf_extrap_convective_op_needs_init;
+
+    int d_lf_extrap_max_num_time_steps = 15;
```
boyceg, line 327: Where does 15 come from? Suggest avoiding "magic" numbers.

include/ibamr/PhaseChangeUtilities.h
```diff
@@ -0,0 +1,325 @@
...
+ * \param ctx is the pointer to SetFluidProperties class object.
+ */
+
+void callSetDensityCallbackFunction(int rho_idx,
+                                    SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > rho_var,
+                                    SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps> hier_math_ops,
+                                    const int cycle_num,
+                                    const double time,
+                                    const double current_time,
+                                    const double new_time,
```
boyceg, line 53: As above --- don't `const` qualify in the interface when passing by value.

include/ibamr/PhaseChangeUtilities.h
```diff
@@ -0,0 +1,325 @@
...
+ * \param ctx is the pointer to SetFluidProperties class object.
+ */
+
+void callSetThermalConductivityCallbackFunction(int kappa_idx,
+                                                SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > kappa_var,
+                                                SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps> hier_math_ops,
+                                                const int cycle_num,
+                                                const double time,
+                                                const double current_time,
+                                                const double new_time,
```
boyceg, line 69: As above --- don't `const` qualify in the interface when passing by value.

include/ibamr/PhaseChangeUtilities.h
```diff
@@ -0,0 +1,325 @@
...
+ * \param ctx is the pointer to SetFluidProperties class object.
+ */
+
+void callSetSpecificHeatCallbackFunction(int specific_heat_idx,
+                                         SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > specific_heat_var,
+                                         SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps> hier_math_ops,
+                                         const int cycle_num,
+                                         const double time,
+                                         const double current_time,
+                                         const double new_time,
```
boyceg, line 85: As above --- don't `const` qualify in the interface when passing by value.

include/ibamr/PhaseChangeUtilities.h
```diff
@@ -0,0 +1,325 @@
...
+ * \param ctx is the pointer to SetFluidProperties class object.
+ */
+
+void callSetViscosityCallbackFunction(int mu_idx,
+                                      SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > mu_var,
+                                      SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps> hier_math_ops,
+                                      const int cycle_num,
+                                      const double time,
+                                      const double current_time,
+                                      const double new_time,
```
boyceg, line 101: As above --- don't `const` qualify in the interface when passing by value.

include/ibamr/PhaseChangeUtilities.h
```diff
@@ -0,0 +1,325 @@
...
+                       const double rho_gas,
+                       const double kappa_liquid,
+                       const double kappa_solid,
+                       const double kappa_gas,
+                       const double Cp_liquid,
+                       const double Cp_solid,
+                       const double Cp_gas,
+                       const double mu_liquid,
+                       const double mu_solid,
+                       const double mu_gas);
```
boyceg, line 127: As above --- don't `const` qualify in the interface when passing by value.

include/ibamr/PhaseChangeUtilities.h
```diff
@@ -0,0 +1,325 @@
...
+                       SAMRAI::solv::RobinBcCoefStrategy<NDIM>* lf_bc_coef,
+                       const double rho_liquid,
+                       const double rho_solid,
+                       const double rho_gas,
+                       const double kappa_liquid,
+                       const double kappa_solid,
+                       const double kappa_gas,
+                       const double Cp_liquid,
+                       const double Cp_solid,
+                       const double Cp_gas);
```
boyceg, line 148: As above --- don't `const` qualify in the interface when passing by value.

boyceg: (Also, I think that `const Pointer<T>` does not do what it seems like it should do.)

boyceg: I won't mark the other instances, but the same comment applies below as well...

boyceg:
> I believe in general specifying SDFs analytically is quite rare. So this is kept outside the library. My previous comment seems to be misinformed.

It seems like it is what is done in most of the examples for the initialization routines?

include/ibamr/EnthalpyHierarchyIntegrator.h
Raamkrishnan: Yes. I will correct them in all the phase change classes.

Raamkrishnan:
> I believe LSLocate stuff was moved into the library to reduce code repeatability in the examples. This was done for multiphase flow examples first and now for the phase change examples. @Raamkrishnan is this right?

We have not moved LSLocate stuff where we specify the initial condition for a level set.

include/ibamr/EnthalpyHierarchyIntegrator.h
boyceg: One thing that you can check is that you can remove `const` in the header but not in the implementation, and everything will be fine.

It is totally fine to `const` qualify arguments that are passed by value in an implementation, but it is not really meaningful to do it in the interface, because (AFAIK) C++ does not enforce that the implementation also uses the same `const` qualification. E.g., I think that this is valid code:

```
int foo(const int);

int main() {
    foo(10);
    return 0;
}

int foo(int x) {
    return x + 10;
}
```
and so is this:
```
int foo(int);

int main() {
    foo(10);
    return 0;
}

int foo(const int x) {
    return x + 10;
}
```
and this:
```
class Foo
{
public:
    Foo() = default;
    ~Foo() = default;
    int foo(const int);
};

int main() {
    Foo f;
    f.foo(10);
    return 0;
}

int Foo::foo(int x) {
    return x + 10;
}
```
The main point is that the compiler doesn't care if you `const` qualify an argument that is passed by value in a function prototype --- you can revoke its `const`-ness in the implementation.

Raamkrishnan: Thanks. I will remove the ```const``` keyword in the prototype and keep it in the implementation even though the ```const``` keyword is redundant in both the interface and the implementation when the function has pass-by-value arguments.

amneetb:
@boyceg : I believe @Raamkrishnan has taken care of all the comments, and this branch is ready to be merged with master. Please have a look.

amneetb:
@boyceg : Ping :-)

amneetb:
@boyceg and @drwells : can we please merge this to master?

drwells:
I don't have the resources these days to review a patch this large - this one is up to Boyce.

amneetb:
My bi-monthly reminder to get this merged :-)

drwells:
@boyceg how do you want to handle this one?

boyceg:
I think it needs to be reviewed before it is merged --- there is a lot of stuff in here. I might be able to make some progress during Spring Break (next week).

amneetb:
@boyceg: Ping :-)

src/navier_stokes/fortran/navier_stokes_surface_tension_forcing3d.f.m4
```diff
@@ -144,6 +144,19 @@ c     Local variables.
...
+c     But the \grad \phi is computed in the ghost cell as well which is
+c     used in the curvature calculation later.
+
+      if (U_gcw .eq. 1.d0) then
+            i_normal = 1.d0
+            i_tangential = 0.d0
+      else
+            i_normal = 2.d0
+            i_tangential = 1.d0
+      endif
```
boyceg: This seems dangerous. It probably would be better to error out if the ghost cell widths are not adequate to evaluate everything. In any case, these should not be double-precision values but integers:
```suggestion
      if (U_gcw .eq. 1) then
            i_normal = 1
            i_tangential = 0
      else
            i_normal = 2
            i_tangential = 1
      endif
```

src/navier_stokes/fortran/navier_stokes_surface_tension_forcing2d.f.m4
```diff
@@ -126,6 +126,19 @@ c     Local variables.
...
+c     But the \grad \phi is computed in the ghost cell as well which is
+c     used in the curvature calculation later.
+
+      if (U_gcw .eq. 1.d0) then
+            i_normal = 1.d0
+            i_tangential = 0.d0
+      else
+            i_normal = 2.d0
+            i_tangential = 1.d0
+      endif
```
boyceg: This seems dangerous. It probably would be better to error out if the ghost cell widths are not adequate to evaluate everything. In any case, these should not be double-precision values but integers:
```suggestion
      if (U_gcw .eq. 1) then
            i_normal = 1
            i_tangential = 0
      else
            i_normal = 2
            i_tangential = 1
      endif
```

examples/phase_change/ex1/Postprocessing/liquid_fraction_visit_export.pdf
boyceg, line 1: Do we need this file in the repo?

examples/phase_change/ex2/LSLocateInterface.h
```diff
@@ -0,0 +1,106 @@
...
+                                           SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps> hier_math_ops,
+                                           double time,
+                                           bool initial_time,
+                                           void* ctx);
+
+class LSLocateInterface
+{
+    /*!
+     * \brief class LSLocateInterface is a utility class which is used to identify
+     * the nterface for level set computations
```
boyceg:
```suggestion
     * the interface for level set computations
```
(This typo might show up elsewhere...)

examples/phase_change/ex4/LSLocateInterface.h
```diff
@@ -0,0 +1,102 @@
...
+                                           SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps> hier_math_ops,
+                                           double time,
+                                           bool initial_time,
+                                           void* ctx);
+
+class LSLocateInterface
+{
+    /*!
+     * \brief class LSLocateInterface is a utility class which is used to identify
+     * the nterface for level set computations
```
boyceg:
```suggestion
     * the interface for level set computations
```

examples/phase_change/ex4/LSLocateInterface.h
```diff
@@ -0,0 +1,102 @@
...
+///////////////////////////// INCLUDES ///////////////////////////////////
+#include <ibamr/AdvDiffHierarchyIntegrator.h>
+
+namespace IBTK
+{
+class HierarchyMathOps;
+}
+
+/*
+ * Pre processing call back function to be hooked into IBAMR:LInitStrategy
```
boyceg:
```suggestion
 * Pre processing call back function to be hooked into IBAMR::LSInitStrategy
```
This typo may also show up in other files.

examples/phase_change/ex6/LSLocateInterface.h
```diff
@@ -0,0 +1,108 @@
...
+///////////////////////////// INCLUDES ///////////////////////////////////
+#include <ibamr/AdvDiffHierarchyIntegrator.h>
+
+namespace IBTK
+{
+class HierarchyMathOps;
+}
+
+/*
+ * Pre processing call back function to be hooked into IBAMR:LInitStrategy
```
boyceg:
```suggestion
 * Pre processing call back function to be hooked into IBAMR::LSInitStrategy
```
(I won't tag more of these.)

examples/phase_change/ex6/LSLocateInterface.h
```diff
@@ -0,0 +1,108 @@
...
+                                           SAMRAI::tbox::Pointer<IBTK::HierarchyMathOps> hier_math_ops,
+                                           double time,
+                                           bool initial_time,
+                                           void* ctx);
+
+class LSLocateInterface
+{
+    /*!
+     * \brief class LSLocateInterface is a utility class which is used to identify
+     * the nterface for level set computations
```
boyceg:
```suggestion
     * the interface for level set computations
```
(I won't tag more of these.)

boyceg (changes requested):
Limited review, but the examples seem OK. Let's try to fix the typos.

Do we need to include PDFs in the repo? I would prefer that we did not embed binary files.

In the main library, for the FORTRAN routines, I might be mistaken ,but it looks like `.d0` is being used for integer values — if so, let's switch those to integers.

It also is not clear to me if the FORTRAN routines that processes values in the ghost cell region "do the right thing" if the ghost cell regions are too small. This seems like an easy way to get hard to track down bugs. It would be better, I think, if the code errored out if the ghost cell widths are inadequate. I realize that this is not something that we do in other FORTRAN kernels, but in those other kernels, it will be easier to catch since the bad output is just in the patch interiors and will show up, for instance, in viz files.

examples/phase_change/ex1/Postprocessing/liquid_fraction_visit_export.pdf
Raamkrishnan: Yes. In this file, I explain how to export the interface position with time in visit. We want to keep this so that the users can refer this to do post processing.

examples/phase_change/ex2/LSLocateInterface.h
Raamkrishnan: Yes, I have corrected all of them in phase_change examples.

examples/phase_change/ex4/LSLocateInterface.h
Raamkrishnan: I have corrected it in all the phase change examples.

src/navier_stokes/fortran/navier_stokes_surface_tension_forcing2d.f.m4
Raamkrishnan: I have changed them into integers.

amneetb:
@Raamkrishnan is this ready to be merged with master?

boyceg:
I would like to take a pass at this branch myself to see if I can reduce code duplication before it is merged.

Raamkrishnan:
@amneetb Yes, I have made the requested changes for this branch. Since we no longer use the auto-tools build, I will remove the Makefile and configuration files from all the examples. I will do this today after work.

@boyceg Sure, please take a look and let me know your suggestions. This branch has already gone through substantial review. Since I graduated two years ago, I did not have access to the nodes to run the code and had to request access to my university through @amneetb. If we can review and merge this branch soon, that would be really helpful.

Raamkrishnan:
@boyceg I will try to specify the initial conditions from input file and remove the initial condition and LSLocateInterface files from the examples. Thank you!

boyceg:
@Raamkrishnan: to be clear, I'm not asking for you to do anything on this right now -- I am going to see if I can do some code reorganization myself.

Raamkrishnan:
> I believe in general specifying SDFs analytically is quite rare. So this is kept outside the library. My previous comment seems to be misinformed.

@amneetb @boyceg We had the conversation about the initial condition files here. Please take a look.

Raamkrishnan:
> @Raamkrishnan: to be clear, I'm not asking for you to do anything on this right now -- I am going to see if I can do some code reorganization myself.

Sure, please let me know if I can help. Thank you!

PR #1632, opened 2023-08-08
fix examples/IB/explicit/ex6

boyceg:
Fixes `examples/IB/explicit/ex6`.

(I think that the usual checklist does not apply.)

### DIFF
--- a/examples/IB/explicit/ex6/IBSimpleHierarchyIntegrator.cpp
+++ b/examples/IB/explicit/ex6/IBSimpleHierarchyIntegrator.cpp
@@ -49,29 +49,13 @@ IBSimpleHierarchyIntegrator::preprocessIntegrateHierarchy(const double current_t
 {
     IBHierarchyIntegrator::preprocessIntegrateHierarchy(current_time, new_time, num_cycles);
 
-    const int coarsest_level_num = 0;
-    const int finest_level_num = d_hierarchy->getFinestLevelNumber();
-    Pointer<IBMethod> p_ib_method_ops = d_ib_method_ops;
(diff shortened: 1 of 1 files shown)


drwells (approved):
Yup - the base class is responsible for allocating and deallocating those values.

examples/IB/explicit/ex6/IBSimpleHierarchyIntegrator.cpp
```diff
@@ -170,28 +154,11 @@ IBSimpleHierarchyIntegrator::postprocessIntegrateHierarchy(const double current_
     IBHierarchyIntegrator::postprocessIntegrateHierarchy(
         current_time, new_time, skip_synchronize_new_state_data, num_cycles);
```
drwells, line 155: This is in the wrong place - this class should postprocess its own data before calling the base class function. It's the opposite order as preprocessing (i.e., inner-outer-outer-inner).

drwells (changes requested):
Actually, on further investigation - while these changes make sense the resulting visualization field has zero velocity and zero structure movement - that can't be right.

boyceg:
That is what this one does — forces are zero and so nothing happens:
```
    // For simplicity, set the Lagrangian force density to equal zero.
    ierr = VecZeroEntries(d_F_data->getVec());
    IBTK_CHKERRQ(ierr);
```
I added this example many years ago for someone who wanted a simple IB time stepping algorithm implementation that students could use as a template for experimenting with alternative IB time stepping schemes.

drwells (approved):

PR #1634, opened 2023-08-14
Fix complex_fluids/ex4.

abarret:
Related to #1632. This came up on the listserv. `complex_fluids/ex4` needs a larger penalty parameter. I updated the example to bring it in line with the write-up. I also went ahead and updated some comments and cleaned the input files for the rest of the complex fluid examples.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/examples/complex_fluids/ex0/example.cpp
+++ b/examples/complex_fluids/ex0/example.cpp
@@ -11,21 +11,8 @@
 //
 // ---------------------------------------------------------------------
 
-// Config files
-#include <SAMRAI_config.h>
-
-// Headers for basic PETSc functions
-#include <petscsys.h>
-
-// Headers for basic SAMRAI objects
-#include <BergerRigoutsos.h>
-#include <CartesianGridGeometry.h>
-#include <LoadBalancer.h>
-#include <StandardTagAndInitialize.h>
-
-// Headers for application-specific algorithm/data structure objects
-#include "ibamr/CFINSForcing.h"
 #include <ibamr/AdvDiffSemiImplicitHierarchyIntegrator.h>
+#include <ibamr/CFINSForcing.h>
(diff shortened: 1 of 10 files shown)


examples/complex_fluids/ex0/example.cpp
```diff
@@ -36,10 +23,19 @@
 #include <ibtk/muParserCartGridFunction.h>
 #include <ibtk/muParserRobinBcCoefs.h>
 
+#include <petscsys.h>
+
+#include <BergerRigoutsos.h>
+#include <CartesianGridGeometry.h>
+#include <LoadBalancer.h>
+#include <SAMRAI_config.h>
```
drwells, line 31:
```suggestion
```
We never need to include this header ourselves - might as well get rid of it. Can you just delete this inclusion now here and the other files instead of putting it in the reorganized lists?

examples/complex_fluids/ex0/input2d
```diff
@@ -33,8 +31,8 @@ CONVECTIVE_OP_TYPE = "PPM"                // convective differencing discretizat
 CONVECTIVE_FORM    = "ADVECTIVE"          // how to compute the convective terms
 NORMALIZE_PRESSURE = TRUE                 // whether to explicitly force the pressure to have mean zero
 CFL_MAX            = 0.3                  // maximum CFL number
-DT_MAX = 0.1/N/0.1
-END_TIME = 4.0
+DT_MAX             = 0.01/N               // maximum time step size
```
drwells, line 34: Do we support AMR with complex fluids? If so I'd like to use `NFINEST` here instead of `N`.

drwells (changes requested):
I have two minor comments.

drwells (approved):

PR #1639, opened 2023-08-29
Add a check for ghost cell consistency in the PoissonFACPreconditionerStrategy

abarret:
Currently, if the provided RHS patch index does not have the same ghost cell width as that provided in the constructor (or database), you get a very obtuse error message about `CoarsenSchedule` consistency. This adds a check in `initializeOperatorState()` to print a more helpful message.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/solvers/impls/PoissonFACPreconditionerStrategy.cpp
+++ b/ibtk/src/solvers/impls/PoissonFACPreconditionerStrategy.cpp
@@ -360,6 +360,19 @@ PoissonFACPreconditionerStrategy::initializeOperatorState(const SAMRAIVectorReal
     d_coarsest_ln = solution.getCoarsestLevelNumber();
     d_finest_ln = solution.getFinestLevelNumber();
 
+#if !defined(NDEBUG)
+    // To prevent very obtuse errors later on, we check that the rhs index has the correct ghost cell width.
+    Pointer<PatchDescriptor<NDIM> > pd = VariableDatabase<NDIM>::getDatabase()->getPatchDescriptor();
+    const IntVector<NDIM>& gcw = pd->getPatchDataFactory(rhs_idx)->getGhostCellWidth();
+    if (gcw != d_gcw)
+    {
+        TBOX_ERROR(
+            d_object_name +
+                "::initializeOperatorState(): RHS index does not have the correct ghost width. RHS has ghost width of "
+            << gcw << ". Expecting a ghost cell width of " << d_gcw << ".\n");
+    }
+#endif
+
     // Perform implementation-specific initialization.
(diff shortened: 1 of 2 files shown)


ibtk/src/solvers/impls/PoissonFACPreconditionerStrategy.cpp
```diff
@@ -360,6 +360,19 @@ PoissonFACPreconditionerStrategy::initializeOperatorState(const SAMRAIVectorReal
...
+    Pointer<PatchDescriptor<NDIM> > pd = VariableDatabase<NDIM>::getDatabase()->getPatchDescriptor();
+    const IntVector<NDIM>& gcw = pd->getPatchDataFactory(rhs_idx)->getGhostCellWidth();
+    if (gcw != d_gcw)
+    {
+        TBOX_ERROR(
+            d_object_name +
+                "::initializeOperatorState(): RHS index does not have the correct ghost width. RHS has ghost width of "
+            << gcw << ". Expecting a ghost cell width of " << d_gcw << ".\n");
+    }
+#endif
```
drwells, line 374: What exactly are the errors this causes? I'm not that familiar with this preconditioner - do we end up applying it to the right-hand side at some point?

Can you document this requirement in the header?

abarret: This just a multigrid preconditioner, so we need to restrict the RHS to coarser levels. The general issue is that `FACPreconditioner` calls the strategy implementations that coarsen or refine with copies or clones of the `b` vector, but `PoissonFACPreconditionerStrategy` caches the coarsen and refine schedules with the `d_scratch_idx`, which may have different ghost cell widths than whatever is inside of the `b` vector. In order to use the cached coarsen/refine schedules with different patch indices, the new patch indices need to be consistent with the cached patch indices. 

This is the simplest fix. The ghost cell width of `d_scratch_idx` is set in the constructor. If we require that the patch index in `b` has the same ghost width of `d_scratch_idx`, then all communication schedules will be consistent.

It's possible that we can relax the condition that `b` has the same ghost cell width as `d_scratch_idx`, but this assumption is asserted later in derived classes of `PoissonFACPreconditionerStrategy` (e.g. in `smoothError()` routines). This fix just pulls that assertion into the base class before any `solveSystem()` call occurs.

abarret: I've added some documentation in the header about this requirement.

amneetb: I have also got burnt by these bugs more than once. It is good to throw error sooner than later.

drwells (approved):

PR #1644, opened 2023-10-16
Update our deprecation strategy.

drwells:
It's time to say goodbye to the old-school SAMRAI logging process. Instead we should use `[[deprecated]]` or `[[deprecated("msg")]]` so that deprecation warnings occur at the step in which we can fix them (at compilation).

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


include/ibamr/BrinkmanPenalizationStrategy.h
```diff
@@ -82,9 +82,8 @@ class BrinkmanPenalizationStrategy : public SAMRAI::tbox::Serializable
 
     /*!
      * \brief Set the Brinkman penalization coefficient.
-     *
-     * @deprecated Use setBrinkmanPenaltyFactor() instead.
      */
+	IBTK_DEPRECATED("Use setBrinkmanPenaltyFactor() instead.")
```
boyceg, line 86: Run clang-format to get rid of tabs?

boyceg:
Does this mean that we need to add configure checks for `deprecated`?

ibtk/include/ibtk/FEDataManager.h
```diff
@@ -195,9 +195,8 @@ class FEData : public SAMRAI::tbox::Serializable
     /*!
      * \brief Set the equations systems object that is associated with the
      * FEData object.
-     *
-     * @deprecated The equation systems object should be set by the constructor.
```
abarret, line 199: Does a deprecated message still show up in the doxygen generated files?

drwells: I thought that doxygen would pick these up but I don't see an option for it, so I'll add these back in.

drwells:
We already check for availability of `[[deprecated]]` and `[[deprecated("msg")]]` via
```cpp
#if defined(__has_cpp_attribute)
#if __has_cpp_attribute(deprecated)
#define IBTK_DEPRECATED(msg) [[deprecated(msg)]]
#else
#define IBTK_DEPRECATED(msg) [[deprecated]]
#endif
#else
#define IBTK_DEPRECATED(msg)
#endif
```
so that much should be fine - we've been using that for awhile to print deprecation warnings for global system names.

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -61,55 +61,12 @@ IBTK_ENABLE_EXTRA_WARNINGS
 #if __has_cpp_attribute(deprecated)
 #define IBTK_DEPRECATED(msg) [[deprecated(msg)]]
 #else
-#define IBTK_DEPRECATED(msg)
+#define IBTK_DEPRECATED(msg) [[deprecated]]
```
boyceg, line 64: I am sure that I am being dumb, but if `__has_cpp_attribute(deprecated)` is false, shouldn't this be:
```suggestion
#define IBTK_DEPRECATED(msg)
```
?

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -61,55 +61,12 @@ IBTK_ENABLE_EXTRA_WARNINGS
...
 #endif
 #else
 #define IBTK_DEPRECATED(msg)
 #endif
 
-#define IBTK_DEPRECATED_CLASS1(deprecated_class_name)                                                                  \
```
boyceg, line 70: One of the rationales for these macros is to standardize output and minimize text in setting up deprecation notices. Could we keep that strategy even if we move to an all-header-based approach?

ibtk/include/ibtk/ibtk_utilities.h
drwells: Huh - looks like that should be the case. It looks like we don't test any compiler new enough to have `__has_cpp_attribute` but not C++14 so it isn't an issue. Now would be a good time to update to C++14.

We've had this in-place for a year and nothing bad has happened so I'm disinclined to change it.

ibtk/include/ibtk/ibtk_utilities.h
drwells: We could, but that makes it harder to read the header itself. I'm ambivalent about this trade-off.

ibtk/include/ibtk/ibtk_utilities.h
boyceg: If we bump up to C++14 can we just use `[[deprecated]]` everywhere instead of using our own macro?

In fact, we might as well consider updating to C++17 since -- AFAIK -- that is now required by libMesh.

drwells: C++17 would be nice - most people now have at least GCC 9 so it isn't a particularly onerous requirement.

We started requiring C++11 in 2018 so requiring C++17 in 2024 is a logical progression.

drwells:
Lets handle this after the release. At that point we can bump the minimum supported C++ dialect which will clean this (and a lot of other things) up.

drwells:
Yes (#1739). I'll take a look.

include/ibamr/IBFEMethod.h
```diff
@@ -749,16 +751,6 @@ class IBFEMethod : public FEMechanicsBase, public IBStrategy
...
-    /*!
-     * Register a load balancer and work load patch data index with the IB
-     * strategy object.
-     *
-     * @deprecated This method is no longer necessary with the current
-     * workload estimation scheme.
-     */
-    void registerLoadBalancer(SAMRAI::tbox::Pointer<SAMRAI::mesh::LoadBalancer<NDIM> > load_balancer,
-                              int workload_data_idx) override;
-
```
boyceg, line 761: Should this be in this PR?

drwells: Yes. That function doesn't do anything besides set some protected variables which are never read from.

drwells: Also (and importantly) the base class function does the same thing, so this isn't a breaking change.

boyceg: 👍

PR #1645, opened 2023-10-17
A few jump options /cleanup

AminKolah:
It's important that we can easily test activating 1) only the pressure correction 2) only the viscous term correction, 3) the pressure and viscous terms corrections but not the interpolation correction, 4) viscous term and interpolation correction but not the pressure correction and 5) all the corrections or none of them. 

Additionally, it helps to be able to provide ad-hoc tolerance values related to the intersection algorithm via the input file, for instance in an application with extreme physical/discretization parameters for which additional digits of precision may become important. 

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/examples/IIM/ex0/input2d
+++ b/examples/IIM/ex0/input2d
@@ -18,8 +18,7 @@ ELEM_TYPE = "TRI3"                               // type of element to use for s
 
 // solver parameters
 IB_DELTA_FUNCTION            = "IB_3"            // the type of smoothed delta function to use for Lagrangian-Eulerian interaction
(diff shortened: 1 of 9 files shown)


drwells:
Can you take a look at the failing tests

AminKolah:
Apparently tightening the tolerance `d_mesh_perturb_tol` from the default value of 10^-4 to 10^-8 has slightly changed the outputs for curved geometry tests. I set the default value back to what we had before but this may need a bit more attention to make sure nothing unusual happens in the intersection algorithm when perturbing the mesh.

drwells (approved):

PR #1647, opened 2023-10-19
avoid compiling boost code that is incompatible with C++17

boyceg:
- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/config/config.h.tmp
+++ b/ibtk/config/config.h.tmp
@@ -66,4 +66,9 @@ _Pragma("GCC diagnostic ignored \"-Wunneeded-internal-declaration\"")
 #include <ibtk/compiler_hints.h>
 #include <ibtk/ibtk_utilities.h>
 
+// Disable deprecated features in boost.
+#ifndef _HAS_AUTO_PTR_ETC
+#define _HAS_AUTO_PTR_ETC 0
+#endif
+
 #endif //#ifndef included_IBTK_config


ibtk/config/config.h.tmp
```diff
@@ -66,4 +66,9 @@ _Pragma("GCC diagnostic ignored \"-Wunneeded-internal-declaration\"")
 #include <ibtk/compiler_hints.h>
 #include <ibtk/ibtk_utilities.h>
 
+// Tell boost to get with the program:
```
drwells, line 69:
```suggestion
// Disable std::auto_ptr inside our bundled copy of boost:
```

drwells (changes requested):

ibtk/config/config.h.tmp
boyceg: Unfortunately this macro does much more than just disable `auto_ptr`. We need it here to disable the use of `unary_function`.

ibtk/configure.ac
```diff
@@ -94,7 +94,9 @@ dnl provided by argument)
 CONFIGURE_LIBMESH
 CONFIGURE_PETSC
 # Configure other libraries that are not themselves dependencies:
-CONFIGURE_BOOST("$ABSOLUTE_SRCDIR/contrib/boost","\$(abs_top_builddir)/contrib/boost")
+if test "$USING_BUNDLED_BOOST" = yes; then
+  CONFIGURE_BOOST("$ABSOLUTE_SRCDIR/contrib/boost","\$(abs_top_builddir)/contrib/boost")
+fi
```
drwells, line 99: Is this correct? We already check for bundled boost in `configure_boost.m4`.

boyceg: I am not sure, but I could not get `configure` to run without adding this check.

ibtk/config/config.h.tmp
drwells: How about `// Disable features which were deprecated in C++11 and removed in C++17 in boost:`

boyceg: Isn't that what I wrote? 😁

drwells: Technically but I didn't know that it was for the other deprecated then removed functions 🙈

boyceg: I think this might be a MSVC flag that Boost has adopted...?

ibtk/configure.ac
boyceg: I will take another look when I get a few minutes.

boyceg:
@drwells I've updated this PR. LMK if you have other comments. I am not sure the usual checklist applies.

drwells (approved):

drwells:
Things pass the CI so this is good - thanks for taking a second look.

PR #1652, opened 2023-11-14
Add an API for getting the CFL number.

drwells:
Presently, the CFL number is only available through the logfile. This PR consolidates where we calculate it (the INS solver is always responsible) and makes it available via `getCurrentCFLNumber()`.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/include/ibamr/INSHierarchyIntegrator.h
+++ b/include/ibamr/INSHierarchyIntegrator.h
@@ -371,6 +371,21 @@ public:
      */
     int getNumberOfCycles() const override;
 
+    /*!
+     * Finish postprocessing the hierarchy by computing the current CFL number.
+     */
+    virtual void postprocessIntegrateHierarchy(double current_time,
+                                               double new_time,
+                                               bool skip_synchronize_new_state_data,
+                                               int num_cycles = 1) override;
+
+    /*!
+     * Return the current CFL number (i.e., the CFL number computed from the
+     * current velocity field). This is typically computed at the end of each
+     * time step.
+     */
+    virtual double getCurrentCFLNumber() const;
+
 protected:
     /*!
      * The constructor for class INSHierarchyIntegrator sets some default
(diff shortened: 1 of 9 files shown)


abarret (approved):
A couple of comments.

include/ibamr/INSHierarchyIntegrator.h
```diff
@@ -422,6 +434,12 @@ class INSHierarchyIntegrator : public IBTK::HierarchyIntegrator
      */
     virtual void regridProjection() = 0;
 
+    /*!
+     * Pure virtual method to update the current CFL number (i.e., at the end of
+     * a timestep).
+     */
+    virtual void updateCurrentCFLNumber(const int data_idx, const double dt) = 0;
```
boyceg, line 441: I wonder about making this a static that "knows what to do" for different data centerings.

I guess the advantage of making it virtual is if we ever get around to implementing INS schemes that use non-interpolatory kernels.

Note also that a generic implementation could be provided that uses functionality from `ArrayDataNormOps` to do this in a somewhat data-centering agnostic way.

drwells: We could make it static but I think a good alternative would be to recycle the external version in IBHierarchyIntegrator which can handle both data centerings.

Data is stored in a weird way for side-centered so I don't think using the array operations would make this significantly simpler.

include/ibamr/INSHierarchyIntegrator.h
```diff
@@ -371,6 +371,18 @@ class INSHierarchyIntegrator : public IBTK::HierarchyIntegrator
...
+                                               double new_time,
+                                               bool skip_synchronize_new_state_data,
+                                               int num_cycles = 1) override;
+
+    /*!
+     * Return the current CFL number (i.e., the CFL number computed from the
+     * current velocity field). This is typically computed at the end of each
+     * time step.
+     */
+    virtual double getCurrentCFLNumber() const;
```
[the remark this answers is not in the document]
drwells: No, but it can't hurt - maybe someone will do something funny to compute this in the future.

PR #1662, opened 2023-12-12
Various mpi bugs

### LINKED ISSUE #1656: Bug in IBTK_MPI::allGather()
We do not handle the corner case of x_in being nullptr or x_in_size being zero properly. A processor may not have anything to contribute to the all gather operation. Presently the code seg faults because we do 

`mpi_type_id(x_in[0])` 

here: https://github.com/IBAMR/IBAMR/blob/master/ibtk/include/ibtk/private/IBTK_MPI-inl.h#L147

 If `x_in` is nullptr or has no memory then there is an error. Possible fix that I did locally:

```    
    T dummy_value;
    MPI_Allgatherv(x_in,
                   size_in,
                   mpi_type_id(dummy_value),
                   x_out,
                   rcounts.data(),
                   disps.data(),
                   mpi_type_id(dummy_value),
                   IBTK_MPI::getCommunicator());
```

MPI_Allgatherv seems to handle the case of x_in == nullptr and x_in_size == 0 properly. It is our mpi_type passing that causes segfault.

drwells:
Fixes #1656 as well as some other minor issues.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/utilities/IBTK_MPI.cpp
+++ b/ibtk/src/utilities/IBTK_MPI.cpp
@@ -16,6 +16,7 @@
 #include <ibtk/IBTK_MPI.h>
 
 #include <tbox/SAMRAI_MPI.h>
+#include <tbox/Utilities.h>
 
 #include <ostream>
 #include <string>
@@ -51,7 +52,8 @@ int
 IBTK_MPI::getNodes()
 {
     int nodes = 1;
-    MPI_Comm_size(IBTK_MPI::getCommunicator(), &nodes);
+    const int ierr = MPI_Comm_size(IBTK_MPI::getCommunicator(), &nodes);
+    TBOX_ASSERT(ierr == MPI_SUCCESS);
     return nodes;
 } // getNodes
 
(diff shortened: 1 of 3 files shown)


ibtk/src/utilities/IBTK_MPI.cpp
```diff
@@ -43,30 +44,33 @@ IBTK_MPI::getSAMRAIWorld()
 #if SAMRAI_VERSION_MAJOR == 2
     return SAMRAI::tbox::SAMRAI_MPI::commWorld;
 #else
-    return SAMRAI::tbox::SAMRAI_MPI::getSAMRAIWorld()
+    return SAMRAI::tbox::SAMRAI_MPI::getSAMRAIWorld();
```
boyceg, line 47: Does this mean you are messing around with SAMRAI 3/4??? 😀

boyceg (approved):

ibtk/src/utilities/IBTK_MPI.cpp
drwells: No, not yet; it's just that what we have clearly won't work without the `;`.

PR #1671, opened 2024-01-10
Introduce doIntegrateHierarchy() virtual function.

### LINKED ISSUE #1636: executeIntegrateHierarchyCallbackFcns behavior in the outermost derived class
Hi @boyceg ,

We have ```PhaseChangeHierarchyIntergrator (PCHI)``` derived from ```AdvDiffSemiImplicitHierarchyIntegrator (ADSIHI)``` class. The derived class PCHI calls the ```ADSIHI::integrateHierarchy()``` in its ```integrateHierarchy()```. The derived class PCHI also calls the ```executeIntegrateHierarchyCallbackFcns()``` at the end of ```integrateHierarchy()```.  This is needed if we want to modify or reset the solution before the next cycle. The problem is the base class ADSIHI also calls the ```executeIntegrateHierarchyCallbackFcns``` function in the ```integrateHierarchy``` function. Eventually, ```executeIntegrateHierarchyCallbackFcns``` is executed twice: One in the ADSIHI and the other in the PCHI. Do you have any suggestions for this?

abarret:
Fixes #1636. This introduces a new protected pure virtual function `doIntegrateHierarchy()` in `HierarchyIntegrator`. Previously, `integrateHierarchy()` was designed to do two things: (1) perform an integration cycle and (2) execute post integration callbacks. The design principle was that the class furthest down the hierarchy ends up calling the post integration callbacks. This causes problems (e.g. in the phase change code from @amneetb and @Raamkrishnan) where child classes want to call the parent class's `integrateHierarchy()` to perform action (1) but not (2).

This changes the design principle so that the parent class `HierarchyIntegrator::integrateHierarchy()` executes post integration callbacks after `doIntegrateHierarchy()` is called. Child classes no longer need to execute the callbacks. I've left `integrateHierarchy()` as a virtual function in case some future application needs to modify the implementation, although I don't think this is necessary.

@boyceg This is the kind of change I was suggesting. This clearly delineates the two purposes of `integrateHierarchy()`. The function `doIntegrateHierarchy()` should _only_ integrate equations.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/HierarchyIntegrator.h
+++ b/ibtk/include/ibtk/HierarchyIntegrator.h
@@ -409,14 +409,10 @@ public:
     virtual void preprocessIntegrateHierarchy(double current_time, double new_time, int num_cycles = 1);
 
     /*!
-     * Pure virtual method to advance data from current_time to new_time.
-     *
-     * Implementations of this virtual function are not required to synchronize
-     * data on the patch hierarchy.  Data synchronization may be done
-     * (optionally) in a specialization of the public virtual member function
-     * postprocessIntegrateHierarchy().
+     * Advance data from current_time to new_time. The current implementation calls
+     * doIntegrateHierarchy() followed by executeIntegrateHierarchyCallbackFcns().
      */
-    virtual void integrateHierarchy(double current_time, double new_time, int cycle_num = 0) = 0;
+    void integrateHierarchy(double current_time, double new_time, int cycle_num = 0);
 
     /*!
      * Method to skip a cycle of the time integration scheme (e.g. for cases in
@@ -697,6 +693,21 @@ public:
     void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db) override;
 
 protected:
+    /*!
+     * Pure virtual function that integrates the hierarchy from current_time to new_time.
+     *
+     * Implementations of this virtual function are not required to synchronize
+     * data on the patch hierarchy.  Data synchronization may be done
+     * (optionally) in a specialization of the public virtual member function
+     * postprocessIntegrateHierarchy().
+     *
+     * This function is called by integrateHierarchy() prior to the execution of callbacks.
+     *
+     * @note Inheriting classes should call their base class versions of this
+     * method.
+     */
+    virtual void integrateHierarchySpecialized(double current_time, double new_time, int cycle_num = 0) = 0;
+
     /*!
      * Perform any necessary work relevant to data owned by the current
      * integrator prior to regridding (e.g., calculating divergences). An
--- a/src/IB/IBExplicitHierarchyIntegrator.cpp
+++ b/src/IB/IBExplicitHierarchyIntegrator.cpp
@@ -156,9 +156,11 @@ IBExplicitHierarchyIntegrator::preprocessIntegrateHierarchy(const double current
 } // preprocessIntegrateHierarchy
 
 void
-IBExplicitHierarchyIntegrator::integrateHierarchy(const double current_time, const double new_time, const int cycle_num)
+IBExplicitHierarchyIntegrator::integrateHierarchySpecialized(const double current_time,
+                                                             const double new_time,
+                                                             const int cycle_num)
 {
-    IBHierarchyIntegrator::integrateHierarchy(current_time, new_time, cycle_num);
+    IBHierarchyIntegrator::integrateHierarchySpecialized(current_time, new_time, cycle_num);
     const double half_time = current_time + 0.5 * (new_time - current_time);
     VariableDatabase<NDIM>* var_db = VariableDatabase<NDIM>::getDatabase();
     const int u_current_idx = var_db->mapVariableAndContextToIndex(d_ins_hier_integrator->getVelocityVariable(),
@@ -357,8 +359,6 @@ IBExplicitHierarchyIntegrator::integrateHierarchy(const double current_time, con
                                              half_time);
     }
 
-    // Execute any registered callbacks.
-    executeIntegrateHierarchyCallbackFcns(current_time, new_time, cycle_num);
     return;
 } // integrateHierarchy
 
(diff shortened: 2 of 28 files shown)


drwells:
> This changes the design principle so that the parent class HierarchyIntegrator::integrateHierarchy() executes post integration callbacks after doIntegrateHierarchy() is called. Child classes no longer need to execute the callbacks. I've left integrateHierarchy() as a virtual function in case some future application needs to modify the implementation, although I don't think this is necessary.

That's definitely the right thing to do. My main concern with this patch is that it breaks basically any every non-IBAMR class which inherits from `HierarchyIntegrator`.

Perhaps we can do things in a more compatible way: I'll take a quick look and report back.

abarret:
> That's definitely the right thing to do. My main concern with this patch is that it breaks basically any every non-IBAMR class which inherits from `HierarchyIntegrator`.

Changing which class is responsible for executing the callback function will break a lot of classes that inherit from `HierarchyIntegrator` anyway. The only real change required for downstream classes is changing the name `integrateHierarchy()` to `doIntegrateHierarchy()`, so the change is minimal. 

We could provide a no-op implementation of `HierarchyIntegrator::doIntegrateHierarchy()` so that classes that want to do everything in `integrateHierarchy()` still can.

abarret:
Another question is should we do the same thing for all the callback functions: `postprocessIntegrateHierarchy`, `preprocessIntegrateHierarchy`, `applyGradientDetector`, and `regridHierarchy`.

drwells:
I'd rather not modify those extension mechanisms (i.e,. we should go outside-in and then inside-out for preprocessing and postprocessing) if we don't have to.

drwells:
How about we rename `doIntegrateHierarchy()` to `integrateHierarchySpecialized()` to match the naming convention? Similarly, I think `integrateHierarchy()` should not be virtual: people should use the documented extension mechanisms instead of making this even more complex (and we're breaking stuff now in an incompatible way).  Other than that (and the CI errors) I think this is fine.

My basic attempt (move `executeIntegrateHierarchyCallbackFcns()` into `HierarchyIntegrator::integrateHierarchy()` didn't work (I'm not sure why) so I don't think there's another straightforward way to fix this. We should also check that derived classes call the parent class `integrateHierarchySpecialized()` at the top of their own version of that function.

ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -409,14 +409,15 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
     virtual void preprocessIntegrateHierarchy(double current_time, double new_time, int num_cycles = 1);
 
     /*!
-     * Pure virtual method to advance data from current_time to new_time.
+     * Virtual method to advance data from current_time to new_time. The current implementation calls
```
drwells, line 412:
```suggestion
     * Advance data from current_time to new_time. The current implementation calls
```

ibtk/include/ibtk/HierarchyIntegrator.h
```diff
@@ -697,6 +698,16 @@ class HierarchyIntegrator : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM
     void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db) override;
 
 protected:
+    /*!
+     * Pure virtual function that integrates the hierarchy from current_time to new_time.
+     *
+     * Implementations of this virtual function are not required to synchronize
+     * data on the patch hierarchy.  Data synchronization may be done
+     * (optionally) in a specialization of the public virtual member function
+     * postprocessIntegrateHierarchy().
```
drwells, line 707: Could you add a `@note` here explaining why we did this, i.e., some discussion of when callbacks are called?

drwells: Could you also mention that derived classes should typically call the parent class `integrateHierarchySpecialized()` at the top of their own implementation of this function?

src/IB/IBExplicitHierarchyIntegrator.cpp
```diff
@@ -156,9 +156,10 @@ IBExplicitHierarchyIntegrator::preprocessIntegrateHierarchy(const double current
 } // preprocessIntegrateHierarchy
 
 void
-IBExplicitHierarchyIntegrator::integrateHierarchy(const double current_time, const double new_time, const int cycle_num)
+IBExplicitHierarchyIntegrator::integrateHierarchySpecialized(const double current_time,
+                                                             const double new_time,
+                                                             const int cycle_num)
 {
-    IBHierarchyIntegrator::integrateHierarchy(current_time, new_time, cycle_num);
```
drwells, line 161: I'm actually surprised that this works. We should definitely still call `IBHierarchyIntegrator::integrateHierarchySpecialized()` here even if that function doesn't do anything.

drwells (changes requested):
Thanks - I have a few more minor suggestions.

I think that it's important that we keep the idiom 'outside-in and then inside-out' working for hierarchy integrators. We use that (or at least we should - I think I fixed all of them) for preprocessing and postprocessing too so that derived classes don't mess around with parent class data.

drwells:
@abarret can you get this done today? If not lets leave it for 0.15.

src/IB/IBExplicitHierarchyIntegrator.cpp
abarret: I added an empty implementation of `HierarchyIntegrator::integrateHierarchySpecialized()` so we can make this call.

Note the only integrator that doesn't call it's immediate base class is `BrinkmanAdvDiffSemiImplicitHierarchyIntegrator`, which inherits from `AdvDiffSemiImplicitHierarchyIntegrator` but calls `AdvDiffHierarchyIntegrator::integrateHierarchySpecialized()`. This is consistent with the class's behavior prior to this PR.

abarret:
> @abarret can you get this done today? If not lets leave it for 0.15.

I've made the requested changes and squashed the commits.

drwells (approved):

PR #1677, opened 2024-01-24
Run clang-format on everything.

drwells:
Part of #1664.

It's just clang-format. The normal list doesn't apply. We'll be able to get rid of this step soon once we canonicalize our clang-format version (which I've done on most IBAMR projects now).


src/navier_stokes/MarangoniSurfaceTensionForceFunction.cpp
```diff
@@ -504,7 +504,7 @@ MarangoniSurfaceTensionForceFunction::setDataOnPatchSide(Pointer<SideData<NDIM,
             .upper(1)
 #endif
 #if (NDIM == 3)
-        F_data->getPointer(0),
+                F_data->getPointer(0),
```
boyceg, line 507: ??

boyceg (approved):

PR #1680, opened 2024-01-29
add additional composite B-spline kernel functions

boyceg:
Adds 5/4 and 6/5 composite B-splines.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
@drwells, @abarret, could one of you help me to figure out why these tests are failing? I cannot reproduce locally 😿

drwells:
Interesting problem. In `exact` mode `interpolate_01` should just print `OK` at the end but it also printed

```

    ##2       ==> 0.48555, 0.390822, 0.346354, 1.52145e-12
```
The last column is the error: `1.52e-12`. The problem is here:

https://github.com/IBAMR/IBAMR/blob/171f7499849321bb99c65fb589f888e56504ecfb/tests/interpolate/interpolate_01.cpp#L207-L210

That line needs to be updated to
```cpp
const double tol = LEInteractor::getStencilSize(interp_fcn) > 4 ? 1e-10 : 1e-12;
```
or something nearby to lower the error tolerance enough for the stencils to pass. This presumes that these stencils can exactly reproduce a bilinear field which I think they should do.

drwells:
I'm not sure why the CI testers exhibit more error than my desktop - these tests pass for me too. Must be different optimization levels.

boyceg:
@drwells tests all pass now.

drwells (approved):

PR #1683, opened 2024-02-07
Remove cdash uploading.

### LINKED ISSUE #1666: Remove our dependency on open.cdash.org
One of our CI runners failed because this website wasn't available. We don't ever use this, nor do we need to - we only merge PRs if 100% of the tests pass so there's nothing to log there.

drwells:
This is not useful since our local CI doesn't pass unless 100% of run tests pass and there are zero warnings.

Fixes #1666. The normal checklist doesn't apply.


drwells:
@boyceg Could you take a quick look and approve? This patch prevents some intermittent CI failures.

boyceg (approved):

PR #1684, opened 2024-02-08
Update attest to work with openmpi 5.

drwells:
I updated things on my local machine and that has necessitated some fixes - I have a few more minor updates for the mailing list and ibsamrai2.

This version of openmpi reads a different set of environment variables. It is infrastructure so the normal list does not apply.

### DIFF
--- a/attest
+++ b/attest
@@ -98,13 +98,21 @@ class Parameters:
 
         # These are the only two required inputs:
         if self.mpiexec == "":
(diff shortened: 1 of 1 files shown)


drwells:
@boyceg Could you take a quick look and approve? This patch is necessary for some CI upgrades (see #1693)

boyceg (approved):

PR #1688, opened 2024-02-22
Make get functions public in IBMethod.

abarret:
Currently, the IB data in `IBMethod` during the middle of a time step is inaccessible outside of `IBMethod`. The data is owned by the class `IBMethod` and therefore can not be retrieved via `LDataManager`. This PR makes the get functions public.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/include/ibamr/IBMethod.h
+++ b/include/ibamr/IBMethod.h
@@ -28,6 +28,7 @@
 
 #include "ibtk/LInitStrategy.h"
 #include "ibtk/LSiloDataWriter.h"
+#include "ibtk/ibtk_enums.h"
 #include "ibtk/ibtk_utilities.h"
 
 #include "GriddingAlgorithm.h"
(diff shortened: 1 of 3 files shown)


include/ibamr/IBMethod.h
```diff
@@ -444,49 +444,69 @@ class IBMethod : public IBImplicitStrategy
      */
     void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db) override;
 
-protected:
     /*!
      * Get the current structure position data.
+     *
+     * data_time must be equal to one of current time, new time, or half time. If this condition is met, X_data is set
+     * to the data at that respective time, otherwise the X_data is unchanged.
```
boyceg, line 451:
```suggestion
     * to the data at that respective time, otherwise the X_data pointers are unchanged.
```

include/ibamr/IBMethod.h
```diff
@@ -444,49 +444,69 @@ class IBMethod : public IBImplicitStrategy
      */
     void putToDatabase(SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> db) override;
 
-protected:
     /*!
      * Get the current structure position data.
+     *
+     * data_time must be equal to one of current time, new time, or half time. If this condition is met, X_data is set
```
boyceg, line 450: Are these time values available from `IBMethod`? If we are using them in this interface, it might be good to have accessors.

include/ibamr/IBMethod.h
```diff
@@ -444,49 +444,69 @@ class IBMethod : public IBImplicitStrategy
...
+     *
+     * If the linearized position data does not exist, it will be created.
      */
     void getLinearizedPositionData(std::vector<SAMRAI::tbox::Pointer<IBTK::LData> >** X_data,
                                    bool** X_needs_ghost_fill);
 
     /*!
      * Get the current interpolation/spreading position data.
+     *
+     * data_time must be equal to one of current time, new time, or half time. If this condition is met, X_LE_data is
```
boyceg, line 468: Same as above.

include/ibamr/IBMethod.h
```diff
@@ -444,49 +444,69 @@ class IBMethod : public IBImplicitStrategy
...
+     * If this class is not set up to use fixed coupling, this returns data from getPositionData().
      */
     void getLECouplingPositionData(std::vector<SAMRAI::tbox::Pointer<IBTK::LData> >** X_LE_data,
                                    bool** X_LE_needs_ghost_fill,
                                    double data_time);
 
     /*!
      * Get the current structure velocity data.
+     *
+     * data_time must be equal to one of current time, new time, or half time. If this condition is met, U_data is set
```
boyceg, line 480: Same as above.

include/ibamr/IBMethod.h
```diff
@@ -444,49 +444,69 @@ class IBMethod : public IBImplicitStrategy
...
      * Get the linearized structure velocity data.
+     *
+     * If the linearized velocity data does not exist, it will be created.
      */
     void getLinearizedVelocityData(std::vector<SAMRAI::tbox::Pointer<IBTK::LData> >** U_data);
 
     /*!
      * Get the current structure force data.
+     *
+     * data_time must be equal to one of current time, new time, or half time. If this condition is met, F_data is set
```
boyceg, line 495: Same as above.

boyceg:
Looks good. I am inclined to keep the interface related to implicit time stepping protected for now, since I think those interfaces are likely to change as we get some effective preconditioners implemented over the next several months. However, if you need access now or soon, I guess that it's OK to go ahead and make it public.

include/ibamr/IBMethod.h
abarret: I've add a `TimePoint` enum so that users can specify time points without knowing the actual time values.

boyceg: Nice. We should be doing this in `IBFEMethod` and other places too. 😁

abarret:
> Is this ready for review again?

Yes

include/ibamr/IBMethod.h
drwells: Everywhere else in the library we've made do with just a `double` to represent one of the three time values. In what context would we use a `TimePoint` instead of a `double` inside the library?

I'd like to avoid having two parallel interfaces which do the same things.

abarret: Actually, most public interfaces to data do not rely on `double`'s. Eulerian data is retrieved via `getCurrentContext()` and `getNewContext()`. Data held by `IBFE` classes and `LDataManager` is retrieved through strings.

Another thing is that this change doesn't change the internals of the class. This only adds an additional public interface to the data owned by the class. The internal private interface stays the same.

> In what context would we use a `TimePoint` instead of a `double` inside the library?

This is useful for classes that don't know (or need to know) the time points used by `IBMethod`. My use case involves computing forces between structures time stepped by different `IBMethod` objects.

boyceg:
@drwells do you want to get this one in the release?

drwells:
I haven't had time to review it. Lets tackle this one next week.

boyceg:
@drwells, do you want to take another look at this before merging?

drwells (approved):
I got super behind in reviewing these - @abarret Any further comments? If not lets merge.

abarret:
I've got nothing to add.

Issue #1650, opened 2023-10-25
The draft, and actually writing some tutorial programs


drwells:
We need to make it easier for new people to get started with IBAMR. A good way to do this is to have a curated set of tutorial programs with inlined explanations of what is going on in the doxygen documentation.

I'm drafting all eligible (i.e., not presently writing their thesis) graduate students and postdocs using IBAMR here at UNC to help me in this task - they will do most of the tutorial writing. The first thing to do here is to resurrect the doxygen deployment scripts which have been offline for quite some time - see #1649.

drwells:
Something to add here is documentation on how to handle paraview support. We got this shell script on the mailing list:

```sh
#!/bin/bash

VizPath="./viz_IB3d_$1/*/"                          #data folder path
OutFileName="./viz_IB3d_$1/paraview.samrai.series"  #output file
declare -a arrdata=()                               #declare array for data
declare -a arrfolder=()                             #declare array for folder names
count=0                                             #counter for probe date
arrdata+=0                                          #add a zero in the beginning for probe data.
arrfolder+=($(ls -d $VizPath))                      #get folder names from VizPath

if test -f "$OutFileName"; then                     #if outputfile exist, delete it.
    rm "$OutFileName"
fi
arrfolder=( $(printf "%s\n" "${arrfolder[@]}" | sort -V) ) ## this is needed to sort the folders in 'natural' order
printf "{\n" >> $OutFileName
printf "  \"file-series-version\" : \"1.0\",\n" >> $OutFileName
printf "  \"files\" : [\n" >> $OutFileName

for i in "${arrfolder[@]}"
do
   var=$(echo $i | rev | cut -d'/' -f-2 | rev) #reformat string with folder names
   var2="${var}summary.samrai"                 #add summary.samrai to folder name strings
   time="${arrdata[$count]}"                   #get time value
   printf "    { \"name\" : \"./$var2\", \"time\" : $count }, \n"  >> $OutFileName #time will be 0,1,2,3 ….
    (( count++ ))
done
printf "  ]\n" >> $OutFileName
printf "}" >> $OutFileName

#done
```
we should clean it up a bit and bundle it in IBAMR proper (with an attribution to Jonas).

drwells:
See also #1642 for some more things which could be better documented.

drwells:
[...]

PR #1692, opened 2024-03-01
Documentation test branch

brynbarker:
Added a github action to run doxygen and update ibamr.github.io by pushing to the IBAMR-docs repository
This action also stores the doxygen log as an artifact

This does not fix any issue fully but is part of issue #1650 in that with this fix, any added tutorials will be correctly reflected on the github website. 

Next step will be to make it possible to build doxygen locally

- [ ] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


.github/workflows/push-master.yml
```diff
@@ -17,21 +17,15 @@ on:
   push:
     branches:
       - documentation-test-branch
```
drwells, line 19:
```suggestion
      - master
```

.github/workflows/push-master.yml
```diff
@@ -17,21 +17,15 @@ on:
...
   deploy_documentation:
     runs-on: ubuntu-latest
-    name: Build IBAMR and run tests
+    name: Compile and deploy doxygen documentation
     env:
-      # TODO: this is a copy-and-paste of the Travis version. Once this is
-      # working we should clean it up.
-      TRAVIS_BUILD_NUMBER: ${{ github.run_number }}
-      TRAVIS_COMMIT: ${{ github.sha }}
+      TRAVIS_BUILD_DIR: ${{ github.workspace }}
```
drwells, line 27: Can we get rid of the rest of the travis variables at this point?

drwells (changes requested):

drwells:
Thanks a ton for this work. There's a lot more to do with the manual but getting things consistently building is the first step.

drwells:
Can you also rebase and squash once everything is done?

abarret:
Just FYI, the IBAMR doc page is empty as of this PR.

drwells:
@brynbarker can you take a look? We need to get that on again ASAP.

.github/workflows/push-master.yml
brynbarker: we need to keep this variable! removing it is what caused the ibamr doc page to be blank

drwells:
Interesting - I don't know why we would need a travis variable when we don't use travis but lets sort that out later.

drwells (approved):

.github/workflows/push-master.yml
drwells: We still need to make this change. Things are correctly rebuilt but we want it to be for `master`.

drwells (changes requested):
We still need to change the branch name.

drwells:
Nice. Would you mind squashing the commits so we have tidy history?

drwells (approved):
Let's deploy this and see what happens 😎

PR #1693, opened 2024-03-04
Clang format

### LINKED ISSUE #1124: Style Guide

### LINKED ISSUE #1299: Start using clang-format

### LINKED ISSUE #1423: Infrastructure Week (end of February)
I mentioned I would do this in another thread and I need to write down some todos somewhere.

We've run into a couple of silly problems recently due to insufficient CI coverage, including:
- examples don't compile with autotools
- examples don't compile with CMake
- bug with older versions of CMake
- things don't compile with older versions of clang (in particular when the compiler does not default to C++11 or newer)

Now that we have docker environments set up we should be able to expand our CI in a modular way (and use some machines at UNC). Some TODOs:
- [x] Set up another docker container that uses autotools
- [x] Verify that the current docker container compiles with optimizations on
- [x] Set up another docker container that uses all oldest supported versions of our dependencies (if we could use GCC 4.8.5 that would be even better)
- [x] enable `-Werror` to prevent silly stuff like unused variables coming through
- [ ] Set up clang-format
- [x] Compile tests and examples without libMesh

drwells:
I think its time to finally do this - we always have outstanding work so lets just get this in now and deal with the fallout afterwards.

Fixes #1124. Fixes #1299. Fixes #1423. 

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
Getting this all working correctly required installing git on each of the CI docker images, which in turn required rebuilding them. I made some small updates.

If desired, I can split this in two parts - CI and clang-format - let me know.

boyceg (approved):

drwells:
There will probably be some hiccups with the other open PRs - I'll keep an eye out.

PR #1696, opened 2024-03-19
Clean up restart db

drwells:
Follow-up to a conversation I had with @marshallrdavey - it turns out that most of the stuff we store in a restart database is never actually read. Removing things we don't ever use, like plot-only variables, shrinks its size by about 50%.

- [x] Was clang-format run?
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
+++ b/src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
@@ -921,13 +921,13 @@ INSStaggeredHierarchyIntegrator::initializeHierarchyIntegrator(Pointer<PatchHier
     {
         d_F_cc_idx = invalid_index;
     }
-    registerVariable(d_Omega_idx, d_Omega_var, no_ghosts, getCurrentContext());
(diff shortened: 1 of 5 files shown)


src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -1638,9 +1669,58 @@ INSStaggeredHierarchyIntegrator::getStableTimestep(Pointer<Patch<NDIM> > patch)
...
+            if (level->checkAllocated(idx)) level->deallocatePatchData(idx);
+        }
+        if (!level->checkAllocated(d_Div_U_idx))
+        {
+            level->allocatePatchData(d_Div_U_idx);
+        }
+    }
+
+    // Ensure that the divergence is allocated:
+
```
abarret, line 1723:
```suggestion
```

abarret (approved):
This is a good change. I've run out of scratch space before because I forgot about restart files. Also this could help speed up regridding slightly.

PR #1703, opened 2024-04-26
More ci cleanup

drwells:
We've exhausted our free CI hours partially because some of the caching isn't right. I'm investigating.


boyceg (approved):
I feel like I am not qualified to review this one, but I guess it looks OK to me. 🤷‍♂️

drwells:
Things worked well-enough here but there are some other tweaks we need to make to make maximal use of CI caching. In other words - getting this right is taking a few steps and the only way to figure it out is to try.

PR #1708, opened 2024-05-17
add IMEX-BDF2-type time stepping

boyceg:
Adding support for BDF2-like time stepping. The goal is to enable "consistent" time stepping for IIM.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
+++ b/src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
@@ -475,6 +475,7 @@ INSStaggeredHierarchyIntegrator::INSStaggeredHierarchyIntegrator(std::string obj
     switch (d_viscous_time_stepping_type)
     {
     case BACKWARD_EULER:
+    case BDF2:
     case FORWARD_EULER:
     case TRAPEZOIDAL_RULE:
         break;
@@ -482,7 +483,7 @@ INSStaggeredHierarchyIntegrator::INSStaggeredHierarchyIntegrator(std::string obj
         TBOX_ERROR(d_object_name << "::INSStaggeredHierarchyIntegrator():\n"
                                  << "  unsupported viscous time stepping type: "
                                  << enum_to_string<TimeSteppingType>(d_viscous_time_stepping_type) << " \n"
-                                 << "  valid choices are: BACKWARD_EULER, FORWARD_EULER, "
+                                 << "  valid choices are: BACKWARD_EULER, BDF2, FORWARD_EULER, "
                                     "TRAPEZOIDAL_RULE\n");
     }
     switch (d_convective_time_stepping_type)
@@ -573,6 +574,7 @@ INSStaggeredHierarchyIntegrator::INSStaggeredHierarchyIntegrator(std::string obj
     d_F_var = INSHierarchyIntegrator::d_F_var;
     d_Q_var = INSHierarchyIntegrator::d_Q_var;
     d_N_old_var = new SideVariable<NDIM, double>(d_object_name + "::N_old");
+    d_U_old_var = new SideVariable<NDIM, double>(d_object_name + "::U_old");
 
     // with our convention fine_boundary_represents_var = false means that will
     // use a conforming discretization (the solution, with bilinear/trilinear
@@ -897,13 +899,27 @@ INSStaggeredHierarchyIntegrator::initializeHierarchyIntegrator(Pointer<PatchHier
 
     // We need previous values of N for Adams-Bashforth so keep it in the
     // restart database.
-    registerVariable(d_N_old_current_idx,
-                     d_N_old_new_idx,
-                     d_N_old_scratch_idx,
-                     d_N_old_var,
-                     side_ghosts,
-                     d_N_coarsen_type,
-                     d_N_refine_type);
+    if (is_multistep_time_stepping_type(d_convective_time_stepping_type))
+    {
+        registerVariable(d_N_old_current_idx,
+                         d_N_old_new_idx,
+                         d_N_old_scratch_idx,
+                         d_N_old_var,
+                         side_ghosts,
+                         d_N_coarsen_type,
+                         d_N_refine_type);
+    }
+
+    if (is_multistep_time_stepping_type(d_viscous_time_stepping_type))
+    {
+        registerVariable(d_U_old_current_idx,
+                         d_U_old_new_idx,
+                         d_U_old_scratch_idx,
+                         d_U_old_var,
+                         side_ghosts,
+                         d_U_coarsen_type,
+                         d_U_refine_type);
+    }
 
     registerVariable(d_U_nc_idx, d_U_nc_var, no_ghosts, getPlotContext());
     d_plot_indices.push_back(d_U_nc_idx);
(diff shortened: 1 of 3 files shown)


boyceg:
The implementation is intended to correspond to VSSBDF2 from [this paper](https://steveruuth.org/wp-content/uploads/2020/10/vsimex_Wang_Ruuth_jcm.pdf).

boyceg:
One thing that I know that I need to fix is regridding the old velocity field.

boyceg:
[Here is a paper](https://www.sciencedirect.com/science/article/pii/S0021999119303237) looking at errors in the initial time step. The recommendation is to use something like TR-BDF2 in the first time step.

boyceg:
@abarret I am not getting perfect second-order convergence rates; I wonder if it is start-up error?

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -2287,27 +2384,43 @@ INSStaggeredHierarchyIntegrator::reinitializeOperatorsAndSolvers(const double cu
...
     }
     PoissonSpecifications U_problem_coefs(d_object_name + "::U_problem_coefs");
-    U_problem_coefs.setCConstant((rho / dt) + K * lambda);
-    U_problem_coefs.setDConstant(-K * mu);
+    U_problem_coefs.setCConstant(K1 * rho / dt + K2 * lambda);
+    U_problem_coefs.setDConstant(-K2 * mu);
     PoissonSpecifications P_problem_coefs(d_object_name + "::P_problem_coefs");
     P_problem_coefs.setCZero();
-    P_problem_coefs.setDConstant(rho == 0.0 ? -1.0 : -1.0 / rho);
+    P_problem_coefs.setDConstant(rho == 0.0 ? -1.0 : -1.0 / (K1 * rho));
```
abarret, line 2423: P is not timestepped.
```suggestion
    P_problem_coefs.setDConstant(rho == 0.0 ? -1.0 : -1.0 / rho);
```

boyceg:
I seem to be getting essentially perfect 2nd-order accuracy with the uniform grid version of `examples/navier_stokes/ex0` with periodic boundary conditions.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -1440,15 +1490,28 @@ INSStaggeredHierarchyIntegrator::setupSolverVectors(const Pointer<SAMRAIVectorRe
...
 #endif
-            const double omega = dt / d_dt_previous[0];
-            d_hier_sc_data_ops->linearSum(N_idx, 1.0 + 0.5 * omega, N_idx, -0.5 * omega, d_N_old_current_idx);
+            const double omega = getTimeStepSizeRatio();
+            double beta1, beta2;
+            if (is_bdf_time_stepping_type(d_viscous_time_stepping_type))
+            {
+                beta1 = 1.0 + omega;
+                beta2 = -omega;
+            }
```
abarret, line 1506: I don't think this is true for `BACKWARD_EULER`. The coefficients depend on the particular BDF scheme.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -1464,23 +1527,57 @@ INSStaggeredHierarchyIntegrator::setupSolverVectors(const Pointer<SAMRAIVectorRe
...
+        double data_time;
+        if (is_bdf_time_stepping_type(d_viscous_time_stepping_type))
+        {
+            data_time = new_time;
+        }
+        else
+        {
+            data_time = half_time;
+        }
+        d_F_fcn->setDataOnPatchHierarchy(d_F_scratch_idx, d_F_var, d_hierarchy, data_time);
```
abarret, line 1539: This might be a problem for body forcing terms that depend on advected quantities, for example I don't know if the complex fluid solver is set up to correctly handle evaluations at `new_time`.

I have switched over to using AB2 for body forces.

boyceg: This change is actually the primary reason that I am doing this --- we want to be able to use the end-time step forcing for IIM.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
boyceg: Don't you want the RHS to be at the end of the time step at least for any BDF with order >= 2?

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
boyceg: Here I have to admit that I have forgotten why we need to include `rho` in the pressure setup.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
abarret: Yes, but we need a better way to inform force functions that the approximation should be using data at `t^{n+1}` instead of `t^{n+1/2}`. For example, this always uses `t^{n+1/2}` regardless of time stepping scheme, which for BDF2 will be first order.
https://github.com/IBAMR/IBAMR/blob/2b40ae33badfd03f2759c8999705cf91060da548/examples/navier_stokes/ex5/BoussinesqForcing.cpp#L76-L86

boyceg: Well, here the implementation (probably by me) is ignoring `data_time`....!

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
abarret: Oh hmmm... I think I misunderstood what this was being used for.

boyceg: Specifically, at least for the test I am running, this seems to make no difference for `TRAPEZOIDAL_RULE` but to shave one iteration off for `BDF2`. Let me see if I can figure out whether we care about this or not in practice.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
abarret: Yes, but `BACKWARD_EULER` is not order >= 2. I guess if you are using first order viscous time stepping, using first order convective time stepping is fine too?

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
boyceg: This is a little obscure, but the projection preconditioner is derived from
```
    //    rho (U - U^*) + G Phi = 0
    //    -D U = F_P
```
and then figuring out how `Phi` relates to `P`. I think that for a BDF scheme, we need to do instead
```
    //    K1 rho (U - U^*) + G Phi = 0
    //    -D U = F_P
```
where `K1` is this factor that is hitting `U^{n+1}` in the BDF time stepping scheme.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
boyceg: Well, what is done now is to use time-step centered convective time stepping with `BACKWARD_EULER`.

I can drop `BACKWARD_EULER` from the list of BDF time stepping schemes if you would prefer that. 😁

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
boyceg: There is a hook in the API that callers can use to inform when things should be evaluated. If the called function cannot evaluate at that time, it should just emit an error message and abort.

abarret: I agree, but there is not a great API to figure out which time point `data_time` corresponds to. But this is an issue with the use of `CartGridFunction`, not the time stepping scheme.

boyceg: I've tried to capture this issue and a possible approach in #1709.

boyceg: Also #1710. It would be nice to be able to handle time extrapolation in a more uniform manner, so that (e.g.) we could easily allow for either evaluation at the end of the time step or extrapolation to the end of the time step in places like this.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
boyceg: I have removed `BACKWARD_EULER` from the list of BDF time stepping schemes for now. Maybe at some point we can rationalize this stuff.

abarret (changes requested):
Some code clarifications. We need to add tests. I think I will have time tomorrow or early next week to do that if you want me to add them.

boyceg:
> Some code clarifications. We need to add tests. I think I will have time tomorrow or early next week to do that if you want me to add them.

If you have the time and interest to take it over, that would be great. I am now working on a couple of grant applications.

include/ibamr/IBFEMethod.h
```diff
@@ -625,6 +626,11 @@ class IBFEMethod : public FEMechanicsBase, public IBStrategy
         const std::vector<SAMRAI::tbox::Pointer<SAMRAI::xfer::RefineSchedule<NDIM> > >& u_ghost_fill_scheds,
         double data_time) override;
 
+    /*!
+     * Indicate that multistep time stepping will be used.
+     */
+    void setUseMultistepTimeStepping(int n_steps = 1) override;
```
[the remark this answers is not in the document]
boyceg: Added some documentation and changed the variable name to `n_previous_steps`.

boyceg:
@abarret can you take a look and see if this has addressed your concerns?

abarret:
> @abarret can you take a look and see if this has addressed your concerns?

This looks good to me. I have some tests, may I push them to this branch?

boyceg:
> > @abarret can you take a look and see if this has addressed your concerns?
> 
> This looks good to me. I have some tests, may I push them to this branch?

That would be wonderful.

abarret (approved):
I've added two tests for BDF2.

boyceg:
@drwells or @amneetb, does one of you want to approve before we merge since @abarret and I are now effectively collaborating on this PR? 😁

boyceg:
@abarret do you feel like squashing?

amneetb:
I can approve but I am not listed as a reviewer on this PR.

abarret:
Squashed

PR #1717, opened 2024-06-08
Fix Omega_idx error in INSVCStaggeredHierarchyIntegrator class

Raamkrishnan:
### IBAMR Pull Request Checklist
- [x] Does the test suite pass?
- [x] Was clang-format (i.e., `make indent`) run? For more information see
      `scripts/formatting/README.md`.
- [x] Were relevant issues cited? Please remember to add `Fixes #12345` to close
      the issue automatically if we are fixing the problem.
- [ ] Is this a change others will want to know about? If so, then has a
      changelog entry been added?
- [ ] If new data structures have been added to a class then have they been set
      up appropriately for restarts? If so, ensure that the restart version
      number is incremented.
- [x] Does this change include a bug fix or new functionality? If so, a new test
      or tests should be added. New tests should run quickly (less than a minute
      in release mode). If possible, an older test should gain a new option so
      that we do not need to compile more test executables.
- [x] Did you (if your account has permission to do so) set relevant labels on
      GitHub for the pull request?

If we deallocate ```d_Omega_idx``` after tagging cells in ```applyGradientDetectorSpecialized``` of ```INSVCStaggeredHierarchyIntegrator``` class, then we need to allocate ```d_Omega_idx``` again in   ```setupPlotDataSpecialized``` to use it for plotting purposes. So I am not deallocating ```d_Omega_idx```. Another way is to allocate and deallocate ```d_Omega_idx``` within the ```setupPlotDataSpecialized``` function. To test this, I modified the ```tests/multiphase_flow/high_density_droplet case``` to use an AMR grid.

PR #1721, opened 2024-06-26
Fix a level number check.

drwells:
The test added in #1717 aborts on `master` because of this typo from the vorticity patches. @Raamkrishnan thanks again for adding this test!

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/navier_stokes/INSVCStaggeredHierarchyIntegrator.cpp
+++ b/src/navier_stokes/INSVCStaggeredHierarchyIntegrator.cpp
@@ -1758,7 +1758,7 @@ INSVCStaggeredHierarchyIntegrator::applyGradientDetectorSpecialized(const Pointe
         d_hier_math_ops->curl(d_Omega_idx, d_Omega_var, d_U_scratch_idx, d_U_var, d_U_bdry_bc_fill_op, error_data_time);
         tagCellsByVorticityMagnitude(level_number, d_Omega_idx, tag_index);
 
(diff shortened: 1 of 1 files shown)


src/navier_stokes/INSVCStaggeredHierarchyIntegrator.cpp
```diff
@@ -1758,7 +1758,7 @@ INSVCStaggeredHierarchyIntegrator::applyGradientDetectorSpecialized(const Pointe
         d_hier_math_ops->curl(d_Omega_idx, d_Omega_var, d_U_scratch_idx, d_U_var, d_U_bdry_bc_fill_op, error_data_time);
         tagCellsByVorticityMagnitude(level_number, d_Omega_idx, tag_index);
 
-        for (int ln = 0; ln < d_hierarchy->getFinestLevelNumber(); ++ln)
+        for (int ln = 0; ln <= d_hierarchy->getFinestLevelNumber(); ++ln)
```
boyceg, line 1761: Why do we need to evaluate the gradient detector on the finest level of the patch hierarchy?

drwells: We could fix this in a follow-up, but `INSVCStaggeredHierarchyIntegrator` assumes that the finest patch level is allocated. I'd prefer to leave that as-is for now - allocating one more vector once isn't that big of a deal.

boyceg (approved):

PR #1725, opened 2024-07-01
change the velocity interpolation correction term.

ryanleaf1000:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
This change should change the results generated by test code, and so you are going to need to update the tests. (If it doesn't cause tests to fail, then that means that we need to expand our IIM test coverage!)

This change also will make it difficult to replicate prior work exactly. So, my suggestion would be to make it an option, and to default to the prior implementation for now. We should do some checks to see if one approach is better than the other and, if there is a clear winner, update the default (if needed).

Finally, please include some comments in the code about what is being implemented here. (It might be helpful also to include comments about what the old implementation is doing, so that the differences are easier to see.)

ryanleaf1000:
Yes, Let me try to add these.

ryanleaf1000:
Hello Boyce, 
I ran a convergence test from taylor-couette 2d example to compare the two interpolation scheme:
The one I added is from:
https://epubs.siam.org/doi/abs/10.1137/080712970?casa_token=LevsVOzF2m8AAAAA:UAgURw3W4dny0SrV2Vv30tt8-tOtlhsuzCCLfair1jLPyIqlL0ob-qnFaaL7oaweopiA2lDOXKo
The one currently in the IBAMR I believe is from:
https://arxiv.org/abs/1812.06840
The convergence test results indicate that using Tan's correction term yields a smaller surface velocity L2 error compared to the existing interpolation scheme. However, there is not much difference in the Eulerian error between using different correction terms in interpolation. Should we keep the existing correction scheme as the default, or switch to Tan's scheme as the default?

boyceg:
For this PR, it is important to add automatic tests for both versions of the interpolation scheme.

Let's just make it a runtime option that can be toggled by the user, and default to the original scheme for now. If we find some compelling differences in terms of accuracy or stability, we can always switch.

The other related thing is if there is a way to use this kind of a scheme to evaluate pressure and shear stress on the boundary. If they can all be handled with the same scheme, then it becomes more compelling to me to switch the default.

boyceg:
@drwells, Am I remembering correctly that we are going to drop CENTOS 7 from the CI? The CI is giving this error for that OS:
```
/usr/bin/docker exec  e7d06e75dc0bb66e7d706aebc3f90a365c381693cca23ff60a2457bdb844c0ca sh -c "cat /etc/*release | grep ^ID"
/__e/node20/bin/node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.[21](https://github.com/IBAMR/IBAMR/actions/runs/9863343499/job/27235905016?pr=1725#step:3:22)' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by /__e/node20/bin/node)
/__e/node20/bin/node: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by /__e/node20/bin/node)
```

src/IB/IIMethod.cpp
```diff
@@ -1317,21 +1317,48 @@ IIMethod::interpolateVelocity(const int u_data_idx,
...
+                                }
 #if (NDIM == 2)
-                                coeff_vec =
-                                    VectorValue<double>(interpCoeff[ic[0]][ic[1]][0], interpCoeff[ic[0]][ic[1]][1]);
-                                Ujump[ic[0]][ic[1]][d] = dx[0] * w[0][ic[0] - ic_lower[0]] * w[1][ic[1] - ic_lower[1]] *
-                                                         (coeff_vec * du_jump);
+                                if (d_use_Tan_velocity_correction)
+                                {
+                                    // Use velocity correction term from
+                                    // https://epubs.siam.org/doi/abs/10.1137/080712970?casa_token=LevsVOzF2m8AAAAA:UAgURw3W4dny0SrV2Vv30tt8-tOtlhsuzCCLfair1jLPyIqlL0ob-qnFaaL7oaweopiA2lDOXKo
```
boyceg, line 1329:
```suggestion
                                    // https://epubs.siam.org/doi/abs/10.1137/080712970?
```
Please also add some information in a comment about how the two approaches differ from each other.

src/IB/IIMethod.cpp
```diff
@@ -4414,6 +4441,9 @@ IIMethod::getFromInput(Pointer<Database> db, bool /*is_from_restart*/)
     if (db->isBool("use_pressure_jump_conditions"))
         d_use_pressure_jump_conditions = db->getBool("use_pressure_jump_conditions");
 
+    if (db->isBool("use_Tan_velocity_correction"))
```
boyceg, line 4444: I think the option name should have something about velocity interpolation in the name. Also, is there a clear geometrical difference between the two approaches? It might be clearer to use that in the option name than the name of the first author of the paper.

src/IB/IIMethod.cpp
ryanleaf1000: Hello Boyce, I am not sure what comment I should include in the code. Could you please give me some suggestions? I could write down my understanding of difference here:
The difference between those two schemes is just the correction terms in velocity interpolation operator:
See Eqn. (3.12)-(3.15) in Sec. 3.3 in
https://epubs.siam.org/doi/epdf/10.1137/080712970
See Eqn. (27)-(28) in Sec. 3.5 in 
https://arxiv.org/pdf/1812.06840
The first version Tan derived base on generalized taylor expansiion. And it is second order according to the author and this is verified by Me and Michael's code.
I don't fully understand Amin's interpolation scheme (the one currently in the IBAMR). But according to what Amin wrote in the paper:
"This modified bilinear interpolation is similar to the approach of Tan et al. [72, 73]. An important difference
is in the value used for the distance from the Cartesian grid locations to the interface in the Taylor series
expansion of the correction terms. The approach by Tan et al. uses the distance between the grid points
and the interfacial point (|r| in Fig. 5). Here, this is modified by projecting the location vector r to the
normal vector n and using |r · n| instead. Taking into account for the normal vector naturally provides a
more accurate approximation to the actual distance between the grid point and the interface"

boyceg: Can you check to see if the current scheme gives the right answer for a piecewise linear field with a derivative discontinuity along the interface?

ryanleaf1000: Sure!

src/IB/IIMethod.cpp
boyceg: You may need to consider different interface positions and orientations.

src/IB/IIMethod.cpp
boyceg: Please make sure to test multiple interface positions and orientations.

ryanleaf1000: Hello Boyce, 
I constructed an interface that not parallel to either axis. Then I constructed Tan's scheme and made sure it give exact answer  for a piecewise linear field with a derivative discontinuity along the interface. 
Then I modified and verified the interpolation scheme according from the paper:
See Eqn. (27)-(28) in Sec. 3.5 in
https://arxiv.org/pdf/1812.06840
which do not seems to give the exact answer.
However if I understand the interpolation code in IBAMR correctly, I think it is not equivalent to the one Amin wrote in the paper neither equivalent to Tan's version. And I did the same tests with my understanding of what is in IBAMR. I got the exact answer with multiple orientation and interface positions. I think the one in the IBAMR use the interface condition as a given condition implicitly in the construction of correction terms which give equivalent answer from Tan's scheme if velocity field only have normal derivative jump across the interface. 
So the difference between the two schemes is:
The exisiting interpolation scheme in IBAMR would be second order accurate and equivalent to Tan's scheme if the velocity field only has normal derivative jumps. Tan's scheme would be second-order accurate with velocity field with jumps in both tangential and normal directions.

boyceg: OK --- I think that we only encounter normal derivative discontinuities, so the current scheme seems OK, but it is unfortunate that it doesn't agree with what is in the paper.

@AminKolah, do you have any time to look at this? My inclination right now is just to switch the implementation to use Tan's scheme and not worry about maintaining the original scheme (even as an option).

AminKolah:
> For this PR, it is important to add automatic tests for both versions of the interpolation scheme.
> 
> Let's just make it a runtime option that can be toggled by the user, and default to the original scheme for now. If we find some compelling differences in terms of accuracy or stability, we can always switch.
> 
> The other related thing is if there is a way to use this kind of a scheme to evaluate pressure and shear stress on the boundary. If they can all be handled with the same scheme, then it becomes more compelling to me to switch the default.

Hi guys, sorry I'm late to the conversation. I also agree—it'd be neat to get rid of our current ad-hoc coefficient for the fluid stress-related quantities (pressure and WSS), relying on the the same consistent interpolation operator as in the velocity interpolation.

AminKolah:
> Hello Boyce, I ran a convergence test from taylor-couette 2d example to compare the two interpolation scheme: The one I added is from: https://epubs.siam.org/doi/abs/10.1137/080712970?casa_token=LevsVOzF2m8AAAAA:UAgURw3W4dny0SrV2Vv30tt8-tOtlhsuzCCLfair1jLPyIqlL0ob-qnFaaL7oaweopiA2lDOXKo The one currently in the IBAMR I believe is from: https://arxiv.org/abs/1812.06840 The convergence test results indicate that using Tan's correction term yields a smaller surface velocity L2 error compared to the existing interpolation scheme. However, there is not much difference in the Eulerian error between using different correction terms in interpolation. Should we keep the existing correction scheme as the default, or switch to Tan's scheme as the default?

This is interesting, Qi.... By 'smaller error in the Lagrangian velocity,' are we only talking about the absolute error or the overall order of accuracy? It makes sense that the errors for Eulerian velocity are not affected, given that the normal derivative is all we care about here. However, for certain applications involving slip velocity, we might want to have a more generalized formula.

src/IB/IIMethod.cpp
AminKolah: @boyceg I think Qi nailed it. The difference is pretty straightforward, but I totally agree the twisted description in the paper made it a bit confusing. It looks like there's an extra normal vector factor in my formulation that needs to go. Sorry about that! Looks like we've fixed at least a couple of errors in the first IIM paper by now, but thankfully, those same mistakes didn't make it into the rigid and flexible papers (lesson learned!)."

AminKolah: I'll go ahead and fix that error in the arxiv version.  There were also a couple of other things that we have already fixed in the arxiv (a sign error in the jump and the inverted surface Jacobian definition). Maybe it makes sense at this point to publish an erratum that just fix those in the Eqs?  @boyceg Let me know what you think. I have some time now to work on this if you and the rest of the co-authors agree.

boyceg: Yeah if there are errors in formulae it might be good to publish an erratum.

ryanleaf1000: @AminKolah Hello Amin, 
As I tried to formulate our new paper with small mfac. I had difficulties with grid convergence tests and reproduce the convergence rate as state in the paper. I am wondering if you could double check and reconfirm the convergence tests results in the paper with finner grid?

AminKolah: Yes, that seems right. We all agree that there is deteriorated convergence for a finer grid at this point, at least when using the same form of penalty parameter as in the original work.

AminKolah: I have something coming up on Monday, but I should be able to add my comments on the draft by next Friday, or just on the grid convergence if that's the only part you want me to look at? To better follow the flow of info in the manuscript, it'd be good to have all the important sections in place before reviewing. Oh, and I believe your overleaf account is not premium, so I'm unable to add edits with track changes.  That is however not a problem. I’ll download the PDF and add my comments directly to the file instead. In any case, we can continue discussing this via email.

MichaelFacci:
Hi all, I am finally back to the land of wifi, eduroam, and civilization. I am going to peruse Qi's changes for velocity interpolation.

drwells:
@ryanleaf1000 This looks mostly done - there are plenty of tests for this change. Can you rebase and wrap this up next week?

ryanleaf1000:
> @ryanleaf1000 This looks mostly done - there are plenty of tests for this change. Can you rebase and wrap this up next week?

Sure. I can work on this.

boyceg:
@ryanleaf1000 looks like something is going wrong here...

ryanleaf1000:
yeah, I am trying to figuring out

boyceg (approved):
Looks good to me. @drwells or @abarret could one of you take a look?

boyceg:
@ryanleaf1000 can you address the checklist? This changes functionality so there should be a changelog entry.

ryanleaf1000:
> @ryanleaf1000 can you address the checklist? This changes functionality so there should be a changelog entry.

Sorry I completely missed this. How do I add a changelog entry?

drwells:
I'm out of the IIM loop, but this PR will significantly change the output of every app using `IIMethod`. Is that what we want to do?

boyceg:
I think that we have convinced ourselves that the method in the PR is more accurate than what is currently provided. It also is the approach that @MichaelFacci is using as he works towards getting multiple intersections working, which is critical for handling near contact.

From the standpoint of reproducibility, however, it might be worthwhile to have an option that controls which method is used?

boyceg:
I'm not sure that it is _always_ better, but I think that this implementation has a clearer theoretical foundation than the scheme that was originally implemented.

MichaelFacci:
Yes I can confirm that this agrees with the Tan et al paper. It is good to merge!

Get Outlook for Android<https://aka.ms/AAb9ysg>

PR #1727, opened 2024-07-08
Delete the CENTOS 7 runner.

drwells:
See also

https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#supported-architectures-and-operating-systems-for-self-hosted-runners

This OS is over 10 years old and glibc 2.27 (and older) are no longer supported by GitHub actions' infrastructure.

N/A, this is just infrastructure.


drwells:
@boyceg yup - that machine needs to be retired. We can move past gcc 4.8 now :)

boyceg (approved):

PR #1728, opened 2024-07-17
start removing templates on NDIM

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
So before we do something like this I think we need to agree on the workflow we're going to use for upgrading to SAMRAI 4 - via the email, I think both Amneet and I would prefer that these kinds of upgrades are reproducible via script. Can you put the commands you used in a script instead of committing things here?

boyceg:
> So before we do something like this I think we need to agree on the workflow we're going to use for upgrading to SAMRAI 4 - via the email, I think both Amneet and I would prefer that these kinds of upgrades are reproducible via script. Can you put the commands you used in a script instead of committing things here?

Yes, this is all being done using a simple script (currently named `modernize.sh`).

drwells:
GitHub's UI is really struggling with a patch this large. Could you put that script in a separate commit?

boyceg:
I am almost done purging `NDIM` --- I'll make a separate branch with just the script and interface header once it is done.

boyceg:
@drwells see #1730 for just the basic scripts and headers.

PR #1735, opened 2024-07-18
Update clang-format scripts.

drwells:
based on updates to the deal.II versions of these scripts.

This is just an infrastructure update so the normal list doesn't apply.

### DIFF
--- a/scripts/formatting/compile-clang-format
+++ b/scripts/formatting/compile-clang-format
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# based on the compile_clang_format deal.II script and licensed under the LGPL
+# based on the compile-clang-format deal.II script and licensed under the LGPL
 
 #
 # This script downloads, compiles and installs the clang-format binary. The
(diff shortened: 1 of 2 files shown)


boyceg (approved):

boyceg:
Is there a canonical `clang-format` version that this will set up automatically?

drwells:
Yup - we require clang-format 16.

boyceg:
Can I get it to build and use the right version automatically? I seem to have clang-format 18.1.8 on my systems.

drwells:
Yup - you can use either one (`compile-clang-format` or `download-clang-format`) to install the correct version and CMake will pick it up.

boyceg:
How do I do it without using CMake? 😁

drwells:
We don't support that. People need to move to CMake.

drwells:
Well, technically, you can run `./scripts/formatting/indent` - the only thing which doesn't work is `make indent`.

drwells:
autotools can also pick it up - see #1700.

boyceg:
> autotools can also pick it up - see #1700.

Now that's what I'm talking about! 😁

PR #1742, opened 2024-08-02
Acoustic streaming branch

amneetb:
This branch is rebased against latest master. I am pushing it here for the CI to do its testing. Once potential bugs related to CI are fixed I will redo the history for an ease of review. The good news is that the acoustic formulation is more or less worked out and this branch contains (almost) stable code. 

### IBAMR Pull Request Checklist
- [ ] Does the test suite pass?
- [ ] Was clang-format (i.e., `make indent`) run? For more information see
      `scripts/formatting/README.md`.
- [ ] Were relevant issues cited? Please remember to add `Fixes #12345` to close
      the issue automatically if we are fixing the problem.
- [ ] Is this a change others will want to know about? If so, then has a
      changelog entry been added?
- [ ] If new data structures have been added to a class then have they been set
      up appropriately for restarts? If so, ensure that the restart version
      number is incremented.
- [ ] Does this change include a bug fix or new functionality? If so, a new test
      or tests should be added. New tests should run quickly (less than a minute
      in release mode). If possible, an older test should gain a new option so
      that we do not need to compile more test executables.
- [ ] Did you (if your account has permission to do so) set relevant labels on
      GitHub for the pull request?

PR #1744, opened 2024-08-06
Acoustic rewrite history

amneetb:
Supersedes #1742. This has commits organized properly.  

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
There's too much in this PR for me to reasonably review - either someone else is going to handle it or it needs to be chopped down to about 1000 lines per PR.

amneetb:
> There's too much in this PR for me to reasonably review - either someone else is going to handle it or it needs to be chopped down to about 1000 lines per PR.

That's our perpetual dilemma :-) Adding new physics does not fit as cleanly in Git as changing existing code/software infrastructure. The commits I have committed are logical in the sense that if someone goes over them one by one in a specific order they will immediately understand the thought I have put in organizing them.

drwells:
We don't agree on this and that's fine - as principal developers both of our viewpoints are valid. That doesn't change the restrictions on my time available for reviewing patches.

boyceg:
@amneetb, there are changes to core classes along with the new stuff. Please break these up into multiple PRs. I request one for each core class (looks like it is just `GeneralOperator`), one for new operator implementations (again looks like it is just one), and then at least one for the new solver.

boyceg:
I've been using Codex to look across this series of PRs, and I would like to help to get them merged. However, there are some API changes that I think are probably too broad. This affects several branches, so here in #1744 seems like the right place to agree on the design.

The main concern is that we now have both `setPoissonSpecifications()` and `setProblemSpecification()`, without a clear relationship between them. Some implementations use only the new specification pointer even though the older setters remain available. The generic pointer also leaves the required specification type, ownership, and lifetime implicit.

My initial preference would be to:

- Keep `GeneralOperator` and `GeneralSolver` unchanged.
- Introduce typed coefficient specifications in the operator and solver families that need them, with separate specifications where operators and preconditioners require different information.
- Preserve the behavior of existing Poisson and Stokes configuration interfaces. If we keep an older setter during migration, it should update the same internal configuration as the new interface.
- Define when configuration must be supplied, how it reaches operators and preconditioners, and which changes require reinitialization. This should work with the existing solver factories.
- Make supported coefficient modes explicit: constants, patch-data indices, and omitted or zero terms where appropriate.

For the specification objects themselves, I would prefer keeping the small coefficient descriptors by value, while documenting the lifetime requirements for the patch data they reference.

Does this seem reasonable? Once we agree on the design, I can use Codex to help revise the affected PRs consistently. We can then address some remaining acoustic-specific API questions.

amneetb:
@boyceg: Seems reasonable. Can we work directly with this branch? https://github.com/IBAMR/IBAMR/tree/acoustic_streaming/stokes_drift

This is what we are using to put and test all of the acoustic code for the past 1.5 years.

boyceg:
Can I (try to) pull those changes here? My goal is to get this all back into "mainline" IBAMR.

amneetb:
Sure! This branch has all the latest changes so it would be good to have them pulled in.

boyceg:
@amneetb, can we try to wrap up any lingering LS stuff (#1977, #1978, #1979) and the phase change updates/tests/examples (I am still working on those...) before tackling this one?

If we don't need #1977, #1978, #1979, we can just close those. I don't know enough about the LS implementation to know if there is a there there...

amneetb:
@boyceg : I looked into these and they look good. @YavkreetS02 : Can you take a look closely into these LS fixes and see if these changes affect any of your work on phase change and/or VOF?

amneetb:
@Raamkrishnan : Can you also take a look into LS related changes?

PR #1745, opened 2024-08-15
acoustic_streaming/PR1

amneetb:
Adding ProblemSpecification to General{Operator,Solver} classes

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/GeneralOperator.h
+++ b/ibtk/include/ibtk/GeneralOperator.h
@@ -39,6 +39,11 @@ class SAMRAIVectorReal;
 } // namespace solv
 } // namespace SAMRAI
 
+namespace IBTK
+{
+struct ProblemSpecification;
+} // namespace IBTK
+
 /////////////////////////////// CLASS DEFINITION /////////////////////////////
 
 namespace IBTK
@@ -122,6 +127,16 @@ public:
      * \brief Get the HierarchyMathOps object used by the operator.
      */
     virtual SAMRAI::tbox::Pointer<HierarchyMathOps> getHierarchyMathOps() const;
+
+    /*!
+     *  \brief Set the ProblemSpecification object used by the operator.
+     */
+    virtual void setProblemSpecification(const ProblemSpecification* problem_spec);
+
+    /*!
+     *  \brief Set the ProblemSpecification object used by the operator.
+     */
+    virtual const ProblemSpecification* getProblemSpecification() const;
 
     /*!
      * \brief Compute \f$y=F[x]\f$.
@@ -303,6 +318,9 @@ protected:
 
     // Logging configuration.
     bool d_enable_logging = false;
+
+    // Problem specs (coefficients) for the operator
+    const ProblemSpecification* d_problem_spec = nullptr;
 
 private:
     /*!
--- a/ibtk/include/ibtk/ProblemSpecification.h
+++ b/ibtk/include/ibtk/ProblemSpecification.h
@@ -1,0 +1,120 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2011 - 2023 by the IBAMR developers
+// All rights reserved.
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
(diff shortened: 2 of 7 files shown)


boyceg:
Thanks, Amneet! I just had knee surgery but will try to take a look before the semester gets too crazy — FDOC is Monday! 😢

boyceg:
@amneetb, if we add this to the `GeneralOperator` interface, then I think that we also should change all of the current implementations to support it instead of whatever operator-specific mechanisms are being used to specify problem coefficients.

For instance, with this change, I think that `LaplaceOperator::setPoissonSpecifications()` should become `GeneralOperator::setProblemSpecification()`. But then we would also need to introduce a new subclass of `ProblemSpecification` to replace the existing `PoissonSpecifications` class. I am not totally opposed to doing this, but it would entail a lot of code changes.

One other consideration is that not every operator is configurable --- for instance, the PETSc operator "wrapper" classes.

I think that right now, my preference would be to keep operator-specific configuration interfaces in the subclasses instead of in the base class unless there is a compelling reason to redo all of the problem specification interfaces in the library.

ibtk/include/ibtk/ProblemSpecification.h
```diff
@@ -0,0 +1,120 @@
...
+{
+};
+
+/*!
+ * \brief Class VCViscousOpSpec is a light-weight structure holding
+ * values of VCViscousOp coefficients.
+ */
+struct VCViscousOpSpec : public ProblemSpecification
+{
+    VCViscousOpSpec()
```
boyceg, line 46: Please put `inline` implementations in `include/ibtk/private/VCViscousOpSpec-inl.h`.

ibtk/include/ibtk/ProblemSpecification.h
```diff
@@ -0,0 +1,120 @@
...
+
+    bool d_D_is_const = false, d_C_is_const = false;
+
+}; // VCViscousOpSpec
+
+/*!
+ * \brief Class VCViscousDilatationalOpSpec is a light-weight structure holding
+ * values of VCViscousDilatationalOp coefficients.
+ */
+struct VCViscousDilatationalOpSpec : public VCViscousOpSpec
```
boyceg, line 79: I would prefer to stick with one class per header.

boyceg (changes requested):
We need either to roll out `ProblemSpecification` implementations for all implementations of `GeneralOperator`, or move this `ProblemSpecification` interface into a subclass.

My preference would be:
1. move it into a subclass and
2. not create a general base class for all problem specifications

I think that having a generic base class for all problem specifications objects might be worthwhile, but if we did it, we should use that everywhere and not only in this one operator.

A specific problem with the current PR is that `LaplaceOperator` will now have both `LaplaceOperator::setProblemSpecification()` as well as `LaplaceOperator::setPoissonSpecifications()`, but only the second one actually controls operator setup.

ibtk/include/ibtk/GeneralOperator.h
```diff
@@ -123,6 +128,16 @@ class GeneralOperator : public virtual SAMRAI::tbox::DescribedClass
      */
     virtual SAMRAI::tbox::Pointer<HierarchyMathOps> getHierarchyMathOps() const;
 
+    /*!
+     *  \brief Set the ProblemSpecification object used by the operator.
+     */
+    virtual void setProblemSpecification(const ProblemSpecification* problem_spec);
```
drwells, line 134: I agree that this should be a feature of derived classes. Furthermore - we need those parameters to actually do anything, right? They should be constructor arguments so we don't have half-constructed classes.

drwells:
@amneetb What's the plan for addressing review comments and moving forward with this one? We need to integrate these features in a way that doesn't make the library more complicated.

amneetb:
I think I need to act on the former suggestion of Boyce, i.e., "We need either to roll out ProblemSpecification implementations for all implementations of GeneralOperator"

Since this is a big undertaking, I am avoiding it in the interest of doing some research and writing a few papers at my end. However, I plan to resume this work once the phase change PRs 6 & 7 are merged (I think PR 6 is done as per Boyce's suggestions).

PR #1747, opened 2024-08-15
acoustic_streaming/PR3

amneetb:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/KrylovLinearSolverPoissonSolverInterface.h
+++ b/ibtk/include/ibtk/KrylovLinearSolverPoissonSolverInterface.h
@@ -69,6 +69,12 @@ public:
     void setPoissonSpecifications(const SAMRAI::solv::PoissonSpecifications& poisson_spec) override;
 
     /*!
+     * \brief Set the IBTK::ProblemSpecification object used to specify
+     * the coefficients for the scalar-valued or vector-valued Laplace operator.
+     */
+    void setProblemSpecification(const IBTK::ProblemSpecification* problem_spec) override;
+
+    /*!
      * \brief Set the SAMRAI::solv::RobinBcCoefStrategy object used to specify
      * physical boundary conditions.
      *
--- a/ibtk/include/ibtk/GeneralOperator.h
+++ b/ibtk/include/ibtk/GeneralOperator.h
@@ -39,6 +39,11 @@ class SAMRAIVectorReal;
 } // namespace solv
 } // namespace SAMRAI
 
+namespace IBTK
+{
+struct ProblemSpecification;
+} // namespace IBTK
+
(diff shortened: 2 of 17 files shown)


ibtk/include/ibtk/KrylovLinearSolverPoissonSolverInterface.h
```diff
@@ -68,6 +68,12 @@ class KrylovLinearSolverPoissonSolverInterface : public PoissonSolver
      */
     void setPoissonSpecifications(const SAMRAI::solv::PoissonSpecifications& poisson_spec) override;
 
+    /*!
+     * \brief Set the IBTK::ProblemSpecification object used to specify
+     * the coefficients for the scalar-valued or vector-valued Laplace operator.
+     */
+    void setProblemSpecification(const IBTK::ProblemSpecification* problem_spec) override;
```
boyceg, line 75: Here I think you can see the problem with having both `setPoissonSpecifications` and `setProblemSpecification` --- I think we could have one or the other but should not use both in the same interface.

boyceg (changes requested):
There will be some changes needed here based on however #1745 is resolved.

ibtk/include/ibtk/KrylovLinearSolverPoissonSolverInterface.h
amneetb: @boyceg: Based on these three PRs that require setting `ProblemSpecification` which derived class should I delegate `setProblemSpecification()` to?

or are you recommending removing `setPoissonSpecifications` entirely? For the latter would you want to re-design the interfaces properly? I can follow your lead and rebase PR1 (and others) on it.

boyceg:
I think there are basically two different routes:
1. Set up a different class hierarchy for your operator(s) and solver(s) under `LinearOperator` / `LinearSolver` and then set up an independent class for its specifications.
2. Set up a unified problem specifications class with specializations for the current and new operators. In this case, I think that the basic approach would be to move `PoissonSpecifications` from `SAMRAI` into `IBTK`, keeping its current interface for now, and add a new base class that could also be derived for your operator(s) and solver(s). Then your operator(s) and solver(s) could go under, e.g., `LaplaceOperator` and `PoissonSolver`.
The second approach seems like it is probably more work but would reduce the number of different interfaces --- which seems good.
Aaron also works a lot with these kinds interfaces --- @abarret, do you have an opinion?

amneetb:
I like the second approach. If @abarret could put out an interface and change the existing solver and operator infrastructure to use `ProblemSpecification` instead of `PoissonSpecification` that would be really great. This will make our existing solvers much much more general, and this is where we are going. Days of constant coefficient solvers are long gone!

PR #1762, opened 2024-09-07
add an option to skip the velocity solve in the projection preconditioner

boyceg:
If viscosity or time step size is very small, then we can skip solving the velocity block in the projection preconditioner without needing to take many more overall Stokes solver iterations.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/navier_stokes/StaggeredStokesProjectionPreconditioner.cpp
+++ b/src/navier_stokes/StaggeredStokesProjectionPreconditioner.cpp
@@ -83,10 +83,13 @@ static Timer* t_deallocate_solver_state;
 
 StaggeredStokesProjectionPreconditioner::StaggeredStokesProjectionPreconditioner(
     const std::string& object_name,
-    Pointer<Database> /*input_db*/,
+    Pointer<Database> input_db,
     const std::string& /*default_options_prefix*/)
-    : StaggeredStokesBlockPreconditioner(/*needs_velocity_solver*/ true,
-                                         /*needs_pressure_solver*/ true)
+    : StaggeredStokesBlockPreconditioner(
+          /*needs_velocity_solver*/ input_db ?
+              (input_db->keyExists("skip_velocity_solve") ? !input_db->getBool("skip_velocity_solve") : true) :
+              true,
+          /*needs_pressure_solver*/ true)
 {
     GeneralSolver::init(object_name, /*homogeneous_bc*/ true);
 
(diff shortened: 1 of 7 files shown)


boyceg:
@drwells, @marshallrdavey, @labdala: I think that this should speed up the heart model. Basic settings are:
```
   stokes_solver_type = "PETSC_KRYLOV_SOLVER"
   stokes_precond_type = "PROJECTION_PRECONDITIONER"
   stokes_solver_db {
      ksp_type = "richardson"  // switch to "fgmres" if it takes more than a couple iterations
   }
   stokes_precond_db {
      skip_velocity_solve = TRUE
   }
```

amneetb:
@boyceg : while you are at solvers, would you want to take a stab at changing the interface from using `PoissonSpecification` to `ProblemSpecification` consistently in the library? Perhaps you can come with a perl script 😀

src/navier_stokes/StaggeredStokesProjectionPreconditioner.cpp
```diff
@@ -215,10 +218,20 @@ StaggeredStokesProjectionPreconditioner::solveSystem(SAMRAIVectorReal<NDIM, doub
...
+    }
+    else
+    {
+        TBOX_ASSERT(!d_U_problem_coefs.cIsZero());
+        d_velocity_data_ops->scale(U_vec->getComponentDescriptorIndex(0),
+                                   1.0 / d_U_problem_coefs.getCConstant(),
+                                   F_U_vec->getComponentDescriptorIndex(0));
+    }
 
     // (2) Solve the pressure sub-problem.
```
amneetb, line 236: @boyceg : are you using a diagonal preconditioning here?

boyceg: Not really, this is just making sure that the the units are correct for evaluating the RHS in the next step.

boyceg: It would be very easy to turn this into a proper diagonal preconditioner in the constant coefficient case by adding an additional term here.

boyceg: It would not be too hard to turn this into a diagonal preconditioner, but it would require looking up the grid spacings etc.

abarret (changes requested):
The option should be specified in the documentation. In particular, it should mention that the viscous term is ignored.

boyceg:
> The option should be specified in the documentation. In particular, it should mention that the viscous term is ignored.

How's this?

abarret (approved):

PR #1763, opened 2024-09-10
enable SAMRAIScopedVectorDuplicate to do type conversions for cell-centered data

boyceg:
@drwells, I am sure that you are going to love this one. 🤣 

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/major/20240910BoyceGriffith
+++ b/doc/news/changes/major/20240910BoyceGriffith
@@ -1,0 +1,3 @@
+New: initial support to convert types stored in SAMRAIVectorReal objects.
+<br>
+(Boyce Griffith, 2024/09/10)


drwells:
While we definitely need to solve this problem (I did similar things in the redbud project) there is a significant mismatch between 'doing linear algebra and type conversions cleanly' and what SAMRAI expects as a data storage format. I'll find some time this week to sit down and think about it some more.

boyceg:
Templating on the floating point type is not helping things.

I will keep experimenting with multiprecision solvers and can just put stuff together into an integration branch to see if it actually helps with performance. If it doesn't --- probably we should just drop this PR.

boyceg:
@drwells or @abarret, I am doing something dumb, but I cannot figure out how to setup the test outputs. Can you see what I have done and tell me what I should change to fix?

boyceg:
Ah yes, I am a moron. Fixed now.

boyceg:
@drwells & @abarret, this is ready for review if either of you want to take a look at it.

Also fine with me to wait and see if the multiprecision solvers are actually faster before taking any action. I'm hoping to have a prototype going by next week in an effort to speed up @labdala's heart simulations.

drwells:
I think we should do some of those experiments first. If we have a mixed-precision solver set up it may provide a better way to handle these conversions: e.g., it can create the new variables itself instead of relying on those being hard-coded inside this class.

boyceg:
> I think we should do some of those experiments first. If we have a mixed-precision solver set up it may provide a better way to handle these conversions: e.g., it can create the new variables itself instead of relying on those being hard-coded inside this class.

It is working now!!! 😁

PR #1766, opened 2024-09-22
conditionally register force for restart even if it is not being output

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
@amneetb, this seems to fix the restart error that shows up with `ConstraintIBMethod`.

I am having trouble getting a restart test to work for ConstraintIB. Could you take a look and add a test to this branch?

boyceg:
@drwells / @abarret , I think that the restart logic for the Eulerian force was buggy in the case that the user did not request output. Can you take a look?

amneetb:
> @amneetb, this seems to fix the restart error that shows up with `ConstraintIBMethod`.
> 
> 
> 
> I am having trouble getting a restart test to work for ConstraintIB. Could you take a look and add a test to this branch?

Sure. Will add a test for restart later tonight or tomorrow.

amneetb:
Are there some examples of testing restart capability with attest in the library? 

I am guessing attest needs to run a simulation and then restart it again from a restore point.

boyceg:
> Are there some examples of testing restart capability with attest in the library?
> 
> I am guessing attest needs to run a simulation and then restart it again from a restore point.

Yes --- input files with `restart` in them test restarts.

The thing that I am having trouble with is that the output for the ConstraintIB test seems to depend (in a way that I was having trouble sorting out) on some files that get written during the test.

amneetb:
I see. Can you point me to any one of it in the library? I can follow that to create a test for ConstraintIB.

boyceg:
Sure: `CIB/cib_plate.restart=5.mpirun=4.input`

amneetb:
@boyceg : I added a test file for checking restart. When I do numdiff on local output and output committed to GitHub, it says the files are the same 

```
~/Softwares/NumDiff/bin/numdiff output oscillating_rigid_cylinder.restart=40.output

+++  Files "output" and "oscillating_rigid_cylinder.restart=40.output" are equal
``` 

However, when I use attest for this test, it says DIFF FAILED. Not sure what is going on.  Can you take a look?

boyceg:
Unfortunately this is what I ran into and why I had trouble putting the test together. 😁 

As I mentioned above, it seems like the test program output itself depends on some of the files that are generated by the test, which made it hard for me to get it to generate reproducible output when restarting.

amneetb:
This also did not work with attest. I am not sure what attest is doing or comparing against with restarted runs.

drwells:
How exactly did you generate the new output file? `attest` will do roughly the following in this case:

```
> cd /tmp/testdir
> mpirun -n 4 ./cib_plate ./cib_plate.restart=5.mpirun=4.input
> mpirun -n 4 ./cib_plate ./cib_plate.restart=5.mpirun=4.input ./restart 5
> numdiff output cib_plate.restart=5.output
> cd -
> rm -r /tmp/testdir
```
so you need to generate the output file from the fifth restart for it to match what attest expects.

amneetb:
This test simply copies the content of two text files into `output`; see
https://github.com/IBAMR/IBAMR/blob/master/tests/ConstraintIB/oscillating_rigid_cylinder.cpp#L351-L366

When the simulation is re-run, the behavior is to append data to the text files rather than create a new one (and wiping out the old content). I tried both ways (appended output vs non-appended output from restarted run), but attest still fails.

amneetb:
At least one of them should have worked.

tests/ConstraintIB/oscillating_rigid_cylinder.restart=40.input
```diff
@@ -0,0 +1,340 @@
...
+
+// visualization dump parameters
+   viz_writer            = "VisIt","Silo"
+   viz_dump_interval     = int(0.01 * (1.0/FREQUENCY)/DT  + 0.5)  //zero to turn off
+   viz_dump_dirname      = "viz_cylinder2d"
+   visit_number_procs_per_file = 1
+
+// restart dump parameters
+   restart_interval      = 20                 // zero to turn off
+   restart_write_dirname = "restart_IB2d"
```
drwells, line 277:
```suggestion
   restart_write_dirname = "restart"
```
The test suite requires that this directory always have the same name (`"restart"`)

boyceg:
> @drwells / @abarret , I think that the restart logic for the Eulerian force was buggy in the case that the user did not request output. Can you take a look?

(As a reminder, this is what we are trying to fix here.)

PR #1779, opened 2025-01-20
Fix some out-of-date autotools files.

drwells:
I'll be frank - dealing with two build systems is not a good use of our time anymore. We should either automate this kind of check or delete autotools.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -186,8 +186,8 @@ am__DIST_COMMON = $(srcdir)/Makefile.in \
 	$(top_srcdir)/config/install-sh $(top_srcdir)/config/ltmain.sh \
 	$(top_srcdir)/config/make.inc.in $(top_srcdir)/config/missing \
 	config/compile config/config.guess config/config.rpath \
-	config/config.sub config/install-sh config/ltmain.sh \
-	config/missing
+	config/config.sub config/depcomp config/install-sh \
+	config/ltmain.sh config/missing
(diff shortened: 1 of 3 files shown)


boyceg:
I think that getting rid of autotools is OK at this point.

drwells:
I'd like to drop it but that definitely shouldn't happen in this release.

drwells:
I can run CMake with Ninja in 11s on my old desktop, whereas running `./configure` takes 43s. The CMake version contains a lot of extra checks which would be extraordinarily tedious to implement with autotools (i.e., it does a lot more useful work in a lot less time).

boyceg (approved):

boyceg:
> I can run CMake with Ninja in 11s on my old desktop, whereas running `./configure` takes 43s. The CMake version contains a lot of extra checks which would be extraordinarily tedious to implement with autotools (i.e., it does a lot more useful work in a lot less time).

I'm OK with dropping autotools after the next release.

boyceg:
(And I am someone who is still using autotools builds. 😅)

amneetb:
We are also using autotools and would like to continue with them.

drwells:
If we are going to keep it then we need to maintain it - e.g., write checks to prevent the bug on this patch (which can break `master`s compilation). Are you willing to do that or provide other maintenance resources?

drwells:
Wanna approve this one? :)

drwells:
Edit: it was approved already!

PR #1781, opened 2025-01-29
Require libMesh 1.7 or newer.

drwells:
This version has been out for two years - it is time to upgrade.


drwells:
@abarret This is the last non-bookkeeping PR for the release - can you take a look? Once this is in we can collate the changelog and tag a release candidate.

abarret (approved):
If we are requiring 1.7, do we want to remove the relevant version checks in the code itself? E.g. in `FEMapping.cpp`
```c++
#if LIBMESH_VERSION_LESS_THAN(1, 4, 0)
    TBOX_ASSERT(d_n_nodes <= 27);
#else
    TBOX_ASSERT(d_n_nodes <= static_cast<int>(libMesh::Elem::max_n_nodes));
#endif
```

drwells:
We need to do that - there's a bunch of stuff which is also not guarded by the processor (like finally switching to `QNodal`).

PR #1786, opened 2025-02-26
Add class that can compute averaged flow quantities

abarret:
This adds the `HierarchyAveragedDataManager` class that utilizes the `SnapshotCache` to compute averages of noisy fields over multiple cycles. This class does not yet support restarts. You can compute an average velocity using the following code
```c++
 while (t < T_final)
 {
     // Advance flow field
     ins_integrator->advanceHierarchy(dt);
     t += dt;

     // Take a snapshot if at the correct time.
     if (t >= next_snap_time)
     {
         auto var_db = VariableDatabase<NDIM>::getDatabase();
         const int u_idx = var_db->mapVariableAndContextToIndex(ins_hierarchy->getVelocityVariable(),
                                                                ins_hierarchy->getCurrentContext());
         HierarchyMathOps hier_math_ops("HierarchyMathOps", patch_hierarchy);
         hier_math_ops.resetLevels(0, patch_hierarchy->getFinestLevelNumber());
         const int wgt_sc_idx = hier_math_ops.getSideWeightPatchDescriptorIndex();
         bool at_steady_state = hier_avg_data_manager.updateTimeAveragedSnapshot(
                  u_idx, t, patch_hierarchy, "CONSERVATIVE_LINEAR_REFINE", wgt_sc_idx, 1.0e-8);
         if (at_steady_state)
             pout << "Determined steady state!\n";
     }
 }
 ```

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### CODE REVIEW

**Comment by drwells** [comment]
Do you want to add restart data to this class?

File: ibtk/include/ibtk/HierarchyAveragedDataManager.h [hunk 1]
```diff
@@ -0,0 +1,293 @@
...
+     * Set of time points that snapshots are determined. These values are between d_t_start and d_t_end.
+     */
+    std::set<double> d_snapshot_time_pts;
+
+    SnapshotCache d_snapshot_cache;
+
+    SAMRAI::tbox::Pointer<SAMRAI::math::HierarchyDataOpsReal<NDIM, double> > d_hier_data_ops;
+
+    // Drawing stuff
+    std::unique_ptr<SAMRAI::appu::VisItDataWriter<NDIM> > d_visit_data_writer;
```
drwells, line 279: Bad news: this isn't going to to work in all cases because `VisItDataWriter` is a de facto singleton since it can only handle a single summary file (see `VisItDataWriter.C` circa line 3724). What happens to the original output when you write this? Does writing one summary file reset the other?

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,293 @@
...
+#ifndef included_IBTK_HierarchyAveragedDataManager
+#define included_IBTK_HierarchyAveragedDataManager
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include <ibtk/SnapshotCache.h>
+#include <ibtk/ibtk_utilities.h>
+
+#include "CellVariable.h"
+#include "HierarchySideDataOpsReal.h"
```
drwells, line 25:
```suggestion
```

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,293 @@
...
+#include "IntVector.h"
+#include "SideVariable.h"
```
drwells, line 27:
```suggestion
```

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,293 @@
...
+                                 bool register_for_restart = true);
+
+    /*!
+     * The constructor for class HierarchyAveragedDataManager sets some default values and determines data centering. In
+     * this constructor, the times at which snapshots are taken are set by arguments.
+     *
+     * The input database is searched for the following keys:
+     *  - 'enable_logging' : Bool used to determine whether to print convergence information to the log file.
+     *  - 'output_data' : Bool used to determine whether to write visualization files for the mean and deviations.
+     *  - 'dir_dump_name' : String used to determine which folder to write visualization files to. Defaults to "".
```
drwells, line 121:
```suggestion
     *  - 'dir_dump_name' : String used to determine which folder to write visualization files to. Defaults to "./".
```
What you have works but we can make it clearer.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,293 @@
...
+        return d_snapshot_cache;
+    }
+
+    /*!
+     * Set the threshold for achieving a steady state. This class determines whether a periodic steady state has been
+     * achieved by checking the whether \f$ \frac{1}{N}||u - u_avg||_2 \f$ is less than the provided threshold.
+     */
+    void setSteadyStateThreshold(const double threshold)
+    {
+        d_periodic_thresh = threshold;
```
drwells, line 226: Why not set this in the constructor?

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,293 @@
...
+     *  - 'output_data' : Bool used to determine whether to write visualization files for the mean and deviations.
+     *  - 'dir_dump_name' : String used to determine which folder to write visualization files to. Defaults to "".
+     *  - 'refine_type' : String to determine which refine operator will be used by default. Defaults to
+     * "CONSERVATIVE_LINEAR_REFINE".
+     */
+    HierarchyAveragedDataManager(std::string object_name,
+                                 SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > var,
+                                 SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db,
+                                 SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy,
+                                 std::set<double> snapshot_time_points,
```
drwells, line 129: FYI: `std::set` technically cannot be used with `double` since `double`s are not totally ordered due to `NaN`s. This code (same with `std::map`) will break in complicated ways if a `NaN` gets in.

ibtk/src/utilities/HierarchyAveragedDataManager.cpp
```diff
@@ -0,0 +1,316 @@
...
+        deallocate_patch_data(d_mean_idx, hierarchy, 0, hierarchy->getFinestLevelNumber());
+        deallocate_patch_data(d_dev_idx, hierarchy, 0, hierarchy->getFinestLevelNumber());
+    }
+    else
+    {
+        // Determine our convergence criteria: ||1/(N + 1) * (u_avg - u)||
+        d_hier_data_ops->linearSum(d_scratch_idx, 1.0 / (N + 1.0), u_idx, -1.0 / (N + 1.0), d_scratch_idx);
+    }
+
+    const double L2_norm = d_hier_data_ops->L2Norm(d_scratch_idx, wgt_idx);
```
drwells, line 276: Will this work correctly if we put in `invalid_index`? I think we always want to use the real norm.

ibtk/src/utilities/HierarchyAveragedDataManager.cpp
```diff
@@ -0,0 +1,316 @@
...
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+/////////////////////////////// INCLUDES /////////////////////////////////////
+#include "ibtk/HierarchyAveragedDataManager.h"
+#include "ibtk/HierarchyGhostCellInterpolation.h"
+#include "ibtk/HierarchyMathOps.h"
```
drwells, line 17:
```suggestion
#include <ibtk/HierarchyAveragedDataManager.h>
#include <ibtk/HierarchyGhostCellInterpolation.h>
#include <ibtk/HierarchyMathOps.h>
```

etc.

drwells (changes requested):

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,293 @@
...
+                                 double t_start,
+                                 double t_end,
+                                 SAMRAI::tbox::Pointer<SAMRAI::hier::GridGeometry<NDIM> > grid_geom,
```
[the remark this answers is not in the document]
abarret: This is needed to construct the `SnapshotCache` when restarting.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,293 @@
...
+     *  - 'output_data' : Bool used to determine whether to write visualization files for the mean and deviations.
+     *  - 'dir_dump_name' : String used to determine which folder to write visualization files to.
+     *  - 'refine_type' : String to determine which refine operator will be used by default.
+     */
+    HierarchyAveragedDataManager(std::string object_name,
+                                 SAMRAI::tbox::Pointer<SAMRAI::hier::Variable<NDIM> > var,
+                                 SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db,
+                                 SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > hierarchy,
+                                 SAMRAI::tbox::Pointer<SAMRAI::hier::GridGeometry<NDIM> > grid_geom,
+                                 bool register_for_restart = true);
```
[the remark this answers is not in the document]
abarret: I am working on supporting restarts.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
abarret: I've done this several times for different projects and have never had a problem, as long as files are written to different directories. Granted, I haven't paid close attention to the details of what's happening. I'll take a closer look.

ibtk/src/utilities/HierarchyAveragedDataManager.cpp
abarret: If we use `invalid_index` then each cell will have `1.0` as the weight. For uniform grids, this returns a quantity that is proportional to the "real" norm. Note we currently don't have an automated way to set edge, node, or face weight indices. In particular, we use `invalid_index` in the test.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
abarret: Are you suggesting that I rewrite this to use `std::vector` and keep it sorted myself? Or just to check for `NaN` on construction?

ibtk/src/utilities/HierarchyAveragedDataManager.cpp
```diff
@@ -0,0 +1,316 @@
...
+    }
+    return false;
+}
+
+double
+HierarchyAveragedDataManager::getTimePt(double time, const double tol)
+{
+    time = map_to_period(d_t_start, d_t_end, time);
+    auto it_up = d_snapshot_time_pts.upper_bound(time);
+    auto it_low = std::next(it_up, -1);
```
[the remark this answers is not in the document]
abarret: Hmm, I think you get undefined behavior if `upper_bound()` returns `d_snapshot_time_pts.begin()`. It should be fine if it returns `end()`. I will fix it.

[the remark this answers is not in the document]
abarret: I don't think this works with `std::set` because it's not guaranteed to be contiguous.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
abarret: As long as the `VisItDataWriter` objects write to different directories, there are no obvious problems. You get two summary files that are correct and no data is overwritten.

abarret:
@drwells I have added support for restarts.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
drwells: I'm not sure what the best solution but I think checking for NaNs suffices.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,288 @@
...
+    inline bool isAtPeriodicSteadyState(double time, const double tol)
+    {
+        time = getTimePoint(time, tol);
+        return d_idx_steady_state_map.at(time);
+    }
+
+    /*!
+     * Return the time points this class is storing.
+     */
+    inline const std::set<double>& getSnapshotTimePts()
```
drwells, line 201:
```suggestion
    inline const std::set<double>& getSnapshotTimePoints()
```

ibtk/include/ibtk/HierarchyAveragedDataManager.h
```diff
@@ -0,0 +1,288 @@
...
+    /*!
+     * Returns the exact data point at which points are being stored in the internal map. Note that if a time point is
+     * not found within the specified tolerance, an unrecoverable error occurs.
+     */
+    double getTimePoint(double time, double tol);
+
+    /*!
+     * Get the SnapshotCache object.
+     */
+    SnapshotCache& getSnapshotCache()
```
drwells, line 215: `inline` is redundant when class declarations also contain definitions but could you add it here so that it is consistent with the others? Alternatively you could remove the others.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
drwells: Hm, I suppose this is OK. We could 'fix' it in ibsamrai2 and then require that everyone update their SAMRAI installation - I'll let you decide on whether or not we should do that.

ibtk/src/utilities/HierarchyAveragedDataManager.cpp
```diff
@@ -0,0 +1,384 @@
...
+map_to_period(const double t_start, const double t_end, double time)
+{
+    const double period = t_end - t_start;
+    if (IBTK::abs_equal_eps(period, 0.0)) return t_start;
+#ifndef NDEBUG
+    // We only deal with positive period lengths.
+    TBOX_ASSERT(period > 0.0);
+#endif
+
+    while (time > t_end)
```
drwells, line 38: Could we use `fmod()` here instead?

ibtk/src/utilities/HierarchyAveragedDataManager.cpp
```diff
@@ -0,0 +1,384 @@
...
+        time += period;
+    }
+#ifndef NDEBUG
+    TBOX_ASSERT(time <= t_end && time >= t_start);
+#endif
+    return time;
+}
+
+void
+allocate_patch_data(const int idx,
```
drwells, line 55: We might actually have another version of these two functions floating around somewhere else in the library: to make sure we don't have symbol clashes (now or later) could you put these source-only functions in an anonymous namespace?

drwells (changes requested):
I apologize that it took me awhile to get back to this. This looks pretty good - I have a few minor QA requests.

ibtk/include/ibtk/HierarchyAveragedDataManager.h
abarret: I added a warning to the documentation about writing to different directories.

abarret:
@drwells I have addressed the comments and squashed the commits.

drwells (approved):

PR #1791, opened 2025-04-02
Cleanup CI caching

### LINKED ISSUE #1774: IBAMR 0.16: January 31, 2025
# IBAMR release tasks

## Summary

We haven't done a ton in the last six months (I was extremely busy getting Dr. Abdala everything she needed to finish on-time) but we should still tag a release at the end of the month.

We have a lot of PRs in the pipeline which are in some partially done state. As always, smaller PRs go through faster. I'll use the last week of this month to put out a release candidate so if you want stuff to make it in to 0.16 then get it done by January 24th.

Some extra TODOs:
- [x] *website dependencies*: Check and strongly consider combining the mac and linux pages.
- [x] *Docker image cleanup*: make sure we have a single Dockerfile available for each CI job and that they are reasonably up-to-date.
- [x] *Increase minimum version of libMesh and update tests*: at some point (1.6?) they rewrote their nodal quadrature rules. Lets just require 1.7 or newer and update our test output. *Edit* Actually... things work again with 1.7? I can't find the commit but it looks like they reversed their weird nodal quadrature update. This is good news - the test suite passes without modifications with 1.7.
- [x] *Eigen updates*: Check if we can get rid of `EigenAlignedVector`. I think that has been implemented correctly in a relatively recent version of Eigen.
- [ ] *CI caching*: use lessons learned in setting up Cardinal to revisit our object caching strategy

## static analysis

- [x] verify that neither `namespaces.h` nor `app_namespaces.h` are in IBTK or
  IBAMR headers
- [x] clear basic compilation warnings with GCC and clang on linux
- [x] compile with the latest libMesh release (which should be configured with
  `--disable-deprecated`) to verify that we do not use any deprecated
  functionality (this will cause compilation errors if we do use them). Add
  version checks: for example, here is one (lightly edited) in
  `IBFEInstrumentPannel.cpp` that uses a macro defined in `libmesh_version.h`:
```cpp
    // new API in 1.4.0
#if LIBMESH_VERSION_LESS_THAN(1, 4, 0)
    boundary_info.build_node_list(nodes, bcs);
#else
    const auto node_list = boundary_info.build_node_list();
    for (const auto& pair : node_list)
    {
        nodes.push_back(std::get<0>(pair));
        bcs.push_back(std::get<1>(pair));
    }
#endif
```
  around APIs to keep backwards compatibility with older versions of libMesh.
- [x] check the test suite with the oldest (3.7.0) and newest (latest release)
  versions of PETSc. Also check with the oldest (1.1.0) and newest (latest
  release) versions of libMesh.
- [x] clear basic compilation warnings with GCC and clang on linux and macOS
- [x] run cppcheck via
```
  cppcheck --force -j4 --enable=all -I./include/ -I./ibtk/include/  \
  $(find ./src ./ibtk/src -name '*.cpp') >cppcheck-results.txt 2>&1
```
- [x] Check that we don't call functions like
  `ReplicatedMesh::active_local_elements_end()` inside `for`-loop declarations
  since the end iterator is expensive to compute.
- [x] Check that we always get unique `HierarchyDataOps` by passing `true` as
  the third argument to `getOperationsDouble` et al. This prevents memory leaks
  where a statically stored `HierarchyDataOps` object keeps a patch hierarchy
  alive that is otherwise inaccessible.
- [x] Run clang-format on the entire code base via `make indent-all`.
- [x] Check that we don't use `Eigen::Index` anywhere: that is not available in
  older versions of Eigen. Instead, write
```cpp
    // Older versions of Eigen don't make Eigen::Index publicly available
#if EIGEN_VERSION_AT_LEAST(3, 3, 0)
    using IndexType = Eigen::Index;
#else
    using IndexType = typename MatrixType::Index;
#endif
```
  where needed.

## testing

- [x] Send out a message to the mailing list about the release asking about the
  status of any open PRs or bugs.
- [x] Run the test suite on a cluster where libMesh and PETSc were built without
  HDF5 support and HDF5 was installed as a user (see #687). Make sure all
  examples can compile and link.
- [x] Check that everything works on macOS where dependencies are installed with
  HomeBrew.
- [x] Double check that all examples compile cleanly with both autotools and
  CMake.
- [x] In general, run the test suite on as many platforms and configurations as
  possible to weed out problems.

## Copyright

- [x] Ensure that the top level file `COPYRIGHT` has the release year in it.
- [x] Check that new files that have been added since the last release have the
  correct copyright header. Try, for example,
```
  git log --diff-filter=A --numstat --pretty='COMMIT: %H %cd' v0.5.1..master
```
  to print all added files between `0.5.1` and `master` or just run
```
grep -L 'by the IBAMR developers' -R ./include/ ./ibtk/include/ ./src/ \
  ./ibtk/src/ | grep -v Makefile.in'
```
  to print all relevant (i.e., files not generated by automake) files which
  currently lack the copyright statement.
- [x] Run `scripts/maintenance/update_copyright.sh` to update copyright for
  standard file types.

## creating the release

- [x] Collate the changelog entries on `master` with
  `scripts/maintenance/generate-changes.sh`. Open a PR for this on GitHub: only
  proceed once it is merged.
- [x] Create a new branch `IBAMR-X.Y` locally off of master and increment the
  version in `VERSION` correctly. Push to GitHub without a PR:
```
  git push https://github.com/ibamr/ibamr.git IBAMR-X.Y
```
  where `X` and `Y` are the version numbers for the release (e.g., for the
  `0.5.0` release they are `0` and `5`). Note that the bugfix number does not go
  into the branch name.
- [x] Increment the version number on `master` to, e.g., `0.6.0-pre` to signify
  that `master` now corresponds to a prerelease state. Put this in a PR.
- [x] At this point the release branch and `master` will have diverged. Create a
  release candidate from the `IBAMR-X.Y` branch on GitHub and tag it as a
  prerelease (the tag should be `vX.Y.Z-rc1`). Be sure to include instructions
  on how to patch and compile SAMRAI.
- [x] If necessary, patch `IBAMR-X.Y` should the release candidate uncover more
  bugs.
- [x] Once `IBAMR-X.Y` is in good shape tag a proper release, i.e., `vX.Y.Z` and
  create a non-draft release on GitHub.
- [x] Send out a message to the mailing list about the release.
- [x] Add a note on the release to the web page.
- [x] Update version numbers on the build instructions web page.
- [x] Celebrate!

drwells:
Fixes #1774 (this is the last item on the release TODO list). Fixing the caching should make PRs run about twice as fast.

Since this is a CI change the normal checklist doesn't apply.

### DIFF
--- a/.github/workflows/push-master.yml
+++ b/.github/workflows/push-master.yml
@@ -20,6 +20,7 @@ on:
 permissions:
   contents: write
 jobs:
+  # doxygen
   deploy_documentation:
     runs-on: ubuntu-latest
     name: Compile and deploy doxygen documentation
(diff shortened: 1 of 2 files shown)


.github/workflows/push-master.yml
```diff
@@ -66,3 +67,141 @@ jobs:
           destination-repository-name: 'IBAMR-docs'
           user-email: 'travis@travis-ci.org'
           target-branch: master
+
+  # build IBAMR with CMake and update the cache
+  build_ibamr_cmake_016:
```
drwells, line 72: While this should all work (this is ported from Cardinal, where we have been using this approach for a year) there isn't a good way to test this except to merge this PR, wait, open another PR, and verify that we pick up the correct cache.

drwells:
Can someone take a look this week?

boyceg (approved):

boyceg:
@drwells: I got this error: https://github.com/IBAMR/IBAMR/actions/runs/14433566237

PR #1792, opened 2025-04-02
Remove old CI scripts.

drwells:
We replaced all of this with GitHub matrix code.

Like the other PR: this is a CI change so the normal checklist doesn't apply.


drwells:
Can someone take a look this week?

boyceg (approved):

PR #1793, opened 2025-04-04
avoid some copies

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/CCLaplaceOperator.h
+++ b/ibtk/include/ibtk/CCLaplaceOperator.h
@@ -155,9 +155,6 @@ private:
     std::vector<HierarchyGhostCellInterpolation::InterpolationTransactionComponent> d_transaction_comps;
     SAMRAI::tbox::Pointer<HierarchyGhostCellInterpolation> d_hier_bdry_fill, d_no_fill;
 
-    // Scratch data.
-    SAMRAI::tbox::Pointer<SAMRAI::solv::SAMRAIVectorReal<NDIM, double> > d_x, d_b;
-
     // Hierarchy configuration.
     SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > d_hierarchy;
     int d_coarsest_ln = IBTK::invalid_level_number, d_finest_ln = IBTK::invalid_level_number;
--- a/ibtk/include/ibtk/SCLaplaceOperator.h
+++ b/ibtk/include/ibtk/SCLaplaceOperator.h
@@ -132,9 +132,6 @@ protected:
     std::vector<HierarchyGhostCellInterpolation::InterpolationTransactionComponent> d_transaction_comps;
     SAMRAI::tbox::Pointer<HierarchyGhostCellInterpolation> d_hier_bdry_fill, d_no_fill;
 
-    // Scratch data.
-    SAMRAI::tbox::Pointer<SAMRAI::solv::SAMRAIVectorReal<NDIM, double> > d_x, d_b;
-
     // Hierarchy configuration.
     SAMRAI::tbox::Pointer<SAMRAI::hier::PatchHierarchy<NDIM> > d_hierarchy;
     int d_coarsest_ln = IBTK::invalid_level_number, d_finest_ln = IBTK::invalid_level_number;
--- a/ibtk/src/solvers/impls/CCLaplaceOperator.cpp
+++ b/ibtk/src/solvers/impls/CCLaplaceOperator.cpp
@@ -175,10 +175,6 @@ CCLaplaceOperator::initializeOperatorState(const SAMRAIVectorReal<NDIM, double>&
     // Deallocate the operator state if the operator is already initialized.
     if (d_is_initialized) deallocateOperatorState();
 
-    // Setup solution and rhs vectors.
-    d_x = in.cloneVector(in.getName());
-    d_b = out.cloneVector(out.getName());
-
     // Setup operator state.
     d_hierarchy = in.getPatchHierarchy();
     d_coarsest_ln = in.getCoarsestLevelNumber();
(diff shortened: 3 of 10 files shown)


drwells:
Nice work. Do we have any explicit tests for this class?

boyceg:
Refine schedules that are used for ghost cell filling that copy the full data from one context to another eat up a lot of runtime. This patch avoids doing them in the side-centered Laplace and Stokes operator implementations. There may be other places where we can do this.

boyceg:
> Nice work. Do we have any explicit tests for this class?

Yes -- there is a specific `SCLaplaceOperator` test along with anything that uses the Stokes solver with the standard preconditioners.

drwells:
Looks like some variable viscosity tests failed. Do you have time to debug it?

boyceg:
> Looks like some variable viscosity tests failed. Do you have time to debug it?

I think I know why this is happening --- will try to fix it.

abarret:
This requires that the indices passed in to `apply()` have sufficient ghost cell width for the application. Previously, this was only required for indices passed in to `initializeOperatorState()`. We could also get rid of the `d_x` and `d_b` variables while we're at it.

This is a good change, but it should be noted that the original behavior of not modifying ghost cells for patch index arguments has changed.

drwells:
This looks good. We should sit down and discuss making some formal definitions of ghosted and unghosted vectors so we can guarantee certain properties (e.g., doing `a X + b Y` does not require ghost data, but `A X` does). I want to reread this whole thing with fresh eyes on Monday before we merge.

abarret:
> I think that isn't quite right --- if you look in the implementation, `initializeOperatorState()` sets up communication for the given `x` and `b` vectors, and it clones `x` for the vector that is used to evaluate `A x`. If you pass in a vector with ghost cells to `initializeOperatorState()` and a vector without ghost cells to `apply()`, then the communications schedules could break, since it would change the data layout.

Yeah you are correct. For some reason I thought you could reset schedules with source patch indices that have smaller ghost cell widths, but that is not true.

> I think that the only major change here is that apply() can now change ghost cell values in the input vector x, whereas before it couldn't. It saves >10% of runtime in the Navier-Stokes tests that I've tried and doesn't break anything, so that seems like a good tradeoff to me...

I agree that it's a good tradeoff. It's just a change in how the operator works, which should be noted somewhere.

boyceg:
@drwells, @abarret: This along with IBAMR/IBSAMRAI2#35 improves runtime performance of `examples/navier_stokes/ex1/main3d` by > 35%.

Also, for that test, it seems like the approximately the best Navier-Stokes solver settings appear to be:
```

   stokes_solver_type = "PETSC_KRYLOV_SOLVER"
   stokes_precond_type = "PROJECTION_PRECONDITIONER"
   stokes_solver_db {
      ksp_type = "fgmres"
   }
   stokes_precond_db {
      skip_velocity_solve = TRUE
   }

   pressure_solver_type = "PETSC_KRYLOV_SOLVER"
   pressure_precond_type = "POINT_RELAXATION_FAC_PRECONDITIONER"
   pressure_solver_db {
      ksp_type = "richardson"
      max_iterations = 2
   }
   pressure_precond_db {
      cycle_type = "V_CYCLE"
      num_pre_sweeps  = 0
      num_post_sweeps = 3
      prolongation_method = "LINEAR_REFINE"
      restriction_method  = "CONSERVATIVE_COARSEN"
      coarse_solver_type  = "HYPRE_LEVEL_SOLVER"
      coarse_solver_rel_residual_tol = 1.0e-2
      coarse_solver_abs_residual_tol = 1.0e-50
      coarse_solver_max_iterations = 1
      coarse_solver_db {
         solver_type           = "PFMG"
         num_pre_relax_steps   = 3
         num_post_relax_steps  = 3
         rap_type              = 1     // specialized to 5/7-point stencils
         skip_relax            = 1     // set for isotropic problems
         initial_guess_nonzero = FALSE // the bottom solver's initial guess is zero
         enable_logging        = FALSE
      }
   }
```
Although we should double check `gmres` vs `fgmres`...

boyceg:
(I would guess that performance will see a larger improvement for IB models that do less frequent regridding, but I haven't checked.)

amneetb:
Were the changes in the VCINS solvers minimal? Do you also have performance numbers for them as well? Ex0 can be used for this purpose. 
https://github.com/IBAMR/IBAMR/tree/master/examples/vc_navier_stokes/ex0

boyceg:
The biggest impact is from aggressively reusing memory allocated by `SAMRAI::tbox::Array` --- that is in IBAMR/IBSAMRAI2#35. That should improve every solver, although the impact will depend on how fast/slow `new` and `delete` are on your system. On my Mac, I was seeing a huge fraction of runtime eaten up in calls to `malloc` and `free`.

All of the changes in this PR are pretty minimal — mostly avoiding some copies in the application of the Stokes-like operators. It means that the ghost regions, Dirichlet boundaries, and "invalid" coarse-grid regions can change in the input vector. I think that the original implementations avoided those kinds of changes because I was worried they could screw up Krylov methods if they inadvertently changed the solution vectors, but it doesn't seem to be a problem in practice.

I have another branch that goes further and also applies this to the convective operator implementations. Unfortunately this seems to break some of the tests --- not sure if I will have time to debug this any time soon.

boyceg:
I am not sure about removing those, since we may need them for resetting communication schedules.

drwells:
Where do we do that? AFAICT we only ever set up these vectors and then destroy them when we reset the operator state.

boyceg:
Let me try to remove them real quick...

boyceg:
@drwells, @abarret, I think this is basically ready to go. I am going to try to streamline synchronization in a different PR.

drwells (approved):

PR #1794, opened 2025-04-06
Fix code to remove nullspace components

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
It looks fine but I think you already know what I am going to ask - do we have simple tests for this function?

boyceg:
@drwells, I think that tests that set `explicitly_remove_nullspace` to `TRUE` will exercise these code paths. There are tests for both the constant-coefficient and variable-coefficient solvers that are configured like this.

boyceg:
No — I think that the mathematical expressions in the original implementation are actually wrong. (@amneetb pointed this out to me a while ago.)

boyceg:
That is a good question.

These vectors are used internally by the PETSc solvers when there are nullspaces attached to the solvers. I think that the vectors are supposed to be orthonormal when passed in, but maybe it ultimately doesn't matter.

boyceg:
I think that testing here would probably require adding a routine to print out whether the basis vectors are orthonormal or not. I guess we could add an accessor function to allow callers to grab the nullspace vectors?

boyceg:
(I guess we could also put in some code that only runs in debug mode that confirms that the vectors are orthonormal to a reasonable precision.)

boyceg:
@drwells I added some testing code to verify that the nullspace vectors are orthonormal. It will already be exercised by tests that have nullspaces.

drwells (approved):
I can't think of a better way to test it (and we know this formula is right) so let's roll with it.

PR #1798, opened 2025-04-13
avoid vector copies in the FAC preconditioner

boyceg:
`FACPreconditioner` has a specialized code path that is used with V-cycles without presmoothing. We can avoid copying one `SAMRAIVectorReal` and duplicating another.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/solvers/impls/FACPreconditioner.cpp
+++ b/ibtk/src/solvers/impls/FACPreconditioner.cpp
@@ -105,21 +105,21 @@ FACPreconditioner::solveSystem(SAMRAIVectorReal<NDIM, double>& x, SAMRAIVectorRe
     // Set the initial guess to equal zero.
     x.setToScalar(0.0, /*interior_only*/ false);
 
-    // Clone the right-hand-side vector to avoid modifying it during the
-    // preconditioning operation.
-    SAMRAIScopedVectorCopy<double> f(b);
-    SAMRAIScopedVectorDuplicate<double> r(b);
-
     // Apply a single FAC cycle.
     if (d_cycle_type == V_CYCLE && d_num_pre_sweeps == 0)
     {
         // V-cycle MG without presmoothing keeps the residual equal to the
         // initial right-hand-side vector f, so we can simply use that vector
         // for the residual in the FAC algorithm.
-        FACVCycleNoPreSmoothing(x, f, d_finest_ln);
+        FACVCycleNoPreSmoothing(x, b, d_finest_ln);
     }
     else
     {
+        // Clone the right-hand-side vector to avoid modifying it during the
+        // preconditioning operation.
+        SAMRAIScopedVectorCopy<double> f(b);
+        SAMRAIScopedVectorDuplicate<double> r(b);
+
         switch (d_cycle_type)
         {
         case F_CYCLE:


drwells (approved):
Nice!

Do we have tests which use both FAC preconditioners?

boyceg:
It looks like everything in the test suite actually uses the optimized code path. We should probably add tests for all of the supported FAC variants.

ibtk/src/solvers/impls/FACPreconditioner.cpp
```diff
@@ -105,21 +105,21 @@ FACPreconditioner::solveSystem(SAMRAIVectorReal<NDIM, double>& x, SAMRAIVectorRe
...
-    SAMRAIScopedVectorDuplicate<double> r(b);
-
     // Apply a single FAC cycle.
     if (d_cycle_type == V_CYCLE && d_num_pre_sweeps == 0)
     {
         // V-cycle MG without presmoothing keeps the residual equal to the
         // initial right-hand-side vector f, so we can simply use that vector
         // for the residual in the FAC algorithm.
-        FACVCycleNoPreSmoothing(x, f, d_finest_ln);
+        FACVCycleNoPreSmoothing(x, b, d_finest_ln);
```
drwells, line 114: Are you sure this is correct? There are a couple of code paths in which we modify the right-hand side here, e.g., in

```cpp
FACPreconditioner::FACVCycleNoPreSmoothing(SAMRAIVectorReal<NDIM, double>& u,
                                           SAMRAIVectorReal<NDIM, double>& f,
                                           int level_num)
{
    if (level_num == d_coarsest_ln)
    {
        // Solve Au = f on the coarsest level.
        d_fac_strategy->solveCoarsestLevel(u, f, level_num);
    }
    else
    {
        // Restrict the residual to the next coarser level.
        d_fac_strategy->restrictResidual(f, f, level_num - 1); // <---- here

        // Recursively call the FAC algorithm.
        FACVCycleNoPreSmoothing(u, f, level_num - 1);

        // Prolong the error from the next coarser level.  Because we did not
        // perform any presmoothing, we do not need to correct the solution on
        // the current level.
        d_fac_strategy->prolongError(u, u, level_num);

        // Smooth error on the current level.
        if (d_num_post_sweeps > 0)
        {
            d_fac_strategy->smoothError(u, f, level_num, d_num_post_sweeps, false, true);
        }
    }
    return;
} // FACVCycleNoPreSmoothing
```
which calls
```cpp
PoissonFACPreconditionerStrategy::restrictResidual(const SAMRAIVectorReal<NDIM, double>& src,
                                                   SAMRAIVectorReal<NDIM, double>& dst,
                                                   int dst_ln)
{
    IBTK_TIMER_START(t_restrict_residual);

    const int src_idx = src.getComponentDescriptorIndex(0);
    const int dst_idx = dst.getComponentDescriptorIndex(0);

    if (src_idx != dst_idx)
    {
        d_level_data_ops[dst_ln]->copyData(dst_idx, src_idx, /*interior_only*/ false);
    }
    xeqScheduleRestriction(dst_idx, src_idx, dst_ln);

    IBTK_TIMER_STOP(t_restrict_residual);
    return;
} // restrictResidual
```
which can do some kind of coarsening.

boyceg: Yes, it will do some coarsening.

Coarsening only impacts the "invalid" region of the coarser levels (i.e. pseudo-DOFs that are covered by true DOFs on finer levels). I look at it as "changing the scratch data", but not the state data, in a vector.

Would it be enough to make the documentation clear that the function can apply coarsening but not refining operations to the RHS vector?

drwells: Yes, that would help.

drwells (approved):

ibtk/src/solvers/impls/FACPreconditioner.cpp
drwells: I forgot to add a comment here when I merged.

Since this patch already does the same thing as the default codepath I think this patch is fine - I added a test for the non-default codepath.

PR #1803, opened 2025-06-09
Boost: don't use features removed in C++17.

drwells:
New versions of Clang are pedantic enough to not include these features any more.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/contrib/boost/boost/functional.hpp
+++ b/ibtk/contrib/boost/boost/functional.hpp
@@ -21,7 +21,7 @@ namespace boost
     namespace functional
     {
         namespace detail {
-#if defined(_HAS_AUTO_PTR_ETC) && !_HAS_AUTO_PTR_ETC
+#if 1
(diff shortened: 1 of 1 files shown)


boyceg:
@drwells should we just update the bundled Boost headers?

boyceg:
I think possibly the only thing that we use at this point in IBAMR is Boost.MultiArray. Can we _easily_ turn those into code that uses Eigen, so that we can drop Boost?

abarret:
> I think possibly the only thing that we use at this point in IBAMR is Boost.MultiArray.

`wave_generation/WaveDampingFunctions.cpp` uses a Newton solver from boost. Although that would be easy to replace.

drwells:
@IBAMR/ibamr-admin-team we still need to merge this patch to get things working with the most recent version of clang.

amneetb:
I think this can be merged with master.

drwells:
Want to approve and merge?

PR #1812, opened 2025-07-02
CI: compile IBAMR on macOS.

### LINKED ISSUE #1811: Set up a new CI runner for macOS
This would have caught the boost problem.

drwells:
Fixes #1811.

This would have found the boost problem.

This will take a few iterations to get right - to start I'll compile ibsamrai2.


drwells:
Looks like we also need to upgrade muparser.

boyceg (approved):

Issue #1821, opened 2025-08-13
Marker data restart not handled correctly by `IBExplicitHierarchyIntegrator`


marshallrdavey:
There are some bugs in the restart implementation marker points handled by `IBExplicitHierarchyIntegrator`. The main issue is that the  marker points are attempted to be initialized before the patch hierarchy is handed to the integrator. 

A potential solution would be to override `initalizePatchHierarchy` to include in the marker point reinitialization there so that the hierarchy will be non `NULL`. Alternatively, this could be handled by a new method called `resetMarkers`, or something similar.

drwells:
Can you share a minimum working example, perhaps one based on `tests/IBFE/explicit_ex4.cpp`?

marshallrdavey:
Like an example with an implemented fix?

marshallrdavey:
Give me an hour to rebuild `IBAMR` with `libMesh`

marshallrdavey:
There are two ways to see this using `explicit_ex4_2d.markers.mpirun=2.restart=80.input` that you can do in serial.

Firstly, run this normally  from the beginning and use `h5dump` on the restart file. `IBExplicitHierarchyIntegrator` on the below line is looking for `group /IBHierarchyIntegrator/IBHierarchyIntegrator::markers`, which doesn't exist. From the restart file the correct call should be `restart_db->keyExists(d_object_name + "::markers")` because `group /IBHierarchyIntegrator::markers` actually exists in the restart database.
https://github.com/IBAMR/IBAMR/blob/8f2bbe262325aa28496b5f61c312c13563662b88/src/IB/IBExplicitHierarchyIntegrator.cpp#L658

Secondly, add a flag for when you get into this if statement. The restart will not make markers because it's looking for the wrong database, so `getNumberOfMarkers() == 0` so it sets the makers with the restarted nodal locations before the time stepping starts, and the test looks like normal.
https://github.com/IBAMR/IBAMR/blob/8f2bbe262325aa28496b5f61c312c13563662b88/tests/IBFE/explicit_ex4.cpp#L560

PR #1823, opened 2025-08-14
Fix marker point restart in IBExplicitHierarchyIntegrator

marshallrdavey:
This is a proposed fix to #1821 that looks in the correct place for the marker restart data and waits until the `PatchHierarchy` has been initialized.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBExplicitHierarchyIntegrator.cpp
+++ b/src/IB/IBExplicitHierarchyIntegrator.cpp
@@ -513,6 +513,39 @@ IBExplicitHierarchyIntegrator::initializeHierarchyIntegrator(Pointer<PatchHierar
     return;
 } // initializeHierarchyIntegrator
 
+void
+IBExplicitHierarchyIntegrator::initializePatchHierarchy(Pointer<PatchHierarchy<NDIM> > hierarchy,
+                                                        Pointer<GriddingAlgorithm<NDIM> > gridding_alg)
+{
+    IBHierarchyIntegrator::initializePatchHierarchy(hierarchy, gridding_alg);
+
+    // Check if there is marker data to load from restart
+    if (RestartManager::getManager()->isFromRestart())
+    {
+        Pointer<Database> restart_db = RestartManager::getManager()->getRootDatabase();
+        if (restart_db->keyExists(d_object_name + "::markers"))
+        {
+            if (d_marker_kernel.size() == 0)
+            {
+                TBOX_ERROR(d_object_name
+                           << "::initializePatchHierarchy():\n To use marker points the IB kernel must be specified in "
+                              "the input database via IB_delta_fcn.");
+            }
+            d_markers = new MarkerPatchHierarchy(d_object_name + "::markers", d_hierarchy, {}, {});
+            // from marker restart
+            d_marker_velocities_set = true;
+            if (d_u_half_idx == IBTK::invalid_index)
+            {
+                VariableDatabase<NDIM>* var_db = VariableDatabase<NDIM>::getDatabase();
+                d_u_half_idx = var_db->registerClonedPatchDataIndex(getVelocityVariable(), d_u_idx);
+                d_ib_data.setFlag(d_u_half_idx);
+            }
+        }
+    }
+
+    return;
+} // initializePatchHierarchy
+
 std::size_t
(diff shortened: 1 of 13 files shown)


tests/IBFE/explicit_ex4.cpp
```diff
@@ -557,7 +557,8 @@ main(int argc, char** argv)
 
         // First test for markers: add them at the start (but don't clobber
         // restart marker data)
-        if (input_db->getBoolWithDefault("test_markers", false) && time_integrator->getNumberOfMarkers() == 0)
+        if (input_db->getBoolWithDefault("test_markers", false) && time_integrator->getNumberOfMarkers() == 0 &&
+            !RestartManager::getManager()->isFromRestart())
```
drwells: In general we try to avoid making control flow depend on whether or not we are in a restarted run - user apps should work without having to add these kinds of checks. Can we fix the bug without adding this?

marshallrdavey: The original flags now work because `getNumberOfMarkers() != 0`--this was just an attempt to make the test more explicit.

marshallrdavey: A check that `loop_time > 0.0` would accomplish the same thing to demonstrate that this is from a restart, or a `tbox::pout << "Initial markers added\n"` which would only show up in the `output` if no restart markers were found.

src/IB/IBExplicitHierarchyIntegrator.cpp
```diff
@@ -655,11 +686,11 @@ IBExplicitHierarchyIntegrator::getFromRestart()
...
-    if (db->isDatabase(d_object_name + "::markers"))
-    {
-        // MarkerPatchHierarchy will access the restart database itself
-        d_markers = new MarkerPatchHierarchy(d_object_name + "::markers", d_hierarchy, {}, {});
-    }
+    // if (db->isDatabase(d_object_name + "::markers"))
+    // {
+    //    // MarkerPatchHierarchy will access the restart database itself
+    //    d_markers = new MarkerPatchHierarchy(d_object_name + "::markers", d_hierarchy, {}, {});
+    // }
```
drwells: I think the original fix you proposed (fixing the database names) is preferable to this. Could you do that and also modify the tests to make it more obvious when particles are added (i.e., for restarted tests the message should not print at the beginning)?

tests/IBFE/explicit_ex4.cpp
drwells: I think the second approach is better.

src/IB/IBExplicitHierarchyIntegrator.cpp
marshallrdavey: What do you mean fix the database names?

drwells: using `restart_db->isDatabase(d_object_name + "::markers")` rather than `db->isDatabase(d_object_name + "::markers")`

marshallrdavey: Yeah, that's the fix I implemented. I don't think we should deal with trying to nest restart data from two different `Serializable` classes.

src/IB/IBExplicitHierarchyIntegrator.cpp
```diff
@@ -513,6 +513,38 @@ IBExplicitHierarchyIntegrator::initializeHierarchyIntegrator(Pointer<PatchHierar
...
+
+    // Check if there is marker data to load from restart
+    if (RestartManager::getManager()->isFromRestart())
+    {
+        Pointer<Database> restart_db = RestartManager::getManager()->getRootDatabase();
+        if (restart_db->keyExists(d_object_name + "::markers"))
+        {
+            if (d_marker_kernel.size() == 0)
+            {
+                TBOX_ERROR(d_object_name << "::setMarkers():\n To use marker points the IB kernel must be specified in "
```
drwells:
```suggestion
                TBOX_ERROR(d_object_name << "::initializePatchHierarchy():\n To use marker points the IB kernel must be specified in "
```

drwells (approved):

PR #1825, opened 2025-08-26
`IBHierarchyIntegrator` etc.: refactor sources derived from `IBStrategy`.

drwells:
This is the first of two PRs which will fix a bug where we didn't correctly update sources in calls to
`INSHierarchyIntegrator::regridProjection()`, which resulted in reading undefined data and therefore arbitrary velocity values. To begin, this commit makes it possible to compute sources outside of the normal timestepping loop and also modifies `IBEulerianSourceFunction` to work on the entire `PatchHierarchy` (instead of single patches) at once.

We don't use this implementation of sources in four-chambered-heart or Cardinal: those apps directly add it to the INS integrator. We do have a test for these sources in IBAMR and I checked that, in addition to passing, it has the same visual output as before.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/tests/IBFE/explicit_ex4_2d.source_test.output
+++ b/tests/IBFE/explicit_ex4_2d.source_test.output
@@ -28,16 +28,24 @@ IBHierarchyIntegrator::integrateHierarchy(): spreading Lagrangian force to the E
 IBHierarchyIntegrator::integrateHierarchy(): computing Lagrangian fluid source strength
 IBHierarchyIntegrator::integrateHierarchy(): spreading Lagrangian fluid source strength to the Eulerian grid
 IBHierarchyIntegrator::integrateHierarchy(): solving the incompressible Navier-Stokes equations
+IBHierarchyIntegrator::integrateHierarchy(): computing Lagrangian fluid source strength
+IBHierarchyIntegrator::integrateHierarchy(): spreading Lagrangian fluid source strength to the Eulerian grid
+IBHierarchyIntegrator::integrateHierarchy(): computing Lagrangian fluid source strength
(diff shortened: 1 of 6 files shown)


tests/IBFE/explicit_ex4_2d.source_test.output
```diff
@@ -28,16 +28,24 @@ IBHierarchyIntegrator::integrateHierarchy(): spreading Lagrangian force to the E
...
+IBHierarchyIntegrator::integrateHierarchy(): spreading Lagrangian fluid source strength to the Eulerian grid
+IBHierarchyIntegrator::integrateHierarchy(): computing Lagrangian fluid source strength
+IBHierarchyIntegrator::integrateHierarchy(): spreading Lagrangian fluid source strength to the Eulerian grid
 INSStaggeredHierarchyIntegrator::integrateHierarchy(): stokes solve number of iterations = 11
-INSStaggeredHierarchyIntegrator::integrateHierarchy(): stokes solve residual norm        = 3.3730300296385233243e-15
+INSStaggeredHierarchyIntegrator::integrateHierarchy(): stokes solve residual norm        = 3.3730300296377628623e-15
+IBHierarchyIntegrator::integrateHierarchy(): computing Lagrangian fluid source strength
+IBHierarchyIntegrator::integrateHierarchy(): spreading Lagrangian fluid source strength to the Eulerian grid
+IBHierarchyIntegrator::integrateHierarchy(): computing Lagrangian fluid source strength
+IBHierarchyIntegrator::integrateHierarchy(): spreading Lagrangian fluid source strength to the Eulerian grid
```
drwells, line 40: I think this was a bug - since sources are time-dependent they should be recomputed here.

PR #1830, opened 2025-09-10
Add automatically Restart_cleaner for fixing issue #571

zoraaaa03:
### Description
This pull request introduces a new RestartCleaner class that automatically manages restart/checkpoint directories to prevent disk space issues in long-running simulations.

This feature was developed as part of a summer project under the guidance of @drwells. Thanks @drwells.

### Key Features
- Keeps only the N most recent restart directories
- More features can be added in the future

### Testing
- Added interface and functional tests

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/major/20260220Chen
+++ b/doc/news/changes/major/20260220Chen
@@ -1,0 +1,5 @@
+New: Added RestartCleaner, a utility class that automatically manages
+restart directories to prevent disk space exhaustion in long-running
+simulations. It supports a "keep most recent N" cleanup strategy.
+<br>
+(Meihan Chen, David Wells, 2026/02/20)


ibtk/include/ibtk/RestartCleaner.h
```diff
@@ -0,0 +1,232 @@
...
+     *
+     * \param restart_base_path  Base directory containing restore folders
+     * \param keep_restart_count Number of recent restore directories to keep
+     * \param strategy          Cleanup strategy ("KEEP_RECENT_N")
+     * \param dry_run           If true, only report what would be deleted without actually deleting
+     */
+    explicit RestartCleaner(const std::string& restart_base_path,
+                            int keep_restart_count,
+                            const std::string& strategy = "KEEP_RECENT_N",
+                            bool dry_run = false);
```
drwells: Does this do the same thing as the other constructor? If so, let's standardize on only using one constructor (the one with an input database)

ibtk/src/utilities/RestartCleaner.cpp
```diff
@@ -0,0 +1,293 @@
...
+      d_keep_restart_count(keep_restart_count),
+      d_strategy(parseStrategy(strategy)),
+      d_enabled(true),
+      d_log_actions(true),
+      d_dry_run(dry_run)
+{
+#if !defined(NDEBUG)
+    TBOX_ASSERT(!restart_base_path.empty());
+    TBOX_ASSERT(keep_restart_count > 0);
+#endif
```
drwells:
```suggestion
    TBOX_ASSERT(!restart_base_path.empty());
    TBOX_ASSERT(keep_restart_count > 0);
```

We commonly check for `NDEBUG` in IBAMR, but we really shouldn't when the assertions are cheap to execute.

ibtk/src/utilities/RestartCleaner.cpp
```diff
@@ -0,0 +1,293 @@
...
+      d_enabled(input_db->getBoolWithDefault("enable_cleaner", false)),
+      d_keep_restart_count(input_db->getIntegerWithDefault("keep_recent_files", 5)),
+      d_log_actions(input_db->getBoolWithDefault("log_cleaning_actions", true)),
+      d_strategy(parseStrategy(input_db->getStringWithDefault("cleanup_strategy", "KEEP_RECENT_N"))),
+      d_dry_run(false)
+{
+#if !defined(NDEBUG)
+    TBOX_ASSERT(input_db);
+#endif
+
```
drwells:
```suggestion
```
The initialization list reads from the input database, so it cannot be null at this point (if it were we would have crashed)

tests/IBTK/restart_cleaner_function_test.cpp
```diff
@@ -0,0 +1,470 @@
...
+//
+// ---------------------------------------------------------------------
+
+// Config files
+
+// Define NDIM for SAMRAI template instantiation
+#ifndef NDIM
+#define NDIM 2
+#endif
+
```
drwells:
```suggestion
```

tests/IBTK/restart_cleaner_interface_test.cpp
```diff
@@ -0,0 +1,272 @@
...
+//
+// ---------------------------------------------------------------------
+
+// Config files
+
+// Define NDIM for SAMRAI template instantiation
+#ifndef NDIM
+#define NDIM 2
+#endif
+
```
drwells:
```suggestion
```

ibtk/include/ibtk/RestartCleaner.h
```diff
@@ -0,0 +1,232 @@
...
+/////////////////////////////// INCLUDES /////////////////////////////////////
+
+#include <ibtk/config.h>
+
+#include "ibtk/IBTK_MPI.h"
+
+#include "tbox/Database.h"
+#include "tbox/Pointer.h"
+
+#include <filesystem>
```
drwells, line 28:
```suggestion
```
This header is only used in the definition, not the declaration, so we can move it to the source file.

ibtk/include/ibtk/RestartCleaner.h
```diff
@@ -0,0 +1,232 @@
...
+     * This constructor integrates with IBAMR's standard configuration system.
+     * The cleaner can be enabled or disabled through the configuration.
+     * The input database is searched for the following keys:
+     * - 'enable_cleaner' : Boolean to enable/disable the cleaner (default: false)
+     * - 'keep_recent_files' : Integer number of files to keep (default: 5)
+     * - 'restart_directory' : String path to restart directory (required if enable_cleaner is true)
+     * - 'log_cleaning_actions' : Boolean to enable logging (default: true)
+     * - 'cleanup_strategy' : String cleanup strategy (default: "KEEP_RECENT_N")
+     *
+     * \note dry_run mode is not supported in this constructor.
```
drwells:
```suggestion
     * \note dry_run mode is not supported in this constructor.
```
Why not?

ibtk/src/utilities/RestartCleaner.cpp
```diff
@@ -0,0 +1,269 @@
...
+    {
+    case CleanupStrategy::KEEP_RECENT_N:
+        keepRecentN();
+        break;
+    default:
+        TBOX_ERROR(d_object_name << "::executeStrategy(): Unknown strategy" << std::endl);
+    }
+} // executeStrategy
+
+int
```
drwells:
```suggestion
std::optional<int>
```
Let's encode a missing directory number with `std::optional`: that way we don't have to use `-1` as a magic value.

ibtk/src/utilities/RestartCleaner.cpp
```diff
@@ -0,0 +1,269 @@
...
+RestartCleaner::parseIterationNum(const std::string& dirname) const
+{
+    std::regex pattern("restore\\.([0-9]{6})");
+    std::smatch match;
+
+    if (std::regex_match(dirname, match, pattern))
+    {
+        return std::stoi(match[1].str());
+    }
+    return -1;
```
drwells:
```suggestion
    return std::nullopt;
```

ibtk/src/utilities/RestartCleaner.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+int
+RestartCleaner::parseIterationNum(const std::string& dirname) const
+{
+    std::regex pattern("restore\\.([0-9]{6})");
+    std::smatch match;
+
+    if (std::regex_match(dirname, match, pattern))
+    {
+        return std::stoi(match[1].str());
```
drwells, line 155:
```suggestion
        return std::make_optional(std::stoi(match[1].str()));
```

ibtk/src/utilities/RestartCleaner.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+    // Sort directories by iteration number (ascending order)
+    std::sort(dirs.begin(),
+              dirs.end(),
+              [this](const std::filesystem::path& a, const std::filesystem::path& b)
+              {
+                  int iter_a = parseIterationNum(a.filename().string());
+                  int iter_b = parseIterationNum(b.filename().string());
+                  return iter_a < iter_b;
+              });
```
drwells: we need to correctly filter-out erroneously named directories here. One possibility would be to use `std::partition()`, with the first half having invalid values for iteration number and the second half having valid values.

tests/IBTK/restart_cleaner_interface_test.cpp
```diff
@@ -0,0 +1,226 @@
...
+            Pointer<MemoryDatabase> db = new MemoryDatabase("EnabledCleanerTest");
+            db->putBool("enable_cleaner", true);
+            db->putString("restart_directory", temp_dir);
+            db->putInteger("keep_recent_files", 3);
+            db->putString("cleanup_strategy", "KEEP_RECENT_N");
+            db->putBool("dry_run", true);
+
+            Pointer<Database> db_ptr = db;
+            RestartCleaner cleaner("EnabledCleanerTest", db_ptr);
```
drwells:
```suggestion
            RestartCleaner cleaner("EnabledCleanerTest", db);
```

tests/IBTK/restart_cleaner_interface_test.cpp
```diff
@@ -0,0 +1,226 @@
...
+        {
+            // Create minimal configuration database
+            Pointer<MemoryDatabase> db = new MemoryDatabase("RestartCleanerConfig");
+            db->putBool("enable_cleaner", false); // Disabled to avoid file operations
+            db->putInteger("keep_recent_files", 5);
+            db->putBool("log_cleaning_actions", false);
+            db->putString("cleanup_strategy", "KEEP_RECENT_N");
+
+            Pointer<Database> db_ptr = db;
+            RestartCleaner cleaner("TestCleaner", db_ptr);
```
drwells:
```suggestion
            RestartCleaner cleaner("TestCleaner", db);
```

ibtk/src/utilities/RestartCleaner.cpp
```diff
@@ -0,0 +1,269 @@
...
+        break;
+    default:
+        TBOX_ERROR(d_object_name << "::executeStrategy(): Unknown strategy" << std::endl);
+    }
+} // executeStrategy
+
+int
+RestartCleaner::parseIterationNum(const std::string& dirname) const
+{
+    std::regex pattern("restore\\.([0-9]{6})");
```
drwells: We may have more than six digits (but we always have at least six!)

drwells: Can you add a test which has restart directories with index greater than one million?

tests/IBTK/restart_cleaner_function_test.cpp
```diff
@@ -0,0 +1,505 @@
...
+    pout << "\n=== Test 1: Directory Pattern Parsing and Filtering ===" << std::endl;
+    {
+        const std::string test_dir = "test_restart_parsing";
+        TestDirGuard guard(test_dir);
+
+        // Create mix of valid and invalid directories
+        std::vector<int> valid_iterations = { 1, 100, 200, 300, 1000, 2500, 3000, 5000, 999999 };
+        std::vector<std::string> invalid_names = {
+            "restore.12345",        // 5 digits
+            "restore.1234567",      // 7 digits
```
drwells: this is actually a valid name for a restart directory.

tests/IBTK/restart_cleaner_function_test.cpp
```diff
@@ -0,0 +1,505 @@
...
+            // Create database configuration for empty directory test
+            Pointer<MemoryDatabase> db1 = new MemoryDatabase("EmptyDirTestConfig");
+            db1->putBool("enable_cleaner", true);
+            db1->putString("restart_directory", empty_dir);
+            db1->putInteger("keep_recent_files", 5);
+            db1->putString("cleanup_strategy", "KEEP_RECENT_N");
+            db1->putBool("dry_run", true);
+
+            Pointer<Database> db_ptr1 = db1;
+            RestartCleaner cleaner("EmptyDirTest", db_ptr1);
```
drwells:
```suggestion
            RestartCleaner cleaner("EmptyDirTest", db1);
```

drwells: I think I found all of these - we don't need the extra explicit cast.

ibtk/src/utilities/RestartCleaner.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+/////////////////////////////// PUBLIC ///////////////////////////////////////
+
+RestartCleaner::RestartCleaner(const std::string& object_name, SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db)
+    : d_object_name(object_name),
+      d_restart_base_path(),
+      d_strategy(parseStrategy(input_db->getStringWithDefault("cleanup_strategy", "KEEP_RECENT_N"))),
+      d_keep_restart_count(input_db->getIntegerWithDefault("keep_recent_files", 5)),
+      d_enabled(input_db->getBoolWithDefault("enable_cleaner", false)),
+      d_log_actions(input_db->getBoolWithDefault("log_cleaning_actions", true)),
```
drwells:
```suggestion
      d_enable_logging(input_db->getBoolWithDefault("enable_logging", true)),
```
while this name isn't as accurate, it is consistent with what we do everywhere else in IBAMR.

ibtk/src/utilities/RestartCleaner.cpp
```diff
@@ -0,0 +1,269 @@
...
+/////////////////////////////// STATIC ///////////////////////////////////////
+
+/////////////////////////////// PUBLIC ///////////////////////////////////////
+
+RestartCleaner::RestartCleaner(const std::string& object_name, SAMRAI::tbox::Pointer<SAMRAI::tbox::Database> input_db)
+    : d_object_name(object_name),
+      d_restart_base_path(),
+      d_strategy(parseStrategy(input_db->getStringWithDefault("cleanup_strategy", "KEEP_RECENT_N"))),
+      d_keep_restart_count(input_db->getIntegerWithDefault("keep_recent_files", 5)),
+      d_enabled(input_db->getBoolWithDefault("enable_cleaner", false)),
```
drwells: I think that this should be, in practice, the same as 'dry run', no?

I think we can implement both with `dry_run` (which should be off by default).

ibtk/src/utilities/RestartCleaner.cpp
```diff
@@ -0,0 +1,269 @@
...
+    {
+        TBOX_ERROR(d_object_name << "::getAllRestartDirs(): Directory does not exist: " << restart_dir << std::endl);
+    }
+
+    if (!std::filesystem::is_directory(restart_dir))
+    {
+        TBOX_ERROR(d_object_name << "::getAllRestartDirs(): Path is not a directory: " << restart_dir << std::endl);
+    }
+
+    std::regex pattern("restore\\.([0-9]{6})");
```
drwells: Here and elsewhere, these restore directories have at least six digits, not exactly six digits.

drwells (changes requested):
We're getting there!

ibtk/src/utilities/RestartCleaner.cpp
zoraaaa03: fixed: change to  std::regex pattern("restore\\.([0-9]{6,})");

tests/IBTK/restart_cleaner_function_test.cpp
zoraaaa03: fixed: move it to the valid iterations

ibtk/src/utilities/RestartCleaner.cpp
zoraaaa03: added 1000000, 1234567, 10000000 to test

drwells:
@zoraaaa03 I added a commit which should fix compilation with the other build system (autotools).

Issue #1806, opened 2025-06-16
IBAMR 0.17: June 30, 2025


drwells:
# IBAMR release tasks

## Summary

Let's clean up the current set of PRs and tag a release at the end of the month. I've already gotten things working (except for a few failing tests which aren't that critical) with libMesh 1.8 and the development version.

# IBAMR release tasks

## Summary

This document describes the process for creating a new release of IBAMR on
GitHub.

## static analysis

- [ ] verify that neither `namespaces.h` nor `app_namespaces.h` are in IBTK or
  IBAMR headers
- [ ] verify that each IBAMR header includes `ibamr/config.h` and each IBTK
  header includes `ibtk/config.h`
- [ ] clear basic compilation warnings with GCC and clang on linux
- [x] compile with the latest libMesh release (which should be configured with
  `--disable-deprecated`) to verify that we do not use any deprecated
  functionality (this will cause compilation errors if we do use them). Add
  version checks: for example, here is one (lightly edited) in
  `IBFEInstrumentPannel.cpp` that uses a macro defined in `libmesh_version.h`:
```cpp
    // new API in 1.4.0
#if LIBMESH_VERSION_LESS_THAN(1, 4, 0)
    boundary_info.build_node_list(nodes, bcs);
#else
    const auto node_list = boundary_info.build_node_list();
    for (const auto& pair : node_list)
    {
        nodes.push_back(std::get<0>(pair));
        bcs.push_back(std::get<1>(pair));
    }
#endif
```
  around APIs to keep backwards compatibility with older versions of libMesh. (#1804)
- [ ] check the test suite with the oldest (3.7.0) and newest (latest release)
  versions of PETSc. Also check with the oldest (1.7.0) and newest (latest
  release) versions of libMesh.
- [ ] clear basic compilation warnings with GCC and clang on linux and macOS
- [ ] run cppcheck via
```
  cppcheck --force -j4 --enable=all -I./include/ -I./ibtk/include/  \
  $(find ./src ./ibtk/src -name '*.cpp') >cppcheck-results.txt 2>&1
```
- [ ] Check that we don't call functions like
  `ReplicatedMesh::active_local_elements_end()` inside `for`-loop declarations
  since the end iterator is expensive to compute.
- [ ] Check that we always get unique `HierarchyDataOps` by passing `true` as
  the third argument to `getOperationsDouble` et al. This prevents memory leaks
  where a statically stored `HierarchyDataOps` object keeps a patch hierarchy
[...]

PR #1837, opened 2025-09-19
CI: turn on more warnings in the macOS runner.

drwells:
I disabled these when setting up this runner to make sure things worked. I forgot to turn them back on. This PR fixes the two warnings and then enables them in the macOS runner.

Related to https://github.com/IBAMR/IBAMR/issues/1806#issuecomment-3310250158


boyceg (approved):

configure.ac
```diff
@@ -20,8 +20,9 @@ AC_CONFIG_AUX_DIR([config])
 AC_CONFIG_MACRO_DIR([m4])
 AC_CONFIG_HEADER([config/IBAMR_config.h.tmp])
 AX_PREFIX_CONFIG_H([config/IBAMR_config.h],[IBAMR],[config/IBAMR_config.h.tmp])
-# Also copy our own CMake compatibility shim header:
-AC_CONFIG_LINKS([config/ibamr/config.h:$srcdir/config/config.h.tmp])
+# We used to have a sub config.h which we don't need any more. Keep compatibility
+# by replacing it with a link to the real config.h
+AC_CONFIG_LINKS([config/ibamr/config.h:$srcdir/include/ibamr/config.h])
```
drwells, line 25: This one is interesting: I don't remember why, five years ago, I added this stub header, but it doesn't have most of the things in `ibamr/config.h`. We can just get rid of it.

PR #1840, opened 2025-10-01
add runtime customization of coarse-fine interface discretization options

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/incompatibilities/20251002BoyceGriffith
+++ b/doc/news/changes/incompatibilities/20251002BoyceGriffith
@@ -1,0 +1,4 @@
+Changed: Added input databases to the constructors for several operators and preconditioners.
+These enable runtime configuration of coarse-fine interface discretization details.
+<br>
+(Boyce Griffith, 2025/10/02)


boyceg:
> How is this going to speed up the solvers?

Oh yeah, I should have said — by simplifying the operations performed in the preconditioners!

It appears that you can use lower-order coupling operators in some cases, which results in reduced communication and less expensive stencils.

In my initial tests, it appears that we can reliably use `"CONSERVATIVE_COARSEN"` as the coarsening operator for both cell-centered and side-centered Laplace operators instead of the cubic version that is currently hard-wired in `master`.

I haven't tested this as carefully, but it also looks like we can switch to using `"CONSTANT"` extrapolation at physical boundaries.

For the cell-centered operators, we can switch from using the full coarse-fine interface coupling scheme to simple linear (or even constant) refining.

For the side-centered Laplace operator, it looks like switching to simple linear refinement at coarse-fine interfaces tanks the performance of the preconditioner.

(Incidentally, in most IB models, I think we probably should turn off the velocity solver altogether in the preconditioner: `-velocity_ksp_type preonly -velocity_pc_type none`. There is not a good way to do this using just the input files — I think — because we don't currently have a good way to specify a "no op" solver.)

> These options should be specified at runtime, but this is a breaking change when `homogeneous_bc` is specified. Most of the solvers/operators have the input database as the second argument, so I think that's fine. But it should be mentioned in the changelog. As a side note, #1475 added the input database as the last argument.

I know this is a breaking change, but I think it is OK — this is not a change that is only going to impact developers.

Also, I think we should consider switching the order of the changes made in #1475 - I think that in most of the constructors, at least when they are present, the order is `object_name` first, `input_db` second, and then other options.

abarret (approved):

abarret:
Looks good to me.

PR #1850, opened 2025-10-24
Test Array::Allocator.

drwells:
Before https://github.com/IBAMR/IBSAMRAI2/pull/52 this didn't work correctly since the underlying allocator was never switched to its 'on' state. This test verifies that we correctly count the total number of Array allocations.

This won't work until we update the minimum version of IBSAMRAI2 which I will get done later today.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/tests/external/array_03.output
+++ b/tests/external/array_03.output
@@ -1,0 +1,3 @@
+total number of Array<int> allocations = 0
+total number of Array<bool> allocations = 4
+total number of Array<Array<bool>> allocations = 2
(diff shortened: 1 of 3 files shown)


tests/external/array_03.output
```diff
@@ -0,0 +1,3 @@
+total number of Array<int> allocations = 0
+total number of Array<bool> allocations = 4
```
drwells, line 2: Without the IBSAMRAI2 patch this was 400000, which shows we fixed the bug.

PR #1860, opened 2025-11-11
Add options for regridding operators in the advection diffusion integrator.

abarret:
Also deletes an extra coarsening step that is done in the base class.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/adv_diff/AdvDiffHierarchyIntegrator.cpp
+++ b/src/adv_diff/AdvDiffHierarchyIntegrator.cpp
@@ -203,6 +203,8 @@ AdvDiffHierarchyIntegrator::registerAdvectionVelocity(Pointer<FaceVariable<NDIM,
     TBOX_ASSERT(u_var);
     TBOX_ASSERT(std::find(d_u_var.begin(), d_u_var.end(), u_var) == d_u_var.end());
 #endif
+    if (d_integrator_is_initialized)
+        TBOX_ERROR(d_object_name + "::registerAdvectionVelocity(): Integrator has already been initialized.\n");
     d_u_var.push_back(u_var);
 
     // Set default values.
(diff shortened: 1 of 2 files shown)


drwells:
Nice - it is always good to remove another unneeded communication step.

Could these also be set by the input file? Also, could you add a basic test which uses this functionality?

abarret:
I've added a test.



> Could these also be set by the input file?

Default values could be set in the input file, but generally, these can't be set until variables are registered as transported quantities with the integrators. I didn't add an option to set default refine or coarsen operators in the input file because the linear refine and coarsen operators should work fine in almost all cases.

src/adv_diff/AdvDiffHierarchyIntegrator.cpp
```diff
@@ -333,6 +333,8 @@ AdvDiffHierarchyIntegrator::registerTransportedQuantity(Pointer<CellVariable<NDI
     d_Q_bc_coef[Q_var] = std::vector<RobinBcCoefStrategy<NDIM>*>(Q_depth, nullptr);
     d_Q_reset_priority.push_back(std::numeric_limits<int>::max());
     d_Q_output[Q_var] = output_Q;
+    d_Q_refine_map[Q_var] = "CONSERVATIVE_LINEAR_REFINE";
+    d_Q_coarsen_map[Q_var] = "CONSERVATIVE_COARSEN";
```
boyceg: Can we use `setRefineAndCoarsenOperators()` here?

src/adv_diff/AdvDiffHierarchyIntegrator.cpp
```diff
@@ -754,6 +756,18 @@ AdvDiffHierarchyIntegrator::setHelmholtzRHSOperatorNeedsInit(Pointer<CellVariabl
     return;
 }
 
+void
+AdvDiffHierarchyIntegrator::setRefineAndCoarsenOperators(Pointer<CellVariable<NDIM, double> > Q_var,
+                                                         std::string Q_refine_type,
+                                                         std::string Q_coarsen_type)
+{
+#if !defined(NDEBUG)
```
boyceg, line 772: Isn't there some point in a computation at which we cannot change these?

abarret: Yeah. I went ahead and added checks for other functions than can't be called after hierarchy initialization.

PR #1873, opened 2026-01-27
INSStaggeredHierarchyIntegrator: fix divergence allocation.

drwells:
I believe this is the same issue which was reported on the mailing list: if we need to regrid but don't previously save divergence data, then that data index isn't available.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/tests/IBFE/explicit_ex4_2d.div-regrid.mpirun=4.input
+++ b/tests/IBFE/explicit_ex4_2d.div-regrid.mpirun=4.input
@@ -1,0 +1,240 @@
+// Verify that we correctly allocate the divergence when we need to project
+// the regridded solution
+
+// additional test parameters
+mesh_file = "explicit_ex4_2d.grid-1.xdr"
+
+// physical parameters
+MU  = 0.01
+RHO = 1.0
+L   = 1.0
+
+// grid spacing parameters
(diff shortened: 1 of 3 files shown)


tests/IBFE/explicit_ex4_2d.div-regrid.mpirun=4.input
```diff
@@ -0,0 +1,240 @@
...
+   tag_buffer                    = TAG_BUFFER
+   output_U                      = OUTPUT_U
+   output_P                      = OUTPUT_P
+   output_F                      = OUTPUT_F
+   output_Omega                  = OUTPUT_OMEGA
+   output_Div_U                  = OUTPUT_DIV_U
+   enable_logging                = ENABLE_LOGGING
+   enable_logging_solver_iterations = FALSE
+
+   regrid_max_div_growth_factor = 1e-10
```
drwells, line 171: @boyceg I'm not sure, though I can reproduce the crash by setting this parameter very low.

boyceg: OK, got it — we are not enforcing `div u = 0` to machine precision, and so probably some weirdness occurs in the remapping at the level of the solver tolerance.

boyceg (approved):

PR #1877, opened 2026-02-20
add kernel options to documentation

marshallrdavey:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/LEInteractor.h
+++ b/ibtk/include/ibtk/LEInteractor.h
@@ -70,6 +70,35 @@ namespace IBTK
  * to interpolate data from Eulerian grid patches onto Lagrangian meshes and to
  * spread values (\em not densities) from Lagrangian meshes to Eulerian grid
  * patches.
+ *
+ * Users specify a kernel for the Lagrangian-Eulerian interaction. These are the
+ * available kernel options with a description.
+ *
+ * | Kernel Name          | Description |
+ * |----------------------|-|
+ * | BSPLINE_3            | |
+ * | BSPLINE_4            | |
+ * | BSPLINE_5            | |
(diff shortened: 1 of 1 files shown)


boyceg (changes requested):
This overlaps with the pending CAV stack, starting at #1955, so let’s keep the descriptions consistent with those (new) conventions:
- Remove `INVALID`; it is for internal use only.
- Describe `BSPLINE_N` as the N-point cardinal B-spline of degree N−1. We may need to distinguish its natural support from the potentially padded stencil returned by `LEInteractor::getStencilSize()`.
- For composite kernels, the first order applies along the selected axis and the second in the remaining directions. For side data, the selected axis is the side-normal direction. Consequently, `COMPOSITE_BSPLINE_32` and `COMPOSITE_BSPLINE_23` are different tensor-product kernels.
- Keep the alias relationships in #1955: `PIECEWISE_CONSTANT` corresponds to `BSPLINE_1`, `PIECEWISE_LINEAR` corresponds to `BSPLINE_2`, and `DISCONTINUOUS_LINEAR` corresponds to `COMPOSITE_BSPLINE_21`.
- For now, this table should be attached to `LEInteractor`. (Although we may want to revisit this once #1955 is merged.)

boyceg:
@marshallrdavey, do you want to finish this one up, or shall we defer it for whatever IB kernel mayhem emerges from my pile of branches?

marshallrdavey:
@boyce we should defer it. I can close this. It would be redundant.

boyceg:
Sounds good. I think that if we can ever get CAV R01d (#1994) merged we can replace `LEInteractor` with a much more performant thing.

PR #1878, opened 2026-02-23
SAMRAI compatibility shim classes to enable updating to SAMRAI 4

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


ibtk/include/samrai_compatibility/algs/SAMRAIPatchBoundaryEdgeSum.h
```diff
@@ -0,0 +1,46 @@
...
+//
+// ---------------------------------------------------------------------
+
+#ifndef included_IBTK_samrai_compatability_algs_SAMRAIPatchBoundaryEdgeSum
+#define included_IBTK_samrai_compatability_algs_SAMRAIPatchBoundaryEdgeSum
+
+#include <ibtk/config.h>
+
+#if defined(__has_include)
+#if __has_include(<SAMRAI/algs/PatchBoundaryEdgeSum.h>)
```
drwells: If all of these headers have the same structure, then we should automatically generate them in the build system instead of checking them in.

drwells: We can also significantly simplify these headers by requiring IBSAMRAI2 install things in `$prefix/include/SAMRAI/`. fixing IBAMR's headers, and then adding the subdirectory paths to user projects to preserve compatibility.

boyceg:
> Maybe it is time to do this. Can you run `make indent` and fix the other minor header problems? It would be great to see how this handles the CI.

Unfortunately I don't have the right version of `clang-format` to run `make indent`. Is there any easy way to deal with this?

ibtk/include/samrai_compatibility/algs/SAMRAIPatchBoundaryEdgeSum.h
boyceg: I will see if I can do it. I think it makes sense to squash these commits — originally the plan was to have them organized somehow or another, but they are all the same so maybe it is not so important to split up into multiple commits.

drwells:
Try `./scripts/formatting/download-clang-format`

ibtk/include/samrai_compatibility/algs/SAMRAIPatchBoundaryEdgeSum.h
boyceg: > We can also significantly simplify these headers by requiring IBSAMRAI2 install things in `$prefix/include/SAMRAI/`. fixing IBAMR's headers, and then adding the subdirectory paths to user projects to preserve compatibility.

So that would simplify some thing and complicate others. Right now, this is using the different locations of the headers to determine if it is SAMRAI 2 or SAMRAI 4. Because this is meant to be a temporary set of shims, I think it is OK to leave them as is. If you want to do this for IBSAMRAI2, then we will need another way to identify which SAMRAI version we are using.

drwells: If we have the infrastructure to switch back-and-forth easily right now then it isn't a big deal.

boyceg: This is now simplified a bit by having a macro that handles the detection.

boyceg:
@drwells I think this is basically done for IBSAMRAI2. There are tests that are failing, but I also see those tests failing on master on my MacBook Pro. What is the strategy in this case?

examples/adv_diff/ex0/UFunction.h
```diff
@@ -19,9 +19,15 @@
...
+#include <ibtk/samrai_compatibility_names.h>
 
 // SAMRAI INCLUDES
-#include <CartesianGridGeometry.h>
+#include <SAMRAICartesianGridGeometry.h>
+#include <SAMRAIGridGeometry.h>
+#include <SAMRAIPatch.h>
+#include <SAMRAIPatchLevel.h>
+#include <SAMRAIPointer.h>
+#include <SAMRAIVariable.h>
```
drwells, line 30: Can we avoid prefixing everything with `SAMRAI`?

boyceg: What would you prefer?

boyceg: Having a prefix helps with avoiding naming collisions and ensuring that everything is converted correctly. I guess that we could have the prefix be the namespace, but it seems good to me to have a different class name for the shim class vs. the original SAMRAI class.

drwells: I'd prefer a namespace, but I'm not that strongly invested in this choice as long as it is relatively easy to change later.

boyceg: Having a different class name makes it 100% clear that the code is using the shim. IMO, it is confusing for Pointer to be both the original class and the shim.

boyceg: Anyway, please tell me what you'd propose. I think that @abarret also has some opinions.

PR #1883, opened 2026-02-27
Use angle brackets for most #include statements

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/minor/20260302GriffithIncludeFormatting
+++ b/doc/news/changes/minor/20260302GriffithIncludeFormatting
@@ -1,0 +1,3 @@
+Adopted angle-bracket include formatting across IBAMR and IBTK sources, headers, examples, and tests, while preserving quoted includes for tests.h and same-directory local headers.
+<br>
+(Boyce Griffith and Codex, 2026/03/02)


scripts/formatting/docs/include_style_rewrite.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Optional custom exception regex
+
+If you need additional quoted exceptions:
+
+```bash
+./scripts/maintenance/rewrite_include_quotes_to_angle_brackets.py --write \
+  --keep-quoted-regex '(^|/)(tests\\.h|my_local_header\\.h)$' \
+  tests
+```
```
drwells, line 86: Can you move these "scripts" into a new folder? They aren't really documentation.

doc/samrai_compatibility_adoption.md
```diff
@@ -0,0 +1,103 @@
...
+Use the rewrite scripts for broad, mechanical updates across many files.
+
+1. Iteratively apply SAMRAI alias/type rewrites until convergence.
+2. Iteratively apply include quote-to-angle rewrites until convergence.
+3. Require pre-format dry-run convergence (`updated=0` and `changed=0`).
+4. Run formatting (`make indent-all` in the build directory).
+5. Re-run both scripts in dry-run mode for post-format reporting (informational).
+6. Build and fix remaining one-off compile issues.
+
+Small example:
```
drwells: This should be in the other PR

examples/adv_diff/ex0/example.cpp
```diff
@@ -103,25 +103,25 @@ main(int argc, char* argv[])
...
             new StandardTagAndInitialize<NDIM>("StandardTagAndInitialize",
                                                time_integrator,
                                                app_initializer->getComponentDatabase("StandardTagAndInitialize"));
-        Pointer<BergerRigoutsos<NDIM> > box_generator = new BergerRigoutsos<NDIM>();
-        Pointer<LoadBalancer<NDIM> > load_balancer =
+        Pointer<BergerRigoutsos<NDIM>> box_generator = new BergerRigoutsos<NDIM>();
+        Pointer<LoadBalancer<NDIM>> load_balancer =
             new LoadBalancer<NDIM>("LoadBalancer", app_initializer->getComponentDatabase("LoadBalancer"));
-        Pointer<GriddingAlgorithm<NDIM> > gridding_algorithm =
+        Pointer<GriddingAlgorithm<NDIM>> gridding_algorithm =
```
drwells, line 116: we will need to rebase.

scripts/formatting/docs/include_style_rewrite.md
boyceg: The scripts are in the scripts directory. I agree the path stated here is potentially confusing. The idea is that you run this in the root directory of the source code tree.

Would you prefer the documentation move over to the scripts directory, or for the documentation to stay here but with a clearer statement on where to run this (e.g. preface by `cd /path/to/IBAMR`)?

doc/samrai_compatibility_adoption.md
boyceg: Whoops. I didn't originally have this as a separate commit but figured you would want it split out like this. 😅

examples/adv_diff/ex0/example.cpp
boyceg: Ugh.

boyceg: I think that this is fixed now.

scripts/formatting/docs/include_style_rewrite.md
boyceg: BTW, I am not sure what you mean that they are not documentation. What is documentation? It seems like this is documentation!

boyceg: If documentation should go somewhere other than `doc` we should document where documentation should go. 😆

boyceg:
@drwells I think this addresses review comments other than the question about where the documentation should go (and the philosophical question about whether it is actually documentation or not).

scripts/formatting/docs/include_style_rewrite.md
abarret: Currently, the scripts located in the `scripts` directory are minimally documented in their own directory. I think it makes sense to keep the scripts and their documentation as close to each other as possible.

doc/news/changes/minor/20260227GriffithIncludeFormatBaseline
```diff
@@ -0,0 +1,3 @@
+Adopted angle-bracket include formatting baseline while preserving quoted includes for tests.h and same-directory local headers, and added CMake rewrite targets for changed-file and full-tree application.
```
drwells: I noticed all these PRs use the word 'baseline': why?

scripts/formatting/docs/include_style_rewrite.md
drwells: These files aren't documentation the sense that `visualization.md` is: there's no circumstance where a user is going to want to read this. It might be useful to log the plans we send to codex, but since they aren't documentation we shouldn't put them in this directory.

CMakeLists.txt
```diff
@@ -1424,3 +1424,15 @@ ADD_CUSTOM_TARGET(indent-all
...
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+  COMMAND ./scripts/maintenance/rewrite-include-style
+  COMMENT "Rewriting include directives in recently changed files"
+  )
+
+ADD_CUSTOM_TARGET(rewrite-include-style-all
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+  COMMAND ./scripts/maintenance/rewrite-include-style-all
+  COMMENT "Rewriting include directives in all files"
+  )
```
drwells:
```suggestion
```
We shouldn't have separate commands for different formatting concerns: they are all encapsulated by `make indent`.

I'm also fine with not committing any of the infrastructure we used to convert to brackets.

scripts/maintenance/tests/test_rewrite_samrai_compatibility_file.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
+
+import importlib.util
+import pathlib
+import sys
+import tempfile
+import unittest
```
drwells: This isn't part of the bracket redo.

scripts/maintenance/rewrite-include-style-all
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+source scripts/formatting/indent_common.sh
+
+rewrite_include_style_file()
+{
+  file="${1}"
+  python3 ./scripts/maintenance/rewrite_include_quotes_to_angle_brackets.py --write "${file}" >/dev/null
+}
+export -f rewrite_include_style_file
```
drwells: If we want this to be part of `make indent` then we need to write it in shell: we don't require python as a dependency.

CMakeLists.txt
boyceg: I think that this is very clearly not indentation. 😀

scripts/maintenance/rewrite-include-style-all
boyceg: I think we should have make targets to implement automated changes. I don't think this angle bracket thing should be in make indent, though — it is not an indentation change.

scripts/maintenance/tests/test_rewrite_samrai_compatibility_file.py
boyceg: I will pull it out.

scripts/maintenance/rewrite-include-style-all
boyceg: The goal with these PRs is to add tools to do automated conversion that are easy for users and developers to adopt.

For clang-format, for instance, it is much easier to say "run `make indent`" than "figure out how to run the canonical version of `clang-format`". I think we should provide the same tool here.

Why do we need this tool? If we follow through with my current SAMRAI 4 plan, we are going to be rewriting a _huge_ number of `#include` statements. I don't want to have to preserve quotes vs. angle brackets in the conversion, but then we wind up with a mix of styles that results in other weirdness. It seems better to adopt it now.

CMakeLists.txt
drwells: True, but the way that system evolved is that it covers all general formatting concerns (such as file permissions and line endings). We could rename it, but we don't need more than one tool to canonicalize our formatting.

scripts/maintenance/rewrite-include-style-all
drwells: This overlaps with the other comment, but I think we agree here: all this formatting stuff should be scripted (which is what we do for most cases). If you'd like we can rename (or alias) it to `make format` to make the intention clearer.

boyceg: OK, I will fold this into `make indent` for now. I guess that we can rename `make indent` something else in another PR.

boyceg: I think this is done.

scripts/maintenance/tests/test_rewrite_samrai_compatibility_file.py
boyceg: I think this is done.

CMakeLists.txt
boyceg: I think this is done.

doc/news/changes/minor/20260227GriffithIncludeFormatBaseline
boyceg: I think these are mostly gone. I didn't interrogate Codex about why it likes baseline so much.

scripts/formatting/docs/include_style_rewrite.md
boyceg: I think this is done.

scripts/formatting/indent
```diff
@@ -31,6 +31,8 @@ checks
 # Process all source and header files:
 #
 
+rewrite_include_style_changed
+
```
drwells, line 35: The rest of the script does one check at a time: can you put the new one at the end and follow the current styling?

scripts/formatting/indent-all
```diff
@@ -31,6 +31,8 @@ checks
 # Process all source and header files:
 #
 
+rewrite_include_style_all
```
drwells, line 34: Same

scripts/formatting/indent_common.sh
```diff
@@ -272,3 +280,33 @@ ensure_single_trailing_newline()
...
+
+rewrite_include_style_all()
+{
+  if [ "${IBAMR_SKIP_INCLUDE_STYLE_REWRITE:-false}" = "true" ]; then
+    return
+  fi
+  process "include src ibtk/include ibtk/src examples ibtk/examples tests" \
+    ".*\.(c|cc|cpp|cxx|h|hh|hpp|hxx|inl|tcc)" rewrite_include_style_file
+}
+export -f rewrite_include_style_all
```
drwells, line 312: This script defines the fixup functions: it is the responsibility of the other one to determine which files these functions are run on.

scripts/formatting/indent_common.sh
```diff
@@ -33,6 +33,14 @@ checks() {
     exit 1
   fi
 
+  if ! [ -x "$(command -v "python3")" ]; then
+    echo "***   Warning: no python3 program found."
+    echo "***   Include-style rewrite will be skipped for this indent run."
+    export IBAMR_SKIP_INCLUDE_STYLE_REWRITE=true
+  else
+    export IBAMR_SKIP_INCLUDE_STYLE_REWRITE=false
+  fi
```
drwells, line 42: Let's rewrite this in shell. While we can't use `sed -i` portably, we should essentially do

```bash
sed -i 's/^#\( *\)include\( *\)"\([^.].*\)"/#\1include\2<\3>/' $1
```
to correctly translate the spacing. We basically need that command plus a tempfile: that's like 10 lines of code (see the trailing whitespace function).

I don't think writing this in python is a good idea. The script codex created is huge relative to all our other fixup functions. Does it work correctly with `fix_or_report_error`? It looks like it presently ignores error codes.

boyceg: It is more complicated than just a `sed` one-liner --- it tries to keep quoted `#include` formatting for local headers and for `tests.h`.

If you want to use a different scripting language, please tell me what you'd like me to use other than python.

scripts/formatting/indent_common.sh
boyceg: Please tell me what you want changed here.

drwells (approved):
I changed my mind: let's merge it first and cleanup the python script second.

The python script handles our funny relative includes correctly in tests and examples, which is helpful - my bash one-liner doesn't.

scripts/formatting/indent_common.sh
drwells: Yes, I just noticed that. Let's leave it as-is for the moment.

drwells: > If you want to use a different scripting language, please tell me what you'd like me to use other than python.

Good question - maybe we should just require Python at this point for such things.

scripts/formatting/indent_common.sh
drwells: We aren't super consistent with using `fix_or_report` so let's leave it as-is for now and I'll open an issue - what we have is close enough.

boyceg:
> @boyceg my apologies if I came off as overly harsh in reviewing this: I want to ensure that the CI correctly picks up these problems. I am not sure that it presently does but since you're the main one working on this I right now I don't think that will be a significant problem.

No problem. We should make sure that the stuff that we commit to `master` passes muster. Also, as I am the main person responsible for our Autotools-based build system, you already know my taste is lacking when it comes to scripting. 😆 

I do want to figure out how to use these tools effectively, however — right now we seem to be developer resource constrained, and although YMMV, I think my experience so far suggests that overall this has the potential to be a big force multiplier.

PR #1886, opened 2026-02-27
Reduce CMake/build warning noise for external dependencies

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,6 +159,86 @@ IF("${IBAMR_FORCE_BUNDLED_MUPARSER}" STREQUAL "")
   SET(IBAMR_FORCE_BUNDLED_MUPARSER "OFF")
 ENDIF()
 
+# Normalize linker flags and library search path items so we can reliably
+# remove duplicates without reordering linkage.
+FUNCTION(IBAMR_CANONICALIZE_LINK_ITEM _input _output)
+  SET(_item "${_input}")
+
+  IF("${_item}" MATCHES "^-Wl,-rpath,(.+)$")
+    SET(_path "${CMAKE_MATCH_1}")
+    STRING(REGEX REPLACE "/+$" "" _path "${_path}")
+    IF("${_path}" STREQUAL "")
+      SET(_path "/")
+    ENDIF()
+    SET(_item "-Wl,-rpath,${_path}")
+  ELSEIF("${_item}" MATCHES "^-L(.+)$")
+    SET(_path "${CMAKE_MATCH_1}")
+    STRING(REGEX REPLACE "/+$" "" _path "${_path}")
+    IF("${_path}" STREQUAL "")
+      SET(_path "/")
+    ENDIF()
+    SET(_item "-L${_path}")
+  ENDIF()
+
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -316,7 +396,7 @@ ELSE()
   SET(Boost_INCLUDE_DIRS
     $<BUILD_INTERFACE:${${PROJECT_NAME}_SOURCE_DIR}/ibtk/contrib/boost>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/contrib/boost>)
-  TARGET_INCLUDE_DIRECTORIES(BUNDLED_BOOST INTERFACE ${Boost_INCLUDE_DIRS})
+  TARGET_INCLUDE_DIRECTORIES(BUNDLED_BOOST SYSTEM INTERFACE ${Boost_INCLUDE_DIRS})
```
drwells, line 399: I'm not opposed to doing this, but we already have `IBTK_DISABLE_EXTRA_WARNINGS` which should catch all such problems. We don't need two subsystems for the same thing. Are there warnings which aren't being silenced by that macro?

CMakeLists.txt
```diff
@@ -725,6 +807,38 @@ libMesh is installed incorrectly or the wrong LIBMESH_METHOD was given to \
...
+        -lheapt_w
+        -lgfortran
+        -lquadmath
+        -lHYPRE
+        -lpetsc
+        -lhdf5
+        -lhdf5_hl
+        -lopenblas
+        -lparmetis
+        -lmetis)
```
drwells, line 837: Is there a problem which this fixes? It is unusual to hard-code a list of libraries like this.

I forsee a couple of disadvantages:
1. CMake will detect that it can't correctly link to a shared object file if there are multiple versions in multiple places. By turning this on we lose that extra check.
2. We are not guaranteed that some of these libraries actually are duplicated by another project (likely PETSc): e.g., metis and parmetis.
3. I don't know how this will interact with our "detect multiple versions of MPI" subsystem

Independent of that: let's tackle one problem at a time. This and the header changes should be in separate PRs.

CMakeLists.txt
```diff
@@ -159,6 +159,86 @@ IF("${IBAMR_FORCE_BUNDLED_MUPARSER}" STREQUAL "")
...
+# Normalize linker flags and library search path items so we can reliably
+# remove duplicates without reordering linkage.
+FUNCTION(IBAMR_CANONICALIZE_LINK_ITEM _input _output)
+  SET(_item "${_input}")
+
+  IF("${_item}" MATCHES "^-Wl,-rpath,(.+)$")
+    SET(_path "${CMAKE_MATCH_1}")
+    STRING(REGEX REPLACE "/+$" "" _path "${_path}")
+    IF("${_path}" STREQUAL "")
+      SET(_path "/")
```
drwells, line 171: Are you sure we want to add `-Wl,-rpath,/` to the list of link flags? That seems like a bad idea.

drwells (changes requested):
These are some interesting ideas but let's tackle them independently.

PR #1887, opened 2026-03-01
Implicit IB milestone 1: Restore implicit IB solver to a functioning state

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/minor/20260301Griffith
+++ b/doc/news/changes/minor/20260301Griffith
@@ -1,0 +1,3 @@
+Stabilization: Restore baseline implicit Stokes-IB integration workflow and promote a finished attest test for the implicit hierarchy path.
+<br>
+(Boyce Griffith + Codex, 2026/03/01)
--- a/plans/implicit-ib/roadmap.md
+++ b/plans/implicit-ib/roadmap.md
@@ -1,0 +1,177 @@
+# Implicit IB Roadmap (High-Level Milestones)
+
+## Summary
+Focus on a staged roadmap to stabilize implicit IB, validate algorithmic correspondence, improve performance, and expand capability (parallel + kernel generalization).
+Use deterministic CMake tests as the main confidence signal, but treat milestone checks as directional goals rather than strict gates.
+
+## Roadmap Record
+- Store and maintain roadmap at: `/Users/boyceg/code/IBAMR/doc/plans/implicit-ib/roadmap.md`
+- Keep this file updated with:
+1. Current reference configs
+2. Milestones and status
+3. Key comparison notes vs MATLAB
+4. PR history and follow-up items
+
+---
+
+## Current State (2026-03-01)
+1. Development is on branch `codex/implicit-ib-milestone-1`.
+2. Milestone 1 work is in-progress and currently under review (not yet merged).
+3. Reference implicit stabilization changes are applied in:
+   - `/Users/boyceg/code/IBAMR/src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp`
+   - `/Users/boyceg/code/IBAMR/src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp`
+   - `/Users/boyceg/code/IBAMR/include/ibamr/INSStaggeredHierarchyIntegrator.h`
+4. A promoted finished reference implicit test is added in:
+   - `/Users/boyceg/code/IBAMR/tests/IB/implicit_stokes_ib_01.cpp`
+   - `/Users/boyceg/code/IBAMR/tests/IB/implicit_stokes_ib_01.baseline.input`
+   - `/Users/boyceg/code/IBAMR/tests/IB/implicit_stokes_ib_01.baseline.output`
+   - `/Users/boyceg/code/IBAMR/tests/CMakeLists.txt` (test registration)
+5. Focused validation currently passing:
+   - `attest -R 'IB/implicit_stokes_ib_01.baseline.input'`
+
+---
+
+## Milestone 1 — Cleanup and stabilization
+### Goal
+Clean up implicit IB implementation and land a focused stabilization PR that restores reference implicit workflows and finished test coverage.
+
+### Scope
+1. Remove accidental/debug artifacts that interfere with reference implicit solves.
+2. Stabilize the `IBImplicitHierarchyIntegrator` execution path and required INS-side plumbing.
+3. Promote core reference implicit tests from unfinished/development status into finished attest-managed tests.
+4. Keep this milestone strictly reference-oriented (no new smoother feature development).
+
+### Status
+- In progress.
+- Completed so far:
+1. Reference stabilization edits applied for `IBImplicitStaggeredHierarchyIntegrator` and `INSStaggeredHierarchyIntegrator` synchronization path.
+2. Reference finished implicit attest test added and passing in focused runs.
+3. Roadmap file created under `doc/`.
+- Remaining before closeout:
+1. Final review and cleanup of staged diffs.
+2. Land commit sequence and changelog entry for milestone 1.
+
+### Explicitly Deferred from M1
+1. Coupling-aware vs geometrical ASM behavior work.
+2. New smoother modes/policies and tuning keys.
+3. Transfer-method feature expansion (RT0/FAC/PETSc tuning work).
+4. MATLAB parity/performance studies for coupling-aware smoothers.
+
+---
+
+## Milestone 2 — Stokes infrastructure verification (without IB coupling)
+### Goal
+Verify that Stokes-only infrastructure corresponds to basic Vanka-style multigrid behavior, including known special cases where strict CAV and classical Vanka should coincide.
+
+### Scope
+1. Define reference Stokes-only cases.
+2. Compare IBAMR behavior with MATLAB reference implementations.
+3. Document where strict CAV should match classical Vanka and where behavior intentionally differs.
+
+### Status
+- Not started.
+
+---
+
+## Milestone 3 — RT0 prolongation validation
+### Goal
+Add confidence that RT0 prolongation is behaving as expected, especially divergence-related behavior for velocity transfer.
+
+### Scope
+1. Add focused tests/diagnostics for transfer invariants.
+2. Check both FAC transfer settings and PETSc-side transfer settings.
+3. Record expected behavior and known caveats in roadmap notes.
+4. Verify the RT0 implementation behavior directly.
+5. Add a deterministic transfer test for `R * Id * P`:
+   - row sums are `1` for velocity DOFs and `0` for pressure DOFs,
+   - column sums are `1` for velocity DOFs and `0` for pressure DOFs.
+
+### Status
+- Not started.
+
+---
+
+## Milestone 4 — MATLAB parity for Stokes + IB
+### Goal
+Build exact-comparison cases between IBAMR and MATLAB for Stokes+IB using programmatic structure generation, then identify and reduce discrepancies.
+
+### Scope
+1. Align geometry/parameters between codebases.
+2. Compare subdomains, iteration traces, and residual behavior.
+3. Track mismatches with likely causes and resolution plan.
+
+### Status
+- Not started.
+
+---
+
+## Milestone 5 — Bespoke Gauss-Seidel performance work
+### Goal
+Improve performance of the custom GS-like smoother while preserving intended solver semantics.
+
+### Scope
+1. Identify dominant runtime costs.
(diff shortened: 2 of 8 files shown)


boyceg:
This is part one of a series of PRs to get @colegruninger97's coupling-aware Vanka preconditioner into IBAMR. I have a roadmap of what I think we should try to do in doc.

tests/IB/implicit_stokes_ib_01.baseline.output
```diff
@@ -0,0 +1,245 @@
...
+IBHierarchyIntegrator::integrateHierarchy_velocity(): interpolating Eulerian velocity to the Lagrangian mesh
+IBHierarchyIntegrator::postprocessIntegrateHierarchy(): interpolating Eulerian velocity to the Lagrangian mesh
+IBHierarchyIntegrator::postprocessIntegrateHierarchy(): CFL number = 0
+IBHierarchyIntegrator::postprocessIntegrateHierarchy(): Eulerian estimate of upper bound on IB point displacement since last regrid = 0
+IBHierarchyIntegrator::advanceHierarchy(): synchronizing updated data
+IBHierarchyIntegrator::advanceHierarchy(): resetting time dependent data
+
+At end       of timestep # 1
+Simulation time is 0.01
++++++++++++++++++++++++++++++++++++++++++++++++++++
```
drwells, line 244: What's the intent with this test? Since it only does one timestep everything is zero.

plans/implicit-ib/roadmap.md
```diff
@@ -0,0 +1,173 @@
...
+Focus on a staged roadmap to stabilize implicit IB, validate algorithmic correspondence, improve performance, and expand capability (parallel + kernel generalization).
+Use deterministic CMake tests as the main confidence signal, but treat milestone checks as directional goals rather than strict gates.
+
+## Roadmap Record
+- Store and maintain roadmap at: `/Users/boyceg/code/IBAMR/doc/implicit-ib-roadmap.md`
+- Keep this file updated with:
+1. Current baseline configs
+2. Milestones and status
+3. Key comparison notes vs MATLAB
+4. PR history and follow-up items
```
drwells, line 13: This is a useful record of what we are doing but (like the other PR) it isn't something which would go in the manual, so I think we should put it somewhere else.

drwells (changes requested):
Nice!

tests/IB/implicit_stokes_ib_01.baseline.output
boyceg: It runs the implicit solver for one time step to make sure it doesn't crash.

plans/implicit-ib/roadmap.md
boyceg: Sure.

boyceg: Where do you want this kind of stuff?

drwells: Some other directory - perhaps `/ai/` or `/plans/`. Given codex's fascination with the word baselines, we could do `/baselines/` too :)

boyceg: Sounds good.

I can tell it to knock it off with this baselines nonsense. 😀

boyceg:
It looks like an `IBInstrumentPanel` is being created somewhere in the implicit solver. Let me see if I can get rid of that...

include/ibamr/INSStaggeredHierarchyIntegrator.h
```diff
@@ -396,6 +396,7 @@ class INSStaggeredHierarchyIntegrator : public INSHierarchyIntegrator
      */
     SAMRAI::tbox::Pointer<StaggeredStokesPhysicalBoundaryHelper> d_bc_helper;
     SAMRAI::tbox::Pointer<IBTK::SideDataSynchronization> d_U_rhs_side_synch_op;
+    int d_U_rhs_synch_data_idx = IBTK::invalid_index;
```
drwells: Could you add some documentation for this value?

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
```diff
@@ -1567,9 +1567,18 @@ INSStaggeredHierarchyIntegrator::setupSolverVectors(const Pointer<SAMRAIVectorRe
...
+    if (rhs_synch_data_idx != d_U_rhs_synch_data_idx)
+    {
+        d_U_rhs_side_synch_op->deallocateOperatorState();
+        d_U_rhs_side_synch_op->initializeOperatorState(rhs_synch_transaction, d_hierarchy);
+        d_U_rhs_synch_data_idx = rhs_synch_data_idx;
+    }
+    else
+    {
+        d_U_rhs_side_synch_op->resetTransactionComponent(rhs_synch_transaction);
+    }
```
drwells: I missed this the first time. Is this going to be run for everything, or just implicit IB? Would it be possible to move this over to the implicit IB integrator?

drwells (changes requested):
Two more comments. Could you make it clear why (and if) we have to modify the INS solver so this implicit IB solver works?

include/ibamr/INSStaggeredHierarchyIntegrator.h
boyceg: I was able to remove this.

src/navier_stokes/INSStaggeredHierarchyIntegrator.cpp
boyceg: I was able to remove this.

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
```diff
@@ -794,8 +781,6 @@ IBImplicitStaggeredHierarchyIntegrator::integrateHierarchy_velocity(const double
     ierr = MatDestroy(&interp_op);
     IBTK_CHKERRQ(ierr);
 
-    // Execute any registered callbacks.
-    executeIntegrateHierarchyCallbackFcns(current_time, new_time, cycle_num);
```
drwells, line 798: I also missed this one before: where are the callbacks now?

scripts/formatting/indent_common.sh
```diff
@@ -283,19 +283,22 @@ ensure_bracket_includes()
...
+    {
+      if ($0 ~ /^#[[:space:]]*include[[:space:]]*<[^>]+>$/) {
+        header = $0
+        sub(/^#[[:space:]]*include[[:space:]]*</, "", header)
+        sub(/>$/, "", header)
+        if (header in local_headers) sub(/<[^>]+>/, "\"" header "\"")
+      }
+      print
+    }
+  ' <(find "$(dirname "$1")" -type f -exec basename {} \;) "${tmpfile1}" > "${tmpfile2}"
```
drwells: Sorry about making a mess here - I forgot that the CI wasn't set up to actually run these scripts on macs.

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
```diff
@@ -203,21 +215,51 @@ IBImplicitStaggeredHierarchyIntegrator::preprocessIntegrateHierarchy(const doubl
...
+        d_eul_sol_vec->addComponent(p_var, p_scratch_idx, wgt_cc_idx, d_hier_pressure_data_ops);
+
+        d_eul_rhs_vec = d_eul_sol_vec->cloneVector(d_object_name + "::eulerian_rhs_vec");
+        d_u_scratch_vec = d_eul_sol_vec->cloneVector(d_object_name + "::u_scratch_vec");
+        d_f_scratch_vec = d_eul_rhs_vec->cloneVector(d_object_name + "::f_scratch_vec");
+    }
+
+    d_eul_rhs_vec->allocateVectorData(current_time);
+    d_u_scratch_vec->allocateVectorData(current_time);
+    d_f_scratch_vec->allocateVectorData(current_time);
```
abarret: Suggest moving this to its own member function.

abarret (commented):
It seems the only code change is to make the solver vec's member variables.

The roadmap seems reasonable. Is RT0 prolongation important for solver performance? Is the intent for the first few milestones to only test the application of the smoother, without prolongation or restrictions?

boyceg:
> The roadmap seems reasonable. Is RT0 prolongation important for solver performance?

RT0 is important. @colegruninger97 can you send the references?

> Is the intent for the first few milestones to only test the application of the smoother, without prolongation or restrictions?

This is mostly getting @colegruninger97's solver from his PhD thesis implemented. The whole thing was broken, though — hence this commit.

All of the prolongation and restriction infrastructure is already in place, but was never carefully tested. Like after this first PR, the implicit solver "works" — it just isn't competitive with an explicit solver.

The goal of this project is to see if we can come up with an implicit solver that is competitive with an explicit one — hopefully faster.

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
boyceg: Done.

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
boyceg: Added back to the common integrateHierarchy() implementation.

At some point I am going to split the two modes (velocity vs. position-based), but not yet.

boyceg:
FWIW, we are set up to try both RT0 and linear prolongation in IBAMR along with the corresponding adjoint restriction.

doc/news/changes/minor/20260301Griffith
```diff
@@ -0,0 +1,3 @@
+Stabilization: Restore baseline implicit Stokes-IB integration workflow and promote a finished attest test for the implicit hierarchy path.
```
drwells, line 1: Can you make this more in line with the way we usually write changelog entries?

I'm not sure we should add a changelog at this point since it will be superseded by "implicit IB now works" later.

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
```diff
@@ -794,11 +750,52 @@ IBImplicitStaggeredHierarchyIntegrator::integrateHierarchy_velocity(const double
...
-    executeIntegrateHierarchyCallbackFcns(current_time, new_time, cycle_num);
     return;
 } // integrateHierarchy_velocity
 
+void
+IBImplicitStaggeredHierarchyIntegrator::setupSolverVectors(const double current_time,
+                                                           const int coarsest_ln,
+                                                           const int finest_ln)
+{
+    Pointer<INSStaggeredHierarchyIntegrator> ins_hier_integrator = d_ins_hier_integrator;
```
drwells, line 761: Since this is also a 'staggered'  class, can we just store a pointer with the correct type?

tests/IB/implicit_stokes_ib_01.cpp
```diff
@@ -0,0 +1,280 @@
...
+            iteration_num += 1;
+        }
+
+        // Cleanup Eulerian boundary condition specification objects (when
+        // necessary).
+        for (unsigned int d = 0; d < NDIM; ++d) delete u_bc_coefs[d];
+
+    } // cleanup dynamically allocated objects prior to shutdown
+
+    return 0;
```
drwells, line 279: Do you anticipate that we will reuse this same test executable for future tests?

drwells (changes requested):
I have a few minor comments.

doc/news/changes/minor/20260301Griffith
boyceg: Before this branch, the solver was crashing when I tried to use it, and so this is bringing things back to a baseline working (if still inefficient) state and adding a test to make sure that it doesn't crash.

src/IB/IBImplicitStaggeredHierarchyIntegrator.cpp
boyceg: I think that I would rather use the more general base class. I think that the position-based implicit method doesn't formally require a staggered solver, and so at some point I think I'd like to split them into two classes — one that needs this specialized staggered-grid multigrid algorithm, and another that (may) not.

tests/IB/implicit_stokes_ib_01.cpp
boyceg: Yes of course. Also, AFAIK, there is no test coverage for this class in `master` at present, so we need to do something.

boyceg:
Milestone branch 5 (#1904) got pretty messy so I decided to do these (#1919 -- #1926).

PR #1891, opened 2026-03-03
CI: enable SILO for Arch Linux.

drwells:
We compiled and installed SILO

https://github.com/IBAMR/docker-files/blob/30e47f39719af9c1c7aa909c9bbb4139e52e1588/dependencies/Dockerfile.archlinux#L51

but didn't turn it on.

### DIFF
--- a/.github/workflows/push-pull.yml
+++ b/.github/workflows/push-pull.yml
@@ -79,7 +79,7 @@ jobs:
         run: |
           mkdir build
           cd build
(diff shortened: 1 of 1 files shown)


boyceg:
I think we should leave Silo disabled for this build. We are treating Silo as optional, and it seems good to check that everything works with / without Silo.

I had forgotten that there was a way to disable Silo warnings for the tests in #1887. Now that I'm using those, the tests work fine with / without Silo.

PR #1911, opened 2026-03-22
Add markImportant() to RestartCleaner for protecting restart directories from deletion

zoraaaa03:
### Description
This pull request adds importance marking to RestartCleaner (PR #1830), allowing users to protect specific restart directories from deletion during cleanup. 

The implementation uses a .important marker file inside the restart directory. This design is stateless and crash-safe: markers persist on the filesystem across job crashes, SLURM restarts, and new simulation runs without any external state files.

This feature was developed under the guidance of @drwells. Thanks @drwells!

### Key Features
- Creates a .important marker file inside the corresponding restore.XXXXXX/ directory.
- keepRecentN() now checks for importance markers before deletion.
- No input file changes required: fully backward compatible with existing configurations.
- Markers are safe to call multiple times and survive across simulations.

### Testing
- Existing Tests pass with unchanged baseline output.
- Added 4 extra functional tests.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/RestartCleaner.h
+++ b/ibtk/include/ibtk/RestartCleaner.h
@@ -46,7 +46,8 @@ namespace IBTK
  * -# Scan a specified directory for subdirectories matching the pattern "restore.NNNNNN..."
  * -# Parse restart restore numbers from these directory names
  * -# Sort directories based on restart restore numbers
- * -# Keep the N most recent directories and delete the rest
+ * -# Keep the N most recent directories plus any marked important, and delete the rest
+ * -# Mark specific directories as important to protect them from cleanup
  *
  * \note This class assumes restart directories follow the naming pattern "restore.NNNNNN..."
  * where the restart restore number is zero-padded to a minimum of 6 digits by IBAMR.
@@ -68,10 +69,11 @@ namespace IBTK
  * // Configuration-driven usage with SAMRAI Database
  * Pointer<Database> restart_db = input_db->getDatabase("RestartCleaner");
  * RestartCleaner cleaner("RestartCleaner", restart_db);
- * cleaner.cleanup();  // Use dry_run=true in config to preview without deleting
  *
- * // Check available restart restore numbers
- * auto restore_numbers = cleaner.getAvailableRestoreNumbers();
+ * // In the restart dump block of the time loop:
+ * RestartManager::getManager()->writeRestartFile(dirname, iteration_num);
+ * cleaner.markImportant(iteration_num);  // protect from deletion
+ * cleaner.cleanup();  // deletes old directories, keeps recent + important
  * \endcode
  */
 class RestartCleaner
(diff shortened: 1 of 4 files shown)


boyceg:
I think there is one MPI issue in `markImportant()` that needs to be fixed before merging.

In the missing-directory branch, rank 0 returns at [RestartCleaner.cpp:107](https://github.com/IBAMR/IBAMR/blob/cddbb1324924042decfab7bdc30301d3de6effdb/ibtk/src/utilities/RestartCleaner.cpp#L107), skipping the barrier at line 130.

The other ranks enter that barrier, so the intended warning-and-skip behavior can hang the simulation.

This can be reproduced with this PR’s implementation: the two-rank existing-directory case completes, while the missing-directory case times out with rank 1 still inside `markImportant()`. The extended functional test passes with `attest`.

Would it be possible to restructure this branch so that marker creation is skipped when the directory is absent, but every rank still reaches the common barrier?

It might be worthwhile to add a two-processor test case that verifies `markImportant()` returns on all ranks when the requested directory does not exist.

boyceg (changes requested):
I think this may need to be revised to work correctly in parallel.

boyceg:
@zoraaaa03, @drwells, do you all want to finish this up, or shall we close this PR?

zoraaaa03:
> @zoraaaa03, @drwells, do you all want to finish this up, or shall we close this PR?

Hi, I’m currently working through the changes based on your suggestions, and I expect to have them completed this week. Thank you for checking in.

boyceg:
Great!!

zoraaaa03:
@boyceg Thanks for catching this!!! I changed the missing-directory branch of markImportant() to an if/else (the marker creation is skipped in the else branch instead of returning early), so rank 0 now reaches the barrier at the end of the function together with the other ranks, while the warning-and-skip behavior stays the same.

For testing and verification, new Test 11 calls markImportant() for a restore number that does not exist and checks via a sum reduction that the call returns on every rank (with the previous implementation it times out with rank 1 stuck inside markImportant(), and with this change it passes). I also added restart_cleaner_function_test.mpirun=2.input so the whole functional test runs on two processors (the test helpers now create and remove the test directories on rank 0 only to make sure Tests 1 to 11 work in parallel). Please let me know if there's anything else that needs to change :)

boyceg:
@zoraaaa03: OK, I think this fixes the issue!

One last thing is that I think that we would like to get rid of the merge commit ([44d1b8e](https://github.com/IBAMR/IBAMR/pull/1911/commits/44d1b8e5db4163794c4941f66742fe4cf8a01876)) before merging. I think that you can accomplish this by rebasing your branch on the current `master` branch and then squashing your branch down to just one commit.

(If you haven't used `git rebase` before, you might want to make a copy of your branch before you rebase it — it is definitely possible to lose work in a misformed rebase! 😬)

boyceg (approved):
Looks good to me. Maybe wait for the CI jobs all to clear before merging.

zoraaaa03:
> @zoraaaa03: OK, I think this fixes the issue!
> 
> One last thing is that I think that we would like to get rid of the merge commit ([44d1b8e](https://github.com/IBAMR/IBAMR/pull/1911/commits/44d1b8e5db4163794c4941f66742fe4cf8a01876)) before merging. I think that you can accomplish this by rebasing your branch on the current `master` branch and then squashing your branch down to just one commit.
> 
> (If you haven't used `git rebase` before, you might want to make a copy of your branch before you rebase it — it is definitely possible to lose work in a misformed rebase! 😬)

Done! I rebased on the current master and squashed everything into one commit with git merge --squash, the same way as for the first RestartCleaner PR. The code is unchanged. Would you like a changelog entry as well?

boyceg:
> > @zoraaaa03: OK, I think this fixes the issue!
> > One last thing is that I think that we would like to get rid of the merge commit ([44d1b8e](https://github.com/IBAMR/IBAMR/pull/1911/commits/44d1b8e5db4163794c4941f66742fe4cf8a01876)) before merging. I think that you can accomplish this by rebasing your branch on the current `master` branch and then squashing your branch down to just one commit.
> > (If you haven't used `git rebase` before, you might want to make a copy of your branch before you rebase it — it is definitely possible to lose work in a misformed rebase! 😬)
> 
> Done! I rebased on the current master and squashed everything into one commit with git merge --squash, the same way as for the first RestartCleaner PR. The code is unchanged. Would you like a changelog entry as well?

It is always good to have a change log entry.

boyceg:
@drwells what do you think?

PR #1916, opened 2026-03-31
CMake: if we are using a new enough version then deduplicate libraries.

drwells:
Alternative to #1886: this is the kind of thing we should let cmake handle. Let's see how well it does!

See

https://cmake.org/cmake/help/latest/policy/CMP0156.html

for more information.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,11 @@ IF("${CMAKE_BINARY_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
           You must delete them, or cmake will refuse to work.")
 ENDIF()
 
+IF(POLICY CMP0156)
+  # New in CMake 3.29: De-duplicate libraries on link lines.
+  cmake_policy(SET CMP0156 NEW)
+ENDIF()
+
 # Version info:
 FILE(STRINGS "${CMAKE_SOURCE_DIR}/VERSION" _version LIMIT_COUNT 1)
 STRING(REGEX REPLACE "^([0-9]+)\\..*" "\\1" IBTK_VERSION_MAJOR "${_version}")


boyceg (approved):

boyceg:
It looks like the remaining macOS duplicates come from overlap between PETSc/libMesh’s library lists and the runtime libraries CMake or the compiler driver adds automatically.

What do you think about something like this just before `ADD_SUBDIRECTORY(ibtk)`?
```
IF(APPLE)
  # IBTK's public dependencies already supply these Fortran runtime libraries.
  FOREACH(_lib IN LISTS CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES)
    IF("-l${_lib}" IN_LIST PETSC_LIBRARIES OR "-l${_lib}" IN_LIST LIBMESH_LIBRARIES)
      LIST(REMOVE_ITEM CMAKE_Fortran_IMPLICIT_LINK_LIBRARIES "${_lib}")
    ENDIF()
  ENDFOREACH()

  # The C++ compiler driver supplies its own implicit libraries.
  FOREACH(_lib IN LISTS CMAKE_CXX_IMPLICIT_LINK_LIBRARIES)
    LIST(REMOVE_ITEM PETSC_LIBRARIES "-l${_lib}")
    LIST(REMOVE_ITEM LIBMESH_LIBRARIES "-l${_lib}")
  ENDFOREACH()
ENDIF()
```
This should preserve the Fortran runtimes in IBTK’s public dependencies, so they remain available to C++-only applications linking static IBAMR.

drwells:
I'm hesitant to start messing around with link lines in that way: that's the kind of thing cmake should handle for us.

I'll rebase. Maybe newer versions of cmake have improved this option.

boyceg:
@drwells shall we merge this one since it gets rid of at least some of the dup'ed libs?

drwells:
I'm ambivalent about adding it since it doesn't fix the original problem:

```
ld: warning: duplicate -rpath '/opt/homebrew/opt/hdf5-mpi/lib' ignored
ld: warning: ignoring duplicate libraries: '-lc++', '-lemutls_w', '-lgfortran', '-lheapt_w', '-lmpi', '-lquadmath'
```
these warnings are harmless so it isn't a substantial issue.

boyceg:
I think we should merge it: it doesn't seem to harm anything either, and maybe support for the policy will improve with time?

PR #1927, opened 2026-04-20
Implicit IB 0 - basic infrastructure

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/utilities/ibtk_utilities.cpp
+++ b/ibtk/src/utilities/ibtk_utilities.cpp
@@ -13,20 +13,80 @@
 
 /////////////////////////////// INCLUDES /////////////////////////////////////
 
+#include <ibtk/IBTK_CHKERRQ.h>
 #include <ibtk/IBTK_MPI.h>
 #include <ibtk/ibtk_utilities.h>
+
+#include <tbox/Database.h>
+
+#include <petscsys.h>
 
 #include <CartesianPatchGeometry.h>
 #include <PatchLevel.h>
 
 #include <algorithm>
 #include <cmath>
+#include <filesystem>
+#include <fstream>
+#include <string>
 
 #include <ibtk/app_namespaces.h>
 
 namespace IBTK
 {
 /////////////////////////////// PUBLIC ///////////////////////////////////////
+
+void
+load_petsc_options_file(Pointer<Database> input_db, int argc, char* argv[], const std::string& source_dir)
+{
+    std::string options_key;
+    if (input_db->keyExists("PETSC_OPTIONS_FILE"))
+    {
+        options_key = "PETSC_OPTIONS_FILE";
+    }
+    else if (input_db->keyExists("petsc_options_file"))
+    {
+        options_key = "petsc_options_file";
+    }
+    else
+    {
+        return;
+    }
+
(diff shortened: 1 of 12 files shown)


ibtk/src/utilities/ibtk_utilities.cpp
```diff
@@ -13,20 +13,80 @@
...
+        options_key = "petsc_options_file";
+    }
+    else
+    {
+        return;
+    }
+
+    const std::string petsc_options_file = input_db->getString(options_key);
+    std::string resolved_options_file = petsc_options_file;
+    auto path_exists = [](const std::string& path) -> bool
```
drwells, line 58: I'll summarize what I wrote about this elsewhere so we have everything in one place.

I don't think this is a good idea because it is too clever: it follows symbolic links back to where the input file exists to try and find a petsc options file. This definitely works with the testsuite but it is inconsistent with how we find and load all other files. Rather than special-casing such things, I recommend following the same pattern that we already use elsewhere:

https://github.com/IBAMR/IBAMR/blob/11fd9c311a1dd03fdace28374aa82a157460d7bf/tests/CIB/cib_plate.cpp#L102-L126

This always works (though we should have a barrier here!) and is consistent with how we handle petsc options files right now (as well as every other file type).

Another alternative would be to amend `attest` to define tests as executable, input, output, and any other optional files with the same prefix (everything up to the `.input` suffix). I think that would be a reasonable extension of `attest` (though we'd have to go back and clean everything up to preserve consistency).

ibtk/src/utilities/ibtk_utilities.cpp
```diff
@@ -13,20 +13,80 @@
...
+    else if (input_db->keyExists("petsc_options_file"))
+    {
+        options_key = "petsc_options_file";
+    }
+    else
+    {
+        return;
+    }
+
+    const std::string petsc_options_file = input_db->getString(options_key);
```
drwells, line 56: Independent of that: no, we should not load from the top-level database. `AppInitializer` uses `Main`.

drwells (changes requested):

ibtk/src/utilities/ibtk_utilities.cpp
boyceg: I would very strongly prefer not to have code that explicitly copies input files embedded in the tests, since we need to parse the input files to know the name of the files that are being copied. It seems much better to just use the path that is available to make sure we can load the files.

boyceg: What if we have a version of AppInitializer -- perhaps TestInitializer? -- that does this nonstandard stuff?

drwells: Sure - feel free to add it to `tests.h`.

What would you recommend we do for vertex and ExodusII files instead of copying them?

boyceg: Good question. Let me see if I can come up with something!

boyceg:
I think that these implicit IB PRs are going to be superseded by a sequence of more focused PRs.

PR #1928, opened 2026-04-21
Use data() to access storage in standard containers

boyceg:
      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/src/math/PETScMatUtilities.cpp
+++ b/ibtk/src/math/PETScMatUtilities.cpp
@@ -222,9 +222,9 @@ PETScMatUtilities::constructPatchLevelCCLaplaceOp(Mat& mat,
                         PETSC_DETERMINE,
                         PETSC_DETERMINE,
                         0,
-                        n_local ? &d_nnz[0] : nullptr,
+                        get_data_or_null(d_nnz),
                         0,
-                        n_local ? &o_nnz[0] : nullptr,
+                        get_data_or_null(o_nnz),
                         &mat);
     IBTK_CHKERRQ(ierr);
 
@@ -269,7 +269,8 @@ PETScMatUtilities::constructPatchLevelCCLaplaceOp(Mat& mat,
                             mat_cols[stencil_index] = (*dof_index_data)(i + stencil[stencil_index], d);
                         }
                     }
-                    ierr = MatSetValues(mat, 1, &dof_index, stencil_sz, &mat_cols[0], &mat_vals[0], INSERT_VALUES);
+                    ierr =
+                        MatSetValues(mat, 1, &dof_index, stencil_sz, mat_cols.data(), mat_vals.data(), INSERT_VALUES);
                     IBTK_CHKERRQ(ierr);
                 }
             }
@@ -372,9 +373,9 @@ PETScMatUtilities::constructPatchLevelSCLaplaceOp(Mat& mat,
                         PETSC_DETERMINE,
                         PETSC_DETERMINE,
                         0,
-                        n_local ? &d_nnz[0] : nullptr,
+                        get_data_or_null(d_nnz),
                         0,
-                        n_local ? &o_nnz[0] : nullptr,
+                        get_data_or_null(o_nnz),
                         &mat);
     IBTK_CHKERRQ(ierr);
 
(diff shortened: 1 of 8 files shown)


abarret:
I don't think `vector::data()` is guaranteed to be `nullptr` if empty. Does PETSc require `nullptr` for empty arrays?

boyceg:
> I don't think `vector::data()` is guaranteed to be `nullptr` if empty. Does PETSc require `nullptr` for empty arrays?

UGH.

boyceg:
I am inclined to add a wrapper to avoid the proliferation of ` ? : ` code. What do you think?

abarret:
> I am inclined to add a wrapper to avoid the proliferation of `? :` code. What do you think?

I don't have much of a preference. It doesn't seem like the extra dozen characters of `? :` is that big of a deal to hide it in a wrapper.

It also seems that gcc returns a nullptr if the vector is empty, so maybe all the reasonable compilers do this?

drwells:
Another case in which `std::vector::data()` is not `nullptr` is if `size() == 0` and `capacity() > 0` (the standard only requires that [`data()`, `data() + size()`) is a valid range). But: why would PETSc dereference a pointer which corresponds to 0 elements? That's pretty obviously a PETSc bug (and we pass `nullptr` to PETSc all the time).

ibtk/src/math/PETScMatUtilities.cpp
```diff
@@ -216,16 +216,8 @@ PETScMatUtilities::constructPatchLevelCCLaplaceOp(Mat& mat,
...
-                        n_local,
-                        PETSC_DETERMINE,
-                        PETSC_DETERMINE,
-                        0,
-                        n_local ? &d_nnz[0] : nullptr,
-                        0,
-                        n_local ? &o_nnz[0] : nullptr,
-                        &mat);
+    ierr = MatCreateAIJ(
+        PETSC_COMM_WORLD, n_local, n_local, PETSC_DETERMINE, PETSC_DETERMINE, 0, d_nnz.data(), 0, o_nnz.data(), &mat);
```
drwells: For example, PETSc states that "if the *_nnz parameter is given then the *_nz parameter is ignored". We could be a little more correct and use `PETSC_DEFAULT` instead of `0` here but I don't think it is required.

There must be `n_local` entries in this vector so the current version is correct with or without `nullptr`.

drwells:
The current implementation isn't wrong (AFAIK!) so another possibility is to mechanically transform all the `&a[0]`s into `a.data()` and leave the ternary ifs.

boyceg:
> The current implementation isn't wrong (AFAIK!) so another possibility is to mechanically transform all the `&a[0]`s into `a.data()` and leave the ternary ifs.

I hate this particular use of the ternary operator. For instance, it requires listing the same variable name twice.

How about an inline helper function `get_data` that returns the data pointer (if non-null) and `nullptr` otherwise?

boyceg:
@drwells & @abarret, what do you think of the revision?

boyceg:
I was going to go crazy and change all usages of `&xxx[0]` to `xxx.data()` but decided to make this a little more narrow for now.

abarret (commented):
It seems that you only switch IB things to `get_data_or_null` while switching Eulerian objects to using just `.data()`. Is that intentional?

boyceg:
The intention was that calls that were "guarded" by `xxx.empty() ? nullptr : &xxx[0]` or equivalent became `get_data_or_null(xxx)`, and that thing that were simply `&xxx[0]` became `xxx.data()`.

abarret:
Ok. The calls in the level solvers and mat utilities that look like `nlocal ? &d_nnz[0] : nullptr` are being replaced by `d_nnz.data()`.

boyceg:
> Ok. The calls in the level solvers and mat utilities that look like `nlocal ? &d_nnz[0] : nullptr` are being replaced by `d_nnz.data()`.

Whoops. I think that is fixed now. At the same time, I decided to drop changes to files that are not going to be touched at least in the initial implicit overhaul. The main thing I'd like to do right now is to get those classes up-to-date so that the access patterns are consistent.

abarret (approved):
Looks fine to me.

boyceg:
> Looks fine to me.

Thanks, @abarret!

PR #1931, opened 2026-05-07
CMake: default to GCC's Fortran mangling.

drwells:
CMake doesn't recognize GCC-16 yet, but it behaves the same way as every other GCC.


drwells:
Can someone take a look? We need this to update the CI.

boyceg (approved):

PR #1932, opened 2026-05-07
CMake: support Eigen versions newer than 3.4.

drwells:
We are also compatible with newer versions.


drwells:
Can someone take a look? We need this to update the CI.

boyceg (approved):

PR #1933, opened 2026-05-20
Redo our bundled boost.

drwells:
Required by #1930. The RTree implementation is huge: I tried to trim things down to what we need, and while I was at it I wrote a script to copy over only the boost headers we need to compile the tests and examples. I also verified that all of our boost headers can be independently preprocessed (i.e., we aren't missing any includes we actually use in C++17-mode)

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
Woah. That is a lot. I sort of wonder if we should go the opposite route: stop bundling boost and have autoibamr handle it if it is not present.

drwells:
I think I'd rather just not bundle it at all: we already have mandatory external dependencies and a way to install them (and boost is commonly installed via a package manager). We'd have to update autoibamr to install it but that wouldn't be that hard.

boyceg:
The whole reason I started bundling things is because I didn't have autoibamr but also didn't want to ask folks to install these kinds of dependencies 'by hand'.

drwells:
Realistically, if people are on a cluster they can use autoibamr (and boost is typically installed anyway), and if they are on a personal machine they can install boost with the system package manager or use autoibamr - I suppose we might as well unbundle it. I'll see how hard it is to build it ourselves.

drwells:
Putting it in autoibamr is pretty easy: see https://github.com/IBAMR/autoibamr/pull/217

I'm in favor of deleting our bundled copy of boost.

drwells:
Let's go a different way.

PR #1934, opened 2026-06-01
CI: update containers.

drwells:
I uploaded new docker images (containing full boost installations) to Dockerhub. Once this is in we can delete our bundled boost installation.


drwells:
Hm, we have a bunch of things to fix both in ibsamrai2 and IBAMR before this passes.

drwells:
Much better. We still need to merge #1941 first.

drwells:
The mac failure is expected so this is ready!

boyceg (approved):

PR #1935, opened 2026-06-03
Fix some out-of-bounds refinement errors.

drwells:
Nowadays gfortran has options like

    -fcheck=bits,bounds,do

which verifies that we do not mess up bitwise operations, exceed loop bounds, or modify do-loop iteration counters. The second one is most relevant to us: we pass around a lot of arrays and array sizes in independent structures.

Since we should convert to a side box before ever calling these functions, and these functions don't take into account the axis, I can't figure out why we ever wanted the upper bound incremented by 1 in these places. The only assertion error I got was in 3d (i.e., the 3d navier stokes test fails) but we have the same logic error in 2d. Interestingly no IB tests caught this assertion.

I updated some test output but this is a pretty sophisticated bug and I think we should dig through the Fortran during review to make sure this makes sense (and the refining functions do what we want).

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/doc/news/changes/minor/20260603DavidWells
+++ b/doc/news/changes/minor/20260603DavidWells
@@ -1,0 +1,4 @@
+Fixed: Side-centered refinement operators will no longer read out-of-bounds
+data.
+<br>
+(David Wells, 2026/06/03)
--- a/ibtk/src/refine_ops/fortran/cart_side_refine2d.f.m4
+++ b/ibtk/src/refine_ops/fortran/cart_side_refine2d.f.m4
@@ -108,7 +108,7 @@ c
          enddo
       enddo
 
-      do i1=ilower1,iupper1+1
+      do i1=ilower1,iupper1
          coarsen_index(i1,i_c1,i_f1,ratio(1))
          do i0=ilower0,iupper0
             coarsen_index(i0,i_c0,i_f0,ratio(0))
@@ -188,7 +188,7 @@ c
 
       do i1=ilower1,iupper1
          coarsen_index(i1,i_c1,i_f1,ratio(1))
-         do i0=ilower0,iupper0+1
+         do i0=ilower0,iupper0
             coarsen_index(i0,i_c0,i_f0,ratio(0))
             i_f0 = i_c0*ratio(0)
 
@@ -205,7 +205,7 @@ c
          enddo
       enddo
 
-      do i1=ilower1,iupper1+1
+      do i1=ilower1,iupper1
          coarsen_index(i1,i_c1,i_f1,ratio(1))
          do i0=ilower0,iupper0
             coarsen_index(i0,i_c0,i_f0,ratio(0))
--- a/ibtk/src/refine_ops/fortran/cart_side_refine3d.f.m4
+++ b/ibtk/src/refine_ops/fortran/cart_side_refine3d.f.m4
@@ -112,7 +112,7 @@ c
          coarsen_index(i2,i_c2,i_f2,ratio(2))
          do i1=ilower1,iupper1
             coarsen_index(i1,i_c1,i_f1,ratio(1))
-            do i0=ilower0,iupper0+1
+            do i0=ilower0,iupper0
                coarsen_index(i0,i_c0,i_f0,ratio(0))
                if ( i0 .eq. i_f0 ) then
                   u0_f(i0,i1,i2) = u0_c(i_c0,i_c1,i_c2)
@@ -129,7 +129,7 @@ c
 
       do i2=ilower2,iupper2
          coarsen_index(i2,i_c2,i_f2,ratio(2))
-         do i1=ilower1,iupper1+1
+         do i1=ilower1,iupper1
             coarsen_index(i1,i_c1,i_f1,ratio(1))
             do i0=ilower0,iupper0
                coarsen_index(i0,i_c0,i_f0,ratio(0))
@@ -146,7 +146,7 @@ c
          enddo
       enddo
 
-      do i2=ilower2,iupper2+1
+      do i2=ilower2,iupper2
          coarsen_index(i2,i_c2,i_f2,ratio(2))
          do i1=ilower1,iupper1
             coarsen_index(i1,i_c1,i_f1,ratio(1))
@@ -240,7 +240,7 @@ c
          coarsen_index(i2,i_c2,i_f2,ratio(2))
          do i1=ilower1,iupper1
             coarsen_index(i1,i_c1,i_f1,ratio(1))
-            do i0=ilower0,iupper0+1
+            do i0=ilower0,iupper0
                coarsen_index(i0,i_c0,i_f0,ratio(0))
 
                w0 = 1.d0-dble(i0-i_f0)/dx0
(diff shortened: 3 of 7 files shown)


boyceg:
I don't think this is right — I think that it will miss some of the side-centered DOFs. There is always one extra side-centered DOF in the axis-oriented direction.

My guess is that the old implementation is OK because the out-of-bounds value was getting hit with a big-ol' `0.d0`.

drwells:
I had the same thought but there is no axis parameter here (nor do we have some special treatment for it at the call site in `CartSideDoubleSpecializedLinearRefine.cpp`), so I don't see how that could be it.

boyceg:
`u0` always has one extra index in the `i0` direction; `u1` always has one extra index in the `i1` direction, etc.

drwells:
oh yeah, we pass in all the components at once: that makes sense. There must be a better fix than this one.

boyceg:
If you skip the last entry it is definitely wrong.

I don't love this solution, but it would be possible to add a guard to the shifted index using `min` to ensure it never accesses anything out of bounds. It also would be possible to add a guard that does not access entries if the weight is zero.

drwells:
The way we set up these calls is very straightforward (based on box indices) so there must be something funky happening if those are off. I'd also rather not clamp things.

boyceg:
It is simple linear interpolation. At the very last value, it is interpolating beyond the edge of the box, but because it is evaluated on the box boundary, nothing too terrible happens, because the weight that is "off the box" is zero.

boyceg (changes requested):
I think the ~current~ original code is OK. The correct fix is either to add if statements to avoid sampling "out of bounds", or to handle the upper edge of the patch separately. Both seem undesirable.

drwells:
I think we need to do a better job checking input boxes (that would be a good followup project), especially for `SideBox`. For now I need some help figuring out how to handle this case (this is one of those cases where we implicitly switch between cell and side indexing and I cannot remember how this works)

#### backtrace

- `cart_side_specialized_linear_refine3d`: cart_side_refine3d.f.m4 line 282
- `CartSideDoubleSpecializedLinearRefine::refine()`: CartSideDoubleSpecializedLinearRefine.cpp line 176
- (SAMRAI stuff)
- `HierarchyGhostCellInterpolation::fillData()`: HierarchyGhostCellInterpolation.cpp line 602
- `INSStaggeredHierarchyIntegrator::setupPlotDataSpecialized()`: INSStaggeredHierarchyIntegrator.cpp line 2137
- `HierarchyIntegrator::setupPlotData()`: HierarchyIntegrator.cpp line 626

#### boxes

- (in the Fortran routine) `u1_f` has index bounds (31, 31, 31) to (96, 97, 96). The out-of-bounds index is (97, 31, 31) (i.e., clearly out-of-bounds for `u1_f`). Based on our conversation yesterday I think this corresponds to interpolating outside the domain and we should set this value to zero. However, the code definitely tries to set it to a nonzero value (the weight, `w1`, is `0.25` and the right-hand side of the assignment is (rounded) `0.025708`). 
- (in CartSideDoubleSpecializedLinearRefine) `fine_box` (AFAICT this is in cells) has index bounds (96, 31, 31) to (96, 96, 96)
- in CartSideDoubleSpecializedLinearRefine)  `fdata_box` (not sure if this is in cells or sides since `fdata` is `SideData`) has index bounds (32, 32, 32) to (95, 95, 95)
- in CartSideDoubleSpecializedLinearRefine) `fdata_gcw` is 1


So, given that we're trying to assign a nonzero value out-of-bounds something is amiss. We do need to fill ghost data here so we can calculate the vorticity, so it makes sense that we need to do that but I'm a bit lost in the indices.

boyceg:
Can you point me to a minimal reproducer?

drwells:
The `navier_stokes_01_3d` test. Compile with gfortran and `-DCMAKE_Fortran_FLAGS="-g -O0 -fcheck=bits,bounds,do -finit-real=snan`.

boyceg:
Sorry in the delay in following up on this.

Indexing with SAMRAI data structures can be tricky.

The C++ wrappers pass the cell-centered `fine_box.lower()` and `fine_box.upper()` directly to these Fortran routines; they do not pass separately converted side boxes. Consequently, each component's loop must extend by one only in that component's normal direction.

Some suggestions:

1. In `ibtk/src/refine_ops/fortran/cart_side_refine2d.f.m4`, revert all three changes in the PR.

2. In `ibtk/src/refine_ops/fortran/cart_side_refine3d.f.m4`, revert the first four changes:

   - Restore `iupper0+1` in the RT0 `u0` loop.
   - Restore `iupper1+1` in the RT0 `u1` loop.
   - Restore `iupper2+1` in the RT0 `u2` loop.
   - Restore `iupper0+1` in the specialized-linear `u0` loop.

3. Keep the PR's removal of `iupper0+1` from the specialized-linear `u1` loop, but also add `+1` to its `i1` loop:

   ```fortran
   do i2=ilower2,iupper2
      coarsen_index(i2,i_c2,i_f2,ratio(2))
      do i1=ilower1,iupper1+1
         coarsen_index(i1,i_c1,i_f1,ratio(1))
         do i0=ilower0,iupper0
   ```

4. Keep the PR's removal of `iupper0+1` from the specialized-linear `u2` loop, but also add `+1` to its `i2` loop:

   ```fortran
   do i2=ilower2,iupper2+1
      coarsen_index(i2,i_c2,i_f2,ratio(2))
      do i1=ilower1,iupper1
         coarsen_index(i1,i_c1,i_f1,ratio(1))
         do i0=ilower0,iupper0
   ```

After these changes, every component must use this domain:

- `u0`: `i0` through `iupper0+1`; `i1` and `i2` through their cell upper bounds
- `u1`: `i1` through `iupper1+1`; `i0` and `i2` through their cell upper bounds
- `u2`: `i2` through `iupper2+1`; `i0` and `i1` through their cell upper bounds

### Test changes

1. Revert all four expected-output changes currently included in the PR:

   - `tests/refine/rt0_refine_01_2d.2.4.output`
   - `tests/refine/rt0_refine_01_3d.2.output`
   - `tests/refine/rt0_refine_01_3d.4.output`
   - `tests/refine/rt0_refine_01_3d.2.4.output`

   RT0 behavior should not change under the corrected patch. If these outputs still change, that indicates another error.

2. Strengthen `tests/refine/rt0_refine_01.cpp` so omitted destination entries cannot retain their exact initial values:

   - Initialize the coarse data normally.
   - Immediately before `refine()`, fill only the fine destination data with a large finite sentinel value.
   - Call `refine()`.
   - Compare every entry in each component's complete destination side box against the expected refined value. The sentinel ensures that any omitted entries produce an obvious failure.
   - Retain the existing exact-solution checks.

3. Add coverage for `CartSideDoubleSpecializedLinearRefine`, since the actual bug is in that operator rather than RT0.

   The test should:

   - Run in 3D.
   - Provide and initialize the one coarse ghost layer required by the limited tangential slopes.
   - Fill the fine destination with a large finite sentinel before refinement.
   - Apply the specialized-linear operator.
   - Compare the complete destination side box for each component against the expected values.
   - Explicitly verify the planes `i0 = iupper0+1` for `u0`, `i1 = iupper1+1` for `u1`, and `i2 = iupper2+1` for `u2`.
   - Use at least a constant field and a globally linear field so the expected refined values are unambiguous.

boyceg:
@drwells I think this is fixed by #1960.

drwells:
yup, thanks!

PR #1936, opened 2026-06-14
Fixing multipatch/level bug in IBHydrodynamicSurfaceForceEvaluator class

amneetb:
This came about from work related to acoustic streaming branch. I have modified the `tests/multiphase_flow/check_hydro_force.mpirun=8.input` file so that it creates a body that spans multiple patches and uses AMR. The test still passes with the old output (which considered the body over 1 or 2 patches). 

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
+++ b/src/IB/IBHydrodynamicSurfaceForceEvaluator.cpp
@@ -224,6 +224,11 @@ IBHydrodynamicSurfaceForceEvaluator::computeHydrodynamicForceTorque(IBTK::Vector
                    << " Forces are evalauted at only current or new time. \n");
     }
 
+    // Get the side weight patch data index that contains the correct volume element for each side-centered degree of
+    // freedom. This is used to compute the area element for each side-centered DoF.
+    Pointer<HierarchyMathOps> hierarchy_math_ops = d_fluid_solver->getHierarchyMathOps();
+    const int wgt_sc_idx = hierarchy_math_ops->getSideWeightPatchDescriptorIndex();
+
     // Allocate required patch data
     Pointer<PatchHierarchy<NDIM>> patch_hierarchy = d_fluid_solver->getPatchHierarchy();
     const int coarsest_ln = 0;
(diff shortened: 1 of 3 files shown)


amneetb:
Some issue with the MacOS CI; also does not seem like `multiphase_flow/check_hydro_force` was included in the test suite on different CIs. It passes locally on my computer.

boyceg:
Can you accomplish the same thing more by using the side-centered control volume values provided by `HierarchyMathOps` to determine the relevant surface area? That should handle patch boundaries correctly and would not require additional storage and communication.

amneetb:
Hmm, didn't think about that. Is the side-centered volume defined only for master side DOFs?

boyceg:
I cannot remember, but I think it is split across the boundary. I think that should do what you want in either case.

I think that you also could just default to using the patch box (not side box) for this computation except for patches that touch an 'upper' domain boundary — in those cases you would extend by one to include the domain boundary DOFs.

I am not 100% sure, but I don't think you should need to use the side data synchronization here.

amneetb:
The surface is approximated as stair steps, which are taken to be cell faces. 

At a patch boundary, there is double counting of the same face.

If the side volume is halved at the patch boundary, then it can work, as it will also half the face area.

boyceg:
In that case, I think that the side-centered control volume will do what you want.

amneetb:
Ok, I'll try that then as it removes additional data structures that I wanted to avoid in the first place.

tests/multiphase_flow/check_hydro_force.mpirun=8.input
```diff
@@ -485,7 +485,7 @@ StandardTagAndInitialize {
    tagging_method = "REFINE_BOXES"
    // tagging_method = "GRADIENT_DETECTOR"
    RefineBoxes {
-      level_0 = [( 0 , 0 ),( Nx - 1 , Ny - 1 )]
+      level_0 = [( Nx/4 , Ny/4 ),( 3*Nx/4 , 3*Ny/4 )]
       level_1 = [( 0 , 0 ),( 2*Nx - 1 , 2*Ny - 1 )]
```
drwells, line 489: Instead of modifying an existing test, would you please add a new test (which runs on just two processors so it is triggered in the CI) with this configuration, if that triggers the bug? Thanks!

amneetb:
@drwells and @boyceg : Made the changes.

drwells (approved):
Thanks!

PR #1937, opened 2026-06-16
Fix IB Redistribution

abarret:
This fixes a bug that can occur during redistributions using `IBMethod`. If local `LNode` data is changed prior to regridding, stale ghost `LNode` data can potentially become local. This updates ghost `LNode` data prior to redistributing Lagrangian data.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/ibtk/include/ibtk/LDataManager.h
+++ b/ibtk/include/ibtk/LDataManager.h
@@ -850,6 +850,16 @@ public:
     void scatterToZero(Vec& parallel_vec, Vec& sequential_vec) const;
 
     /*!
+     * \brief Synchronize node-attached payloads for the ghost nodes that are
+     * already present in the current Lagrangian mesh.
+     *
+     * This routine updates the node data stored on existing ghost LNodes
+     * without changing ghost topology or PETSc indexing.
+     */
+    void synchronizeExistingGhostNodes(int coarsest_ln = invalid_level_number,
+                                       int finest_ln = invalid_level_number);
+
+    /*!
      * \brief Start the process of redistributing the Lagrangian data.
      *
      * This method uses the present location of each Lagrangian mesh node to
--- a/ibtk/include/ibtk/LDataManager.h
+++ b/ibtk/include/ibtk/LDataManager.h
@@ -856,8 +856,7 @@ public:
      * This routine updates the node data stored on existing ghost LNodes
      * without changing ghost topology or PETSc indexing.
      */
-    void synchronizeExistingGhostNodes(int coarsest_ln = invalid_level_number,
-                                       int finest_ln = invalid_level_number);
+    void synchronizeGhostLNodes(int coarsest_ln = invalid_level_number, int finest_ln = invalid_level_number);
 
     /*!
      * \brief Start the process of redistributing the Lagrangian data.
--- a/ibtk/src/lagrangian/LDataManager.cpp
+++ b/ibtk/src/lagrangian/LDataManager.cpp
@@ -1346,6 +1346,81 @@ LDataManager::scatterToZero(Vec& parallel_vec, Vec& sequential_vec) const
 } // scatterToZero
 
 void
+LDataManager::synchronizeExistingGhostNodes(const int coarsest_ln_in, const int finest_ln_in)
+{
+    const int coarsest_ln = (coarsest_ln_in == invalid_level_number) ? d_coarsest_ln : coarsest_ln_in;
+    const int finest_ln = (finest_ln_in == invalid_level_number) ? d_finest_ln : finest_ln_in;
+
+#if !defined(NDEBUG)
+    TBOX_ASSERT(coarsest_ln >= d_coarsest_ln && coarsest_ln <= d_finest_ln);
+    TBOX_ASSERT(finest_ln >= d_coarsest_ln && finest_ln <= d_finest_ln);
+#endif
+
+    const int num_procs = IBTK_MPI::getNodes();
+    for (int level_number = coarsest_ln; level_number <= finest_ln; ++level_number)
+    {
+        if (!d_level_contains_lag_data[level_number]) continue;
+
+        const Pointer<LMesh> mesh = getLMesh(level_number);
+        if (!mesh) continue;
+
+        using LNodeTransactionComponent = LNodeTransaction::LTransactionComponent;
+        const std::vector<LNode*>& local_nodes = mesh->getLocalNodes();
+        const std::vector<LNode*>& ghost_nodes = mesh->getGhostNodes();
+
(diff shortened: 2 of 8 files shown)


ibtk/include/ibtk/LDataManager.h
```diff
@@ -849,6 +849,16 @@ class LDataManager : public SAMRAI::mesh::StandardTagAndInitStrategy<NDIM>, publ
...
     void scatterToZero(Vec& parallel_vec, Vec& sequential_vec) const;
 
+    /*!
+     * \brief Synchronize node-attached payloads for the ghost nodes that are
+     * already present in the current Lagrangian mesh.
+     *
+     * This routine updates the node data stored on existing ghost LNodes
+     * without changing ghost topology or PETSc indexing.
+     */
+    void synchronizeExistingGhostNodes(int coarsest_ln = invalid_level_number,
```
boyceg:
```suggestion
    void synchronizeGhostLNodes(int coarsest_ln = invalid_level_number,
```
?

ibtk/src/lagrangian/LDataManager.cpp
```diff
@@ -1345,6 +1345,148 @@ LDataManager::scatterToZero(Vec& parallel_vec, Vec& sequential_vec) const
...
+        if (!mesh) continue;
+
+        using LNodeTransactionComponent = LNodeTransaction::LTransactionComponent;
+        const std::vector<LNode*>& local_nodes = mesh->getLocalNodes();
+        const std::vector<LNode*>& ghost_nodes = mesh->getGhostNodes();
+
+        std::vector<int> num_local_nodes_proc(num_procs, 0);
+        IBTK_MPI::allGather(static_cast<int>(local_nodes.size()), &num_local_nodes_proc[0]);
+        std::vector<int> node_offsets(num_procs + 1, 0);
+        std::partial_sum(num_local_nodes_proc.begin(), num_local_nodes_proc.end(), node_offsets.begin() + 1);
```
boyceg: We perform this kind of operation a lot, but probably should implement it 1) in one place and 2) more efficiently without calling `IBTK_MPI::allGather`.

ibtk/src/lagrangian/LDataManager.cpp
```diff
@@ -1345,6 +1345,148 @@ LDataManager::scatterToZero(Vec& parallel_vec, Vec& sequential_vec) const
...
+
+        const int request_data_size = static_cast<int>(request_data.size());
+        const int all_request_data_size = IBTK_MPI::sumReduction(request_data_size);
+        std::vector<int> all_request_data(all_request_data_size);
+        // Share only request data globally; LNode payloads are sent later only by the ranks that own the requested
+        // ghost nodes.
+        IBTK_MPI::allGather(request_data_size > 0 ? &request_data[0] : nullptr,
+                            request_data_size,
+                            all_request_data_size > 0 ? &all_request_data[0] : nullptr,
+                            all_request_data_size);
```
boyceg: Can we reasonably avoid `IBTK_MPI::allGather` here?

boyceg (changes requested):
Looks good. Some minor comments. I think we should be able to avoid using `allGather`, but I am not 100% sure. It might not be worth it at this point — I suspect we have bigger fish to fry right now in the parallelization front.

ibtk/src/lagrangian/LDataManager.cpp
abarret: We can by doing a direct point-to-point request so that only owner ranks are informed of required ghost nodes. Gather calls make my brain hurt... but it shouldn't be that hard.

ibtk/src/lagrangian/LDataManager.cpp
abarret: Actually I think we already have that data as a member variable. I'll fix that.

boyceg: I am more guilty than most, but we probably should have code to compute this in exactly one place, and I am 99% sure it can be implemented without using all-to-all communication.

abarret: We now look this up using `VecGetOwnershipRanges`, which is probably the best way to do it. I'm assuming PETSc will cache these things for us. This is used a few places in `LDataManager` already, although we should probably search for MPI Gather calls in other places.

ibtk/src/lagrangian/LDataManager.cpp
abarret: This now uses two `Alltoall` calls instead of the `allGather`.

boyceg:
@abarret is this ready to go?

abarret:
> @abarret is this ready to go?

Yes this is finished.

ibtk/src/lagrangian/LDataManager.cpp
```diff
@@ -1345,6 +1345,171 @@ LDataManager::scatterToZero(Vec& parallel_vec, Vec& sequential_vec) const
...
+            requested_node_data[dst_rank].emplace_back(Pointer<LNode>(local_nodes[local_idx], false), Point::Zero());
+        }
+
+        Schedule lnode_data_mover;
+        std::vector<std::vector<Pointer<Transaction>>> transactions(num_procs,
+                                                                    std::vector<Pointer<Transaction>>(num_procs));
+        for (int src_proc = 0; src_proc < num_procs; ++src_proc)
+        {
+            for (int dst_proc = 0; dst_proc < num_procs; ++dst_proc)
+            {
```
boyceg, line 1527: Ultimately this seems like it is going to kill scaling — although I guess the all-to-all above would as well. Can we avoid doing `(num_procs)^2` work here?

The prudent thing probably is to kick the can down the road...

abarret: Most of these transactions are no ops. Do you know how `communicate()` handles no op transactions?

For the all-to-all, some number of the all-to-all sends and recv's will be empty. I think we have the information to break it down to direct send and recv calls. Is the overhead of an empty synch worth eliminating?

boyceg: I'm not sure, but even if you just blink in the inner loop, you are doing to be doing `O(num_proc^2)` work.

abarret: I've replaced that double loop with one that only appends transactions that have non-zero communications. I've also replaced the second all-to-all with non-blocking direct send and recv calls. There's still an all-to-all with send counts that can't reasonably be replaced.

boyceg (approved):

boyceg:
@abarret shall we merge this since it is fixing an actual `IBMethod` bug?

PR #1939, opened 2026-06-20
adds VelocityCartGridFunction to save and load velocity field for tra…

tmweigand:
@boyceg Here is some code to allow for saving a velocity field to a binary file from an IBAMR simulation and then loading the velocity field for a transport solve in a different simulation. There is a lot that should be added here including check to ensure the meshes are identical, support for interpolation between time-steps, etc. But this is just to get it started. 

I added adv_diff/ex9 (ased of of adv_diff/ex_2) just to ensure this works. I am only grabbing the velocity field at time=0 in my example. 

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/examples/adv_diff/ex9/example.cpp
+++ b/examples/adv_diff/ex9/example.cpp
@@ -152,7 +152,7 @@ main(int argc, char* argv[])
 
         // Tim
         Pointer<VelocityCartGridFunction> velocity_function =
-            new VelocityCartGridFunction("velocity", grid_geometry, -1);
+            new VelocityCartGridFunction("velocity", grid_geometry, -1, "saveVelocityDB");
 
         Pointer<FaceVariable<NDIM, double>> u_adv_var = new FaceVariable<NDIM, double>("u_adv");
         time_integrator->registerAdvectionVelocity(u_adv_var);
@@ -197,14 +197,14 @@ main(int argc, char* argv[])
         double loop_time = time_integrator->getIntegratorTime();
         const double dt = time_integrator->getMaximumTimeStepSize(); // fix time step
         double loop_time_end = time_integrator->getEndTime();
+
         while (!IBTK::rel_equal_eps(loop_time, loop_time_end))
         {
             u_init->setDataOnPatchHierarchy(u_side_index, u_side_adv_var, patch_hierarchy, loop_time);
-            velocity_function->save_velocity_data(patch_hierarchy, save_dir, loop_time, iteration_num);
+            velocity_function->saveVelocity(patch_hierarchy, iteration_num, loop_time);
             loop_time += dt;
             iteration_num += 1;
         }
-
         // Deallocate initialization objects.
         app_initializer.setNull();
 
@@ -233,8 +233,8 @@ main(int argc, char* argv[])
             pout << "At beginning of timestep # " << iteration_num << "\n";
             pout << "Simulation time is " << loop_time << "\n";
 
-            auto vel_time = velocity_function->read_velocity_data(patch_hierarchy, save_dir, iteration_num);
-            TBOX_ASSERT(loop_time == vel_time);
+            auto vel_time_db = velocity_function->readVelocity(patch_hierarchy, iteration_num);
+            TBOX_ASSERT(loop_time == vel_time_db);
 
             time_integrator->advanceHierarchy(dt);
             loop_time += dt;
--- a/examples/adv_diff/ex9/example.cpp
+++ b/examples/adv_diff/ex9/example.cpp
@@ -1,0 +1,311 @@
+// ---------------------------------------------------------------------
+//
+// Copyright (c) 2017 - 2026 by the IBAMR developers
+// All rights reserved.
+//
+// This file is part of IBAMR.
+//
+// IBAMR is free software and is distributed under the 3-clause BSD
+// license. The full text of the license can be found in the file
+// COPYRIGHT at the top level directory of IBAMR.
+//
+// ---------------------------------------------------------------------
+
+// Config files
+#include <SAMRAI_config.h>
+
+// Headers for basic PETSc functions
+#include <petscsys.h>
+
+// Headers for basic SAMRAI objects
+#include <BergerRigoutsos.h>
+#include <CartesianGridGeometry.h>
+#include <LoadBalancer.h>
+#include <StandardTagAndInitialize.h>
+
+// Headers for application-specific algorithm/data structure objects
+#include <ibamr/AdvDiffPredictorCorrectorHierarchyIntegrator.h>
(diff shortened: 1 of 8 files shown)


drwells:
Sorry for the long delay in reviewing this. Why not use `Patch::putToDatabase()` and `HDFDatabase`? We already have a complete serialization mechanism for handling this kind of situation.

tmweigand:
Hi @drwells. The main reason was ignorance. However, the use case for this is also a bit odd. We want to extract the velocity field at some point and then use that for a transport solve. I found that the `restartManager` is unhappy when trying to add transport and stopping the NS solve to an existing simulation so I was skirting that issue. 

Is there a better route? Perhaps creating a new database and reading/writing to that.

drwells:
If you use that function (`Patch::putToDatabase()`) then you can put information in an arbitrary `Database` object: it doesn't need to be the restart database. The same goes for reading from `Patch::getFromDatabase()`.

tmweigand:
Thanks @drwells. I am now using the `HDFDatabase` to store the velocity field and using the `patch` functionality.

drwells:
I didn't notice it before but since there is also `PatchHierarchy::putToDatabase()` and `PatchHierarchy::getFromDatabase()`. I think you can serialize what you want, outside of the restart database, with two database setups and two function calls.

src/utilities/VelocityCartGridFunction.cpp
```diff
@@ -0,0 +1,270 @@
...
+                                       int iteration_num,
+                                       double loop_time)
+{
+    // create <base_name>/<iteration_num>/
+    ensureDirectoryExists(getIterationDir(iteration_num));
+
+    const std::string filename = getIterationFilename(iteration_num);
+
+    Pointer<HDFDatabase> hdf_db = new HDFDatabase(d_object_name + "::hdf_db");
+    hdf_db->create(filename);
```
boyceg, line 177: Could we make the output rank-specific, or otherwise use a distributed hierarchy serialization mechanism? `PatchLevel::Iterator` loops over locally owned patches, but every MPI rank currently calls `HDFDatabase::create()` with the same filename. In a multi-rank run, the ranks can truncate or overwrite the same file, leaving an incomplete or corrupted database. The directory creation also appears to have a check-then-create race. Existing hierarchy-output routines in IBAMR include the MPI rank in each filename.

src/utilities/VelocityCartGridFunction.cpp
```diff
@@ -0,0 +1,270 @@
...
+                                                  const double data_time,
+                                                  const bool /*initial_time*/,
+                                                  int coarsest_ln,
+                                                  int finest_ln)
+{
+    copy_side_to_face(data_idx, d_velocity_index, hierarchy);
+}
+
+void
+VelocityCartGridFunction::setDataOnPatch(const int /*data_idx*/,
```
boyceg, line 157: This class implements `setDataOnPatchHierarchy()`, but `setDataOnPatch()` is empty. The Godunov predictor-corrector path calls the inherited `setDataOnPatchLevel()`, which dispatches to `setDataOnPatch()`, so selecting `solver_type = "GODUNOV"` appears to leave the requested face-centered velocity unset. Could `setDataOnPatch()` be implemented, or could `setDataOnPatchLevel()` be overridden, so both supported integrators receive the saved velocity?

src/utilities/VelocityCartGridFunction.cpp
```diff
@@ -0,0 +1,270 @@
...
+    : CartGridFunction(object_name),
+      d_grid_geometry(grid_geometry),
+      d_velocity_index(velocity_index),
+      d_db_base_name(db_base_name)
+{
+    ensureDirectoryExists(d_db_base_name);
+}
+
+bool
+VelocityCartGridFunction::isTimeDependent() const
```
boyceg, line 139: Could we clarify and implement the intended time-selection behavior here? The object reports that it is time-dependent, but `setDataOnPatchHierarchy()` ignores `data_time`. In the semi-implicit integrator, the function is evaluated at both the current and new times; because the example loads only the current snapshot before advancing, both evaluations receive that same snapshot. Even if interpolation is deferred, exact saved times should select their corresponding snapshots—or, if the intended use is a single frozen field, the object should probably not report itself as time-dependent.

src/utilities/VelocityCartGridFunction.cpp
```diff
@@ -0,0 +1,270 @@
...
+    const int finest_ln = patch_hierarchy->getFinestLevelNumber();
+
+    for (int ln = 0; ln <= finest_ln; ++ln)
+    {
+        Pointer<PatchLevel<NDIM>> level = patch_hierarchy->getPatchLevel(ln);
+        for (PatchLevel<NDIM>::Iterator p(*level); p; p++)
+        {
+            Pointer<Patch<NDIM>> patch = level->getPatch(p());
+
+            const std::string key = "patch_data_" + std::to_string(ln) + "_" + std::to_string(p());
```
boyceg, line 190: The database keys identify patches only by level number and iterator index. That appears to assume identical patch numbering, processor ownership, and MPI size in the producing and consuming simulations. Could we either serialize enough hierarchy and box metadata to match data by patch geometry and diagnose mismatches, or explicitly enforce and document the stronger identical-decomposition requirement? This seems especially important for the stated use in a different simulation.

examples/adv_diff/ex9/example.cpp
```diff
@@ -0,0 +1,311 @@
...
+        {
+            iteration_num = time_integrator->getIntegratorStep();
+            loop_time = time_integrator->getIntegratorTime();
+
+            pout << "\n";
+            pout << "+++++++++++++++++++++++++++++++++++++++++++++++++++\n";
+            pout << "At beginning of timestep # " << iteration_num << "\n";
+            pout << "Simulation time is " << loop_time << "\n";
+
+            auto vel_time_db = velocity_function->readVelocity(patch_hierarchy, iteration_num);
```
boyceg, line 236: Could we add a narrow save/reload test that directly verifies the restored velocity? For example, fill the side data with known values, save it, overwrite or clear it, reload it, and compare the restored values on every patch. Running that test with more than one MPI rank would also exercise the distributed-file behavior. The present example demonstrates the workflow, but its final transport error does not isolate whether the serialized velocity itself was restored correctly.

boyceg (commented):
Sorry for the slow reply. This looks like a useful capability! Before we can merge, however, I think we need to resolve the parallel-file behavior, ensure that both supported advection-diffusion integrators receive the saved velocity correctly, and define how requested times select saved snapshots. A direct save/reload test would make those requirements much easier to verify.

boyceg (changes requested):

Issue #1938, opened 2026-06-17
ibsamrai2 needs some more patches to make things work on macs


drwells:
If you look at any of the recent CI runs, the mac tester fails because SAMRAI has code like `CellVariable<DIM>(NULL, NULL)`. This was undefined behavior in any version of C++ and was banned in C++23, see

https://open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2166r1.html

Would someone with a mac put together an IBSAMRAI2 patch for this so we can update the CI? Without patching this we can't compile on very new mac toolchains. Thanks!

PR #1940, opened 2026-07-16
Get rid of snprintf() et al 

### LINKED ISSUE #1602: Get rid of snprintf() et al
We should just use `std::ostringstream` everywhere.

jason-sli:
Replaced all instances of snprintf() and sprintf() with modern C++ alternatives. As suggested in the issue, most replacements use std::ostringstream for string formatting and construction. Two utility functions were created that encode the standard file-naming scheme. Additionally, a test was created that makes sure the file names produced are correct.
Fixes #1602 

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?

### DIFF
--- a/examples/ConstraintIB/eel2d/example.cpp
+++ b/examples/ConstraintIB/eel2d/example.cpp
@@ -433,9 +433,7 @@ output_data(Pointer<PatchHierarchy<NDIM>> patch_hierarchy,
 
     // Write Cartesian data.
     string file_name = data_dump_dirname + "/" + "hier_data.";
-    char temp_buf[128];
-    std::snprintf(temp_buf, sizeof(temp_buf), "%05d.samrai.%05d", iteration_num, IBTK_MPI::getRank());
-    file_name += temp_buf;
+    file_name += generateSAMRAIName(iteration_num);
     Pointer<HDFDatabase> hier_db = new HDFDatabase("hier_db");
     hier_db->create(file_name);
     VariableDatabase<NDIM>* var_db = VariableDatabase<NDIM>::getDatabase();
@@ -457,8 +455,7 @@ output_data(Pointer<PatchHierarchy<NDIM>> patch_hierarchy,
     VecDuplicate(X_petsc_vec, &X_lag_vec);
     l_data_manager->scatterPETScToLagrangian(X_petsc_vec, X_lag_vec, finest_hier_level);
     file_name = data_dump_dirname + "/" + "X.";
-    std::snprintf(temp_buf, sizeof(temp_buf), "%05d", iteration_num);
-    file_name += temp_buf;
+    file_name += generateIterationName(iteration_num);
     PetscViewer viewer;
     PetscViewerASCIIOpen(PETSC_COMM_WORLD, file_name.c_str(), &viewer);
     VecView(X_lag_vec, viewer);
--- a/examples/complex_fluids/ex0/example.cpp
+++ b/examples/complex_fluids/ex0/example.cpp
@@ -386,9 +386,7 @@ output_data(Pointer<PatchHierarchy<NDIM>> patch_hierarchy,
     plog << "writing hierarchy data at iteration " << iteration_num << " to disk" << endl;
     plog << "simulation time is " << loop_time << endl;
     string file_name = data_dump_dirname + "/" + "hier_data.";
-    char temp_buf[128];
-    std::snprintf(temp_buf, sizeof(temp_buf), "%05d.samrai.%05d", iteration_num, IBTK_MPI::getRank());
-    file_name += temp_buf;
+    file_name += generateSAMRAIName(iteration_num);
     Pointer<HDFDatabase> hier_db = new HDFDatabase("hier_db");
     hier_db->create(file_name);
     VariableDatabase<NDIM>* var_db = VariableDatabase<NDIM>::getDatabase();
--- a/ibtk/contrib/boost/boost/lexical_cast/detail/converter_lexical_streams.hpp
+++ b/ibtk/contrib/boost/boost/lexical_cast/detail/converter_lexical_streams.hpp
@@ -32,6 +32,8 @@
 #include <string>
 #include <cstring>
 #include <cstdio>
+#include <iomanip>
+#include <algorithm>
 #include <boost/limits.hpp>
 #include <boost/type_traits/conditional.hpp>
 #include <boost/type_traits/is_pointer.hpp>
@@ -278,38 +280,68 @@ namespace boost {
             bool shl_real_type(float val, char* begin) {
                 using namespace std;
                 const double val_as_double = val;
-                finish = start +
-#if defined(_MSC_VER) && (_MSC_VER >= 1400) && !defined(__SGI_STL_PORT) && !defined(_STLPORT_VERSION)
-                    sprintf_s(begin, CharacterBufferSize,
-#else
-                    sprintf(begin,
-#endif
-                    "%.*g", static_cast<int>(boost::detail::lcast_get_precision<float>()), val_as_double);
+                
+                ostringstream oss;
+                oss << setprecision(static_cast<int>(boost::detail::lcast_get_precision<float>()))
+                    << val_as_double;
+
+                if (oss.fail())
+                    return false;
+
+                string str = oss.str();
+
+                if (str.size() >= CharacterBufferSize)
+                    return false;
+
+                copy(str.begin(), str.end(), begin);
+                begin[str.size()] = '\0';
+
+                finish = start + str.size();
                 return finish > start;
             }
 
             bool shl_real_type(double val, char* begin) {
                 using namespace std;
-                finish = start +
-#if defined(_MSC_VER) && (_MSC_VER >= 1400) && !defined(__SGI_STL_PORT) && !defined(_STLPORT_VERSION)
-                    sprintf_s(begin, CharacterBufferSize,
-#else
-                    sprintf(begin,
-#endif
-                    "%.*g", static_cast<int>(boost::detail::lcast_get_precision<double>()), val);
+
+                ostringstream oss;
+                oss << setprecision(static_cast<int>(boost::detail::lcast_get_precision<double>()))
+                    << val;
+
+                if (oss.fail())
+                    return false;
+
+                string str = oss.str();
+
+                if (str.size() >= CharacterBufferSize)
+                    return false;
+
+                copy(str.begin(), str.end(), begin);
+                begin[str.size()] = '\0';
+
+                finish = start + str.size();
                 return finish > start;
             }
 
 #ifndef __MINGW32__
             bool shl_real_type(long double val, char* begin) {
                 using namespace std;
-                finish = start +
-#if defined(_MSC_VER) && (_MSC_VER >= 1400) && !defined(__SGI_STL_PORT) && !defined(_STLPORT_VERSION)
-                    sprintf_s(begin, CharacterBufferSize,
-#else
-                    sprintf(begin,
-#endif
-                    "%.*Lg", static_cast<int>(boost::detail::lcast_get_precision<long double>()), val );
+
+                ostringstream oss;
+                oss << setprecision(static_cast<int>(boost::detail::lcast_get_precision<long double>()))
+                    << val;
+
+                if (oss.fail())
+                    return false;
+
+                string str = oss.str();
+
+                if (str.size() >= CharacterBufferSize)
+                    return false;
+
+                copy(str.begin(), str.end(), begin);
+                begin[str.size()] = '\0';
+
+                finish = start + str.size();
                 return finish > start;
             }
 #endif
--- a/ibtk/contrib/muparser/src/muParserDLL.cpp
+++ b/ibtk/contrib/muparser/src/muParserDLL.cpp
@@ -208,7 +208,10 @@ API_EXPORT(const muChar_t*) mupGetVersion(muParserHandle_t a_hParser)
 		muParser_t* const p(AsParser(a_hParser));
 
 #ifndef _UNICODE
-		snprintf(s_tmpOutBuf, count_of(s_tmpOutBuf), "%s", p->GetVersion().c_str());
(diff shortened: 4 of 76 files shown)


tests/IBFE/explicit_ex2.cpp
```diff
@@ -492,9 +492,10 @@ output_data(Pointer<PatchHierarchy<NDIM>> patch_hierarchy,
 
     // Write Cartesian data.
     string file_name = data_dump_dirname + "/" + "hier_data.";
-    char temp_buf[128];
-    std::snprintf(temp_buf, sizeof(temp_buf), "%05d.samrai.%05d", iteration_num, IBTK_MPI::getRank());
-    file_name += temp_buf;
+    std::ostringstream oss;
+    oss << std::setw(5) << std::setfill('0') << iteration_num << ".samrai." << std::setw(5) << std::setfill('0')
+        << IBTK_MPI::getRank();
+    file_name += oss.str();
```
drwells: Nearly all of these are the same, and they encode SAMRAI's default naming scheme. How about we add a utility function which does this instead of copying and pasting this bit of code?

drwells (changes requested):
Thanks for working on this. Can you take a look at the error messages in the CI? There's something here which doesn't compile.

tests/IBFE/explicit_ex2.cpp
jason-sli: I'll work on putting it into a function. I also see why the Ubuntu 20.04 check is failing and know how to fix it. However, I don't know why the macOS check is failing. The error that it is showing me is not in a file I think I edited. Any ideas why that's going wrong?

I also ran the same CI checks on a branch in my fork where the only change was adding a blank line, by opening a pull request against my fork's master. The macOS check failed there as well, so I'm wondering if that failure is unrelated to my changes.

drwells: The macOS failure is due to something else - just ignore it. See also #1938.

jason-sli: I removed the repeated code and replaced it with calls to two utility functions that I created. The compilation error should also be resolved.

jason-sli: You may have missed the updates I made to my code since I did it all over the weekend. It still shows up as 2 commits because I erased the previous history before changing the code. It's ready for review whenever you have time.

ibtk/contrib/boost/boost/lexical_cast/detail/converter_lexical_streams.hpp
```diff
@@ -278,38 +280,68 @@ namespace boost {
...
+
+                string str = oss.str();
+
+                if (str.size() >= CharacterBufferSize)
+                    return false;
+
+                copy(str.begin(), str.end(), begin);
+                begin[str.size()] = '\0';
+
+                finish = start + str.size();
```
drwells: I appreciate the proactivity but we should not change our dependencies.

ibtk/contrib/muparser/src/muParserDLL.cpp
```diff
@@ -208,7 +208,10 @@ API_EXPORT(const muChar_t*) mupGetVersion(muParserHandle_t a_hParser)
 		muParser_t* const p(AsParser(a_hParser));
 
 #ifndef _UNICODE
-		snprintf(s_tmpOutBuf, count_of(s_tmpOutBuf), "%s", p->GetVersion().c_str());
+		std::string version = p->GetVersion();
+		std::size_t versionSize = std::min(version.size(), count_of(s_tmpOutBuf) - 1);
+		std::copy_n(version.begin(), versionSize, s_tmpOutBuf);
+		s_tmpOutBuf[versionSize] = '\0';
```
drwells, line 214: Same thing here.

examples/complex_fluids/ex0/example.cpp
```diff
@@ -386,9 +386,7 @@ output_data(Pointer<PatchHierarchy<NDIM>> patch_hierarchy,
     plog << "writing hierarchy data at iteration " << iteration_num << " to disk" << endl;
     plog << "simulation time is " << loop_time << endl;
     string file_name = data_dump_dirname + "/" + "hier_data.";
-    char temp_buf[128];
-    std::snprintf(temp_buf, sizeof(temp_buf), "%05d.samrai.%05d", iteration_num, IBTK_MPI::getRank());
-    file_name += temp_buf;
+    file_name += generateSAMRAIName(iteration_num);
```
drwells: This comment applies to all of these changes but I'll leave this one here. Yes, we are removing `snprintf()` and that is good, but the point of the existing code was to get the name of the already existing output directory. As you can see, these are always the same (we start with the known data directory and then add `hier_data.`, the iteration number, and the rank). 

Since they are always the same we can do

```cpp
hier_db->create(format_hierarchy_data_output_filename(iteration_num));
```
or some other equivalently named function which does what we need.

ibtk/include/ibtk/ibtk_utilities.h
```diff
@@ -408,6 +413,28 @@ checked_dereference(SAMRAI::tbox::Pointer<T>& p)
     return *p;
 }
 
+/*!
+ * Generate a name for SAMRAI data files.
```
drwells: The other half of it is here: we save a lot of SAMRAI files. Let's be specific and describe what this does and then describe why in the next sentence.

examples/ConstraintIB/eel2d/example.cpp
```diff
@@ -457,8 +455,7 @@ output_data(Pointer<PatchHierarchy<NDIM>> patch_hierarchy,
     VecDuplicate(X_petsc_vec, &X_lag_vec);
     l_data_manager->scatterPETScToLagrangian(X_petsc_vec, X_lag_vec, finest_hier_level);
     file_name = data_dump_dirname + "/" + "X.";
-    std::snprintf(temp_buf, sizeof(temp_buf), "%05d", iteration_num);
-    file_name += temp_buf;
+    file_name += generateIterationName(iteration_num);
```
drwells: We can do the same thing here: all of these call sites have the same formatting (though sometimes we use `"X"` and sometimes something else: that can be an input argument).

drwells (dismissed):
Thanks again for working on this! Let's define the two new functions first and then make all the changes second so we don't have to keep rewriting things. Sometimes this is called "make the change easy, and then make the easy change".

boyceg:
@jason-sli, this is a somewhat peculiar git-ism, but to synch with master, what you want to do is to rebase, not to merge. (@drwells do we have this clearly documented somewhere? If we don't, we should. I am going to try to create a "developer's guide" this fall in any case...)

boyceg:
(@drwells, can you help @jason-sli sort out the rebase - vs - merge workflow here?)

jason-sli:
@boyceg I think I changed it to rebase now. Does the history look correct to you?

boyceg:
Yes, I think that did it! You also can take a look at https://github.com/IBAMR/IBAMR/network and see if your branch is branches off of `master` at the position you think it should. (`git log` also will show you this information if you know where to look for it.)

@drwells can you take a look at @jason-sli's revisions?

boyceg:
Thanks for working through this large cleanup!

I have two changes I’d like you to make before we merge this:

1. Please revert the changes to `ibtk/contrib/muparser/src/muParserDLL.cpp`. That file belongs to a bundled third-party dependency, and we try to keep such dependencies unchanged. Reverting it will also resolve the whitespace errors currently reported by `git diff --check`.

2. Please update the two new helper functions to follow IBAMR’s naming and API conventions:

   - `formatSAMRAIOutputFilename()` → `format_samrai_output_filename()`
   - `formatIterationOutputFilename()` → `format_iteration_output_filename()`
   - Remove `const` from the pass-by-value `int iteration_num` parameters in the public signatures.

After updating the call sites, please run `make indent` and check that `git diff --check` produces no output.

A small test that checks the exact zero-padded filenames would also be useful, since many examples now depend on these helpers.

I think that once these changes are made, this will be ready to merge—or at least very close to it.

jason-sli:
@boyceg I changed the function names, reverted the changes to muParserDLL.cpp, and created a simple test. Let me know if I need to do anything else.

boyceg:
Thanks for updating this!

I think there are a few remaining issues to address:
- In `examples/IBFE/explicit/ex7/example.cpp`, the variable `file_name` needs to be declared as a `std::string` when assigning the mesh filename. It looks like your refactoring removed its earlier declaration, so this example fails to compile in this branch.
- This is not done completely uniformly in the library, but my preference is to use the "public" headers only for the programming interface specification, and to keep implementations elsewhere. If this isn't too annoying, could you move the inlined implementation to a new file, `ibtk/private/ibtk_utilities-inl.h`? I will give a draft of what this looks like in a follow-up message.
- Use `plog`, configured to write to `output` in the new test. I will include a suggestion on how to do this below as well.

boyceg:
Here is a draft for the new private header file (`ibtk/include/ibtk/private/ibtk_utilities-inl.h`):
```
//
// Copyright (c) 2026 by the IBAMR developers
// All rights reserved.
//
// This file is part of IBAMR.
//
// IBAMR is free software and is distributed under the 3-clause BSD
// license. The full text of the license can be found in the file
// COPYRIGHT at the top level directory of IBAMR.
//

/////////////////////////////// INCLUDE GUARD ////////////////////////////////

#ifndef included_IBTK_ibtk_utilities_inl_h
#define included_IBTK_ibtk_utilities_inl_h

/////////////////////////////// INCLUDES /////////////////////////////////////

#include <ibtk/config.h>

#include <ibtk/IBTK_MPI.h>
#include <ibtk/ibtk_utilities.h>

#include <iomanip>
#include <sstream>
#include <string>

/////////////////////////////// FUNCTION DEFINITIONS /////////////////////////

namespace IBTK
{
inline std::string
format_samrai_output_filename(const int iteration_num,
                             const std::string& data_dump_dirname,
                             const std::string& prefix)
{
    std::ostringstream oss;
    oss << data_dump_dirname << "/" << prefix << ".";
    oss << std::setw(5) << std::setfill('0') << iteration_num << ".samrai."
        << std::setw(5) << IBTK_MPI::getRank();
    return oss.str();
}

inline std::string
format_iteration_output_filename(const int iteration_num,
                                const std::string& data_dump_dirname,
                                const std::string& prefix)
{
    std::ostringstream oss;
    oss << data_dump_dirname << "/" << prefix << ".";
    oss << std::setw(5) << std::setfill('0') << iteration_num;
    return oss.str();
}
} // namespace IBTK

//////////////////////////////////////////////////////////////////////////////

#endif // #ifndef included_IBTK_ibtk_utilities_inl_h
```

To enable these to be used, in `ibtk_utilities.h`, keep the documentation and the function declarations, but remove the function implementations:

```
inline std::string
format_samrai_output_filename(int iteration_num,
                             const std::string& data_dump_dirname,
                             const std::string& prefix);

inline std::string
format_iteration_output_filename(int iteration_num,
                                const std::string& data_dump_dirname,
                                const std::string& prefix);
```

Finally, at the bottom of `ibtk_utilities.h`, after the closing `} // namespace IBTK` and before the final `#endif`, add:

```
/////////////////////////////// INLINE ///////////////////////////////////////

#include <ibtk/private/ibtk_utilities-inl.h> // IWYU pragma: keep
```

(If you grep for other instances of "private" or "-inl", you will see some other examples.)

boyceg:
For `tests/IBTK/filename_utilities.cpp`, replace `<fstream>` with `<tbox/PIO.h>`, replace `std::ofstream out("output");` with `SAMRAI::tbox::PIO::logOnlyNodeZero("output");`, and write through `SAMRAI::tbox::plog`.

Here is a draft:
```
#include <ibtk/IBTKInit.h>
#include <ibtk/ibtk_utilities.h>

#include <tbox/PIO.h>

#include <string>

int
main(int argc, char* argv[])
{
    IBTK::IBTKInit ibtk_init(argc, argv);

    SAMRAI::tbox::PIO::logOnlyNodeZero("output");

    const std::string iteration_filename =
        IBTK::format_iteration_output_filename(42, "output", "data");
    SAMRAI::tbox::plog << "Iteration filename:\n";
    SAMRAI::tbox::plog << iteration_filename << "\n\n";

    const std::string samrai_filename =
        IBTK::format_samrai_output_filename(42, "output", "data");
    SAMRAI::tbox::plog << "SAMRAI filename:\n";
    SAMRAI::tbox::plog << samrai_filename << "\n";

    return 0;
}
```
(These changes should allow the test to run correctly in both serial and parallel.)

drwells:
@boyceg can we skip those changes to the test? There's nothing wrong with creating a single `std::ofstream` for a serial test (we do it in other places too).

boyceg:
> @boyceg can we skip those changes to the test? There's nothing wrong with creating a single `std::ofstream` for a serial test (we do it in other places too).

Fine with me!

jason-sli:
I moved the filename functions to a separate file and also removed an unnecessary include from the test.

boyceg:
Thanks! This looks good to me. @drwells, is there anything else to fix before we merge?

boyceg:
@drwells, any comments before we merge?

boyceg (approved):

jason-sli:
@boyceg I see that you approved the changes, but below it still says that there are conflicts that must be resolved. Is this something I need to fix on my end and then repush?

boyceg:
Not on my end — but because @drwells requested changes he still needs to approve.

@drwells are there further changes needed here?

boyceg:
@jason-sli: @drwells gave me permission offline to dismiss his revision requests, so I have done that. It looks like there is now a conflict with `master` that needs to be resolved — can you rebase onto the updated `master` branch? After that is done, I think this should be ready to merge!

jason-sli:
@boyceg That should do it.

boyceg:
@jason-sli: If it shows up for you, do you want to push the merge button? 😁

jason-sli:
@boyceg Unfortunately, it says I am "not authorized to push to this branch". You do the honors.

boyceg:
Congratulations on your first merged PR, @jason-sli! 🎊

PR #1948, opened 2026-08-29
[CAV R00] Resolve PETSc options files portably

boyceg:
Read PETSc options from the IBAMR input database or an input-relative options file while preserving command-line precedence.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
This was just a test commit. Removing now.

ibtk/src/utilities/ibtk_utilities.cpp
```diff
@@ -13,20 +13,80 @@
...
+        options_key = "petsc_options_file";
+    }
+    else
+    {
+        return;
+    }
+
+    const std::string petsc_options_file = input_db->getString(options_key);
+    std::string resolved_options_file = petsc_options_file;
+    auto path_exists = [](const std::string& path) -> bool
```
drwells: Why not use `std::filesystem::exists()`? If you used an LLM it is mixing C++ version idioms here.

ibtk/src/utilities/ibtk_utilities.cpp
```diff
@@ -13,20 +13,80 @@
...
+    if (!path_exists(resolved_options_file) && argc > 1)
+    {
+        std::filesystem::path input_filename = argv[1];
+        std::error_code error_code;
+        input_filename = std::filesystem::weakly_canonical(input_filename, error_code);
+        if (error_code) input_filename = argv[1];
+        const std::filesystem::path input_dir = input_filename.parent_path();
+        if (!input_dir.empty())
+        {
+            const std::string candidate = (input_dir / petsc_options_file).string();
```
drwells: Could we either use all `std::filesystem::path`s or all `std::string`s in this function? Mixing and matching them is hard to follow.

ibtk/src/utilities/AppInitializer.cpp
```diff
@@ -84,12 +85,8 @@ AppInitializer::AppInitializer(int argc, char* argv[], const std::string& defaul
...
     InputManager::getManager()->parseInputFile(input_filename, d_input_db);
 
-    // Set custom PETSc options file when one is specified.
-    if (d_input_db->keyExists("petsc_options_file"))
-    {
-        std::string petsc_options_file = d_input_db->getString("petsc_options_file");
-        PetscOptionsInsertFile(PETSC_COMM_WORLD, nullptr, petsc_options_file.c_str(), PETSC_TRUE);
-    }
+    // Set a custom PETSc options file when one is specified.
+    load_petsc_options_file(d_input_db, argc, argv);
```
drwells: Could I convince you to have the PETSc options file in the `Main` subdatabase? That's where all of `AppInitializer`s stuff is supposed to live.

ibtk/src/utilities/ibtk_utilities.cpp
```diff
@@ -13,20 +13,80 @@
...
+#include <string>
 
 #include <ibtk/app_namespaces.h>
 
 namespace IBTK
 {
 /////////////////////////////// PUBLIC ///////////////////////////////////////
+
+void
+load_petsc_options_file(Pointer<Database> input_db, int argc, char* argv[], const std::string& source_dir)
```
drwells: It looks like the only place we need argc and argv is the working directory for the original input file. Why not just pass that in instead? We know the directory where the input file is (otherwise we couldn't open it).

I'm having trouble differentiating between source_dir and argc + argv here. Can we combine them?

drwells (changes requested):
I find the split between `AppInitializer` and the utility function confusing. `load_petsc_options_file()` looks for things in the database parsed by `AppInitializer` in a specific way: why not just put that in `AppInitializer`? Do we anticipate ever trying to load a PETSc options file from some place which isn't an `AppInitializer`?

I'd also like to consider the alternative of simply inlining the flags as strings into the input file. For example, we could do

```cpp
Main {

// log file parameters
   log_file_name               = "cylinder2d.log"
   log_all_nodes               = FALSE

// etc.
   petsc_options_SP = "-SP_ksp_type gmres -SP_ksp_rtol 1.0e-08 -SP_ksp_pc_side right -SP_ksp_gmres_restart 60 -SP_ksp_max_it  1000"
   petsc_options_Linv = "-LInv_ksp_type richardson -LInv_ksp_rtol 1.0e-08 -LInv_ksp_max_it 1 -LInv_ksp_norm_type none"
}
```
or, even more cleanly

```cpp
Main {
// log file parameters
   log_file_name               = "cylinder2d.log"
   log_all_nodes               = FALSE

// etc.
   petsc_options_SP {
      ksp_type          = "gmres"
      ksp_rtol          = 1.0e-08
      ksp_pc_side       = "right"
      ksp_gmres_restart = 60
      ksp_max_it        = 1000
   }

   petsc_options_Linv {
      ksp_type      = "richardson"
      ksp_rtol      = 1.0e-08
      ksp_max_it    = 1
      ksp_norm_type = "none"
   }
}
```
so that we don't have to maintain multiple files. (using subdatabases this way is a Cardinal-ism I like a lot)

knepley:
> I find the split between `AppInitializer` and the utility function confusing. `load_petsc_options_file()` looks for things in the database parsed by `AppInitializer` in a specific way: why not just put that in `AppInitializer`? Do we anticipate ever trying to load a PETSc options file from some place which isn't an `AppInitializer`?
> 
> I'd also like to consider the alternative of simply inlining the flags as strings into the input file. For example, we could do
> 
> ```c++
> Main {
> 
> // log file parameters
>    log_file_name               = "cylinder2d.log"
>    log_all_nodes               = FALSE
> 
> // etc.
>    petsc_options_SP = "-SP_ksp_type gmres -SP_ksp_rtol 1.0e-08 -SP_ksp_pc_side right -SP_ksp_gmres_restart 60 -SP_ksp_max_it  1000"
>    petsc_options_Linv = "-LInv_ksp_type richardson -LInv_ksp_rtol 1.0e-08 -LInv_ksp_max_it 1 -LInv_ksp_norm_type none"
> }
> ```
> 
> or, even more cleanly
> 
> ```c++
> Main {
> // log file parameters
>    log_file_name               = "cylinder2d.log"
>    log_all_nodes               = FALSE
> 
> // etc.
>    petsc_options_SP {
>       ksp_type          = "gmres"
>       ksp_rtol          = 1.0e-08
>       ksp_pc_side       = "right"
>       ksp_gmres_restart = 60
>       ksp_max_it        = 1000
>    }
> 
>    petsc_options_Linv {
>       ksp_type      = "richardson"
>       ksp_rtol      = 1.0e-08
>       ksp_max_it    = 1
>       ksp_norm_type = "none"
>    }
> }
> ```
> 
> so that we don't have to maintain multiple files.

Note also that we can accept JSON input if that is easier to work with.

drwells:
It's good to see you here! I think we are stuck with SAMRAI's own input database format for now - it would be nice to support standard stuff like JSON (or, in my dreams, lua) but that's pretty far off for us.

Edit: clarification: my point is I'd like this to more resemble IBAMR's input format, rather than having stuff in two places with different file formats.

boyceg:
Matt, David is going objecting to having two different kinds of input files.A thing I am not crazy about is having very long lists of PETSc options as single strings in SAMRAI-formatted input files — can SAMRAI actually parse stings that include new lines?On Sep 1, 2026, at 5:27 PM, David Wells ***@***.***> wrote:﻿drwells left a comment (IBAMR/IBAMR#1948)
It's good to see you here! I think we are stuck with SAMRAI's own input database format for now - it would be nice to support standard stuff like JSON (or, in my dreams, lua) but that's pretty far off for us.

knepley:
> It's good to see you here! I think we are stuck with SAMRAI's own input database format for now - it would be nice to support standard stuff like JSON (or, in my dreams, lua) but that's pretty far off for us.
> 
> Edit: clarification: my point is I'd like this to more resemble IBAMR's input format, rather than having stuff in two places with different file formats.

Lua! I used to use a desktop written in Lua (Awesome WM). One of the best stories in South American tech.

I would be more than happy to support a custom IBAMR input format too.

boyceg:
I think the cleanest thing to do is something like:
```
petsc_options {
   ksp_type = "fgmres"
   ksp_rtol = 1.0e-12
}
```
etc. and then insert those options.

However, we also have some existing solvers that use external `petsc_options.dat` files; I don't think we should abandon those just yet.

ibtk/src/utilities/ibtk_utilities.cpp
boyceg: I think that the directory is to handle tests — if we inline all of the options it becomes easy to drop it.

boyceg:
Agreed. The implementation now lives privately in AppInitializer. Options can be supplied as typed entries in `petsc_settings` or `petsc_settings_<prefix>` subdatabases, under Main or alongside the relevant object configuration. The new name avoids collisions with existing `petsc_options_*` settings. Following our interpretation of PETSc conventions, command-line options take precedence over settings supplied in the input file.

ibtk/src/utilities/ibtk_utilities.cpp
boyceg: Changed to the nonthrowing `std::filesystem::exists(path, error_code)` overload. PETSc handles parsing the selected file.

ibtk/src/utilities/ibtk_utilities.cpp
boyceg: Path resolution now uses `std::filesystem::path` throughout, converting to strings only for diagnostics and the PETSc/MPI interfaces.

ibtk/src/utilities/AppInitializer.cpp
boyceg: Settings blocks can appear anywhere in the input database tree, so they can sit alongside the objects they configure. Their location is organizational; only an explicit tag supplies an options prefix. The existing root-level file keys remain available for backward compatibility, but cannot be combined with inline settings.

Two definitions producing the same final PETSc option name are rejected, case-insensitively, even if their values agree. The error identifies both input paths, and validation happens before any settings are inserted.

ibtk/src/utilities/ibtk_utilities.cpp
boyceg: Removed the source-directory argument and fallback. The private resolver receives the known input filename directly and resolves relative options files against its canonical directory. Copies of argc/argv remain only in the PETSc insertion step to preserve command-line precedence—not to infer the input path.

ibtk/src/utilities/AppInitializer.cpp
```diff
@@ -45,6 +59,293 @@ namespace IBTK
...
+};
+
+void
+broadcast_string(std::string& value)
+{
+    int size = IBTK_MPI::getRank() == 0 ? static_cast<int>(value.size()) : 0;
+    size = IBTK_MPI::bcast(size, 0);
+    value.resize(size);
+    if (size > 0) IBTK_MPI::bcast(value.data(), size, 0);
+}
```
boyceg: Could/should this be provided as a public, reusable helper?

ibtk/src/utilities/AppInitializer.cpp
```diff
@@ -45,6 +59,293 @@ namespace IBTK
...
+
+std::string
+lowercase(const std::string& value)
+{
+    std::string result = value;
+    for (char& c : result)
+        if (c >= 'A' && c <= 'Z') c += 'a' - 'A';
+    return result;
+}
```
boyceg: I feel like we are accumulating bespoke typographical helpers. Could/should this be provided as a public, reusable helper?

boyceg: Or can we leverage a standard implementation?

ibtk/src/utilities/AppInitializer.cpp
```diff
@@ -45,6 +59,293 @@ namespace IBTK
...
+
+bool
+is_ascii_letter(const char c)
+{
+    return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z');
+}
```
boyceg: As above: Is there a standard implementation to leverage? If not, could/should this be provided as a public, reusable helper?

ibtk/src/utilities/AppInitializer.cpp
```diff
@@ -45,6 +59,293 @@ namespace IBTK
...
+
+bool
+is_ascii_digit(const char c)
+{
+    return c >= '0' && c <= '9';
+}
```
boyceg: As above: Is there a standard implementation to leverage? If not, could/should this be provided as a public, reusable helper?

ibtk/src/utilities/AppInitializer.cpp
```diff
@@ -45,6 +59,293 @@ namespace IBTK
...
+template <class T>
+std::string
+floating_point_to_string(const T value, const std::string& path)
+{
+    if (!std::isfinite(value)) throw std::invalid_argument("IBTK_PETSC_SETTINGS_NONFINITE: nonfinite value at " + path);
+    std::ostringstream stream;
+    stream.imbue(std::locale::classic());
+    stream << std::setprecision(std::numeric_limits<T>::max_digits10) << value;
+    return stream.str();
+}
```
boyceg, line 116: As above: Is there a standard implementation to leverage? If not, could/should this be provided as a public, reusable helper?

ibtk/src/utilities/AppInitializer.cpp
boyceg: Added a reusable `IBTK_MPI::bcast(std::string&, int root)` overload and used it for both broadcasts here. It follows the existing communicator convention, checks the MPI count conversion, and preserves empty strings and embedded NULs. The existing MPI test now checks these cases with a nonzero root.

ibtk/src/utilities/AppInitializer.cpp
boyceg: Kept only a small private by-value wrapper around the standard `std::ctype<char>::tolower` range overload with `std::locale::classic()`. No new public text utility, Boost dependency, or hand-written character conversion is needed.

ibtk/src/utilities/AppInitializer.cpp
boyceg: Replaced the custom predicate with the standard locale-taking character classification functions, using the classic locale to preserve the tag grammar independently of the process locale.

ibtk/src/utilities/AppInitializer.cpp
boyceg: Removed this helper too. The tag check now uses `std::isalpha` and `std::isalnum` with `std::locale::classic()`, retaining the existing rules for underscores.

ibtk/src/utilities/AppInitializer.cpp
boyceg: I checked floating-point `std::to_chars`: it requires libstdc++ from GCC 11 or libc++ 14, whereas this branch only requires C++17. I retained the classic-locale stream with `max_digits10` rather than raising toolchain requirements or adding a compatibility layer. The finite-value check and input-path diagnostic are unchanged.

References: [GCC 11 release notes](https://gcc.gnu.org/gcc-11/changes.html), [libc++ C++17 status](https://libcxx.llvm.org/Status/Cxx17.html). Leaving this thread open for the compatibility disposition.

boyceg: Updated in 5f4065155bd2400e96552795273208c90e806f8a to use C++17 std::to_chars with std::chars_format::general and max_digits10. This supersedes my earlier compatibility reply: we already require C++17 and will use its standard-library facilities. The finite-value check and input-path diagnostics are retained, and conversion errors are checked. The existing quick tests pass twice in both Debug and Release, including the floating-point value check; no new test executable was needed.

tests/IBTK/petsc_options_file.inline.mpirun=2.input
```diff
@@ -0,0 +1,46 @@
...
+    petsc_options_prefix = "also_not_SP_"
+    petsc_options { ignored_old = 1, 2 }
+    petsc_options_SP { ignored_tag = "unused" }
+    petsc_settings_SP {
+        ksp_max_it = 341
+        enabled = TRUE
+        cli_value = 11
+        programmatic_value = 22
+        options_file = "ordinary-value.opts"
+    }
```
drwells, line 29: We don't need to go from 'no support' to this many supported options. I'll put together a little PR with what I think we can for databases which is much smaller than this.

boyceg: Why do you want to limit the options that are supported? This just treats _everything_ in the `petsc_settings` sub-database as a PETSc option.

boyceg: (Different parts of PETSc support different options, and so I don't think we want to prescribe what options are supported or not.)

boyceg: There are only two choices here IMO:

* We have a way of recognizing and capturing "arbitrary" keys in the input file as PETSc options or
* We only support external options files.

If we have a prescribed list of recognized options, then I think that chances are good that at some point we are going to run into options that are not supported by the parser and wind up back in a situation where we are combining multiple input files.

drwells: By 'options' I meant the number of things app initializer can do. I don't want to limit the number of PETSc options supported, but the generated code supports way more ways of specifying those options than I think we need to do (at least for starters).

boyceg: In what way?

boyceg: The only recognized tags are `petsc_settings` and `petsc_settings_{OPTIONS_PREFIX}`.

I can remove the `petsc_options` keys from the test — that was the original key name, but some parts of IBAMR already read in `petsc_options_prefix`, and so I didn't want to have to worry about special cases.

In contrast, `petsc_settings` and `petsc_settings_X` are not used (as far as I can tell) in IBAMR and so are fair game. 

I think this particular test is Codex over-correcting for that initial implementation using `petsc_options`.

drwells: We can streamline everything a lot (like reduce the number of lines of code in this PR by 50%). One trick we use in cardinal is to look for keys which are databases and start with, e.g., `petsc_options`. From there, if such a key ends in `_XY` then the PETSc prefix is `XY_`. Otherwise it has no prefix. That covers every option which can be set at the command line as a bool, int, double, or string. That really should be it for, e.g., specifying every `.dat` file in IBAMR as an input database.

boyceg: Yes that is what this PR does as well.

boyceg: I think that a lot of the length here is related to error checking. There also is some complexity related to ensuring that any provided command line options take precedence over settings that are in the input file.

drwells:
One thing I missed early on is that we already support loading PETSc options files in `AppInitializer` (via the toplevel `petsc_options_file`). Let's keep that (backwards compatibility is important). I wouldn't have proposed an alternative solution if I knew we already had that in place (though it is my preference to cut down on the total number of input files and syntaxes).

### readthrough

I've gotten lost a few times when trying to read this PR so I took some time to reread everything. The most substantial issue there is error handling (exceptions do not work with MPI): we should use `TBOX_ERROR()`. I also don't think its validation is correct. For example it does

```cpp
                const auto& locale = std::locale::classic();
                const bool valid = !tag.empty() && std::isalpha(tag.front(), locale) &&
                                   std::isalnum(tag.back(), locale) &&
                                   std::all_of(tag.begin(),
                                               tag.end(),
                                               [&locale](const char c) { return std::isalnum(c, locale) || c == '_'; });
                if (!valid) throw std::invalid_argument("IBTK_PETSC_SETTINGS_TAG: invalid tag at " + entry_path);
                prefix = tag + "_";
```
PETSc string options can be more than alphanumeric characters plus underscores (for example, some viewer options contain colons). Why are we doing this validation at all? PETSc can parse whatever it wants to parse - that's not up to us and it will complain if we hand it something nonsensical. Sorting the keys is weird but not wrong. The other main issue with the parsing is that it happens in a completely separate stage from loading which makes it a lot harder to understand.

One place the AI tool is probably getting lost is that SAMRAI database keys must already contain only alphanumeric characters or underscores. We don't have to check that again. I'm pretty sure it would be faster for me to just write this correctly than to try to corral the AI to do the right thing here because it is so far off.

### my recommendation

What was the original issue here? Was it to run

```sh
> mpirun -n 8 ./main3d ./tests/test1.input
```
and have it, e.g., load `tests/test1.dat` for the PETSc options? We don't load any other files with relative paths in that way. If we want to make an exception then we need to only do something like (relative to `master`, not this PR)

```cpp
    // Set custom PETSc options file when one is specified.
    if (d_input_db->keyExists("petsc_options_file"))
    {
        const std::string filename = d_input_db->getString("petsc_options_file");
        const auto relative_filepath =
            std::filesystem::path(input_filename).parent_path() / std::filesystem::path(filename);
        std::string resolved_filename;
        if (std::filesystem::exists(filename))
            resolved_filename = filename;
        else if (std::filesystem::exists(relative_filepath))
            resolved_filename = relative_filepath.string();
        else
            TBOX_ERROR("Neither the PETSc input file '"
                       << filename << "' nor the path resolved relative to the input file '"
                       << relative_filepath.string() << "' exist, so PETSc options cannot be loaded.");

        const int ierr = PetscOptionsInsertFile(PETSC_COMM_WORLD, nullptr, resolved_filename.c_str(), PETSC_TRUE);
        IBTK_CHKERRQ(ierr);
    }
```
Alternatively: do you want to use these PETSc option files in tests and have them automatically loaded by AppInitializer? We do things like that all the time in the test suite (e.g., every test which uses `.vertex` or `.ex2` files). We can add a simple copy function to `tests.h` and copy whatever we need (it has to be named `petsc_options_file` in the top-level DB already: we know where to look) into the test directory without modifying in the library.

If it is the second one, I don't have a draft of that implemented we can ask Codex to use the input filename (`argv[1]`) to create a `MemoryDatabase`, get the value of `petsc_options_file`, and copy `SOURCE_DIR` + that value into the current working directory. `SOURCE_DIR` is a macro defined for tests which is always the source directory of the test (so, by definition, a test's corresponding PETSc input file, vertex files, etc. must be available in that directory or one inside it). That's a small utility function and a one-liner in any relevant test, which is way easier than canonicalizing paths.

We can copy the files after running `IBTKInit` and before setting up `AppInitializer` and everything should work.

boyceg:
Here is an example of a kind of input file that we want to support.
```
Main {
   petsc_settings {
      ksp_type = "fgmres"
   }
   petsc_settings_SP {
      pc_type = "none"
   }
   petsc_settings_poisson {
      rtol = 1.0e-12
      atol = "0.0"
   }
}

PoissonKrylovSolver {
   petsc_options_prefix = "poisson"
   petsc_settings {
      poisson_ksp_type = "cg"
      poisson_max_it = 30
   }
}

PoissonMultigridPreconditioner {
   petsc_options_prefix = "poisson_mg"
   petsc_settings_poisson_mg {
      ksp_type = "preonly"
      max_it = 1
   }
}
```

boyceg:
@drwells what you say?

ibtk/src/utilities/AppInitializer.cpp
```diff
@@ -45,6 +54,151 @@ namespace IBTK
 {
 /////////////////////////////// STATIC ///////////////////////////////////////
 
+namespace
+{
+constexpr int mpi_root = 0;
+constexpr std::size_t floating_point_format_overhead = 16;
```
boyceg: Let's make a rule that these types of constants have variable names that are UPPER_CASE_SNAKE_CASE please.

Also, for constants that are only used in one place, let's make a rule that the scope should be more narrow than the file-level scope — I think it is OK if the definition of the constant comes at the place it is used.

boyceg: Updated in f1c0cd477add8cc7cc7fde5137a7133eb24d2cc8. The two PETSc-settings names remain translation-unit constants because they are shared by multiple functions; the root-rank and floating-format constants now have their narrower assigned scopes, and the test keeps its shared root-rank constant in the unnamed namespace. All use UPPER_CASE_SNAKE_CASE. The root-rank constant is spelled ROOT_RANK because MPI_ROOT is already a macro in mpi.h.

ibtk/include/ibtk/AppInitializer.h
```diff
@@ -38,6 +38,13 @@ namespace IBTK
 /*!
  * \brief Class AppInitializer provides functionality to simplify the
  * initialization code in an application code.
+ *
+ * PETSc options may be supplied by `petsc_settings` or `petsc_settings_<tag>`
+ * databases placed beside component configuration. They accept scalar Boolean,
+ * integer, floating-point, and string values; a tagged key `k` produces
+ * `-<tag>_<k>`. Alternatively, the top-level `PETSC_OPTIONS_FILE` or
+ * `petsc_options_file` key selects an input-relative options file. Command-line
+ * options take precedence.
```
drwells: Could you provide a complete example (via `@code` `@endcode`) in here so it's clear how this is intended to be used?

tests/CIB/cib_double_shell.input
```diff
@@ -24,7 +24,19 @@ L = N*DX                       // length of domain
...
+   ksp_pc_side = "right"
+   ksp_gmres_restart = 60
+   ksp_max_it = 1000
+}
+petsc_settings_LInv {
+   ksp_type = "richardson"
+   ksp_rtol = 1.0e-08
+   ksp_max_it = 2
+   ksp_norm_type = "none"
+}
```
drwells, line 39: First half: can we keep these in `Main`? We shouldn't let objects access top-level things (i.e., one of our rules is that objects are configured by a subdatabase of the top database).

tests/IBTK/petsc_options_file.duplicate.expect_error=true.input
```diff
@@ -0,0 +1,29 @@
...
+        ksp_type = "fgmres"
+    }
+    petsc_settings_SP {
+        pc_type = "none"
+    }
+    petsc_settings_poisson {
+        ksp_type = "cg"
+        rtol = 1.0e-12
+        atol = "0.0"
+    }
```
drwells, line 4: This is the second half: if we support doing it here we don't need to also look for such databases in the top-level database.

This (once we get rid of the `ksp_type` in `petsc_settings_poisson`) would be a good example to use in the documentation so it is clear how things get copied between databases.

tests/CMakeLists.txt
```diff
@@ -322,8 +323,8 @@ ADD_CUSTOM_COMMAND(TARGET tests
   POST_BUILD
   COMMAND ${CMAKE_COMMAND} -E create_symlink ${CMAKE_SOURCE_DIR}/attest ${CMAKE_BINARY_DIR}/attest)
 
-# Set up the input and output files. Since the input and output files aren't
-# really used by the build system we use a shell script to find them every time
+# Set up the input, output, and PETSc options files. Since these files aren't
+# used by the build system we use a shell script to find them every time
```
drwells, line 327: ✔️

tests/IBTK/petsc_options_file.inline.mpirun=2.input
boyceg: I think this is basically fixed in the update.

ibtk/src/utilities/AppInitializer.cpp
```diff
@@ -45,6 +54,151 @@ namespace IBTK
...
+           key.compare(0, TAGGED_PETSC_SETTINGS_PREFIX.size(), TAGGED_PETSC_SETTINGS_PREFIX.data()) == 0;
+}
+
+template <class T>
+std::string
+floating_point_to_string(const T value)
+{
+    constexpr std::size_t FLOATING_POINT_FORMAT_OVERHEAD = 16;
+    // Allow space for the sign, decimal point, and exponent as well as the significant digits.
+    char buffer[std::numeric_limits<T>::max_digits10 + FLOATING_POINT_FORMAT_OVERHEAD];
```
drwells, line 111: Let's use `std::to_string()` for all such conversions instead of rolling our own conversion code.

ibtk/src/utilities/AppInitializer.cpp
```diff
@@ -45,6 +54,151 @@ namespace IBTK
...
+                value = settings_db->getBool(key) ? "true" : "false";
+                break;
+            case Database::SAMRAI_INT:
+                value = std::to_string(settings_db->getInteger(key));
+                break;
+            case Database::SAMRAI_FLOAT:
+                value = floating_point_to_string(settings_db->getFloat(key));
+                break;
+            case Database::SAMRAI_DOUBLE:
+                value = floating_point_to_string(settings_db->getDouble(key));
```
drwells, line 180: let's use `std::to_string()` for everything.

ibtk/src/utilities/AppInitializer.cpp
```diff
@@ -84,20 +238,33 @@ AppInitializer::AppInitializer(int argc, char* argv[], const std::string& defaul
...
     if (d_input_db->isDatabase("Main"))
     {
         main_db = d_input_db->getDatabase("Main");
     }
 
+    // Configure PETSc options and then reapply normal PETSc sources so command-line options win.
+    std::set<std::string> option_names;
+    const bool found_settings = insert_petsc_settings(d_input_db, option_names);
+    const bool has_options_file =
+        d_input_db->keyExists("PETSC_OPTIONS_FILE") || d_input_db->keyExists("petsc_options_file");
```
drwells, line 252: Can we not add the all-caps version? It also isn't documented.

drwells (commented):
I have some minor comments but I'm happy to approve it once we document how this actually works and provide an example. We should also describe the symbolic linking (I'd prefer not to do any path resolution but it isn't a big deal).

tests/CIB/cib_double_shell.input
boyceg: I don't agree. I think that it is OK to have settings in the top level.

ibtk/src/utilities/AppInitializer.cpp
boyceg: At least in my input files, top level keys are usually ALL_CAPS. Let's just document this.

tests/CIB/cib_double_shell.input
drwells: I don't feel that strongly about it so let's keep it.

ibtk/src/utilities/AppInitializer.cpp
boyceg: `std::to_string()` doesn't appear to take any formatting inputs, and so it is easy to get it to give you `0.000....` when you really want something like `1.0e-12`. That is why this uses `std::to_chars()` instead.

ibtk/src/utilities/AppInitializer.cpp
boyceg: see above

ibtk/include/ibtk/AppInitializer.h
boyceg: Added a complete @code example showing untagged and tagged settings, organizational placement, literal prefixes, supported scalar values, command-line precedence, and the mutually exclusive options-file alternative. The two file-key spellings and lookup locations are now documented too.

tests/IBTK/petsc_options_file.duplicate.expect_error=true.input
boyceg: Simplified this to two settings blocks that both generate -poisson_ksp_type. The documentation uses a separate valid, nonduplicated example and explains that database location is organizational.

tests/CMakeLists.txt
boyceg: Thanks.

ibtk/src/utilities/AppInitializer.cpp
boyceg: Kept std::to_string() for integers. Floating-point settings use checked C++17 std::to_chars() with general format and max_digits10 because std::to_string() rounds a tolerance such as 1.0e-12 to 0.000000; the source now records that rationale.

ibtk/src/utilities/AppInitializer.cpp
boyceg: Likewise, integer settings use std::to_string(), while float and double settings retain checked max_digits10 std::to_chars() conversion so small tolerances are not rounded away.

ibtk/src/utilities/AppInitializer.cpp
boyceg: Documented both top-level spellings and that PETSC_OPTIONS_FILE takes precedence when both are present; petsc_options_file remains the backward-compatible spelling.

boyceg:
@drwells I think this addresses your review comments.

ibtk/src/utilities/AppInitializer.cpp
drwells: You're right. Before C++26
```cpp
#include <string>
#include <iostream>

int main()
{
    std::cout << std::to_string(1e-08) << std::endl;
}
```
prints `0.0000`, which is wrong.

drwells (approved):

drwells:
Thanks!

PR #1949, opened 2026-08-29
CI: repair macOS Homebrew dependencies

boyceg:
Fix macOS CI by installing Boost after #1945 and correcting stale Homebrew PETSc dependency paths in a temporary metadata file.

The usual checklist does not apply to this workflow-only change.


boyceg:
@drwells, I think we have to fix IBSAMRAI2 first.

.github/workflows/push-master.yml
```diff
@@ -178,10 +178,16 @@ jobs:
...
+          for DEPENDENCY in boost ccache cmake curl gcc hdf5-mpi hypre ninja open-mpi petsc
           do
             brew list "$DEPENDENCY" 1>/dev/null 2>&1 || brew install "$DEPENDENCY"
           done
+          # Match Homebrew's pkg-config relocation for PETSc's other metadata.
+          PETSC_VARIABLES="$(brew --prefix petsc)/lib/petsc/conf/petscvariables"
+          sed -i '' -E \
+            -e "s#$(brew --cellar open-mpi)/[^/ ]+#$(brew --prefix open-mpi)#g" \
+            -e "s#$(brew --cellar gcc)/[^/ ]+#$(brew --prefix gcc)#g" \
+            "$PETSC_VARIABLES"
```
drwells: Thanks for working on the CI :)

If this is necessary, could we integrate it into cmake somehow instead? We shouldn't require users to edit PETSc's configuration files.

boyceg: Good point. I was thinking that this was something that was specific to the GitHub CI setup, but I guess it actually is a thing that can go wrong with Homebrew-based installations.

I will see if this is something that can be done in a reasonable way in our build system.

In the meantime, perhaps we can get IBAMR/IBSAMRAI2#63 sorted out and update the IBAMR CI to use the updated IBSAMRAI2?

boyceg:
@drwells: see if this looks better to you.

.github/workflows/push-master.yml
boyceg: So since this is a Homebrew issue and since we don't recommend that folks use Homebrew to get PETSc, I think we should just work around it in the CI.

The other option is to hope that Homebrew fixes it. [I don't see anything in the pipeline, however.](https://github.com/Homebrew/homebrew-core/pulls?q=sort:updated-desc+is:pr+petsc+in:title)

(Maybe we should let them know about the issue?)

boyceg:
@drwells: see if this looks better to you. 😅

.github/workflows/push-master.yml
drwells: We talked about this offline yesterday and we came to the conclusion that we shouldn't patch our buildsystem to fix homebrew bugs. IIRC they had the same bug earlier with PETSc and fixed it eventually.

I think this patch is fine but let's bump the ibsamrai patch version first so we know this compiles and links.

drwells (approved):
I think this is fine (and it is needed by #1956)

PR #1950, opened 2026-08-30
[CAV R01a] Extend IB interpolation kernels

boyceg:
Support face-normal and face-tangential IB interpolation kernels, including composite B-spline pairs and correct odd-width stencil placement.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


ibtk/include/ibtk/PETScMatUtilities.h
```diff
@@ -143,39 +147,73 @@ class PETScMatUtilities
...
     /*!
-     * \brief Standard one-dimensional Peskin 4-pt delta function.
+     * \brief Construct a parallel PETSc Mat object corresponding to the
+     * side-centered IB interpolation operator with distinct one-dimensional
+     * kernels in the velocity-component and transverse directions.
      *
-     * \param r Normalized distance (by grid space h) between the IB point and
-     * the lowermost stencil location.
+     * Each direction uses its own stencil width and centering. The callback
+     * convention and odd-width tie rule are the same as in the scalar overload.
```
boyceg: Where is the scalar overload? I think this could be more clear.

include/ibamr/ibamr_enums.h
```diff
@@ -325,6 +325,75 @@ enum_to_string<TimeSteppingType>(TimeSteppingType val)
...
+{
+    PIECEWISE_CONSTANT,
+    BSPLINE_1 = PIECEWISE_CONSTANT,
+    PIECEWISE_LINEAR,
+    BSPLINE_2 = PIECEWISE_LINEAR,
+    BSPLINE_3,
+    BSPLINE_4,
+    BSPLINE_5,
+    BSPLINE_6,
+    COMPOSITE_BSPLINE_32,
```
boyceg, line 345: We should also add COMPOSITE_BSPLINE_{m}{m+1} — those are useful for quantities that are naturally curls.

include/ibamr/IBImplicitStrategy.h
```diff
@@ -142,6 +142,18 @@ class IBImplicitStrategy : public IBStrategy
                                    int dof_index_idx,
                                    double data_time) = 0;
 
+    /*!
+     * Construct the IB interpolation operator for anisotropic kernels.
```
boyceg: Not necessarily anisotropic — could be isotropic.

However, should we have an isotropic version that only takes in one kernel function instead of two?

ibtk/include/ibtk/PETScMatUtilities.h
boyceg: Clarified the two overloads: the single-kernel form applies the same kernel in the face-normal and face-tangential directions, while the two-kernel form selects them independently.

include/ibamr/ibamr_enums.h
boyceg: Added COMPOSITE_BSPLINE_23, _34, _45, and _56 to the public enum/string conversions. The first order is face-normal and the second is face-tangential; the interpolation test now checks all 18 round trips.

include/ibamr/IBImplicitStrategy.h
boyceg: Replaced the ambiguous anisotropic/component-transverse terminology at this seam with face-normal and face-tangential, including the immediate formal and local names and the changelog/PR description.

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
```diff
@@ -0,0 +1,250 @@
...
+#include <cmath>
+
+namespace IBTK
+{
+inline void
+PETScMatUtilities::piecewise_constant_delta_fcn(const double /*r*/, double* const w)
+{
+    w[0] = 1.0;
+    return;
+}
```
drwells, line 31: These kernels are interesting. Unlike the Fortran kernels they compute the full stencil.

Would you consider returning a `std::array` of the correct size instead? That would avoid aliasing and make things a bit cleaner. These kernels also aren't related to managing `Mat`s so can we put them somewhere else?

ibtk/include/ibtk/PETScMatUtilities.h
```diff
@@ -143,39 +148,75 @@ class PETScMatUtilities
...
-        w[0] = 0.125 * (5.0 - 2.0 * r - q);
-        w[1] = 0.125 * (5.0 - 2.0 * r + q);
-        w[2] = 0.125 * (-1.0 + 2.0 * r + q);
-        w[3] = 0.125 * (-1.0 + 2.0 * r - q);
-        return;
-    } // ib_4_interp_fcn
-
-    static const int ib_4_interp_stencil = 4;
+    static void constructPatchLevelSCInterpOp(Mat& mat,
+                                              void (*face_normal_interp_fcn)(double r_lower, double* w),
```
drwells, line 163: There's a substantial overhead to call by pointer, but most of the time we can inline things. How about we make this a template function on the two interpolation functions? We will need to convert them to functors.

include/ibamr/ibamr_enums.h
drwells: Don't we already have this enumeration in `LEInteractor` (albeit it is only in the source file)? Let's combine them.

drwells (changes requested):
I have some comments on the kernels. I'm not familiar enough with the algorithms yet to really evaluate the rest of it.

boyceg:
@drwells I think this addresses your review comments.

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
boyceg: Done.

ibtk/include/ibtk/PETScMatUtilities.h
boyceg: Done.

ibtk/include/ibtk/IBKernelTensorProductEvaluator.h
```diff
@@ -0,0 +1,92 @@
...
+{
+public:
+    /*! \brief Construct an isotropic product of a copyable scalar evaluator. */
+    explicit IBKernelTensorProductEvaluator(NormalEvaluator evaluator);
+
+    /*! \brief Construct a product of normal and tangential kernel evaluators. */
+    IBKernelTensorProductEvaluator(NormalEvaluator normal_evaluator, TangentialEvaluator tangential_evaluator);
+
+    /*! \brief Return the stencil width in each coordinate direction. */
+    template <int Axis, std::size_t Dim>
```
abarret: Should we default `Dim` to `NDIM`?

ibtk/include/ibtk/private/IBKernelTensorProductEvaluator-inl.h
```diff
@@ -0,0 +1,114 @@
...
+IBKernelTensorProductEvaluator<NormalEvaluator, TangentialEvaluator>::get_stencil_widths()
+{
+    static_assert(Dim == 2 || Dim == 3, "Tensor products support two or three dimensions");
+    static_assert(Axis >= 0 && Axis < static_cast<int>(Dim), "Invalid tensor-product axis");
+    constexpr int normal_width = std::tuple_size<NormalWeights>::value;
+    constexpr int tangential_width = std::tuple_size<TangentialWeights>::value;
+    static_assert(normal_width > 0 && tangential_width > 0, "Kernel stencils must be nonempty");
+    static_assert(std::is_same_v<NormalWeights, std::array<double, normal_width>> &&
+                      std::is_same_v<TangentialWeights, std::array<double, tangential_width>>,
+                  "Kernel evaluators must return std::array<double, N>");
```
abarret: Is there a reason we don't call these assertions in the constructor?

ibtk/include/ibtk/private/IBKernelTensorProductEvaluator-inl.h
```diff
@@ -0,0 +1,114 @@
...
+}
+
+template <class NormalEvaluator, class TangentialEvaluator>
+template <int Axis, std::size_t Dim>
+inline auto
+IBKernelTensorProductEvaluator<NormalEvaluator, TangentialEvaluator>::evaluateFactors(
+    const std::array<double, Dim>& r) const
+{
+    // Instantiate the stencil checks even when the consumer only requests factors.
+    (void)get_stencil_widths<Axis, Dim>();
```
abarret: Assertion checking in the constructor would remove this call.

ibtk/include/ibtk/private/IBOperatorRegistry-inl.h
```diff
@@ -0,0 +1,61 @@
...
+    {
+        if (kernel[d] == IBKernel::UNKNOWN)
+            TBOX_ERROR("IBOperatorRegistry::register_interpolation_matrix_sc(): unspecified kernel " << kernel << '\n');
+    }
+    auto& builders = get_builders();
+    if (builders.find(kernel) != builders.end())
+    {
+        TBOX_ERROR("IBOperatorRegistry::register_interpolation_matrix_sc(): kernel " << kernel
+                                                                                     << " is already registered\n");
+    }
```
abarret: Why do we need to error? Can we just silently skip?

ibtk/include/ibtk/IBOperatorRegistry.h
```diff
@@ -0,0 +1,104 @@
...
+ * names appear in the kernel catalog.
+ */
+class IBOperatorRegistry
+{
+public:
+    /*!
+     * \brief Register a side-centered interpolation-matrix evaluator.
+     *
+     * The evaluator must satisfy the requirements of
+     * PETScMatUtilities::constructPatchLevelSCInterpOp(). It is moved into
```
abarret: What are the requirements on the evaluator? As far as I can tell, there are no checks ran on the evaluator in this function, other than it's not `UNKNOWN`. This should be spelled out and checked as early as possible to hit errors early.

It also seems that `get_builders()` registers several IB builders already. When would a user need to call this function?

ibtk/include/ibtk/IBOperatorRegistry.h
```diff
@@ -0,0 +1,104 @@
...
+
+    /*!
+     * \brief Construct a side-centered interpolation matrix using a registered evaluator.
+     *
+     * A single factor is isotropic; two factors are face-normal and
+     * face-tangential, respectively. Missing registration is a fatal error.
+     * No factor may be IBKernel::UNKNOWN.
+     * Matrix layout, replacement, and boundary limitations are described in
+     * PETScMatUtilities::constructPatchLevelSCInterpOp().
+     */
```
abarret, line 79: It is worth noting that this constructs an interpolation matrix on a specified level.

ibtk/include/ibtk/IBOperatorRegistry.h
```diff
@@ -0,0 +1,104 @@
...
+ * \brief Register and construct IB interpolation matrices.
+ *
+ * Supplied registrations include B-splines through the order selected by the
+ * IBTK_MAX_BSPLINE_ORDER CMake setting, IB_3 through IB_6, and their ordered
+ * normal/tangential combinations. Explicit application registrations are not
+ * restricted by this build setting.
+ * Other kernels require an application-provided evaluator, even if their
+ * names appear in the kernel catalog.
+ */
+class IBOperatorRegistry
```
abarret, line 53: Note that `IBOperatorRegistry` is completely separate from `LEInteractor`, which could be a source of confusion.

Will this only be used to construct interpolation matrices, or is there additional intent for this class? You may want to consider renaming this to something like `IBInterpMatRegistry`.

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
```diff
@@ -0,0 +1,145 @@
...
+    SAMRAI::hier::Index<NDIM> d_domain_lower;
+    //! Number of local IB points and first local matrix row.
+    int d_n_local_points = 0, d_row_lower = 0;
+    //! Local patches and component stencil boxes for each IB point.
+    std::vector<int> d_patch_numbers;
+    std::vector<std::vector<SAMRAI::hier::Box<NDIM>>> d_stencil_boxes;
+    //! Borrowed hierarchy data used to read global column indices.
+    SAMRAI::tbox::Pointer<SAMRAI::hier::PatchLevel<NDIM>> d_level;
+    int d_dof_index_idx;
+};
```
abarret: We should include `struct` definitions in public header files, not private included header files.

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
```diff
@@ -0,0 +1,145 @@
...
+template <int Axis, class Evaluator>
+inline void
+PETScMatUtilities::construct_sc_interp_op_axis(SCInterpOpData& data, const Evaluator& evaluator)
+{
+    using namespace SAMRAI;
+    for (int point = 0; point < data.d_n_local_points; ++point)
+    {
+        const double* const X = &data.d_positions[NDIM * point];
+        const hier::Box<NDIM>& box = data.d_stencil_boxes[point][Axis];
+        const auto& lower = box.lower();
```
abarret: Only use `auto` if the type name is clear.
```suggestion
        const hier::Index<NDIM>& lower = box.lower();
```

ibtk/include/ibtk/PETScMatUtilities.h
```diff
@@ -128,54 +129,37 @@ class PETScMatUtilities
...
+     *
+     * Odd widths use the nearest grid point, choosing the higher index at a
+     * tie. Even widths bracket the point using the component's grid centering.
+     * No kernel registration is required. The evaluator is borrowed for this call.
+     * An existing mat is destroyed and replaced; the caller owns the new matrix.
      *
-     * \warning This routine does not properly handle delta functions for which
-     * interp_stencil is odd, nor does it properly handle physical boundary
-     * conditions.
+     * \warning Physical boundary conditions are not handled.
```
abarret, line 164: Are periodic boundaries handled correctly?

ibtk/src/math/PETScMatUtilities.cpp
abarret:
```suggestion
for (unsigned int d = 0; d < NDIM; ++d)
{
d_dx[d] = dx0[d] / static_cast<double>(ratio(d));
}
```

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
```diff
@@ -0,0 +1,145 @@
...
+    /*! \brief Disallow assigning borrowed array access. */
+    SCInterpOpData& operator=(const SCInterpOpData&) = delete;
+    /*! \brief Finish matrix assembly. */
+    void assemble();
+
+    //! Caller-owned matrix handle.
+    Mat& d_mat;
+    //! Borrowed position vector and its array access.
+    Vec d_X;
+    double* d_positions = nullptr;
```
abarret: This is fine as this is used only internally in the class, but it should be noted that `d_positions` is only a valid pointer for the duration of the lifetime of `d_X`.

ibtk/src/math/PETScMatUtilities.cpp
abarret:
```suggestion
for (unsigned int d = 0; d < NDIM; ++d)
{
d_dx[d] = dx0[d] / static_cast<double>(ratio(d));
}
```

include/ibamr/IBImplicitStrategy.h
```diff
@@ -133,11 +135,18 @@ class IBImplicitStrategy : public IBStrategy
         double data_time) = 0;
 
     /*!
-     * Construct the IB interpolation operator.
+     * \brief Construct a matrix mapping side-centered velocity to IB points at data_time.
+     *
+     * A single kernel factor is isotropic; a pair selects face-normal and
+     * face-tangential kernels, respectively. The kernel must have an evaluator
```
abarret: This comment is confusing here. You only specify an `IBKernelTensorProduct`, not a single kernel or a pair. Isotropy is a decision made before this function is called so is not relevant here.

abarret (changes requested):
I think we should reconsider how the `IBOperatorRegistry` creates builders. Do we need IB operators pre-registered or can we do it lazily?

ibtk/include/ibtk/IBKernelTensorProductEvaluator.h
boyceg: Fixed.

(I also added a rule to the WIP `AGENTS.md` to avoid defining dimension separately from `NDIM` unless there is a need for the dimension to be different from `NDIM`.)

ibtk/include/ibtk/private/IBKernelTensorProductEvaluator-inl.h
boyceg: Nope. Moved up.

ibtk/include/ibtk/IBOperatorRegistry.h
boyceg: Comments in the PR attempt to more clearly address the Evaluator requirements. Let me know if the updated header isn't clear.

`get_builders()` generates a set of matrix-builders for "standard" kernels implemented in this PR. The design is meant to be extensible, so that users can get high-performance implementations for their own custom kernels without having to modify the library.

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
boyceg: Agreed. Fixed and made a new `AGENTS.md` rule.

ibtk/include/ibtk/PETScMatUtilities.h
boyceg: I think so. There is a test. 😁  But I would expect bugs until the full solver is assembled, which could take a while...

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
boyceg: Updated.

ibtk/src/math/PETScMatUtilities.cpp
boyceg: Updated.

include/ibamr/IBImplicitStrategy.h
boyceg: Please see if these comments are now more clear.

boyceg:
@abarret I think this addresses your comments.

boyceg:
> I think we should reconsider how the `IBOperatorRegistry` creates builders. Do we need IB operators pre-registered or can we do it lazily?

I think it is now being done lazily.

ibtk/include/ibtk/private/IBOperatorRegistry-inl.h
boyceg: Right now there is no way to ensure that the same evaluator is used in the two calls. We could ensure that by adding a separate registry for IBKernel evaluators.

I am not a fan of registries in the first place — I am doing it here because it gives a way to try to address @drwells' request for inlined evaluation without templating everything on the IB kernel function type.

If there is another way to make the evaluation inline that you guys would prefer, let me know, but really I think we should just get this implicit solver framework fully implemented and working and then fix up any warts introduced along the way.

ibtk/include/ibtk/IBOperatorRegistry.h
boyceg: 1. I am hoping that some day this will replace `LEInteractor`. 😁 
2. I think that there may be other kinds of interpolation matrices constructed, and possibly also spreading operators. Something that @colegruninger97 and I have been working on sorting out is "what interpolation/spreading operator" should be used for "what kind of quantity" — it is _not_ one-size-fits-all.

doc/cmake.md
```diff
@@ -147,6 +147,14 @@ In particular, to use compilers specified by modules, you need to provide paths
...
 
 ### Configuring the build
+- `IBTK_MAX_BSPLINE_ORDER` sets the highest B-spline order automatically registered
+  for IB interpolation matrices. It defaults to `8` and accepts positive decimal
+  integers without leading zeros. For example, configure with
+  `-DIBTK_MAX_BSPLINE_ORDER=12` and rebuild IBAMR to supply orders 1 through 12.
+  Explicit application registrations are independent of this setting. Larger
+  values increase compilation time and storage requirements: including the four
+  supplied IB kernels, a maximum of `M` generates `(M + 4)^2` ordered kernel pairs.
+  Compiler, matrix-index, and available-memory limits still apply.
```
drwells, line 157: > Compiler, matrix-index, and available-memory limits still apply.

What is a matrix-index limit? There are a lot of problems with this text: e.g., it doesn't define most of the things it mentions (e.g., with a link to a class). For now, I'd strongly prefer that we not let agents write documentation because this is so vague it isn't really helpful.

Additionally we never prefix options with `IBTK` (they are always `IBAMR` and internally converted into `IBAMR` and `IBTK`).

ibtk/include/ibtk/IBKernelEvaluators.h
```diff
@@ -0,0 +1,113 @@
...
+ */
+template <std::size_t N>
+class IBKernelEvaluatorBSpline
+{
+    static_assert(N > 0, "B-spline order must be positive");
+
+public:
+    //! Consecutive weights in the kernel stencil.
+    using Weights = std::array<double, N>;
+
```
drwells, line 51:
```suggestion
```

ibtk/include/ibtk/IBKernelEvaluators.h
```diff
@@ -0,0 +1,113 @@
...
+    /*! \brief Return the stencil weights described in \ref IBKernelEvaluators. */
+    Weights operator()(double r) const;
```
drwells, line 53:
```suggestion
    static std::array<double, N> operator()(double r) const;
```

ibtk/include/ibtk/IBKernelEvaluators.h
```diff
@@ -0,0 +1,113 @@
...
+};
+
+/*!
+ * \brief Evaluate the 3-point IB kernel.
+ */
+class IBKernelEvaluatorIB3
```
drwells, line 59: Let's do the same thing here as for B-spline: make the order a template argument instead of adding three classes. We can switch stencils in `operator()` via `if constexpr()`.

ibtk/include/ibtk/IBKernelEvaluators.h
```diff
@@ -0,0 +1,113 @@
...
+{
+public:
+    //! Consecutive weights in the kernel stencil.
+    using Weights = std::array<double, 3>;
+
+    /*! \brief Return the stencil weights described in \ref IBKernelEvaluators. */
+    Weights operator()(double r) const;
```
drwells, line 66:
```suggestion
    /*! \brief Return the stencil weights described in \ref IBKernelEvaluators. */
    static std::array<double, N> operator()(double r) const;
```

ibtk/include/ibtk/IBKernelEvaluators.h
```diff
@@ -0,0 +1,113 @@
...
+
+#include <ibtk/config.h>
+
+#include <array>
+#include <cstddef>
+
+namespace IBTK
+{
+/*!
+ * \defgroup IBKernelEvaluators IB kernel evaluators
```
drwells, line 25: It's telling that this creates a group with documentation and the constituent classes have nearly identical names. How about instead we do

```cpp
namespace IBTK
{
  namespace Kernels
  {
    template <std::size_t N>
    struct BSpline {/*etc.*/};

    struct IB4 {/*etc.*/};

    struct IB6 {/*etc.*/};
  }
}
```
and then we can make this blurb the documentation of `Kernels`?

ibtk/include/ibtk/IBKernelEvaluators.h
```diff
@@ -0,0 +1,113 @@
...
+ * \brief Evaluate consecutive weights of a one-dimensional IB kernel.
+ *
+ * The argument is the displacement from the lower stencil point to the IB
+ * point, divided by the grid spacing. Entry i is the weight at displacement
+ * r - i. The result contains the natural number of stencil points, without
+ * padding. Geometry, grid spacing factors, and tensor products are not part
+ * of these evaluators.
```
drwells, line 32: Similarly this should documented at the relevant class method.

> The result contains the natural number of stencil points, without padding

I don't think that's what we actually compute? Let's write this documentation ourselves.

ibtk/include/ibtk/IBKernelEvaluators.h
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * For an odd width N, (N - 2)/2 <= r < N/2; for an even width N,
+ * N/2 - 1 <= r <= N/2. Stencil placement is the caller's responsibility.
```
drwells, line 35: > Stencil placement is the caller's responsibility.

What does 'placement' mean in this context? We need to precisely define these terms or not use them (see my top-level review comment).

ibtk/include/ibtk/private/IBKernelEvaluators-inl.h
```diff
@@ -0,0 +1,132 @@
...
+    // subject to roundoff. The calculation uses O(N^2) work and O(N) storage.
+    const double t = r - 0.5 * (static_cast<double>(N) - 2.0);
+    Weights w = {};
+    w[0] = 1.0;
+    for (std::size_t degree = 1; degree < N; ++degree)
+    {
+        double saved = 0.0;
+        for (std::size_t i = 0; i < degree; ++i)
+        {
+            const double term = w[i] / static_cast<double>(degree);
```
drwells, line 48: Let's do something to get rid of this divison.

ibtk/include/ibtk/private/IBKernelEvaluators-inl.h
```diff
@@ -0,0 +1,132 @@
...
+    // "On calculating with B-splines", J. Approx. Theory 6 (1972), 50-62,
+    // Section 2, Eq. (27), doi:10.1016/0021-9045(72)90080-9.
+    //
+    // With w_0^(0) = 1 and missing entries zero, the degree-d weights satisfy
+    // w_i^(d) = ((i+1-t)*w_i^(d-1) + (t+d-i)*w_(i-1)^(d-1))/d.
+    // Here t is in [0,1]; degree N-1 gives the N weights in stencil order.
+    // Each old weight is split into nonnegative contributions whose
+    // coefficients sum to one. This avoids cancellation between weights
+    // and preserves their sum in exact arithmetic. Floating-point sums are
+    // subject to roundoff. The calculation uses O(N^2) work and O(N) storage.
```
drwells, line 39: > Floating-point sums are subject to roundoff

Could we avoid writing general statements like this in comments?

ibtk/include/ibtk/IBKernelEvaluators.h
drwells: Similarly, we can just rename this file `kernels.h`.

ibtk/include/ibtk/IBKernelTensorProductEvaluator.h
```diff
@@ -0,0 +1,99 @@
...
+ * IBKernelTensorProductEvaluator evaluator{IBKernelEvaluatorIB4{}, IBKernelEvaluatorIB3{}};
+ * std::array<double, NDIM> r;
+ * r.fill(1.0);
+ * r[1] = 1.5;
+ * const std::array<double, NDIM == 2 ? 12 : 36> weights = evaluator.evaluate<1>(r);
+ * // weights contains 3 * 4 coefficients in 2D, or 3 * 4 * 3 in 3D.
+ * \endcode
+ */
+template <class NormalEvaluator, class TangentialEvaluator = NormalEvaluator>
+class IBKernelTensorProductEvaluator
```
drwells, line 53: Let's avoid redundant words like 'Evaluator'. How about we do

```suggestion
template <class NormalKernel, class TangentialKernel = NormalKernel>
class TensorProductKernel
```

drwells: ~Similarly, we can rename this file `tensor_product_kernel.h`. It's a style preference, but since we are in `IBAMR`/`IBTK` specifying `IB` as a prefix seems redundant.~

Edit: it isn't a great name but since we already have `IBKernelTensorProduct` let's keep the `Evaluator` suffix.

ibtk/include/ibtk/IBKernelTensorProductEvaluator.h
```diff
@@ -0,0 +1,99 @@
...
+{
+public:
+    /*! \brief Construct an isotropic product of a copyable scalar evaluator. */
+    explicit IBKernelTensorProductEvaluator(NormalEvaluator evaluator);
+
+    /*! \brief Construct a product of normal and tangential kernel evaluators. */
+    IBKernelTensorProductEvaluator(NormalEvaluator normal_evaluator, TangentialEvaluator tangential_evaluator);
```
drwells, line 60: If we make `operator()` static then we can skip these constructors: the types completely specify the functions we want to evaluate. Since we already take these kernels as template types we might as well do it.

ibtk/include/ibtk/IBKernelTensorProductEvaluator.h
```diff
@@ -0,0 +1,99 @@
...
+    //! Normal kernel weights.
+    using NormalWeights = std::invoke_result_t<const NormalEvaluator&, double>;
+    //! Tangential kernel weights.
+    using TangentialWeights = std::invoke_result_t<const TangentialEvaluator&, double>;
+
+    static_assert(std::tuple_size<NormalWeights>::value > 0 && std::tuple_size<TangentialWeights>::value > 0,
+                  "Kernel stencils must be nonempty");
+    static_assert(std::is_same_v<NormalWeights, std::array<double, std::tuple_size<NormalWeights>::value>> &&
+                      std::is_same_v<TangentialWeights, std::array<double, std::tuple_size<TangentialWeights>::value>>,
+                  "Kernel evaluators must return std::array<double, N>");
```
drwells, line 88: This is a very weird way to do it but it isn't wrong 🙈

ibtk/include/ibtk/private/IBKernelTensorProductEvaluator-inl.h
```diff
@@ -0,0 +1,131 @@
...
+#include <utility>
+
+namespace IBTK
+{
+template <class NormalEvaluator, class TangentialEvaluator>
+inline IBKernelTensorProductEvaluator<NormalEvaluator, TangentialEvaluator>::IBKernelTensorProductEvaluator(
+    NormalEvaluator evaluator)
+    : d_normal(evaluator), d_tangential(std::move(evaluator))
+{
+    static_assert(std::is_same_v<NormalEvaluator, TangentialEvaluator>, "An isotropic product uses one evaluator type");
```
drwells, line 30:
```suggestion
```
This is just too bizarre to keep.

ibtk/include/ibtk/IBKernelTensorProductEvaluator.h
```diff
@@ -0,0 +1,99 @@
...
+
+    //! Owned normal kernel evaluator.
+    NormalEvaluator d_normal;
+    //! Owned tangential kernel evaluator.
+    TangentialEvaluator d_tangential;
+};
```
drwells, line 94: We can significantly simplify the implementation by making `operator()` static and getting rid of these: we only need the types (which we always have since they are template arguments).

ibtk/include/ibtk/IBKernelTensorProductEvaluator.h
```diff
@@ -0,0 +1,99 @@
...
+     * Each scalar evaluator is called once per corresponding direction.
+     */
+    template <int Axis>
+    auto evaluate(const std::array<double, NDIM>& r) const;
+
+private:
+    /*! \brief Return a tuple of one-dimensional weight arrays in coordinate order. */
+    template <int Axis>
+    auto evaluateFactors(const std::array<double, NDIM>& r) const;
+
```
drwells, line 78: Could you document `Axis` here and elsewhere?

ibtk/src/CMakeLists.txt
```diff
@@ -240,6 +241,9 @@ IF(IBAMR_HAVE_LIBMESH)
     )
 ENDIF()
 
+SET_PROPERTY(SOURCE math/IBOperatorRegistry.cpp APPEND PROPERTY COMPILE_DEFINITIONS
+  IBTK_MAX_BSPLINE_ORDER=${IBTK_MAX_BSPLINE_ORDER})
+
```
drwells, line 246: Let's add this via `#cmakedefine` to the configuration headers instead of doing it this way (i.e., let's use the standard configuration mechanisms)

include/ibamr/IBImplicitStaggeredHierarchyIntegrator.h
```diff
@@ -84,6 +84,12 @@ class IBImplicitStaggeredHierarchyIntegrator : public IBHierarchyIntegrator
      * some default values, reads in configuration information from input and
      * restart databases, and registers the integrator object with the restart
      * manager when requested.
+     *
+     * The minimum ghost width supplied by ib_method_ops must cover the kernel
+     * selected by \c jacobian_delta_fcn, which may differ from the coupling
```
drwells, line 89: What is `jacobian_delta_fcn`? I don't see it documented anywhere else in this file (I believe my top-level comment about documentation applies here as well).

ibtk/include/ibtk/IBKernelTensorProductEvaluator.h
```diff
@@ -0,0 +1,99 @@
...
+public:
+    /*! \brief Construct an isotropic product of a copyable scalar evaluator. */
+    explicit IBKernelTensorProductEvaluator(NormalEvaluator evaluator);
+
+    /*! \brief Construct a product of normal and tangential kernel evaluators. */
+    IBKernelTensorProductEvaluator(NormalEvaluator normal_evaluator, TangentialEvaluator tangential_evaluator);
+
+    /*! \brief Return the stencil width in each coordinate direction. */
+    template <int Axis>
+    static constexpr std::array<int, NDIM> get_stencil_widths();
```
drwells, line 64: Let's consistently use `std::size_t` or `int` everywhere for stencil widths. It is confusing to do both.

tests/CMakeLists.txt
```diff
@@ -203,6 +203,9 @@ SETUP(ConstraintIB oscillating_rigid_cylinder.cpp IBAMR2d)
 # IB:
 SETUP(IB explicit_ex0.cpp IBAMR2d)
 SETUP(IB explicit_ex1.cpp IBAMR2d)
+SETUP(IB implicit_stokes_ib_solver_components_01.cpp IBAMR2d)
+TARGET_COMPILE_DEFINITIONS(tests-IB_implicit_stokes_ib_solver_components_01 PRIVATE
+  IBTK_MAX_BSPLINE_ORDER=${IBTK_MAX_BSPLINE_ORDER})
```
drwells, line 208: Same thing as my other cmake comment: we should not have multiple configuration mechanisms.

ibtk/include/ibtk/private/IBKernelTensorProductEvaluator-inl.h
```diff
@@ -0,0 +1,131 @@
...
+    for (std::size_t d = 0; d < NDIM; ++d)
+    {
+        widths[d] = d == Axis ? normal_width : tangential_width;
+    }
+    return widths;
+}
+
+template <class NormalEvaluator, class TangentialEvaluator>
+template <int Axis>
+inline auto
```
drwells, line 59: Let's not use `auto` as the return type unless we are doing complex generic programming.

ibtk/include/ibtk/IBOperatorRegistry.h
```diff
@@ -0,0 +1,111 @@
...
+     * \brief Register a side-centered interpolation-matrix evaluator.
+     *
+     * The evaluator must satisfy the requirements of
+     * PETScMatUtilities::constructPatchLevelSCInterpOp(). The registry owns the
+     * evaluator and invokes it as const; move-only evaluators are supported.
+     *
+     * Register each kernel combination on every MPI rank that uses it.
+     * Duplicate registration is a fatal error.
+     * No factor may be IBKernel::UNKNOWN. The evaluator must implement the
+     * mathematical kernel named by kernel.
```
drwells: I don't think this comment is correct: it isn't possible to create an invalid `IBKernelTensorProduct` without that being caught earlier.

ibtk/include/ibtk/IBOperatorRegistry.h
```diff
@@ -0,0 +1,111 @@
...
+     * No factor may be IBKernel::UNKNOWN.
+     * Matrix layout, replacement, and boundary limitations are described in
+     * PETScMatUtilities::constructPatchLevelSCInterpOp().
+     */
+    static void construct_interpolation_matrix_sc(Mat& mat,
+                                                  const IBKernelTensorProduct& kernel,
+                                                  Vec X,
+                                                  const std::vector<int>& num_dofs,
+                                                  int dof_idx,
+                                                  SAMRAI::tbox::Pointer<SAMRAI::hier::PatchLevel<NDIM>> level);
```
drwells, line 85: Could we avoid adding another registry and global variables? We should be able to dispatch based on the `IBKernelTensorProduct` internally without storing a bunch of temporary data.

ibtk/include/ibtk/IBOperatorRegistry.h
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * Supplied registrations include B-splines through the order selected by the
+ * IBTK_MAX_BSPLINE_ORDER CMake setting, IB_3 through IB_6, and their ordered
+ * normal/tangential combinations. Other kernels require an application-provided
+ * evaluator, even if their names appear in the kernel catalog. Explicit
+ * registrations are not restricted by this build setting.
+ *
+ * This registry constructs interpolation matrices; it does not register
+ * interpolation or spreading functions with LEInteractor.
+ */
```
drwells, line 52: Is this class really necessary? The three files which define `IBOperatorRegistry` are about 300 lines of code. Can we just make everything we need template functions and remove all the global variables?

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
```diff
@@ -0,0 +1,128 @@
...
+                {
+                    if (width <= 0)
+                    {
+                        return false;
+                    }
+                }
+            }
+            return true;
+        }(),
+        "Evaluator stencil widths must be positive");
```
drwells, line 68:
```suggestion
```
This code is also too wacky to keep. We already enforce this elsewhere.

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
```diff
@@ -0,0 +1,128 @@
...
+    SCInterpOpData data(mat, X_vec, widths, num_dofs_per_proc, dof_index_idx, patch_level);
+    construct_sc_interp_op_axis<0>(data, evaluator);
+    construct_sc_interp_op_axis<1>(data, evaluator);
+#if (NDIM == 3)
+    construct_sc_interp_op_axis<2>(data, evaluator);
+#endif
+    data.assemble();
+}
+
+template <int Axis, class Evaluator>
```
drwells, line 78: This is the other half of my comment from `IBOperatorRegistry`. The relevant functions are already templated on the evaluator: do we really need a whole registry of functions to get here?

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
```diff
@@ -0,0 +1,128 @@
...
+        for (int width : Evaluator::template get_stencil_widths<Axis>())
+        {
+            size *= width;
+        }
+        return size;
+    }();
+    using Values = decltype(std::declval<const Evaluator&>().template evaluate<Axis>(
+        std::declval<const std::array<double, NDIM>&>()));
+    static_assert(std::is_same_v<Values, std::array<double, nvalues>>,
+                  "A const evaluator must return std::array<double, N> with N equal to the stencil size");
```
drwells, line 95:
```suggestion
```
Let's get rid of this junk.

drwells (commented):
There's a lot of unusual codegen which intermixes `std::tuple` utility functions and `std::array`. It isn't wrong so (in the spirit of agentic development) let's leave it alone.

A recurring theme here is that, even if I understand the intent of the source code, the documentation all feels very off - it includes a lot of tangents and technical terms which are not well-defined. We can use these agentic tools to write code which conforms to an interface but the generated prose leaves much to be desired (i.e., I don't know what it is talking about 25% of the time). We are doing our future selves a disservice if we leave vague and misleading comments.

How about we make one of our rules that we should not use agentic tools for documentation? That would also help us clarify the designs.

ibtk/include/ibtk/IBKernelEvaluators.h
boyceg: I don't want to do that — these are different because there is not really a unifying construction, like there is for the B-spline family of kernel functions.

ibtk/include/ibtk/IBOperatorRegistry.h
boyceg: I don't know. This is to address your comment about inlining these functions.

My goals here are:
1. provide "first class" support for user-defined kernel functions and
2. avoid templating the entire solver hierarchy on the kernel function (especially since we may want to use different kernels in different places).

If there is another way to do that, let me know and I will try to do that instead.

ibtk/include/ibtk/IBOperatorRegistry.h
boyceg: See comment above. The purpose here is to enable us to inject inlinable functors into the inner loops of the solvers without templating the solvers on the kernel function. If there is a better way to do it, I would be happy to get rid of this registry class.

ibtk/include/ibtk/private/PETScMatUtilities-inl.h
boyceg: See above.

drwells:
Another possibility is to update to C++20 and define a `concept` for the kernels: that would encode the restrictions in a way which works with the type system (and would prevent these agents from doing error-checking code in every single function)

boyceg:
> Another possibility is to update to C++20 and define a `concept` for the kernels: that would encode the restrictions in a way which works with the type system (and would prevent these agents from doing error-checking code in every single function)

Sounds good to me. We are in the year 2026 after all.

ibtk/include/ibtk/private/IBOperatorRegistry-inl.h
drwells: I just saw this comment. I do a lot of template-dispatch trickery: I can come up with an alternative if you'd like, but since the registry is 300 lines I suspect just leaving everything in headers and removing the static variables will have basically the same effect.

ibtk/include/ibtk/IBOperatorRegistry.h
drwells: I'll sketch something out.

boyceg: Great. The key thing is that I want to be able to have the ability to use different kernels in different parts of the solver, and so it seems like the templating needs to be pretty 'localized'.

drwells: Using template functions and duck-typing (or being modern and using concepts) will also make it even easier to use user-specified kernels: people can plug whatever they want into `IBTensorProductKernelEvaluator` and it will just work.

boyceg: (As a concrete example, it may make sense to use different kernels for operators or preconditioners, or for different components of forces/velocities, or ...)

ibtk/include/ibtk/IBKernelEvaluators.h
boyceg: What is wrong with giving the weights a name? We may want to use different storage types in the future. (I am thinking about GPU implementations. 😓)

boyceg: As a concrete example — single or half precision in preconditioners.

drwells: oh yeah, if we want to switch to `Kokkos::Array` then we will have to change this later. Let's keep it.

Making things work with lower-precision is also a good motivator to write a concept which lets us return something which is a fixed-size array of some kind of floating-point number.

ibtk/include/ibtk/IBKernelEvaluators.h
drwells: I'd still like to make it `static` since we always use these as template types (i.e., we don't need instances)

ibtk/include/ibtk/IBKernelEvaluators.h
boyceg: For B-splines, you can truly put any positive `N` and — at least in principle — get a reasonable function.

ibtk/include/ibtk/IBOperatorRegistry.h
boyceg: Let's update to C++20 and use concepts.

ibtk/include/ibtk/IBKernelEvaluators.h
boyceg: I believe this is a C++23-ism?

boyceg:
Superseded by #1996, #1997, and #1998.

PR #1955, opened 2026-09-03
[CAV R01-preA] Unify IB kernel definitions

boyceg:
Use shared, extensible kernel identities and ordered factor descriptions in LEInteractor while preserving existing kernel behavior.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,97 @@
...
+
+    //! Reconstruct the canonical name for configuration or diagnostics, not dispatch.
+    std::string getName() const;
+
+    bool operator==(const IBKernel& other) const;
+
+    bool operator!=(const IBKernel& other) const;
+
+private:
+    std::array<std::uint64_t, 2> d_name;
```
boyceg: Should we make the size of the array (and hence the number of characters allowed) a compile-time parameter that defaults to 2? Alternatively we could set the number of characters and derive the array size?

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,81 @@
...
+    //! Repeat the scalar factor in every slot.
+    explicit IBKernelTensorProduct(const IBKernel& kernel);
+
+    //! Store the first factor followed by repetitions of the second.
+    IBKernelTensorProduct(const IBKernel& first, const IBKernel& second);
+
+#if (NDIM == 3)
+    //! Specify all three ordered factors.
+    IBKernelTensorProduct(const IBKernel& first, const IBKernel& second, const IBKernel& third);
+#endif
```
boyceg: Shouldn't this just take in a tuple and not have these 1, 2, and 3 argument specializations?

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,81 @@
...
+
+    //! Specify all ordered factors.
+    explicit IBKernelTensorProduct(const std::array<IBKernel, NDIM>& factors);
```
boyceg: Why limit to NDIM? I think we should let it be some number of inputs — not specified in the API.

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,81 @@
...
+
+    //! Access a factor by slot; throws std::out_of_range for slot >= NDIM.
+    const IBKernel& getKernel(unsigned int slot) const;
+
+    bool isIsotropic() const;
```
boyceg, line 61: This needs documentation.

ibtk/include/ibtk/IBKernelTensorProduct.h
boyceg: In this case we should include an accessor to find out how many "slots" have been set.

ibtk/src/utilities/IBKernel.cpp
```diff
@@ -0,0 +1,200 @@
...
+                                                                       encode_literal("IB_6"),
+                                                                       encode_literal("USER_DEFINED") } };
+
+std::string
+uppercase_name(std::string name)
+{
+    for (char& c : name)
+        if (c >= 'a' && c <= 'z') c = static_cast<char>(c - 'a' + 'A');
+    return name;
+}
```
boyceg: We seem to be collecting bespoke typographical manipulators — any that are not provided by the C++17 standard should go into a common header. ibtk_utilities.h is one possibility, but may not be the best one.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,107 @@
...
+
+    bool operator==(const IBKernel& other) const;
+
+    bool operator!=(const IBKernel& other) const;
+
+private:
+    static constexpr std::size_t MAX_NAME_LENGTH = 24;
+    static constexpr std::size_t DIGITS_PER_WORD = 12;
+    static constexpr std::uint64_t ENCODING_BASE = 38;
+    static constexpr std::size_t NAME_WORD_COUNT = (MAX_NAME_LENGTH + DIGITS_PER_WORD - 1) / DIGITS_PER_WORD;
```
boyceg: I think we should avoid using the term "word" here since it is going to be confused with the "word" corresponding to the encoded kernel name string.

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,76 @@
...
+    /*!
+     * \brief Interpret an existing kernel name or a custom scalar identity.
+     *
+     * Scalar names supply one factor. Composite names supply the two ordered
+     * factors documented in \ref ib_kernel_catalog. No dimensional expansion occurs.
+     *
+     * \throws std::invalid_argument if the scalar name violates the canonical
+     * name rules in \ref ib_kernel_catalog.
+     */
+    static IBKernelTensorProduct from_name(const std::string& name);
```
boyceg: Should this just be a constructor that takes in a `std::string`?

ibtk/src/utilities/IBKernel.cpp
```diff
@@ -0,0 +1,165 @@
...
+// Zero means this is not a recognized composite spelling.
+int
+composite_orders(const std::string& name)
+{
+    if (name == "DISCONTINUOUS_LINEAR") return 21;
+    const int pairs[] = { 12, 21, 23, 32, 34, 43, 45, 54, 56, 65 };
+    for (const int pair : pairs)
+        if (name == "COMPOSITE_BSPLINE_" + std::to_string(pair)) return pair;
+    return 0;
+}
```
boyceg: Should all handling of tensor-product kernel names go in `IBKernelTensorProduct`? I think that `IBKernel` is meant only to provide a way to name scalar kernel functions.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,107 @@
...
+ * using ASCII rules. Canonical scalar names contain 1 to 24 characters from
+ * A-Z, 0-9, and underscore; whitespace and other bytes are not accepted.
+ * Naming a custom kernel does not register an evaluator. USER_DEFINED retains
+ * its separate legacy callback meaning; it is not an alias for custom names.
+ *
+ * IBKernelTensorProduct::from_name() also recognizes COMPOSITE_BSPLINE_XY for
+ * XY = 12, 21, 23, 32, 34, 43, 45, 54, 56, 65, supplying ordered factors
+ * {BSPLINE_X, BSPLINE_Y}. DISCONTINUOUS_LINEAR aliases 21. These
+ * product names cannot be used as scalar names. Execution support is a
+ * separate property of the consuming operation.
```
boyceg: IBKernel should not be documenting IBKernelTensorProduct. Keep it focused on IBKernel (which in this design is used to give names to scalar kernel functions only).

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,107 @@
...
+    IBKernel(Builtin kernel);
+
+    /*!
+     * \brief Construct a scalar identity.
+     *
+     * \throws std::invalid_argument if the canonical name is empty, exceeds
+     * 24 characters, contains an invalid character, or is a recognized
+     * composite name (including DISCONTINUOUS_LINEAR). Aliases are resolved
+     * before checking the canonical scalar length. Use
+     * IBKernelTensorProduct::from_name() to interpret those composite names.
```
boyceg: Same comment as above — this is letting tensor-product details leak into IBKernel.

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,76 @@
...
+/*!
+ * \brief Value describing a tensor product of scalar IBKernel factors.
+ *
+ * Stores exactly the supplied nonempty ordered sequence, of any length.
+ * Equality compares both count and ordered identities: {K} differs from {K,K}.
+ * Construction and copying may allocate. Numerical consumers take descriptions
+ * by reference. Expansion and directional meaning belong to the consuming operation;
+ * see \ref le_interactor_kernel_mapping for LEInteractor's convention.
+ * See \ref ib_kernel_catalog for built-ins and aliases. A description does not
+ * imply that a particular operation can execute it.
```
boyceg: Let's not talk about LEInteractor details here; this is just giving names to tensor-product kernel functions.

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,76 @@
...
+{
+public:
+    //! Own the supplied factors; throws std::invalid_argument for an empty collection.
+    explicit IBKernelTensorProduct(std::vector<IBKernel> factors);
+
+    //! Number of supplied factors, without dimensional expansion.
+    std::size_t getNumberOfKernels() const;
+
+    //! Access a factor; throws std::out_of_range for slot >= getNumberOfKernels().
+    const IBKernel& getKernel(std::size_t slot) const;
```
boyceg: I wonder if it would be better to overload operator[] here? Make the interface look like a simplified version of vector<T>?

tests/IBTK/interpolation_utilities.cpp
```diff
@@ -30,8 +31,180 @@
...
+require_unsupported(F operation)
+{
+    bool rejected = false;
+    try
+    {
+        operation();
+    }
+    catch (const std::invalid_argument& error)
+    {
+        rejected = std::string(error.what()) == "LEInteractor: unsupported kernel description";
```
boyceg: What does this have to do with LEInteractor?

ibtk/src/lagrangian/LEInteractor.cpp
```diff
@@ -37,6 +37,8 @@
 #include <SideData.h>
 #include <SideGeometry.h>
 
+#include <stdexcept>
```
boyceg: IBAMR style largely avoids exceptions. I think we should not change that at this point.

ibtk/include/ibtk/private/IBKernel-inl.h
```diff
@@ -0,0 +1,58 @@
...
+
+#ifndef included_IBTK_IBKernel_inl
+#define included_IBTK_IBKernel_inl
+
+namespace IBTK
+{
+constexpr std::array<std::uint64_t, IBKernel::NAME_BLOCK_COUNT>
+IBKernel::encode_name(std::string_view name)
+{
+    // Zero pads each block; 38^12 < 2^63. Call only with canonical names.
```
boyceg: This comment is out of date?

ibtk/include/ibtk/private/IBKernel-inl.h
```diff
@@ -0,0 +1,58 @@
...
+inline constexpr IBKernel IBKernel::BSPLINE_3(IBKernel::encode_name("BSPLINE_3"));
+inline constexpr IBKernel IBKernel::BSPLINE_4(IBKernel::encode_name("BSPLINE_4"));
+inline constexpr IBKernel IBKernel::BSPLINE_5(IBKernel::encode_name("BSPLINE_5"));
+inline constexpr IBKernel IBKernel::BSPLINE_6(IBKernel::encode_name("BSPLINE_6"));
+inline constexpr IBKernel IBKernel::PIECEWISE_CUBIC(IBKernel::encode_name("PIECEWISE_CUBIC"));
+inline constexpr IBKernel IBKernel::IB_3(IBKernel::encode_name("IB_3"));
+inline constexpr IBKernel IBKernel::IB_4(IBKernel::encode_name("IB_4"));
+inline constexpr IBKernel IBKernel::IB_4_W8(IBKernel::encode_name("IB_4_W8"));
+inline constexpr IBKernel IBKernel::IB_5(IBKernel::encode_name("IB_5"));
+inline constexpr IBKernel IBKernel::IB_6(IBKernel::encode_name("IB_6"));
```
boyceg: Would these be cleaner just to pass in the string instead of `IBKernel::encode_name(...)`?

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,109 @@
...
+    static const IBKernel BSPLINE_3;
+    static const IBKernel BSPLINE_4;
+    static const IBKernel BSPLINE_5;
+    static const IBKernel BSPLINE_6;
+    static const IBKernel PIECEWISE_CUBIC;
+    static const IBKernel IB_3;
+    static const IBKernel IB_4;
+    static const IBKernel IB_4_W8;
+    static const IBKernel IB_5;
+    static const IBKernel IB_6;
```
boyceg, line 52: Let's consistently list in alphabetical order

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,109 @@
...
+
+    /*!
+     * \brief Construct a scalar identity.
+     *
+     * The name must contain 1 to 24 ASCII letters, digits, or underscores.
+     * Scalar aliases are resolved before constructing the identity.
+     */
+    explicit IBKernel(const std::string& name);
```
boyceg, line 62: Constructor needs to say what the constraints are on `name`.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,109 @@
...
+
+    //! Construct a scalar identity from a nonnull C string.
+    explicit IBKernel(const char* name);
```
boyceg: As above.

ibtk/include/ibtk/private/IBKernel-inl.h
boyceg: Also, should we have an assertion that the provided name is canonical?

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,109 @@
...
+/*!
+ * \brief Value identifying a scalar one-dimensional IB kernel.
+ *
+ * Copies and equality use an exact two-block value, without string operations
+ * or allocation. The same canonical name has the same value on every process,
+ * independently of construction order. Resolve names at configuration
+ * boundaries, not in numerical loops. Store names, not internal values, in
+ * input and restart data.
+ * The compile-time name capacity is currently 24 characters; changing it
+ * requires rebuilding both the library and its applications consistently.
```
boyceg: Let's point to the variable that controls this. Also I think this comment block needs to be reflowed.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * The public values below are the standard scalar catalog. Standard names
+ * are case-insensitive; PIECEWISE_CONSTANT and PIECEWISE_LINEAR alias BSPLINE_1
+ * and BSPLINE_2. All names, including custom names, are normalized to uppercase
+ * using ASCII rules. Canonical scalar names contain 1 to 24 characters from
+ * A-Z, 0-9, and underscore; whitespace and other bytes are not accepted.
+ * Naming a custom kernel does not register an evaluator.
```
boyceg: Can we attach this comment to the block of public kernels? Put them in a Doxygen comment grouping?

ibtk/include/ibtk/IBKernel.h
boyceg: Also 24 is the default value of `MAX_NAME_LENGTH`. Let's avoid 'magic numbers' in comments — they can get stale.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,109 @@
...
+    explicit IBKernel(const char* name);
+
+    //! Whether name satisfies the scalar kernel-name grammar.
+    static bool isValidName(const std::string& name);
+
+    /*!
+     * \brief Return the standard scalar kernels.
+     *
+     * The order is for iteration only: it is not an ordinal, serialization,
+     * or communication format.
```
boyceg: Is this needed? At the same time, would it hurt to make the kernels partially ordered and put them in a set or store them in an ordered vector? Comparison is cheap and easy using the coded name.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,109 @@
...
+    std::string getName() const;
+
+    bool operator==(const IBKernel& other) const;
+
+    bool operator!=(const IBKernel& other) const;
+
+private:
+    static constexpr std::size_t MAX_NAME_LENGTH = 24;
+    static constexpr std::size_t DIGITS_PER_BLOCK = 12;
+    static constexpr std::uint64_t ENCODING_BASE = 38;
```
boyceg, line 99: Let's include comments on this block of variables that explains what we are doing more precisely.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,109 @@
...
+    static constexpr std::size_t NAME_BLOCK_COUNT = (MAX_NAME_LENGTH + DIGITS_PER_BLOCK - 1) / DIGITS_PER_BLOCK;
+
+    static constexpr std::array<std::uint64_t, NAME_BLOCK_COUNT> encode_name(std::string_view name);
```
boyceg, line 114: Needs documentation, including any requirements on name.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,109 @@
...
+
+    explicit constexpr IBKernel(const std::array<std::uint64_t, NAME_BLOCK_COUNT>& name) : d_name(name)
+    {
+    }
```
boyceg: Should this go in the private header?

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,99 @@
...
+    IBKernelTensorProduct(const std::string& name);
+
+    //! Implicitly interpret a nonnull legacy scalar or composite C string.
+    IBKernelTensorProduct(const char* name);
+
+    //! Whether name can construct a tensor-product description.
+    static bool isValidName(const std::string& name);
+
+    //! Number of meaningful factors, one or two.
+    std::size_t size() const;
```
boyceg: Number of _distinct_ factors?

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,99 @@
...
+
+    //! Access a factor. Valid indices satisfy slot < size().
+    const IBKernel& operator[](std::size_t slot) const;
+
+    //! Whether the value has one all-directions factor.
```
boyceg: Awkward wording.

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,99 @@
...
+    //! This does not imply that a consumer supports the kernel.
+    bool isIsotropic() const;
+
+    bool operator==(const IBKernelTensorProduct& other) const;
+
+    bool operator!=(const IBKernelTensorProduct& other) const;
```
boyceg, line 67: As with IBKernel should we allow for a partial ordering? Would be pretty cheap to evaluate.

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,99 @@
...
+
+private:
+    struct CanonicalFactors
+    {
+        std::array<IBKernel, 3> factors;
```
boyceg: 1. Why 3?
2. Struct needs documentation?

boyceg: Let's please avoid magic numbers.

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,99 @@
...
+        std::size_t size;
+    };
+
+    explicit IBKernelTensorProduct(CanonicalFactors factors);
+
+    static CanonicalFactors canonicalize(std::initializer_list<IBKernel> factors);
+
+    static CanonicalFactors parse_name(const std::string& name);
```
boyceg, line 99: Documentation?

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,99 @@
...
+
+    std::array<IBKernel, 3> d_factors;
```
boyceg: magic number

ibtk/include/ibtk/LEInteractor.h
```diff
@@ -71,10 +73,35 @@ namespace IBTK
...
+ * numerical output. Masked operations support only IB_4 and the legacy
+ * USER_DEFINED callback.
+ *
+ * \anchor le_interactor_kernel_mapping
+ * One factor applies on every axis. In a two-factor value, slot zero applies
+ * on the distinguished component axis and slot one on every transverse axis.
+ * The distinguished axis is the side-normal axis for side data, the edge axis
+ * for edge data, and axis zero for cell/node data regardless of data component.
+ * Use get_kernel_factor() to resolve this convention, including in other
+ * consumers.
```
boyceg: Let's clean up these comments. Do we need to say all of this? Is it the best place to say it?

ibtk/include/ibtk/LEInteractor.h
```diff
@@ -71,10 +73,35 @@ namespace IBTK
...
  */
 class LEInteractor
 {
 public:
+    /*!
+     * \brief Resolve an ordered factor for a Cartesian axis using
```
boyceg: What is an "ordered factor" here?

ibtk/include/ibtk/LEInteractor.h
```diff
@@ -92,18 +119,26 @@ class LEInteractor
...
-     * specified kernel function.
-     */
-    static int getStencilSize(const std::string& kernel_fcn);
-
     /*!
      * \brief Return whether or not the provided string corresponds to a known
      * kernel function.
+     * Invalid or unsupported names return false; name-validation errors do
+     * not escape this capability query.
      */
```
boyceg, line 113:
```suggestion
    * \brief Return whether or not the provided string corresponds to a known
    * kernel function.
    * 
    * Invalid or unsupported names return false; name-validation errors do
    * not escape this capability query.
    */
```
Make sure new paragraphs are separated by a blank line in the comment. If a new paragraph is not intended, this needs a reflow.

ibtk/src/utilities/IBKernel.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+IBKernel::IBKernel(const char* name) : IBKernel(name ? std::string(name) : std::string())
+{
+}
+
+const std::vector<IBKernel>&
+IBKernel::getStandardKernels()
+{
+    static const std::vector<IBKernel> kernels = { BSPLINE_1,       BSPLINE_2, BSPLINE_3,
```
boyceg: Let's consistently list standard kernels in alphabetical order.

ibtk/src/utilities/IBKernelTensorProduct.cpp
```diff
@@ -0,0 +1,163 @@
...
+namespace
+{
+int
+composite_orders(const std::string& name)
+{
+    if (name == "DISCONTINUOUS_LINEAR") return 21;
+    const int pairs[] = { 12, 21, 23, 32, 34, 43, 45, 54, 56, 65 };
+    for (const int pair : pairs)
+        if (name == "COMPOSITE_BSPLINE_" + std::to_string(pair)) return pair;
+    return 0;
```
boyceg: This should be composite bspline orders — not just composite orders?

boyceg:
@drwells I think this is ready. It is just a way of describing kernels.

boyceg:
describing / naming kernels.

tests/IBTK/ib_kernel_3d.output
```diff
@@ -0,0 +1,6 @@
+Scalar names and aliases: PASS
+Ordered and isotropic products: PASS
+Custom canonical identities and two-block value copies: PASS
+Exact encoding, maximum-length round trip, and rank order/subset independence: PASS
+Canonical axis-relative products, equality, and diagnostics: PASS
+Scalar/composite parsing ownership and validity predicates: PASS
```
drwells: we don't need to fix it now, but we should teach agents how to write decent tests which have reasonable output. We don't need to keep printing 'PASS'.

tests/IBTK/interpolation_utilities.cpp
```diff
@@ -32,6 +32,15 @@
...
 #include <ibtk/app_namespaces.h>
 
+namespace
+{
+void
+require_kernel_check(bool valid, const char* message)
+{
+    if (!valid) TBOX_ERROR(message << '\n');
+}
+} // namespace
```
drwells: Also worth fixing (not necessarily here, but in `AGENTS.md`: a simple `TBOX_ASSERT()` would suffice. We need to keep entropy under control.

boyceg: Replaced the one-off Boolean-abort wrappers with direct TBOX_ASSERT checks throughout the preA-owned test additions.

tests/IBTK/ib_kernel_3d.output
boyceg: Removed the per-check PASS narration. The ordinary assertion-only cases now create empty success output, while the expected-error baselines retain their diagnostics.

drwells:
I did some preliminary testing and this makes LEInteractor a little slower (about 2%): it isn't much but we should double-check the conversion code in LEInteractor with callgrind.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,134 @@
...
+
+    /*!
+     * \brief Construct a scalar identity.
+     *
+     * \param name Name containing at most MAX_NAME_LENGTH ASCII letters,
+     * digits, or underscores. Scalar aliases are accepted.
+     */
+    explicit IBKernel(const std::string& name);
+
+    /*! \brief Construct a scalar identity from a nonnull C string. */
```
boyceg: We need to list the (same) constraints on `name` for both constructors.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,134 @@
...
+    explicit IBKernel(const char* name);
+
+    /*! \brief Return whether \p name satisfies the scalar kernel-name grammar. */
```
boyceg: grammar —> requirements?

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,125 @@
...
+    //! Maximum number of factors in the current public description.
+    static constexpr std::size_t MAX_ACTIVE_FACTORS = 2;
+
+    //! Inline factor capacity, including reserved headroom for a future role.
+    static constexpr std::size_t INLINE_FACTOR_CAPACITY = 3;
+
+    //! Canonical factor storage returned by construction helpers.
+    struct CanonicalFactors
+    {
+        //! Inline factor values; entries beyond size are not observable.
```
boyceg: not observable seems odd

boyceg:
@drwells I think the problem is that the kernels were being passed by the callers as strings at all times. I think the performance issue goes away if we encode the kernels only when initializing everything. Please check this out.

ibtk/src/utilities/IBKernelTensorProduct.cpp
```diff
@@ -0,0 +1,261 @@
...
+bool
+equal_ignoring_case(std::string_view lhs, std::string_view rhs)
+{
+    // Kernel names use ASCII, so case conversion does not require a locale.
+    if (lhs.size() != rhs.size()) return false;
+    for (std::size_t i = 0; i < lhs.size(); ++i)
+    {
+        const char lhs_char = lhs[i] >= 'a' && lhs[i] <= 'z' ? static_cast<char>(lhs[i] - 'a' + 'A') : lhs[i];
+        if (lhs_char != rhs[i]) return false;
+    }
```
abarret: Doesn't this assume `rhs` is already upper-case? If intentional, the function name should be changed to make this clear.

How about
```c++
std::equal(
    lhs.begin(), lhs.end(),
    rhs.begin(), rhs.end(),
    [](char l, char r -> bool {
        return std::tolower(static_cast<unsigned char>(l))
             == std::tolower(static_cast<unsigned char>(r));
    }
);
```

tests/CMakeLists.txt
```diff
@@ -249,6 +249,12 @@ IF(${IBAMR_HAVE_LIBMESH})
 ENDIF()
 
 # IBTK:
+SETUP(IBTK ib_kernel.cpp IBAMR2d)
+TARGET_SOURCES(tests-IBTK_ib_kernel PRIVATE IBTK/ib_kernel_static_initialization.cpp)
```
abarret: ?
```suggestion
SETUP_2D(IBTK ib_kernel.cpp)
TARGET_SOURCES(tests-IBTK_ib_kernel_2d PRIVATE IBTK/ib_kernel_static_initialization.cpp)
```

abarret (commented):
I think @drwells will have more comments than I do but looks find to me. Just a couple of comments.

tests/CMakeLists.txt
boyceg: Fixed.

ibtk/src/utilities/IBKernelTensorProduct.cpp
boyceg: Sure.

boyceg:
@abarret I think this addresses your comments.

abarret (approved):

boyceg:
OK; this has been squashed. @drwells do you want to take another look before I merge?

doc/news/changes/minor/20260901Boyce_R01preA
```diff
@@ -0,0 +1,3 @@
+Improved: Unify kernel identification in LEInteractor.
```
drwells, line 1: Can you be more specific? What does this mean for users of `LEInteractor`?

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,140 @@
...
+{
+/*!
+ * \brief Class IBKernel identifies a one-dimensional IB kernel by name.
+ *
+ * Names are case-insensitive. PIECEWISE_CONSTANT and PIECEWISE_LINEAR are
+ * aliases for BSPLINE_1 and BSPLINE_2, respectively. Applications may define
+ * additional names, but must provide the corresponding kernel implementations.
+ *
+ * UNKNOWN denotes an unspecified kernel and is not included in the standard
+ * kernel list.
```
drwells, line 35: Can we get rid of `UNKNOWN`? Making illegal states impossible would simplify a lot of control flow.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * The same name has the same encoded value on every MPI process. Use getName()
+ * to write kernel names to input or restart files.
```
drwells, line 38: Like my comment elsewhere: this isn't really useful information for anyone. Let's not add comments like "you need to do the same thing on every MPI process".

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,140 @@
...
+    static const IBKernel BSPLINE_3;
+    static const IBKernel BSPLINE_4;
+    static const IBKernel BSPLINE_5;
+    static const IBKernel BSPLINE_6;
+    static const IBKernel IB_3;
+    static const IBKernel IB_4;
+    static const IBKernel IB_4_W8;
+    static const IBKernel IB_5;
+    static const IBKernel IB_6;
+    static const IBKernel PIECEWISE_CUBIC;
```
drwells, line 59: For this class: will there ever be user code which creates `IBKernel`s which are not one of these predetermined values? If not, then we can use the "multi-ton" pattern (a multiple singleton), like we do in deal.II, and assign a unique integer to each of these and get rid of the complex encoding logic.

I'm in favor of having a predetermined list of kernels. If people want more unusual kernels they can patch the list. That would also significantly simplify the upcoming PRs where we construct tensor products of evaluators.

ibtk/include/ibtk/IBKernel.h
```diff
@@ -0,0 +1,140 @@
...
+    explicit IBKernel(const std::string& name);
+
+    /*!
+     * \brief Construct a kernel from its name.
+     *
+     * \param name Nonnull C string satisfying the name requirements of
+     * IBKernel(const std::string&).
+     */
+    explicit IBKernel(const char* name);
+
```
drwells, line 77:
```suggestion
```

Is there a situation where we need a C-string and cannot use `std::string`?

ibtk/include/ibtk/IBKernelTensorProduct.h
```diff
@@ -0,0 +1,133 @@
...
+{
+/*!
+ * \brief Class IBKernelTensorProduct describes a tensor product of scalar IB kernels.
+ *
+ * A single factor applies in every coordinate direction. With two factors,
+ * the first applies along an axis selected by the calling code and the second
+ * applies in the remaining directions. Two equal factors are stored as a single
+ * factor.
+ *
+ * Factors may be IBKernel::UNKNOWN to describe an unspecified kernel.
```
drwells, line 36: Can we ban this? We can simplify downstream usage a lot by not creating objects we cannot use.

ibtk/include/ibtk/LEInteractor.h
```diff
@@ -93,16 +103,23 @@ class LEInteractor
...
-     * specified kernel function.
+     * \brief Return whether LEInteractor supports the named kernel.
+     *
+     * Invalid or unsupported names return false.
      */
-    static int getStencilSize(const std::string& kernel_fcn);
+    static bool isKnownKernel(const std::string& kernel_fcn);
+
+    //! Return whether LEInteractor supports the named kernel; invalid names return false.
+    static bool isKnownKernel(const char* kernel_fcn);
```
drwells, line 113: Like elsewhere - is there something we cannot do with `std::string`?

ibtk/include/ibtk/LEInteractor.h
```diff
@@ -71,6 +73,14 @@ namespace IBTK
...
  * spread values (\em not densities) from Lagrangian meshes to Eulerian grid
  * patches.
+ *
+ * Kernel factors follow IBKernelTensorProduct's axis-relative convention. The
+ * distinguished axis is the side-normal axis for side data, the edge axis for
+ * edge data, and axis zero for cell and node data.
+ *
+ * Legacy string names implicitly construct a resolved description at the API
+ * boundary. Code that repeatedly applies the same kernel should construct and
+ * reuse an IBKernelTensorProduct.
```
drwells, line 83: What does 'legacy' mean here? This comment only makes sense if you know about the old revisions of this file.

drwells (changes requested):
How do you feel about removing the user defined and invalid kernels? Maintaining a list of all the kernels we support in IBAMR would make basically every downstream patch a lot simpler.

Have people used the user-define kernels?

boyceg:
> How do you feel about removing the user defined and invalid kernels? Maintaining a list of all the kernels we support in IBAMR would make basically every downstream patch a lot simpler.

I want these to be things that can be defined without having to recompile the library.

> Have people used the user-define kernels?

I have — that is why it exists in the first place.

But, more generally, as I have already written elsewhere in this discussion — we have new criteria for defining kernel functions, and I would like to be able to experiment with different ones without having to recompile all of IBAMR.

ibtk/include/ibtk/LEInteractor.h
boyceg: No, that is not true — legacy string names are the ways that we used to pass around kernel function specifications.

ibtk/include/ibtk/IBKernel.h
boyceg: I disagree — it simplifies control flow because it gives a way of saying "something went wrong". For instance if you have a kernel generator that barfs, should it return a valid kernel? That seems like it will complicate control flow, not simplify it.

ibtk/include/ibtk/IBKernel.h
boyceg: I very strongly disagree. This says that the encoding is guaranteed to give you consistent values across MPI processes!

ibtk/include/ibtk/IBKernel.h
boyceg: No, the goal is to have an extensible and efficient way to specify these guys.

With the old Fortran subroutines, sure, there is a fixed set of functions that we can use. But we don't have to use Fortran for everything...!

ibtk/include/ibtk/IBKernel.h
boyceg: This seemed to be needed to guarantee that old code that uses default arguments like "IB_4" or whatever works.

ibtk/include/ibtk/IBKernelTensorProduct.h
boyceg: No.

ibtk/include/ibtk/LEInteractor.h
boyceg: This seemed to be needed to guarantee that old code that uses default arguments like "IB_4" or whatever works.

ibtk/include/ibtk/IBKernel.h
boyceg: For instance, you could imagine an encoding that is based on the order that the names appear in the process — that would not be guaranteed to give you consistent encodings that you could pass around to different MPI processes.

Or you could encode based on a pointer value in a data structure. That would similarly not be guaranteed to give you a consistent name.

doc/news/changes/minor/20260901Boyce_R01preA
boyceg: It means that the `LEInteractor` interface has changed in this PR.

ibtk/include/ibtk/IBKernel.h
drwells: If it doesn't work why wouldn't we call `TBOX_ABORT()` or `TBOX_ERROR()`? The other case I can think of is delayed initialization, and nowadays we can use `std::optional` for that (which the implementation already does in a few places internally)

ibtk/include/ibtk/IBKernel.h
drwells: Let's keep it. I think we'll just have to agree to disagree on this one.

ibtk/include/ibtk/IBKernel.h
drwells: That might have been necessary at some intermediate step in converting LEInteractor but is it still true?

ibtk/include/ibtk/LEInteractor.h
drwells: We know that, but it isn't possible to obtain that information from reading any of the documentation right now since the API was updated.

ibtk/include/ibtk/IBKernel.h
boyceg: If you have a function that has to return something, even if there is an error before you get to return, I would prefer to have a return value that somehow signals an "invalid" value rather than have it return a "valid" thing.

A concrete example is the B-spline generator in #1950 — if you ask for a B-spline function that is too-high-order, I agree there should be an error — but there still needs to be something that is provided as the returned value, doesn't there? I would prefer that thing be "invalid" or "unknown".

Are you objecting to the name or to the concept? We commonly have values in `enum`s that correspond to "invalid" values. This seems like it is the same thing as that.

ibtk/include/ibtk/IBKernel.h
boyceg: Not necessarily in the library itself but it could appear in user code that has hard-coded these identifiers.

boyceg: The goal was to make this seamless if you are using "legacy" descriptors.

boyceg: BTW - this is all in response to your request to have a common description of kernel names.

ibtk/include/ibtk/LEInteractor.h
boyceg: So your complaint is that this does not explicitly state "kernels used to be identified by character arrays, but now we identify them as strings that are converted to this type"?

ibtk/include/ibtk/IBKernel.h
drwells: If something can fail to parse we could use `std::optional` to return 'nothing' instead of an invalid value. This would definitely help simplify #1950 (which checks for `UNKNOWN` in several places) and make it easier to generically specify new kernels (one of our goals!) which have to be valid.

ibtk/include/ibtk/LEInteractor.h
drwells: Yup - by referencing an old API without defining it we don't really help users who are trying to understand this file.

ibtk/include/ibtk/LEInteractor.h
drwells: I was curious so I tried it out. The only thing which fails to compile if we delete this function is the new test code which explicitly calls this function with string literals (it fails because of the new non-explicit `IBKernelTensorProduct(const char *)`): i.e., `LEInteractor::interpolate(Q_data, Q_depth, X_data, X_depth, q_data, patch, interp_box, "IB_4");` is totally fine.

ibtk/include/ibtk/IBKernel.h
drwells: see https://github.com/IBAMR/IBAMR/pull/1955#discussion_r3993081933

It is good that we added these interfaces. We can deprecate `IBKernelTensorProduct(const char *)` so everyone upgrades to the new interface.

ibtk/include/ibtk/LEInteractor.h
boyceg: I don't see the harm in keeping it but (as with the constructor above) marking it as deprecated?

boyceg:
Superseded by #1986.

PR #1956, opened 2026-09-03
CI: increment the IBSAMRAI2 version number for macs.

### LINKED ISSUE #1938: ibsamrai2 needs some more patches to make things work on macs

drwells:
Fixes #1938. This is a CI change so we don't need the checklist.

### DIFF
--- a/.github/workflows/push-pull.yml
+++ b/.github/workflows/push-pull.yml
@@ -195,7 +195,7 @@ jobs:
         run: |
           mkdir -p ~/unpack-samrai
           cd ~/unpack-samrai
-          export SAMRAI_VERSION=2026.06.21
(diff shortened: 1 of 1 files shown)


boyceg (approved):

drwells:
we finally have a clean CI run! Thanks for fixing #1938!

PR #1965, opened 2026-09-10
[CAV R05] Fix 3D IB_5 spreading

boyceg:
Fix 3D spreading with the IB_5 kernel, which accumulated every first-coordinate stencil weight into a single grid cell, and update the existing `spread_01_3d.ib_5` baseline.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


drwells:
Can we skip the extra test added in this PR? Updating the existing regression test should suffice for showing that this bug is fixed.

boyceg:
> Can we skip the extra test added in this PR? Updating the existing regression test should suffice for showing that this bug is fixed.

Indeed — I was actually having Codex audit test coverage with the goal of reducing unneeded tests last night when I, ummm, ran out of tokens.

But it is **_definitely_** taking our standard guidance, "Does this change include a bug fix or new functionality? If so, a new test or tests should be added," way too seriously. I think we need to make this more precise.

Perhaps this should be more sharply "the major features" introduced or fixed get tests?

PR #1984, opened 2026-09-12
Require C++20

boyceg:
Require C++20 to build IBAMR and applications that use its public headers.

### IBAMR Pull Request Checklist
- [ ] Does the test suite pass?
- [x] Was clang-format (i.e., `make indent`) run? For more information see
      `scripts/formatting/README.md`.
- [ ] Were relevant issues cited? Please remember to add `Fixes #12345` to close
      the issue automatically if we are fixing the problem.
- [x] Is this a change others will want to know about? If so, then has a
      changelog entry been added?
- [ ] If new data structures have been added to a class then have they been set
      up appropriately for restarts? If so, ensure that the restart version
      number is incremented.
- [ ] Does this change include a bug fix or new functionality? If so, a new test
      or tests should be added. New tests should run quickly (less than a minute
      in release mode). If possible, an older test should gain a new option so
      that we do not need to compile more test executables.
- [x] Did you (if your account has permission to do so) set relevant labels on
      GitHub for the pull request?

PR #1972, opened 2026-09-10
Add unified pointwise Cartesian grid functions

boyceg:
Add a unified functor-based API for initializing and transforming scalar, vector, and tensor patch data, with centering selected from a SAMRAI variable at construction. Supersedes #1640 and fixes #1595.

### IBAMR Pull Request Checklist
- [ ] Does the test suite pass?
- [x] Was clang-format (i.e., `make indent`) run? For more information see
      `scripts/formatting/README.md`.
- [x] Were relevant issues cited? Please remember to add `Fixes #12345` to close
      the issue automatically if we are fixing the problem.
- [x] Is this a change others will want to know about? If so, then has a
      changelog entry been added?
- [x] If new data structures have been added to a class then have they been set
      up appropriately for restarts? If so, ensure that the restart version
      number is incremented.
- [x] Does this change include a bug fix or new functionality? If so, a new test
      or tests should be added. New tests should run quickly (less than a minute
      in release mode). If possible, an older test should gain a new option so
      that we do not need to compile more test executables.
- [x] Did you (if your account has permission to do so) set relevant labels on
      GitHub for the pull request?

PR #1985, opened 2026-09-12
Unify Cartesian data centering

boyceg:
Introduce shared Cartesian centering traits for `CartGridPointwiseFunction` and `CartGridFunctionSet`. Based on #1984; revises the unreleased API from #1972.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


ibtk/include/ibtk/CartesianCentering.h
```diff
@@ -0,0 +1,170 @@
...
+{
+/*! \brief Cartesian patch-data centerings; UNKNOWN also denotes an unsupported scalar type. */
+enum class DataCentering
+{
+    CELL = 0,
+    NODE = 1,
+    SIDE = 2,
+    FACE = 3,
+    EDGE = 4,
+    UNKNOWN = -1
```
drwells: Like elsewhere, I'm reasonably sure that we don't have to add this any more. If we don't have a known value then we can just throw an exception (and that will make every subsequent piece of logic simpler). If things can fail then we can use `std::optional` and let the caller handle it (that class also has nice monadic helpers for help optionally unpacking values and chaining function calls).

ibtk/include/ibtk/CartesianCentering.h
```diff
@@ -0,0 +1,170 @@
...
+};
+
+template <>
+inline DataCentering string_to_enum<DataCentering>(const std::string& value);
```
drwells:
```suggestion
DataCentering string_to_enum<DataCentering>(const std::string& value);
```

ibtk/include/ibtk/CartesianCentering.h
```diff
@@ -0,0 +1,170 @@
...
+
+template <>
+inline std::string enum_to_string<DataCentering>(DataCentering value);
```
drwells:
```suggestion
std::string enum_to_string<DataCentering>(DataCentering value);
```

ibtk/include/ibtk/CartesianCentering.h
```diff
@@ -0,0 +1,170 @@
...
+                                                     SAMRAI::math::PatchSideDataBasicOps<NDIM, T>,
+                                                     SAMRAI::math::PatchFaceDataBasicOps<NDIM, T>,
+                                                     SAMRAI::math::PatchEdgeDataBasicOps<NDIM, T>>>;
+
+    using Index = std::tuple_element_t<static_cast<int>(C),
+                                       std::tuple<SAMRAI::pdat::CellIndex<NDIM>,
+                                                  SAMRAI::pdat::NodeIndex<NDIM>,
+                                                  SAMRAI::pdat::SideIndex<NDIM>,
+                                                  SAMRAI::pdat::FaceIndex<NDIM>,
+                                                  SAMRAI::pdat::EdgeIndex<NDIM>>>;
```
drwells, line 109: This is pretty clever! Codex really loves tuple and tuple accessories for some reason.

ibtk/include/ibtk/CartesianCentering.h
```diff
@@ -0,0 +1,170 @@
...
+template <>
+inline std::string enum_to_string<DataCentering>(DataCentering value);
+
+/*!
+ * \brief Types and geometry for one Cartesian data centering.
+ *
+ * The enum selects the SAMRAI data, factory, index, and patch arithmetic types.
+ * Scalar type and data depth are independent of centering. Side, face, and edge
+ * data have separate arrays for each orientation; cell and node data do not.
+ * For oriented data, axis must be in [0, NDIM); cell and node operations ignore it.
```
drwells: We don't use 'oriented' to describe these arrays (it appears nowhere in SAMRAI's documentation). Could you use more specific terminology for this concept for staggered grids?

ibtk/include/ibtk/CartesianCentering.h
```diff
@@ -0,0 +1,170 @@
...
+
+    using Index = std::tuple_element_t<static_cast<int>(C),
+                                       std::tuple<SAMRAI::pdat::CellIndex<NDIM>,
+                                                  SAMRAI::pdat::NodeIndex<NDIM>,
+                                                  SAMRAI::pdat::SideIndex<NDIM>,
+                                                  SAMRAI::pdat::FaceIndex<NDIM>,
+                                                  SAMRAI::pdat::EdgeIndex<NDIM>>>;
+
+    /*! \brief Whether this centering has oriented arrays. */
+    static constexpr bool is_oriented();
```
drwells: same thing.

ibtk/include/ibtk/CartesianCentering.h
```diff
@@ -0,0 +1,170 @@
...
+    template <typename T>
+    static bool has_axis(const Data<T>& data, int axis);
+
+    /*! \brief Iterate over an orientation of the given cell box. */
+    static typename Data<double>::Iterator begin(const SAMRAI::hier::Box<NDIM>& box, int axis);
+
+    /*! \brief Return the offset from a cell's lower corner in units of its widths. */
+    static VectorNd offset(int axis);
+
+    /*! \brief Convert a native data index to Cartesian coordinate order. */
```
drwells: Similarly, what is 'native'?

ibtk/include/ibtk/private/CartesianCentering-inl.h
```diff
@@ -0,0 +1,199 @@
...
+bool
+CartesianCentering<C>::has_axis(const Data<T>& data, const int axis)
+{
+    if constexpr (C == DataCentering::SIDE)
+    {
+        return data.getDirectionVector()(axis) != 0;
+    }
+    else
+    {
+        return true;
```
drwells, line 94: Is this correct for edge data? I thought that was the same as side in 2d (I didn't check).

Could you document that this is `true` for all non-staggered centerings?

ibtk/include/ibtk/CartesianCentering.h
```diff
@@ -0,0 +1,170 @@
...
+    static SAMRAI::hier::Index<NDIM> cartesian_index(const Index& index);
+};
+
+/*! \brief A Cartesian layout supplying data types, iteration, and cell-relative offsets. */
+template <typename Layout>
+concept Centering = requires(const typename Layout::template Data<double>& data,
```
drwells: I'm not convinced that we actually need to add this concept. `CartesianCentering` encodes all the possible centerings and we can already use enums for this purpose:

```cpp
#include <iostream>

enum class DataCentering
{
A, B, C, D
};

template <DataCentering Layout>
void
foo()
{
    if constexpr (Layout == DataCentering::A)
        std::cout << "A\n";
}

int main()
{
    foo<DataCentering::A>();
    foo<DataCentering::B>();
}
```
which is another advantage of not permitting `DataCentering::UNKNOWN`.

ibtk/src/utilities/CartGridFunctionSet.cpp
```diff
@@ -145,180 +134,76 @@ CartGridFunctionSet::setDataOnPatchLevel(const int data_idx,
...
-#if !defined(NDEBUG)
-    TBOX_ASSERT(cc_var || ec_var || fc_var || nc_var || sc_var);
-#endif
-    VariableDatabase<NDIM>* var_db = VariableDatabase<NDIM>::getDatabase();
-    const int cloned_data_idx = var_db->registerClonedPatchDataIndex(var, data_idx);
-    level->allocatePatchData(cloned_data_idx);
+    const auto add_functions = [&]<Centering Layout>()
+    {
+        using Data = typename Layout::template Data<double>;
+        typename Layout::template PatchOps<double> patch_ops;
```
drwells:
```suggestion
    const auto add_functions = [&]<DataCentering Centering>()
    {
        // etc.
```
we need some minor tweaks to work with just the enum and not a concept.

drwells (changes requested):
I ended up having less time to work on this today than I thought I would. I think this is a nice step forward (and it is telling that even with a new feature we are down 300 lines of code). I want to think some more about how we name these helper functions.

One thing I don't see in the utilities or in this class is depth: will that be handled somewhere else?

ibtk/include/ibtk/CartesianCentering.h
boyceg: As you might have gathered, I like to have a value available to signal that something has gone wrong. 🤣

Two things:

I’m under the impression that there are performance improvements that can be made if we know that we don’t have exceptions — and also I guess I don’t know how exceptions play with (eg) GPUs.

Also, I really don’t like using the name “optional” for things that are actually “errors”.

Codex seems to have a nice workaround for both using C++23. Let me see if I can dig it up when I am home from kids’ soccer practice…

boyceg: Looks like neither Kokkos or CUDA like exceptions FWIW.

boyceg: OK; I removed `UNKNOWN` after all. I did not know about `[[noreturn]]`!!!

ibtk/include/ibtk/CartesianCentering.h
boyceg: Done.

ibtk/include/ibtk/CartesianCentering.h
boyceg: Done.

ibtk/include/ibtk/CartesianCentering.h
boyceg: Renamed `is_oriented()` to `is_staggered()` and updated accordingly. The documentation also now identifies side/face axes as normal directions and edge axes as tangent directions.

ibtk/include/ibtk/CartesianCentering.h
boyceg: Fixed.

ibtk/include/ibtk/CartesianCentering.h
boyceg: Tried to clarify that these are SAMRAI cell/node/side/face/edge indices. `cartesian_index()` returns their coordinates in Cartesian order; the face case undoes `FaceIndex`'s permutation with `toCell(1)`, without shifting the normal coordinate.

ibtk/include/ibtk/private/CartesianCentering-inl.h
boyceg: Edge data allocate every coordinate direction, whereas side data can restrict allocation through their direction vector. In 2D, their locations correspond with exchanged axis labels because edge axes are tangent and side axes are normal.

`has_axis()` now documents this distinction and that cell/node data ignore the axis and always return true.

ibtk/include/ibtk/CartesianCentering.h
boyceg: Removed.

ibtk/src/utilities/CartGridFunctionSet.cpp
boyceg: Both lambdas now take `<DataCentering C>`. The side-specific constructor is selected with `if constexpr (C == DataCentering::SIDE)`.

boyceg:
> One thing I don't see in the utilities or in this class is depth: will that be handled somewhere else?

Depth is already handled in `CartGridPointwiseFunction::applyPointwise()`: it reads `data.getDepth()`, evaluates scalar callbacks at every depth, and handles complete vectors/tensors with the corresponding shape checks. (Depth is not centering.) `CartGridFunctionSet` preserves depth when cloning and uses SAMRAI operations to add all components.

ibtk/include/ibtk/CartesianCentering.h
boyceg: We can get rid of the training calls to `std::abort()` if we merge IBAMR/IBSAMRAI2#70.

ibtk/include/ibtk/private/CartGridPointwiseFunction-inl.h
```diff
@@ -118,354 +120,180 @@ CartGridPointwiseFunction<Value, Data, Function>::setDataOnPatch(
...
-                    Values::load(q, data, index, depth, d_tensor_storage);
-                    Values::assign_result(
-                        result,
-                        std::invoke(d_function, std::as_const(q), std::as_const(x), time, depth, axis),
-                        data.getDepth());
+                    if (d_tensor_storage == TensorStorage::SYMMETRIC && !result.isApprox(result.transpose()))
+                    {
+                        TBOX_ERROR(
+                            "CartGridPointwiseFunction: symmetric storage requires a symmetric callback result\n");
+                    }
```
drwells: we should only check this in debug mode.

ibtk/include/ibtk/CartGridPointwiseFunction.h
```diff
@@ -54,16 +50,27 @@ enum class TensorStorage
...
+std::string enum_to_string<TensorStorage>(TensorStorage value);
+
+/*! \brief Supported collocated pointwise values. */
+template <typename Value>
+concept PointwiseValue = std::same_as<Value, double> || std::same_as<Value, VectorNd> ||
+                         std::same_as<Value, VectorXd> || std::same_as<Value, MatrixNd>;
+
+/*! \brief A functor matching exactly one pointwise callback signature. */
+template <typename Function, typename Value>
+concept PointwiseCallback =
```
drwells: These aren't really callback functions so we should use a different name.

ibtk/include/ibtk/CartGridPointwiseFunction.h
```diff
@@ -54,16 +50,27 @@ enum class TensorStorage
...
+TensorStorage string_to_enum<TensorStorage>(const std::string& value);
 
 template <>
-inline std::string enum_to_string<TensorStorage>(TensorStorage value);
+std::string enum_to_string<TensorStorage>(TensorStorage value);
+
+/*! \brief Supported collocated pointwise values. */
+template <typename Value>
+concept PointwiseValue = std::same_as<Value, double> || std::same_as<Value, VectorNd> ||
+                         std::same_as<Value, VectorXd> || std::same_as<Value, MatrixNd>;
```
drwells: I'll also be a little picky here: these are types, not values. If we want to use concepts we should have some naming guides so we stay consistent. For example, we commonly use `TYPE` here for the underlying scalar type and I think it is fine to analogously extend that to what this function does as well. Similarly, I'd name the concept `OutputType`.

ibtk/include/ibtk/CartGridPointwiseFunction.h
```diff
@@ -54,16 +50,27 @@ enum class TensorStorage
...
+
+/*! \brief A functor matching exactly one pointwise callback signature. */
+template <typename Function, typename Value>
+concept PointwiseCallback =
+    PointwiseValue<Value> && (std::is_invocable_r_v<Value, Function&, const VectorNd&, double, int, int> !=
+                              std::is_invocable_r_v<Value, Function&, const Value&, const VectorNd&, double, int, int>);
```
drwells: I think this checks that things can be invoked in one of two ways but I don't understand why.

ibtk/src/utilities/CartGridFunctionSet.cpp
```diff
@@ -145,180 +134,78 @@ CartGridFunctionSet::setDataOnPatchLevel(const int data_idx,
...
-    const int cloned_data_idx = var_db->registerClonedPatchDataIndex(var, data_idx);
-    level->allocatePatchData(cloned_data_idx);
+    const auto add_functions = [&]<DataCentering C>()
+    {
+        using Layout = CartesianCentering<C>;
+        using Data = typename Layout::template Data<double>;
+        typename Layout::template PatchOps<double> patch_ops;
+        VariableDatabase<NDIM>* var_db = VariableDatabase<NDIM>::getDatabase();
+        const int cloned_data_idx = var_db->registerClonedPatchDataIndex(var, data_idx);
+        level->allocatePatchData(cloned_data_idx);
```
drwells: Let's not modify multiple files in one PR if we can avoid it (though this is a good example of how this functionality considerably reduces our boilerplate).

ibtk/include/ibtk/CartesianCentering.h
```diff
@@ -0,0 +1,154 @@
...
+    template <typename T>
+    static bool has_axis(const Data<T>& data, int axis);
+
+    /*! \brief Iterate over the given cell box in the requested coordinate direction. */
+    static typename Data<double>::Iterator begin(const SAMRAI::hier::Box<NDIM>& box, int axis);
+
+    /*! \brief Return the offset from a cell's lower corner in units of its widths. */
+    static VectorNd offset(int axis);
+
+    /*! \brief Return the cell, node, side, face, or edge index coordinates in Cartesian order. */
```
drwells: Can you add some details on what this conversion does and why? Is the input implicitly cell-centered?

ibtk/include/ibtk/CartesianCentering.h
```diff
@@ -0,0 +1,154 @@
...
+     * Side data use their direction vector. Face and edge data allocate every
+     * direction. Cell and node data ignore axis and always return true.
+     */
+    template <typename T>
+    static bool has_axis(const Data<T>& data, int axis);
+
+    /*! \brief Iterate over the given cell box in the requested coordinate direction. */
+    static typename Data<double>::Iterator begin(const SAMRAI::hier::Box<NDIM>& box, int axis);
+
+    /*! \brief Return the offset from a cell's lower corner in units of its widths. */
```
drwells: This needs some more details. Furthermore, for side-centered data, is this always the lower value?

boyceg:
I am going to try something out...

boyceg:
@codex review

PR #1919, opened 2026-04-17
Implicit IB 1 - solver operator foundation

boyceg:
### IBAMR Pull Request Checklist
- [x] Does the test suite pass?
- [x] Was clang-format (i.e., `make indent`) run? For more information see
      `scripts/formatting/README.md`.
- [x] Were relevant issues cited? Please remember to add `Fixes #12345` to close
      the issue automatically if we are fixing the problem.
- [x] Is this a change others will want to know about? If so, then has a
      changelog entry been added?
- [x] If new data structures have been added to a class then have they been set
      up appropriately for restarts? If so, ensure that the restart version
      number is incremented.
- [x] Does this change include a bug fix or new functionality? If so, a new test
      or tests should be added. New tests should run quickly (less than a minute
      in release mode). If possible, an older test should gain a new option so
      that we do not need to compile more test executables.
- [x] Did you (if your account has permission to do so) set relevant labels on
      GitHub for the pull request?

PR #1986, opened 2026-09-13
[CAV R01-preA-alt] Introduce named IB kernel specifications

boyceg:
Accept scalar and normal/tangential kernel specifications in LEInteractor.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


boyceg:
@abarret, @drwells: This is an alternative to #1955. Let me know what you think.

ibtk/src/lagrangian/LEInteractor.cpp
```diff
@@ -1921,194 +1926,169 @@ enum KernelType
...
+    // intentionally blank
+    return;
+}
+
+void
+LEInteractor::printClassData(std::ostream& os)
+{
+    os << "LEInteractor::printClassData():\n";
+    return;
+}
```
drwells: Why are these here?

drwells:
Putting up an alternative implementation helps. At the same time I think it's a hint this patch is doing too much (circa 1000 lines of non-test code is about my upper limit, which is approximately the size of this PR).

Before I start my review, let's check in on the high-level goals. This functionality seems like the right place to start and I want to have a general idea of the plan before we get going (we can reference other PRs, but I don't have the capacity to read all of them to gather this information myself). Is there a writeup of this somewhere? If not let's do that here.

The ultimate goal is implicit IB, right? To get there we need smoothers, and to get smoothers we need a more general implementation of kernels. This patch implements some generic infrastructure for defining composite kernels in a much more flexible way than a de-facto internally hardcoded list of strings (it encodes them with a couple of integers).

What do you anticipate will happen to `LEInteractor` and all the one-kernel-at-a-time Fortran code? We can't use it for the smoother code (since we need the values of the kernels so we can do local Vanka solves). Do you anticipate that we will be able to completely replace it by the new C++ versions of kernels? If so, I think we can avoid a lot of extra churn by just ignoring `LEInteractor` for now.

It would help if you could point me to the specific places we will use `IBKernel` and `IBKernelTensorProduct`. We should be careful with names: #1986 introduces `ib_kernels.h` which is pretty confusing nomenclature. Do we just need these objects to dispatch to `IBOperator` (or some equivalent) to build the correct collections of Vanka matrices?

boyceg:
The reason this patch and R01a exist is in response to comments like:

* https://github.com/IBAMR/IBAMR/pull/1919/changes#r3117460128
* https://github.com/IBAMR/IBAMR/pull/1919/changes#r3117475846
* https://github.com/IBAMR/IBAMR/pull/1919/changes#r3117532070

R01-preA gives an extensible way of 'naming' IB kernel functions, and R01a gives a common set of definitions for the kernels that are provided by `LEInteractor`. The names and implementations are intentionally kept separate since I can imagine imagine having different implementations of the same kernel for different use cases — that is why `ib_kernels.h` exists separately from `IBKernel.h` — it is where we have implementations of the "standard" kernels that are "named by default" in the library.

This PR ensure that names that correspond to existing kernels supported by `LEInteractor` "just work". I could split that out into a separate PR if you want, but the changes to `LEInteractor` are literally just changing `std::string` to the new name. However, in this stack of development branches, the implicit solver ***uses `LEInteractor`*** to evaluate the ***action*** of the operator. (We are building preconditioners, not solvers!) So it is useful to have a common way of naming the kernels.

The stuff in PR01a--d turns out to be very useful for building much faster versions of `LEInteractor`, and so I think we should keep it.

boyceg:
Anyway, the one and only goal of this PR is to give an extensible and efficient way of naming scalar and tensor-product IB kernels that plays nicely with `LEInteractor` and that can be used later to specify pieces of the implicit solver construction.

boyceg:
(I.e. let's not use either `std::string` — we know it is too slow — or `enum` — it is not extensible — to name these things.)

boyceg:
This also addresses:
* https://github.com/IBAMR/IBAMR/pull/1950#discussion_r3908638472

abarret (approved):
I have no comments or suggestions. Seems fine to me.

PR #2004, opened 2026-09-21
Fix loss of the Stokes-IB FAC coarse solver on partial level reset

boyceg:
Fix `StaggeredStokesIBLevelRelaxationFACOperator` losing its coarse solver when `setResetLevels()` excludes the coarsest level, which made `solveCoarsestLevel()` abort in Debug builds and use a null solver in Release builds. The coarse solver is now kept and re-initialized with the recreated solution and rhs vectors. Adds `stokes_ib_fac_partial_reset`, a two-level regression test that fails without the fix.

This does not cover retained intermediate-level solvers in hierarchies with three or more levels, or the `rediscretize_stokes = FALSE` coarse-solver path, which already fails on a double `MatDestroy` at first initialization.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


tests/IB/stokes_ib_fac_partial_reset.input
```diff
@@ -0,0 +1,88 @@
...
+
+StandardTagAndInitialize {
+   tagging_method = "REFINE_BOXES"
+   RefineBoxes {
+      level_0 = [ (0, 0), (N - 1, N - 1) ]
+   }
+}
+
+LoadBalancer {
+   processor_mapping = "SFC"
```
drwells: This isn't a load balancer option. Let's not add things like this to the library.

PR #2005, opened 2026-09-21
Fix IBMethod::getForceData(TimePoint) returning position data

boyceg:
`IBMethod::getForceData()` called with a `TimePoint` forwarded to `getPositionData()`, so it returned position data; it now forwards to the double-time `getForceData()`.

      changelog entry been added?
      number is incremented.
      GitHub for the pull request?


tests/IB/ib_body_force.cpp
```diff
@@ -136,12 +136,42 @@ main(int argc, char* argv[])
...
+            for (const auto time_pt : { TimePoint::CURRENT_TIME, TimePoint::HALF_TIME })
+            {
+                std::vector<Pointer<LData>>* F_data;
+                ib_method_ops->getForceData(&F_data, &needs_ghost_fill, time_pt);
+                ib_method_ops->getPositionData(&X_data, &needs_ghost_fill, time_pt);
+                if (F_data == X_data || (*F_data)[finest_ln] == (*X_data)[finest_ln])
+                {
+                    TBOX_ERROR("IBMethod::getForceData() returned position data\n");
+                }
+            }
```
drwells, line 163: This isn't a very good test because it tests that we no longer have an obviously wrong implementation, rather than we actually did things correctly. Can you modify this test to check that we really do return the correct force data?

drwells (changes requested):

tests/IB/ib_body_force.cpp
boyceg: Good point — updated to call computeLagrangianForce() at CURRENT_TIME and HALF_TIME and print the actual per-vertex force components read back through the TimePoint overload (same values IBStandardForceGen::setUniformBodyForce() was given), compared via numdiff against a real expected output, instead of a pointer/identity check. Confirmed it fails (prints the vertex positions, not the forces) with the one-line fix reverted.

boyceg: Updated. See if this is better.

boyceg:
@drwells does this work?

